Issue #36, opened 2014-11-10
Some Felix rules may be unnecessary in a non-bridged network


plwhite:
Two particular cases in the code.

* There are many physdev rules in Felix. It's entirely possible that, since Calico does not bridge the network, these rules are not required and we could rationalise replacing the physdev rules with a simple match on interface. That would be a big win in terms of number of rules, and so might well be worthwhile.

* There is a rule that prevents an endpoint acting as a DHCP server. In a non-bridged network, this might be impossible anyway, in which case we should remove the rule for simplicity.

PR #59, opened 2014-12-08
Remove unneeded dropping of DHCP packets.

plwhite:
Endpoints are on an L3 routed network; this rule is not required.

Fixes final parts of issue #36

### DIFF
--- a/calico/felix/frules.py
+++ b/calico/felix/frules.py
@@ -294,26 +294,14 @@ def set_ep_specific_rules(suffix, iface, type, localips, mac):
     index += 1
 
     if type == IPV4:
-        # Allow outgoing DHCP packets.
+        # Allow outgoing v4 DHCP packets.
         rule = fiptables.Rule(type, "RETURN")
         rule.protocol = "udp"
         rule.create_udp_match("68", "67")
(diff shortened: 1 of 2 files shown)


Lukasa:
This PR now fails unit test.

plwhite:
I bet it does. It won't merge properly, I bet. I'm going to do some investigation then sort it out though.

Lukasa:
Actually, i think it merges fine, it just changes the iptables rules which breaks the expectations of the tests.

plwhite:
Grabbing this until it's in a fit state to be merged.

plwhite:
I tested this by using some snoop to see if any DHCP packets could make it from the host to the VM on an OpenStack deployment. The packets duly arrived on the host, but were not forwarded to the VM. I think that means I am confident that this isn't a hole unless somebody has done some explicit reconfiguring of the host to allow it (in which case they have made their own problem if it is a problem for them!). Hence I want to clean up the merge and check in.

plwhite:
Right, this did not merge for reasons that make no sense. I have sorted that out completely by the nuclear method of destroying and recreating the branch with the same changes in it, and now believe it is ready to be merged. Cory, over to you.

PR #64, opened 2014-12-11
pip requirements files

### LINKED ISSUE #63: There does not appear to be a pip requirements file
Adding install_requires=… to the setup.py would make it a lot easier to start installing/developing Calico. 

cc /@liljenstolpe

Lukasa:
This pull request adds requirements files for the various Calico components that can be interpreted by `pip`. It also extends `setup.py` to set those requirements as install-time requirements. Finally, it adds an environment variable, `$CALICODEPS`, that allows users to avoid installing *all* the requirements for every Calico component if they're only using one.

Currently I'm missing the dependencies for the OpenStack plugin because I can't work out how to install neutron via `pip`, so I don't know what its package name is yet. I also need to consult with @neiljerram, @plwhite and @matthewdupre to confirm that these requirements are in fact correct. We shouldn't merge until I've done that digging.

Resolves #63.


neiljerram:
Yes, all looks good to me now - thanks.

PR #66, opened 2014-12-12
Cl3 bird filter

liljenstolpe:
This set of changes does a few things:

1) Sets up two  community strings for "calico export via bgp" and "calico deny export via bgp" based on the AS_NUMBER that is set as part of the Bird configuration setup and two values set in the shell script that does the configuration (666 for deny, 777 for allow).  

2) Installs a filter on the import of direct routes that assigns one of the two community strings on each route imported.  The criteria is on the ifname where the route is learned.  Currently routes learned via tap*, veth*, and dummy1 are tagged with export, the rest tagged with !export.  

Dummy1 is a specific case.  lo* routes in Linux can't be used external to local machine.  dummy* interfaces can be.  There will be some routes that we want to be local machine only, and others that we may want to be reachable via the fabric, or wider (i.e. a dns resolver, API endpoint, etc).  Services that aren't in guests should therefore have their service address aliased to either dummy0 or dummy1 depending on if they want their service advertised beyond the local machine.

3) Installs a filter for BGP export that will ONLY export routes tagged with the "calico export via bgp" community string.

THESE CHANGES HAVE NOT BEEN TESTED - I don't have a testing infra here.  They look correct, and I've double checked them, but someone needs to test them before accepting the pull.

Christopher

### DIFF
--- a/etc/bird/calico-bird.conf.template
+++ b/etc/bird/calico-bird.conf.template
@@ -10,6 +10,17 @@ router id @MY_IP_ADDRESS@;
 
 
 filter tag_direct {
+# The ifname if match will match any route from a "calico advertisable interface"
+# and tag it with the advertise community string.  All other routes will also be 
+# accepted, but will be taged with the no advertise string
+# We start off unsetting all community strings on the route as a safety measure.
+# 
+# Update the ifname list if we have more interface types in Calico.  The dummy1
+# match is important.  There are "loopback" addresses we may want to advertise,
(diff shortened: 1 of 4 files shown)


Lukasa:
ok to test

Lukasa:
ok to test

Lukasa:
ok to build

Lukasa:
Closing this for now because we're swamped under pull requests. Please re-raise if you think it's vitally important, or try to get a dev to commit to getting this into a specific release.

Issue #35, opened 2014-11-10
Workaround for python-iptables bug should be retired


plwhite:
We currently have a workaround for a python-iptables bug in the code.
https://github.com/ldx/python-iptables/issues/111
That bug has been fixed, and we should get rid of that workaround once we are confident that our install process picks up the fix.

Look for "111" in the code to find the relevant places.

PR #69, opened 2014-12-17
Add unit tests for Felix

plwhite:
This constitutes a reasonable level of cover for Felix (86% is being reported to me), though requires some work to get it cleaned up and to 100% coverage (notably in `test_felix`, which is just too long but still does not have enough coverage).

The biggest issue with it is that where I have stubbed out code, I have often used global variables to track system state. That's ugly, but has no major downsides for now, so I proposing leaving as is for now (sorting it out would require some code refactoring, but is not a huge job).

Sample test output is the following.

    C:\work\git\calico>nosetests --with-coverage 2>&1 | more
    ...............'iptables' is not recognized as an internal or external command, operable program or batch file.
    'iptables' is not recognized as an internal or external command, operable program or batch file.
    'ip6tables' is not recognized as an internal or external command, operable program or batch file.
    'iptables' is not recognized as an internal or external command, operable program or batch file.
    'ip6tables' is not recognized as an internal or external command, operable program or batch file.
    .'iptables' is not recognized as an internal or external command, operable program or batch file.
    ..'iptables' is not recognized as an internal or external command, operable program or batch file.
    ...........
    Name                    Stmts   Miss  Cover   Missing
    -----------------------------------------------------
    calico.felix                0      0   100%
    calico.felix.config        76      0   100%
    calico.felix.devices       40      0   100%
    calico.felix.endpoint      85     18    79%   46, 130-132, 147-148, 159, 163-168, 173-175, 178-181, 209-214
    calico.felix.felix        291     88    70%   173-176, 195, 220-227, 246-250, 262-267, 306-308, 338, 351-387, 402-404, 428-431, 448-452, 457-458, 481, 499-503,
    508, 558-561, 564-565, 568-569, 572-574, 627-649, 657-658, 661-662, 668, 680, 701, 706-728, 731
    calico.felix.frules       342     21    94%   103, 133, 412, 584, 603, 623-626,637-640, 648, 657-675
    calico.felix.fsocket      105     10    90%   78-80, 122-124, 144-146, 163
    calico.felix.futils        36      0   100%
    calico.felix.ipsets        25      0   100%
    -----------------------------------------------------
    TOTAL                    1000    137    86%
    ----------------------------------------------------------------------
    Ran 29 tests in 0.224s

    OK

The tedious complaints about `iptables` not being present are benign, and are due to the hack in place to deal with issue #35, which hopefully we can retire shortly.


README.md
```diff
@@ -34,7 +34,10 @@ the native networking implementation – but with higher security,
 
 To get started, you first need a working installation of
 [OpenStack](http://www.openstack.org/). Then download and install the latest
-stable build of Calico [here](http://www.projectcalico.org/download/).
+stable build of Calico following the instructions [here](http://www.projectcalico.org/download/).
+
+Technical documentation is [here](https://github.com/Metaswitch/calico-docs/wiki); if you are going to contribute to the project, you'll also need to run the
+[tests](doc/CalicoUTs.md).
```
plwhite: I agree the punctuation and spacing is a bit odd, but I didn't write or change most of it! I've cleaned up though.

calico/felix/test/stub_fiptables.py
```diff
@@ -103,21 +111,36 @@ def __eq__(self, other):
...
-    def __init__(self, name, rules=[]):
+    def __init__(self, name):
         self.name = name
-        self.rules = rules
+        self.rules = []
+        self.type = None # Not known until put in table.
 
     def flush(self):
-        self.rules = []
+        del self.rules[:]
```
plwhite: I agree that's possible, but it might be confusing because it is possible for a caller to get hold of rules from a chain (and in some cases the rules are directly accessed this way). It seems less confusing if `chain.rules` is a reference to the rules of the chain rather than a reference to the rules before they were flushed.

Lukasa: I *suppose* that's plausible. From my perspective, however, it looks a bit like a footgun that exists for no reason. It should be easy for me to know if `flush` has been called in between now-and-then. I don't care enough to block on this though.

calico/common.py
```diff
@@ -24,9 +24,17 @@
 import logging
 import logging.handlers
 import os
+import re
 import sys
 import errno
 
+# Various useful regular expressions.
+IPV4_REGEX = re.compile("^\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}$")
+IPV6_REGEX = re.compile("^[a-f0-9]+:[:a-f0-9]+$")
```
plwhite: I have replaced these with utility functions, but kept the regexes within the functions. I think that is good enough for now - we could make it better, but it feels like premature optimisation to load a library or write any more complex code to get a 0.0001% use case.

calico/felix/ipsets.py
```diff
@@ -0,0 +1,79 @@
...
+        # ipset list failed - either does not exist, or an error. Either way,
+        # try creation, throwing an error if it does not work.
+        futils.check_call(["ipset", "create", name, typename, "family", family])
+
+def destroy(name):
+    """
+    Destroy an ipset if it exists (and do nothing if not).
+    """
+    if futils.call_silent(["ipset", "list", name]) == 0:
+        futils.check_call(["ipset", "destroy", name])
```
plwhite: I'm confused. If `create` fails, then it throws an exception (because `check_call` throws an uncaught exception). That brings the house down.

I think you might mean that it would be better to spot the difference between "destroy failed" and "list failed" - but it seems slightly more robust to assume that list will fail only very rarely for reasons other than "does not exist", while "destroy" might really fail.

calico/felix/felix.py
```diff
@@ -273,10 +276,10 @@ def handle_endpointcreated(self, message):
         """
         log.debug("Received endpoint create: %s", message.fields)
 
+        # TODO: Ought to firewall missing mandatory fields here.
         endpoint_id = message.fields['endpoint_id']
-        resync_id   = message.fields['resync_id']
-        issued      = message.fields['issued']
         mac         = message.fields['mac']
+        resync_id   = message.fields.get('resync_id')
```
plwhite: I was going with being tolerant of missing fields, at least for now. There's a TODO above to tidy this stuff up, and put in some better handling of missing fields.

plwhite:
I think we should leave as is for now, and resolve when we resolve #35 - that hopefully can be soon (i.e. this week).

calico/felix/ipsets.py
Lukasa: Yeah, it's no longer clear to me what I meant anyway. Clearly not important.

calico/felix/frules.py
```diff
@@ -653,16 +653,19 @@ def update_ipsets(type,
...
-                    icmp_code = 0
-                suffix = ",%s:%s/%s" % (protocol, icmp_type, icmp_code)
-                ipset  = tmp_ipset_port
             else:
-                # Not an integer ICMP type - must be a string code name.
-                suffix = ",%s:%s" % (protocol, icmp_type)
-                ipset  = tmp_ipset_port
+                try:
+                    # Assume integer ICMP type.
+                    int(str(icmp_type))
```
Lukasa: If `icmp_type` is either an integer or a string, then this is really weird. Just call `int()` on it: if it's a non-integer string it'll fail with the `ValueError` you're catching later on. The double coercion is unnecessary, because we clearly don't expect a floating point value.

Lukasa:
One small note.

Lukasa:
Alright, I'm happy. Let's do this.

PR #81, opened 2015-01-05
Allow loggers to be turned off

lwr20:
Allow a new log level: NONE which disables that log handler.

### DIFF
--- a/calico/common.py
+++ b/calico/common.py
@@ -88,7 +88,7 @@ def complete_logging(logfile=None,
     logging is available as early in execution as possible, i.e. before the
     config file has been parsed.
 
-    This function must only be called once, after 
+    This function must only be called once, after
     :meth:`default_logging() <calico.common.default_logging>`
     has been called.
     """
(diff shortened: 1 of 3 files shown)


calico/common.py
```diff
@@ -96,14 +96,20 @@ def complete_logging(logfile=None,
...
         elif isinstance(handler, logging.StreamHandler):
-            handler.setLevel(stream_level)
+            if stream_level is None:
+                root_logger.removeHandler(handler)
+            else:
+                handler.setLevel(stream_level)
 
     # If we've been given a log file, log to file as well.
-    if logfile is not None:
+    if logfile is not None and file_level is not None:
```
Lukasa: I suggest broadening this check to `if not logfile and file_level is not None:`. The reason for doing this is that a user may accidentally put the following line in a logfile:

```python
LogFilePath = 
```

I don't *know* that this will lead to the `logfile` value being the empty string, but I'd certainly want to guard against the possibility.

Lukasa:
Suggest rewording your commit message to say what the change is, rather than what issue this is for.

Lukasa:
LGTM, let's see what @plwhite thinks.

Issue #73, opened 2014-12-19
Daemons should allow binding to a specific address


nbartos:
Currently daemon bind statements do not allow specifying a specific IP address to bind to.  This is necessary/desirable on some machines with multiple IP addresses that serve a variety of functions.

Example output of netstat:
tcp        0      0 0.0.0.0:9902            0.0.0.0:*               LISTEN      32153/python2.7     
udp        0      0 0.0.0.0:47502           0.0.0.0:*                           32153/python2.7     
tcp        0      0 0.0.0.0:9905            0.0.0.0:*               LISTEN      367/python2.7       
tcp        0      0 0.0.0.0:9906            0.0.0.0:*               LISTEN      367/python2.7  

Relevant code sections:
calico/openstack/mech_calico.py:        self.felix_router_socket.bind("tcp://*:%s" % PLUGIN_ENDPOINT_PORT)
calico/openstack/mech_calico.py:        self.acl_get_socket.bind("tcp://*:%s" % PLUGIN_ACLGET_PORT)
calico/openstack/mech_calico.py:        self.acl_pub_socket.bind("tcp://*:%s" % PLUGIN_ACLPUB_PORT)
calico/felix/fsocket.py:            self._zmq.bind("tcp://*:%s" % self.port)
calico/acl_manager/acl_publisher.py:        self.router_socket.bind("tcp://*:%s" % MANAGER_ACLGET_PORT)
calico/acl_manager/acl_publisher.py:        self.pub_socket.bind("tcp://*:%s" % MANAGER_ACLPUB_PORT)

plwhite:
Resolved by pull request #82

Lukasa:
This was fixed in #82. Thanks for the report @nbartos.

PR #82, opened 2015-01-06
Resolve issue #73 (allow explicit specification of bind address)

plwhite:
Applies to both Felix and ACL Manager.

### DIFF
--- a/calico/acl_manager/acl_manager.py
+++ b/calico/acl_manager/acl_manager.py
@@ -42,6 +42,7 @@ def main():
     config.read(args.config_file or 'acl_manager.cfg')
 
     plugin_address = config.get('global', 'PluginAddress')
+    bind_address = config.get('global', 'BindAddr')
     log_file_path = config.get('log', 'LogFilePath')
     log_file_level = config.get('log', 'LogSeverityFile')
     log_syslog_level = config.get('log', 'LogSeveritySys')
@@ -77,7 +78,7 @@ def main():
     acl_store = ACLStore()
     network_store = NetworkStore()
 
-    publisher = ACLPublisher(context, acl_store)
+    publisher = ACLPublisher(context, acl_store, bind_address)
     acl_store.start(publisher)
 
     processor = RuleProcessor(acl_store, network_store)
--- a/calico/acl_manager/acl_manager.py
+++ b/calico/acl_manager/acl_manager.py
@@ -44,7 +44,6 @@ def main():
 
     plugin_address = config.get('global', 'PluginAddress')
     local_address = config.get('global', 'LocalAddress')
-    except ConfigParser.NoOptionError:
     log_file_path = config.get('log', 'LogFilePath')
     log_file_level = config.get('log', 'LogSeverityFile')
     log_syslog_level = config.get('log', 'LogSeveritySys')
--- a/calico/acl_manager/acl_manager.py
+++ b/calico/acl_manager/acl_manager.py
@@ -35,14 +35,16 @@ def main():
     log_defaults = {'LogFilePath': None,
                     'LogSeverityFile': 'INFO',
                     'LogSeveritySys': 'ERROR',
-                    'LogSeverityScreen': 'ERROR'}
+                    'LogSeverityScreen': 'ERROR',
+                    'LocalAddress': '*'   }
 
     # Read config file.
     config = ConfigParser.ConfigParser(log_defaults)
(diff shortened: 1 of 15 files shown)


plwhite:
Note that this does not cover the calico-chef repository, where the config files need to be updated (with comments, not actual values). I'll do that after this is merged.

calico/acl_manager/acl_manager.py
```diff
@@ -37,6 +37,7 @@ def main():
     config.read(args.config_file or 'acl_manager.cfg')
     plugin_address = config.get('global', 'PluginAddress')
     log_file_path = config.get('log', 'LogFilePath')
+    bind_address = config.get('global', 'BindAddr')
```
Lukasa: `BindAddr` is a weird name for this. `LocalAddress`?

calico/felix/config.py
```diff
@@ -163,6 +166,14 @@ def validate_cfg(self):
         self.validate_addr("PluginAddress", self.PLUGIN_ADDR)
         self.validate_addr("ACLAddress", self.ACL_ADDR)
 
+        #*********************************************************************#
+        #* Bind address must be * or an IP address.                          *#
+        #*********************************************************************#
+        if (self.BIND_ADDR != "*" and
+                not common.validate_ipv4_addr(self.BIND_ADDR)):
```
Lukasa: Allow IPv6 as well.

Lukasa: Actually, I think this can be a hostname. I think `localhost` is a perfectly valid string to place here.

calico/felix/test/data/felix_bad_bindaddr.cfg
```diff
@@ -0,0 +1,27 @@
...
+#EndpointRetryTimeMillis = 500
+# Time between complete resyncs
+#ResyncIntervalSecs = 1800
+# Plugin and ACL manager addresses
+PluginAddress = localhost
+ACLAddress    = localhost
+# Metadata IP (or host) and port. If no metadata configuration, set to None
+#MetadataAddr  = 127.0.0.1
+#MetadataPort  = 9697
+BindAddr = localhost
```
Lukasa: Per the above comment, I don't think this is a problem at all.

etc/acl_manager.cfg
```diff
@@ -1,6 +1,8 @@
 [global]
 # Plugin address
 PluginAddress = controller
+# Local address to bind to; IP address or "*" (the default).
```
Lukasa: Inconsistent comments.

Lukasa:
Some comments, back to you. =D

calico/felix/config.py
plwhite: We do not support IPV6 for `0MQ`. If we were to do so, then this would have to change (the point is that `0MQ` is IPv4 only unless you explicitly enable IPv6 by disabling the `ZMQ_IPV4ONLY` option (which none of our code does). That means that an IPv6 address wouldn't work, so we should not permit it.

calico/acl_manager/acl_manager.py
plwhite: Maybe. My rationale for "BindAddr" is that this is the address that the process binds to - which is also a local address.

I think I don't really care, and it is trivial to change. Do you still want me to change it, or have a convinced you that you don't care either?

etc/acl_manager.cfg
plwhite: In what way?

calico/felix/config.py
plwhite: I think it probably does make sense to specify a hostname, though it feels a bit odd - so I will change that.

etc/acl_manager.cfg
Lukasa: Doesn't match the comment used in the other config file, despite being identical fields.

calico/acl_manager/acl_manager.py
Lukasa: I think `LocalAddress` compares more intuitively with `PluginAddress` than `BindAddr` does.

calico/felix/config.py
Lukasa: Fair enough, but in that case we have another bug, which is that we don't support an IPv6 only backend, but we should.

plwhite:
I have raised issue #83 for the lack of support if a pure IP v6 deployment, and addressed (I think!) all the other issues. Back to you.

Lukasa:
This no longer cleanly merges: do you mind rebasing on master?

Issue #77, opened 2014-12-22
Calico Neutron Agent does not allow setting host parameter


nbartos:
Neutron agents (neutron-dhcp-agent, neutron-metadata-agent, etc) have a setting in their config file called "host", which is defined in neutron/common/config.py and defaults to utils.get_hostname().

Setting this parameter works fine for the neutron agents that ship with neutron, however it does not work for the Calico agent.

As our environment does not have hostnames which are resolvable from other nodes, we set the Neutron/Nova/etc hostnames to the IP address so all nodes can access each other.

plwhite:
Neil, I assigned this to you because it seems more likely that you can make progress on it, and it's on our list to fix.

Lukasa:
@plwhite I disagree. Felix reports back its hostname to the plugin in [its resyncstate request](https://github.com/Metaswitch/calico-docs/wiki/Calico-API-Proposal#request-resyncstate). This is the information that needs to be changed, so Felix needs configuration to do it.

plwhite:
Agreed - think I misread it. Taking it back for fixing.

Lukasa:
@nbartos This functionality was merged in #84. Thanks so much for reporting it!

PR #75, opened 2014-12-20
Log out daemon name to syslog.

Lukasa:
This ensures that we match the syslog format more closely. It resolves #71.

This needs functional testing to confirm it works.

The only problem with this fix is that I've hardcoded the binary executable names. That's fine for the 99% use-case, but doesn't necessarily scale well. It might be worth writing some code that can reliably determine the binary name from `sys.argv[0]`, but I didn't want to do that for this initial fix. Shout if anyone thinks I should.

PR #84, opened 2015-01-06
Resolve issue #77 (allow Felix hostname to be explicitly set)

plwhite:
Includes a fix for a bug introduced with the fix to #75 (which broke Windows unit tests).

This is not complete in itself, though it is logically consistent.

* Calico-chef sample config files need updating.
* Something somewhere needs to actually set the value correctly in the install.


Lukasa:
LGTM. :cake:

PR #86, opened 2015-01-06
Fix incorrect spelling of LocalAddress in comments.

plwhite:


plwhite:
Tedious mistake, but probably worth just fixing now to avoid confusion.

PR #87, opened 2015-01-07
Fix up missing dst file in metadata rule (introduced by testing fixes).

plwhite:
There are two bugs here.

* `dst` was not being set up in rules correctly, following the restructuring for testing. This led to a bad iptables rule.
* We failed to insert the rule at the start.

This can have annoying side effects, so worth getting closed down quickly if possible.

### DIFF
--- a/calico/felix/fiptables.py
+++ b/calico/felix/fiptables.py
@@ -95,6 +95,14 @@ class Rule(object):
         self._rule.src = value
 
     @property
+    def dst(self):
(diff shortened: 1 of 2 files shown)


Lukasa:
LGTM. :cake:

Lukasa:
Would be nice to have a regression test for this, but as the fix is urgent I'll expedite the PR. Please open a new issue to track adding a test for it.

Issue #79, opened 2014-12-30
Calico driver reports success to neutron when it knows it failed.


Lukasa:
As reported by @nbartos [on the mailing list](http://lists.projectcalico.org/mail-archive/msg00103.html).

The summary is: when the VM creation event occurs and the driver can't find the host to program, it should not subsequently return a success code. The same is true for any port update events: Neutron should know that we failed.

Relevant logs are [here](https://gist.github.com/nbartos/c593e048f58b357d6148). Assuming it hasn't been changed, see the error behaviour on line 280 of the log.

I think this bug is moderate-priority.

neiljerram:
This is, I think, a particular case of the more general discussion at https://github.com/Metaswitch/calico/issues/19.

neiljerram:
OK, I think what we need here are the following steps.

- Extend https://github.com/Metaswitch/calico-docs/wiki/Calico-API-Proposal to say that the intent of the Endpoint API's support for Felix-plugin disconnection is only to support transient disconnections - e.g. for either component crashing or being reconfigured or upgraded - and is not intended to allow the plugin to accept new port bindings 'on spec' when it doesn't yet have any connection to a Felix on the relevant compute host.

- Add a comment saying the some thing in #19, and then close that issue.

- In send_endpoint (in mech_calico.py), after

       LOG.info("No connection to this host, bail out")

  raise an Exception.  Preferably something that is easy to identify if we later decide that some places might need to handle it specially - does that mean a new Exception class?  I'm sure @Lukasa can advise there.  I believe raising an Exception will be sufficient for Neutron to realize that the attempted port creation has failed.

- Add an entry to the 'Test cases' section of https://bitbucket.org/metaswitch/calico-internal-docs/src/59a2a74e6ca764265584bf280b80eae7d5d9f9b9/nj/plugin-ut-plan.org?at=master to record that we should have UT cases to cover this, i.e. launching a VM when plugin has no Felix connection; also updating a port when no Felix; deleting a port when no Felix.

- Check that FV still passes.

- Report fixed on the mailing list.

@Lukasa, @plwhite Does that sound right?  Have I missed anything?

@plwhite, how would you feel about taking all this on?  (As I don't have time myself at the moment.)

Lukasa:
This sounds right to me. A new subclass of `Exception` is the way to go.

plwhite:
[...]

PR #90, opened 2015-01-08
Handle errors from failure to connect to Felix

plwhite:
Fixes bug #79.

After this fix, failure to connect to Felix leads to an error that is passed back to neutron.


neiljerram:
Looks good - thanks!

PR #92, opened 2015-01-08
Various code tidies, including logging fixes

plwhite:
Fixes various logs, including some that had invalid parameters.
Avoids using "id" where we mean "endpoint id suffix".


plwhite:
Cory, you might as well take this.

plwhite:
I just added another simple and uncontroversial fix to this pull request - we had a leak of unused ipsets, so I fixed it.

Lukasa:
LGTM. :cake:

PR #93, opened 2015-01-08
Acl manager ut

matthewdupre:
ACL Manager UT for the rule processing module, including a bugfix to allow multiple security groups on one endpoint.


matthewdupre:
Markups made.

PR #94, opened 2015-01-09
Delete old (pre-Felix) Calico agent

neiljerram:


neiljerram:
When merging this, we should also update our Chef-based FV system so that it installs and runs nova-metadata-api, and (I think) does not run neutron-metadata-agent.

After doing both those things, we can then revert any metadata-related changes from our Neutron fork.

neiljerram:
Hi @Lukasa, assigning this to you for review - and I guess for the Chef change? - now that Jenkins has OK'd it.

Lukasa:
Chef change went in as Metaswitch/calico-chef@3a3cc0d647de7a3d8f4ce005bafe48c84da05e6e.

Lukasa:
Now just waiting on getting the FV rig back.

neiljerram:
Thanks!  Should MetadataPort in felix.cfg be 8775, though?

Otherwise this change LGTM.

------------------

PR #96, opened 2015-01-12
Remove workaround for (now fixed) python-iptables bug.

plwhite:
Remove the workaround, but also require version 0.7.0 or better of python-iptables everywhere.

### DIFF
--- a/calico/felix/frules.py
+++ b/calico/felix/frules.py
@@ -116,27 +116,10 @@ def set_global_rules(config):
         fiptables.insert_rule(rule, chain, 0)
         truncate_rules(chain, 1)
 
-    #*************************************************************************#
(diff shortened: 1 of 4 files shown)


plwhite:
Oh dear. This works for the `INPUT` and `FORWARD` tables, but not the `felix-INPUT` and `felix-FORWARD` tables (that end up with lots of duplicates). Need to take another look at why that is.

plwhite:
I have now fixed the problems with duplicate rules. This is ready for review.

plwhite:
Fixes bug #35. Well, obviously, but had not put a link in yet.

Lukasa:
LGTM. :cake: :star:

PR #101, opened 2015-01-13
Do not hang in termination in zmq.

plwhite:
`sys.exit(1)` was hanging in some zmq library, according to gdb. We cannot afford to hang in termination, so us `os._exit` instead.

This might be controversial, but is at least pragmatic.

### DIFF
--- a/calico/felix/felix.py
+++ b/calico/felix/felix.py
@@ -24,8 +24,8 @@ The main logic for Felix.
 import argparse
 import collections
 import logging
+import os
 import socket
-import sys
 import time
 import uuid
 import zmq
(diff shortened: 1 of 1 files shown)


Lukasa:
Arg.

Ideally we'd do our best to tidy these threads up. However, we can't really do that, so I guess this will have to do, at least for now.

Lukasa:
However, in my grand tradition of nitpicky code-reviews, `sys` is now an unused import. Mind removing it?

plwhite:
I've applied that markup (and merged the commits). If you are happy with that, then please merge.

Lukasa:
Hurrah! :cake:

Issue #80, opened 2014-12-31
Felix requires ACCEPT-by-default on iptables


Lukasa:
As discussed in the Calico mailing list, Felix doesn't cope with iptables being set to DROP by default on the INPUT or FORWARD chains. Traffic does not get to VMs. That's obviously a problem.

We should look at enhancing Felix to program the appropriate rules in this situation.

plwhite:
This is trivial to fix - probably worth just fixing it ASAP to get rid of it.

plwhite:
We don't want to go nuking a compute host's security rules because that could have some serious side effects - and so we expect to document this as a manual requirement (and add an alarm so that if you get it wrong you can diagnose the problem).

To get past this on an OpenStack system you can run the following commands.

    iptables -A FORWARD -i tap+ -j ACCEPT
    iptables -A FORWARD -o tap+ -j ACCEPT
    iptables -A INPUT -i tap+ -j ACCEPT

Doing so means that traffic from / to tap interfaces (i.e. VMs) is accepted if Felix permits and no earlier rule has barred it.

neiljerram:
In my Fuel work, I've had to add these incantations on the controller:

    # Allow BGP through the Fuel firewall
    iptables -I INPUT 1 -p tcp --dport 179 -j ACCEPT

    # Allow Calico API connections through the Fuel firewall
    iptables -I INPUT 1 -p tcp --dport 9901 -j ACCEPT
    iptables -I INPUT 1 -p tcp --dport 9903 -j ACCEPT
    iptables -I INPUT 1 -p tcp --dport 9904 -j ACCEPT
    iptables -I INPUT 1 -p tcp --dport 9905 -j ACCEPT
    iptables -I INPUT 1 -p tcp --dport 9906 -j ACCEPT

and these on each compute node:

    # Allow BGP through the Fuel firewall
    iptables -I INPUT 1 -p tcp --dport 179 -j ACCEPT

    # Allow Calico API connections through the Fuel firewall
    iptables -I INPUT 1 -p tcp --dport 9902 -j ACCEPT

This is because Fuel installs chains for INPUT that end in DENY.

plwhite:
Key to the fix is that the chains have been restructured. The original structure was as follows.

* We used to have a `felix-INPUT` and `felix-FORWARD` chain. All INPUT packets were sent to `felix-INPUT` and all forwarded packets to `felix-FORWARD`.
[...]

PR #103, opened 2015-01-14
Restructure rules so that DROP default still works.

plwhite:
Fixes bug #80. See extensive comment in that issue for details on why I think this is the right approach.


plwhite:
Not at all a scary fix. This needs some CI runs to make sure it is OK before the next release.

PR #105, opened 2015-01-14
Add support for port ranges

plwhite:
Includes some extra tests to get better coverage of area changed.

Also fixes issue where variable suffix was used twice in same function to mean different things.

### DIFF
--- a/calico/felix/frules.py
+++ b/calico/felix/frules.py
@@ -538,7 +538,7 @@ def update_ipsets(type,
     """
     for rule in rule_list:
         if rule.get('cidr') is None:
-            log.error("Invalid %s rule without cidr for %s : %s",
+            log.error("Invalid %s rule without cidr for %s : %s" %
                       (descr, suffix, rule))
             continue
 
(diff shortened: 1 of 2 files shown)


calico/felix/frules.py
```diff
@@ -538,7 +538,7 @@ def update_ipsets(type,
     """
     for rule in rule_list:
         if rule.get('cidr') is None:
-            log.error("Invalid %s rule without cidr for %s : %s",
+            log.error("Invalid %s rule without cidr for %s : %s" %
```
Lukasa: Why the change to always format?

plwhite: It used to read `log.error(blah %s and %s", (x, y))`. That fails because there are too few arguments. It needed to change to one of `log.error(blah %s and %s", x, y)` or `log.error(blah %s and %s" % (x, y))`. I chose the latter for no good reason whatsoever; I should be consistent, but it's really too late now. PEP8 probably says something about this, come to think of it. Next time I'll check

Lukasa: PEP8 isn't the bit that matters actually.

The reason you care is that using the latter will do the string interpolation every time the line is hit, regardless of whether the log will actually get emitted or not. In this case it doesn't matter much (error level logs will almost certainly get emitted, and it's not too expensive to do this particular reporting), but in some situations it can be extremely expensive.

For that reason, it's a slight performance enhancement to use the first case over the second.

plwhite: Thanks. So we have our policy decided then! I'll be consistent from now on.

I've just added testing for gazillions of logs, so I am no longer so terrified of changing them and breaking things.

PR #107, opened 2015-01-15
Bring python version number up to date.

Lukasa:
This change brings the Python version number of the package up to date.

We need to update our release process to avoid this happening again: I'll open an issue to track it.


Lukasa:
Updated issue in #108.

PR #112, opened 2015-01-15
Fix Felix keepalives

plwhite:
Fixes bug #110

PR #113, opened 2015-01-15
Call send_multipart correctly when responding to heartbeats

neiljerram:


Lukasa:
Might be better if this got moved onto the `connections` branch so that the entire fix appears in #112.

neiljerram:
Done.

PR #124, opened 2015-01-20
General cleanups of ACL manager code.

Lukasa:
Saw these while I was playing around in the code.

Should be no relevant functional change here.


matthewdupre:
Whitespace has already been removed, rebased the other commit.

PR #128, opened 2015-01-20
Change from VIF_TYPE_ROUTED to VIF_TYPE_TAP

neiljerram:
Part of https://github.com/Metaswitch/calico/issues/127


Lukasa:
LGTM. Holding off on a merge until the other changes are proposed, then we can merge all three at once.

Issue #121, opened 2015-01-19
Felix does not send periodic resyncs


plwhite:
Felix does not send resyncs periodically, even though it should.

plwhite:
OK, I think this is an artefact of an outdated Felix install. Cannot repro on latest release, so closing.

plwhite:
I have figured out how to repro. It stops sending resyncs if you ever have a resync including with a zero endpoint_count. Ouch

PR #129, opened 2015-01-20
Ensure that felix handles resyncs with 0 endpoints

plwhite:
Fixes issue #121

The test code is a bit ugly, and could be (and in due course should be) improved and extended. However, the tests do confirm the fix.


Lukasa:
LGTM. Holding off on merge until FV is green.

Issue #133, opened 2015-01-21
Felix does not resubscribe if it recreates the ACL REQ connection


plwhite:
Thus it is no longer subscribed at that point for any endpoints, and gets no more ACL updates for them.

Likely a duplicate of issue #55, though that is somewhat speculation.

PR #134, opened 2015-01-21
If we recreate the ACL SUB connection, resubscribe to everything.

plwhite:
Fixes issue #133

### DIFF
--- a/calico/felix/felix.py
+++ b/calico/felix/felix.py
@@ -653,8 +653,15 @@ class FelixAgent(object):
                 elif sock.type == Socket.TYPE_ACL_REQ:
                     self.acl_queue.clear()
 
-                # Finally, recreate the socket.
+                # Recreate the socket.
                 sock.communicate(self.hostname, self.zmq_context)
+
+                # If this is the ACL SUB socket, then subscribe for all
+                # endpoints.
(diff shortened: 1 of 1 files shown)


neiljerram:
Note that this PR is for merging into master (not stable1).  Is that what you wanted?

------------------

plwhite:
Note that I have nothing to test this on - calico01 has fallen over yet again, though I will retry. However, the fix is logically correct.

plwhite:
@neiljerram I think I want to merge into stable1 too. Damn. I am not going to sort that out right now, since I am just getting very tired and am in the middle of testing it.

Is that an easy thing to change / do?

neiljerram:
Yes, it's easy - we can sort it out in the morning.

Nothing that you merge now will have any effect on the demo - wherever it's merged to - unless you also do a new release to binaries.projectcalico.org.  And I presume you're not contemplating that now - right?

------------------

plwhite:
OK, so this needs to be rebased onto stable1 to create the right pull request. I am not doing that now; it's too late, I am too tired, and I am just going to screw it up if I try. Let's do that stuff in the morning.

neiljerram:
That sounds sensible to me!

------------------

Issue #139, opened 2015-01-22
Felix can fail to send a resync of a connection drops with a keepalive outstanding


plwhite:
Suppose you send a keepalive, and then it is time for a resync. The resync queues up waiting for the connection to clear, but the connection then drops.

We leave ourselves with a resync on the queue even though the resync has completed. The resync is then forgotten, but Felix thinks it is outstanding, and so no resyncs are ever sent ever again.

plwhite:
Fixed as part of the fix to issue #37.

plwhite:
Fixed in PR #155.

PR #140, opened 2015-01-22
Fix issue #139 - do not stop resyncing

plwhite:


plwhite:
I am 90% confident in this fix, but have done no testing whatsoever. Until tests are available it must not be merged; it's here in case somebody else needs to do an emergency release of it.

plwhite:
The fix https://github.com/Metaswitch/calico-docker-prototype/pull/31 is a workaround for this fix.

plwhite:
This bug has been fixed elsewhere, in a way that actually involved testing it - so my proposal is withdrawn.

PR #154, opened 2015-01-29
Plugin UT: checkpoint for first review and check in

neiljerram:
Please let me know what you think!

### DIFF
--- a/calico/openstack/test/test_plugin.py
+++ b/calico/openstack/test/test_plugin.py
@@ -1,0 +1,458 @@
+# -*- coding: utf-8 -*-
+# Copyright 2014 Metaswitch Networks
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+openstack.test.test_plugin
+~~~~~~~~~~~
+
+Top level tests for the Calico/OpenStack Plugin.
+"""
+import logging
+import mock
+import socket
+import sys
+import time
+import unittest
+import uuid
+import eventlet
+import traceback
+import json
+import inspect
+
+sys.modules['neutron'] = m_neutron = mock.Mock()
+sys.modules['neutron.common'] = m_neutron.common
+sys.modules['neutron.openstack'] = m_neutron.openstack
+sys.modules['neutron.openstack.common'] = m_neutron.openstack.common
+sys.modules['neutron.plugins'] = m_neutron.plugins
+sys.modules['neutron.plugins.ml2'] = m_neutron.plugins.ml2
+sys.modules['neutron.plugins.ml2.drivers'] = m_neutron.plugins.ml2.drivers
+
+class DriverBase(object):
+    def __init__(self, agent_type, vif_type, vif_details):
+        pass
+
+m_neutron.plugins.ml2.drivers.mech_agent.SimpleAgentMechanismDriverBase = DriverBase
+
+#*****************************************************************************#
+#* Load calico.felix.devices and calico.felix.test.stub_devices, and the     *#
+#* same for ipsets; we do not blindly override as we need to avoid getting   *#
+#* into a state where tests of these modules cannot be made to work.         *#
+#*****************************************************************************#
+import calico.openstack.mech_calico as mech_calico
+
+REAL_EVENTLET_SLEEP_TIME = 0.2
+
+class TestPlugin(unittest.TestCase):
+    @classmethod
+    def setUpClass(cls):
+        pass
+
+    @classmethod
+    def tearDownClass(cls):
+        pass
+
+    def setUp(self):
+        pass
+
+    def tearDown(self):
+        pass
+
+    def test_startup(self):
+
+        #*********************************************************************#
+        #* Set of addresses that we have sockets bound to.                   *#
+        #*********************************************************************#
+        self.sockets = set()
+
+        #*********************************************************************#
+        #* The simulated time (in seconds) that has passed since the         *#
+        #* beginning of the test.                                            *#
+        #*********************************************************************#
+        self.current_time = 0
+
+        #*********************************************************************#
+        #* Dict of current sleepers.  In each dict entry, the key is an      *#
+        #* eventlet.Queue object and the value is the time at which the      *#
+        #* sleep should complete.                                            *#
+        #*********************************************************************#
+        self.sleepers = {}
+
+        #*********************************************************************#
+        #* When a socket is created, print a message to say so, and hook its *#
+        #* bind method.                                                      *#
+        #*********************************************************************#
+        def socket_created(tp):
+            print "New socket type %s" % tp
+
+            #*****************************************************************#
+            #* Create a new mock socket.                                     *#
+            #*****************************************************************#
+            socket = mock.Mock()
+
+            #*****************************************************************#
+            #* Hook its bind and connect methods, so we can remember the     *#
+            #* address that it binds or connects to.                         *#
+            #*****************************************************************#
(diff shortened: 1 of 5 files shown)


calico/openstack/test/test_plugin.py
```diff
@@ -0,0 +1,458 @@
...
+sys.modules['neutron.openstack.common'] = m_neutron.openstack.common
+sys.modules['neutron.plugins'] = m_neutron.plugins
+sys.modules['neutron.plugins.ml2'] = m_neutron.plugins.ml2
+sys.modules['neutron.plugins.ml2.drivers'] = m_neutron.plugins.ml2.drivers
+
+class DriverBase(object):
+    def __init__(self, agent_type, vif_type, vif_details):
+        pass
+
+m_neutron.plugins.ml2.drivers.mech_agent.SimpleAgentMechanismDriverBase = DriverBase
```
Lukasa: Might be worth commenting to say why you're doing this.

calico/openstack/test/test_plugin.py
```diff
@@ -0,0 +1,458 @@
...
+
+#*****************************************************************************#
+#* Load calico.felix.devices and calico.felix.test.stub_devices, and the     *#
+#* same for ipsets; we do not blindly override as we need to avoid getting   *#
+#* into a state where tests of these modules cannot be made to work.         *#
+#*****************************************************************************#
```
Lukasa: This comment seems spurious.

calico/openstack/test/test_plugin.py
```diff
@@ -0,0 +1,458 @@
...
+        def log_debug(msg):
+            print ">>>>>>DEBUG %s" % msg
+            return None
+        def log_warn(msg):
+            print ">>>>>>>WARN %s" % msg
+            return None
+        def log_exception(msg):
+            print ">>EXCEPTION %s" % msg
+            traceback.print_exc()
+            return None
```
Lukasa: None of this should be necessary: `nosetests` will intercept logging and display it for failed tests.

calico/openstack/test/test_plugin.py
```diff
@@ -0,0 +1,458 @@
...
+        self.felix_endpoint_socket.send_json.reset_mock()
+        self.felix_endpoint_socket.rcv_queue.put_nowait(
+            {'type': 'HEARTBEAT'})
+        self.real_eventlet_sleep(REAL_EVENTLET_SLEEP_TIME)
+        self.real_eventlet_sleep(REAL_EVENTLET_SLEEP_TIME)
+
+        print "Mainline test code finished"
+
+        #*********************************************************************#
+        #* Yield to allow anything pending on other friends to come out.     *#
```
Lukasa: 'friends'?

runtests.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+#rm -rf env
+virtualenv env
+. env/bin/activate
+pip install -e .
+pip install nose mock coverage
+nosetests --with-coverage --nocapture calico.openstack.test.test_plugin:TestPlugin
```
Lukasa: You almost certainly want to remove 'nocapture' here. This is also a really confusing file to add to the repository because it only tests one thing, rather than testing the full set. It would be better to add a Makefile with a `test` target that covers what we do in Jenkins.

calico/openstack/test/test_plugin.py
```diff
@@ -0,0 +1,458 @@
...
+        #* Hook sleeping.                                                    *#
+        #*********************************************************************#
+        self.real_eventlet_sleep = eventlet.sleep
+        mech_calico.eventlet.sleep = simulated_time_sleep
+
+        #*********************************************************************#
+        #* Tell the driver to initialize.                                    *#
+        #*********************************************************************#
+        self.driver.initialize()
+
```
Lukasa: Everything above here is logically setup code, and will likely get duplicated *a lot* for other tests. You should consider pulling it into `TestPlugin.setUp()`.

calico/openstack/test/test_plugin.py
```diff
@@ -0,0 +1,458 @@
...
+        #* Check RESYNCSTATE response was sent.                              *#
+        #*********************************************************************#
+        self.felix_router_socket.send_multipart.assert_called_once_with(
+            ['felix-1',
+             '',
+             json.dumps({'type': 'RESYNCSTATE',
+                         'endpoint_count': 0,
+                         'rc': 'SUCCESS',
+                         'message': 'Здра́вствуйте!'}).encode('utf-8')])
+        self.felix_router_socket.send_multipart.reset_mock()
```
Lukasa: This test is brittle. Python dictionaries aren't guaranteed to be ordered, and so when deserialized their order may change. You may need a subtler test here to ensure that it doesn't break intermittently.

calico/openstack/test/test_plugin.py
```diff
@@ -0,0 +1,458 @@
...
+        #* Tell the driver to initialize.                                    *#
+        #*********************************************************************#
+        self.driver.initialize()
+
+        #*********************************************************************#
+        #* Check that there's a socket bound to port 9901, and get it.       *#
+        #*********************************************************************#
+        bound_sockets = {socket for socket in self.sockets
+                         if socket.bound_address == "tcp://*:9901"}
+        assert len(bound_sockets) == 1
```
Lukasa: Don't use naked `assert` statements, `nosetests` doesn't do well with them and you don't get much introspection. Use the unittest `self.assert*` methods [as documented here](https://docs.python.org/2/library/unittest.html#test-cases).

calico/openstack/test/test_plugin.py
Lukasa: That applies to any test here that uses `json.dumps`.

calico/openstack/test/test_plugin.py
```diff
@@ -0,0 +1,458 @@
...
+        self.felix_endpoint_socket.send_json.assert_called_once_with(
+            {'type': 'HEARTBEAT'},
+            mech_calico.zmq.NOBLOCK)
+        self.felix_endpoint_socket.send_json.reset_mock()
+        self.felix_endpoint_socket.rcv_queue.put_nowait(
+            {'type': 'HEARTBEAT'})
+        self.real_eventlet_sleep(REAL_EVENTLET_SLEEP_TIME)
+        self.real_eventlet_sleep(REAL_EVENTLET_SLEEP_TIME)
+
+        print "Mainline test code finished"
```
Lukasa: This is unnecessary: it should be pretty clear when the test has finished!

Lukasa:
Alright, plenty of initial feedback. =)

Note also that the test has failed. Looks like a mocking problem to me.

calico/openstack/test/test_plugin.py
neiljerram: I find it really useful, during development, to see the logging interleaved with output from the test code.  I definitely wasn't seeing that, before I implemented these hooks.

runtests.sh
neiljerram: I tried a Makefile, but it doesn't work, because all the above commands from '. env/bin/activate' onwards have to be in the same shell.

So, I've kept this as a script but tidied it up and removed the nocapture and plugin-specific testing - let me know what you think when you see the new version.

calico/openstack/test/test_plugin.py
neiljerram: Agreed in principle, but can I leave this for now, until I have more data...

neiljerram:
Thanks @Lukasa.  I've pushed markups for most of your comments - everything except where I've commented more specifically above.  Please let me know what you think about any remaining points.  I'm going to investigate the build failure now.

neiljerram:
@Lukasa I believe this is ready for possible merging now.  Could you merge if you're happy with it?  Obviously I have lots more implementation to do, and there's still the point about brittle dict checking, but I'd like to have this in as a base camp, if you agree.  For example, I think run-unit-test.sh is useful, and also my reorg of .coveragerc.

calico/openstack/test/test_plugin.py
Lukasa: So long as 'more data' is 'even one test failure caused by out-of-order dictionaries', then yes. =)

Lukasa:
Fine by me. =)

calico/openstack/test/test_plugin.py
neiljerram: Absolutely, yes.

Issue #153, opened 2015-01-29
Felix can lose a message when a connection drops


plwhite:
This may apply to `RESYNCSTATE`, but definitely does apply to `GETACLSTATE`. If the message is sent, and no response is ever received, Felix forgets about it. It tidies up the connection after it times out, but we could be left in a situation where we have failed to request an immediate ACL update for a new endpoint. That opens up a timing window where the ACL Manager knows all about state before Felix first subscribes - and yet the GETACLSTATE gets lost and so ACL Manager never tells Felix.

This is actually a pretty unlikely window in practice, but since I found it while refactoring the code, I am going to tidy it up now.

plwhite:
Fixed in PR #155.

Issue #97, opened 2015-01-13
Felix can fail to restart a connection and terminate


plwhite:
The scenario is as follows.

* Felix times out its EP REP socket.
* Felix closes and recreates the socket.
* When ZMQ closes a socket, it tidies it in a background thread, and when Felix tries to reopen it, the close has not completed.

This is timing related, only happening on very slow servers (usually when Felix is running in a container on a slow VM). However, we ought to fix it.

Typical logs are below.

    2015-01-13 10:55:12,703 [WARNING] calico.felix.felix 628: Socket EP REP timed out
    2015-01-13 10:55:12,703 [INFO] calico.felix.fsocket 86: Creating socket to entity 172.17.32.101:9902
    2015-01-13 10:55:12,704 [ERROR] calico.felix.felix 728: Felix exiting after uncaught exception
    Traceback (most recent call last):
      File "/usr/lib/python2.7/dist-packages/calico/felix/felix.py", line 725, in main
        agent.run()
      File "/usr/lib/python2.7/dist-packages/calico/felix/felix.py", line 650, in run
        sock.communicate(self.hostname, self.zmq_context)
      File "/usr/lib/python2.7/dist-packages/calico/felix/fsocket.py", line 92, in communicate
        self._zmq.bind("tcp://*:%s" % self.port)
      File "socket.pyx", line 434, in zmq.backend.cython.socket.Socket.bind (zmq/backend/cython/socket.c:3928)
      File "checkrc.pxd", line 21, in zmq.backend.cython.checkrc._check_rc (zmq/backend/cython/socket.c:6058)
    ZMQError: Address already in use

plwhite:
Worth noting that Felix probably ought never to shut down a socket for timeout if it binds (i.e. listens) rather than connects on that socket. That is probably the right fix.

The logic is that if you are connecting to the far end and it goes away, then you might have a bad socket that is half closed. If you are binding and you get into this state, it's fine - the far end can and must reconnect.

matthewdupre:
Seen by Chris running the docker prototype.

spikecurtis:
I think I'm seeing this as well, or at least something pretty similar.

```
2015-01-26 14:13:45,017 [ERROR] calico.felix.felix 70: Felix starting
2015-01-26 14:13:45,018 [INFO] calico.felix.fsocket 87: Creating socket to entity sjc-calico-00:9902
[...]

Issue #37, opened 2014-11-10
Should refactor fsocket somewhat for simplicity


plwhite:
From NJ's markups :

I think we may have missed a nice encapsulation opportunity here.  I
don't have all the details (obviously), but I'm thinking of a RequestSocket
subclass of felix.fsocket.Socket, which:

- encapsulates the queue, and whether the last sent request is still waiting for a reply
- has a send() method that implements queuing if another request is still
  outstanding, or else actually sends (instead of current
  FelixAgent.send_request method)
- has a receive() method that automatically sends the next queued message
  after a receipt (instead of doing that here)
- automatically clears its queue when closed.

plwhite:
Having seen the spaghetti that has occurred in `felix.py` to handle connection restart, I think we just have to do this.

plwhite:
Fixed in PR #155.

PR #155, opened 2015-01-29
Issue37refactor

plwhite:
Fixes the following issues : 

#37, #153, #139, #97 

Also some drive by fixes, since I had to rewrite so much code anyway.


calico/felix/felix.py
```diff
@@ -643,44 +610,33 @@ def run(self):
...
-                    self.acl_queue.clear()
+                elif sock.type == Socket.TYPE_EP_REQ:
+                    if self.resync_id is not None:
+                        #*****************************************************#
+                        #* The resync we were doing has officially           *#
+                        #* failed. We'll retry when the connection comes     *#
+                        #* back.                                             *#
+                        #*****************************************************#
+                        self.complete_endpoint_resync(False)
+                        endpoint_resync_needed = True
```
Lukasa: How does a timeout on the EP_REQ socket have anything to do with whether the resync has failed?

calico/felix/felix.py
```diff
@@ -411,9 +393,7 @@ def handle_endpointdestroyed(self, message, sock):
             return
 
         # Unsubscribe endpoint.
-        sock = self.sockets[Socket.TYPE_ACL_SUB]
-        sock._zmq.setsockopt(zmq.UNSUBSCRIBE, delete_id.encode('utf-8'))
-
+        self.sockets[Socket.TYPE_ACL_SUB].unsubscribe(delete_id.encode('utf-8'))
```
Lukasa: This line is overlong (81 characters).

calico/felix/felix.py
```diff
@@ -538,8 +518,7 @@ def _create_endpoint(self, endpoint_id, mac):
         self.endpoints[endpoint_id] = endpoint
 
         # Start listening to the subscription for this endpoint.
-        sock = self.sockets[Socket.TYPE_ACL_SUB]
-        sock._zmq.setsockopt(zmq.SUBSCRIBE, endpoint_id.encode('utf-8'))
+        self.sockets[Socket.TYPE_ACL_SUB].subscribe(endpoint_id.encode('utf-8'))
```
Lukasa: This line is overlong (81 characters).

calico/felix/fsocket.py
```diff
@@ -79,37 +117,61 @@ def close(self):
             self._zmq.close()
             self._zmq = None
 
+        # We do this again in communicate, but no harm in doing it here too.
+        if self._send_queue is not None:
+            self._send_queue.clear()
```
Lukasa: Just call `self.clear_queue()`.

calico/felix/felix.py
Lukasa: Separately, I believe this function is bugged. If we're asked to destroy an endpoint we don't think exists, we should either reply with an error code or resync. Currently, as far as I can see, we do neither of those things, we just finish executing.

Lukasa: We should check the other handle_* functions to see if they make the same mistake.

calico/felix/fsocket.py
```diff
@@ -60,16 +62,52 @@ class Socket(object):
...
+        elif type == Socket.TYPE_EP_REP:
+            self.descr = ("plugin (bound socket %s:%s, %s)" %
+                          (self.config.LOCAL_ADDR, self._port, self.type))
+        elif type == Socket.TYPE_ACL_REQ:
+            self.descr = ("ACL Manager %s:%s (request connection %s)" %
+                          (self._remote_addr, self._port, self.type))
         else:
-            self.remote_addr = self.config.ACL_ADDR
+            self.descr = ("ACL Manager %s:%s (subscription connection %s)" %
+                          (self._remote_addr, self._port, self.type))
```
Lukasa: This `else` is asking for bugs in future. I know I did it when I first wrote the code, but we should remove it now: replace with `elif` and have the `else` throw an exception.

calico/felix/fsocket.py
```diff
@@ -79,37 +117,61 @@ def close(self):
...
-        self.last_activity = futils.time_ms()
+        self._last_activity = futils.time_ms()
 
         # We do not have a request outstanding.
-        self.request_outstanding = False
+        self._request_outstanding = False
+
+        # Whatever is on the queues is gone.
+        if self._send_queue is not None:
+            self._send_queue.clear()
```
Lukasa: Again, use `self.clear_queue()`.

calico/felix/fsocket.py
```diff
@@ -213,6 +296,9 @@ def __init__(self, type, fields, endpoint_id=None):
         else:
             self.descr = type
 
+        #: Whether this is a keepalive
+        self.keepalive = keepalive
+
```
Lukasa: This seems to be utterly unused. What's it for?

Lukasa:
Notes inline.

calico/felix/fsocket.py
plwhite: Oops. An earlier version used this field, then I realised it was a bad idea and failed to tidy it up. So I have tidied it up now.

calico/felix/felix.py
plwhite: We lost a message in the exchange to do with the resync; we need to tidy up that resync and try again. I tweaked the comment to this; let me know what you think.

                    #*********************************************************#
                    #* A resync was in progress, but we may have lost the    *#
                    #* RESYNCSTATE request or response or (potentially) an   *#
                    #* ENDPOINTCREATED message due to a lost                 *#
                    #* connection. That means we have to give up on this     *#
                    #* resync, tidy up and retry.                            *#
                    #*********************************************************#

calico/felix/felix.py
plwhite: It is OK to get ACL updates when the endpoint does not exist; I also suspect it may be possible to get a destroy for an endpoint which does not exist. We might (for example) get the destroy but then find that it arrives after the completion of a resync which gets rid of the same endpoint. Or we might have lost the creation because of an earlier issue for which a resync has been initiated.

I therefore think we should just log it as surprising and continue.

Lukasa:
@plwhite Looks like you made an error in a catch-up merge and messed up the branch. Want to fix it?

Lukasa:
Alright, I'm happy to merge this if we can clean the branch up, otherwise I'll cherry-pick in by hand.

PR #156, opened 2015-01-29
ACL Manager stops listening for network updates silently when a rule ref...

matthewdupre:
...erences an empty security group

This fixes the bug and adds a unit test case for it.  The symptom is that ACL Manager stops processing any new information from the mechanism driver (it will still respond to Felix with the information it held at the time of failure).


matthewdupre:
This is a pretty straightforward fix for review.

The crash occurs in line 83, when trying to get the group members of an unknown (or empty) group from group_members.

It's the second fix for issue 55.

Lukasa:
LGTM. :cake:

PR #159, opened 2015-01-30
Ensure ACL Manager exits cleanly with a log when worker threads crash

matthewdupre:
This is still a work in progress - I've created the request just to chop the review up.

ACL Manager does most of its work in child threads, but (before this PR), if one of those crashes the rest of ACL Manager will often silently continue to operate, producing incorrect outputs.  This makes such issues particularly hard to debug.

This PR adds a try-except block around each worker thread which will log any exceptions and kill the whole process.  The service should then be restarted automatically.

### DIFF
--- a/calico/acl_manager/test/test_acl_store.py
+++ b/calico/acl_manager/test/test_acl_store.py
@@ -13,16 +13,21 @@
 # limitations under the License.
 
 import unittest
+import time
+from mock import Mock
 from copy import deepcopy
 
 from stub_acl_publisher import StubACLPublisher
 from stub_processor import StubRuleProcessor
 
 from calico.acl_manager.acl_store import ACLStore
+import calico.acl_manager.utils as utils
 
 
 class TestACLStore(unittest.TestCase):
     """Unit tests for the ACLStore class."""
+
+
 
     def setUp(self):
         self.acl_store = ACLStore()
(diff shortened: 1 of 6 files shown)


calico/acl_manager/utils.py
```diff
@@ -0,0 +1,28 @@
...
+import logging
+
+log = logging.getLogger(__name__)
+
+def terminate():  # pragma: nocover
+    """
+    Log an Exception and terminate ACL Manager
+    """
+    log.critical("ACL Manager terminating - exception %s",
+                 traceback.format_exc())
```
Lukasa: Consider replacing this with `logging.exception()`.

Lukasa: This also needs a big, big warning that you cannot call this from outside an exception handler.

calico/acl_manager/test/test_acl_store.py
```diff
@@ -13,17 +13,22 @@
...
 
 from calico.acl_manager.acl_store import ACLStore
+import calico.acl_manager.utils as utils
 
 
 class TestACLStore(unittest.TestCase):
     """Unit tests for the ACLStore class."""
 
+
+
```
Lukasa: Probably don't need these extra lines.

calico/acl_manager/utils.py
```diff
@@ -0,0 +1,28 @@
...
+#    License for the specific language governing permissions and limitations
+#    under the License.
+
+import traceback
+import os
+import logging
+
+log = logging.getLogger(__name__)
+
+def terminate():  # pragma: nocover
```
Lukasa: Should this maybe take an exit code as an optional argument?

calico/acl_manager/test/test_acl_store.py
```diff
@@ -95,5 +100,21 @@ def test_case2(self):
...
+        """
+        # Mock the terminate function so the UT process doesn't exit
+        utils.terminate = Mock()
+
+        self.processor.test_update_endpoint_acls('e1', self.acls)
+        self.acl_pub.test_set_expected_acls('e1', self.acls)
+        self.acl_pub.test_raise_exception()
+
+        # Allow a second for the worker thread to call the mocked terminate
+        time.sleep(1)
```
Lukasa: This isn't a great option, because it makes the tests timing sensitive. Consider using the [`event` primitive](https://docs.python.org/2/library/threading.html#event-objects) instead.

Lukasa:
Notes inline.

Lukasa:
LGTM. :cake:

PR #169, opened 2015-02-09
Make the Neutron driver respect the Neutron config bind_host setting, for its IP address bindings

neiljerram:


run-unit-test.sh
```diff
@@ -15,5 +15,6 @@ virtualenv env
 . env/bin/activate
 pip install -e .
 pip install nose mock coverage
+rm .coverage
```
Lukasa: This should be unnecessary, I've never seen coverage.py object to having an old coverage directory present.

neiljerram: That's not the point, though.  The point is that .coverage is cumulative, and so - unless it's deleted before each test run - it's easy to make a change that _loses_ some existing code coverage, and not to notice that that coverage has been lost.

neiljerram: (Also, FWIW, it's a file, not a directory.)

Lukasa: It's not cumulative, [see here](http://nedbatchelder.com/code/coverage/cmd.html#h_data_file).

> By default,each run of your program starts with an empty data set.

plwhite:
I'm comfortable with this - I think you should review the (trivial) comments, and then I'll merge.

run-unit-test.sh
neiljerram: I'm sorry, but experience showed that it _was_ cumulative.  I discovered that when Ed asked me why I had reported 53% coverage, while http://calico-jenkins/coverage/ says 38% :-)

I think the text you cited refers to the data set in memory, not in that file.  It is written rather ambiguously, IMO.

Lukasa: Hurrah, conclusion reached! This is `nose`'s fault. In [their docs](http://nose.readthedocs.org/en/latest/plugins/cover.html#cmdoption--cover-erase), they document the `--cover-erase` flag. This means they by default use the equivalent of the `-a` flag to `coverage`, which means they deploy a non-default behaviour.

I recommend turning this off in nose with the `--cover-erase` flag.

neiljerram:
Build has failed, but it appears that all our pull request builds have failed since yesterday, so probably not specific to this change.

Issue #147, opened 2015-01-26
Enhance diags-collection script.


Lukasa:
Credit for these ideas to Nick Bartos on the mailing list:

- Often, using grep without line anchors is not what you want.  For example: `whoami | grep -q "root"` will succeed with a username who is `i_am_groot`.  Also since you're forcing bash above, you can just do `[ "$LOGNAME" != "root" ]`.  `$LOGNAME` works in most shells, I believe.  It's better form to not fork & exec when you don't have to, in any language.
- Quoting variables in shell is a _really_ good idea.  For example if `NEUTRON_DIR="/var/log/neu tron"` then `cp -a $NEUTRON_DIR .` will try and copy /var/log/neu and tron (in the current working directory)
- The output of `iptables-save` will grab the info for all tables/chains, even ones that you didn't mention in your script which could potentially affect things.  For example we make use of the 'raw' table.  IMHO, the output of `iptables-save` is a lot easier to read than the output of `iptables -L`, as it shows the full options that created the rule.
- Grabbing `/var/log/*log` may not be all that you want.  Most non-debian based systems use `/var/log/messages` as the main log, not `/var/log/syslog`.
- When you create a gzipped tar archive, it should end in `.tar.gz`, not just `.gz`.
- If someone has a lot of logs and they end up trying to e-mail this file, it may exceed their mail size limit.  Therefore it might be nice to try and use xz compression first (maybe testing if xz is there and falling back to bzip2 if it's not).
- Also you may consider dumping the rule database (`ip rule list`).  We use this in some installs (although I'm not currently doing it in the calico install) and I bet others do as well.  This can affect routing.  You also might want to dump the routing table that each rule references.

We should look at implementing as much of this as is feasible.

plwhite:
We also need to document this script and install it on servers (which is why the original issue was still open).

plwhite:
Resolved in PR #170 (except for using bzip2, where I decided that the extra time to compress overwhelmed the reduction in size - though we should revisit further because our diags are far too large).

PR #170, opened 2015-02-09
Various minor diags fixups.

plwhite:
Largely issue #147 based, but also other minor tidies (including critically the `dpkg` change that triggered this - getting fed up not being able to work out what version we are running!).


lwr20:
LGTM

PR #174, opened 2015-02-09
Add PyCharm support files

fasaxc:
Chaps, are you interested in this change?  Our team typically use PyCharm and checking in the (hidden) config files gives us a consistent base to work from and lets us share run configs (allowing us to run the UTs from within the IDE, for example).  I checked the PyCharm docs to make sure I was checking in the files that were intended to be shared via version control.

I also made a tweak to setup.py to allow it to work with "setup.py test" (and "setup.py nosetests").


Lukasa:
ok to test

Lukasa:
No objections from me, but I want at least one other person to ok/not-ok. /cc @neiljerram @plwhite @lwr20 @DaveLangridge @matthewdupre

Lukasa:
retest this please

Lukasa:
retest this please

Lukasa:
retest this please

Lukasa:
retest this please

fasaxc:
Ugh, my change to setup.py seems to have exposed a bug in pip: https://github.com/pypa/pip/issues/2416

We could work around by swapping the install order in the Jenkins config so that nose gets installed first but that feels pretty unsatisfactory.  Shall I just remove that part of the pull req?

Lukasa:
I don't know that it's `pip`'s fault, actually. I wonder if nose's setup.py does something different if it's installed from test_requires.

For now, yank it out. =)

fasaxc:
I still think this a good idea but it seems like it's causing more trouble than its worth so I'm going to drop it for now and resubmit when my PyCharm config is nice and stable.

Issue #176, opened 2015-02-09
FelixAgent looks to access undefined self.uuid field


fasaxc:
Having a (PyCharm-assisted) code read of the Felix code, FelixAgent appears to access a non-existent field on one of its error paths:

            log.error("Missing addrs or IP in addrs for endpoint %s, data %s",
                      self.uuid, fields)

PR #179, opened 2015-02-10
Fix up typo in felix.py

plwhite:
No tests of this code - because it's not clear that testing this is higher priority than filling in other holes in our testing and coverage.

Fixes bug #176.

### DIFF
--- a/calico/felix/felix.py
+++ b/calico/felix/felix.py
@@ -547,21 +547,27 @@ class FelixAgent(object):
             for addr in fields['addrs']:
                 addresses.add(Address(addr))
         except KeyError:
-            log.error("Missing addrs or IP in addrs for endpoint %s, data %s",
-                      self.uuid, fields)
-            raise InvalidRequest("No valid address for endpoint %s")
+            log.error("Missing addr or IP in addrs for endpoint %s : %s",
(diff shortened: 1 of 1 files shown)


calico/felix/felix.py
```diff
@@ -547,21 +547,27 @@ def _update_endpoint(self, endpoint, fields):
...
+            log.error("No state for endpoint %s : %s",
+                      endpoint.uuid,
+                      fields)
+            raise InvalidRequest("No state for endpoint %s" % endpoint.uuid)
 
         if state not in Endpoint.STATES:
-            log.error("Invalid state %s for endpoint %s : %s" %
-                      (state, endpoint.uuid, state))
+            log.error("Invalid state %s for endpoint %s : %s",
+                      state, endpoint.uuid, fields)
```
Lukasa: Is this change right?

plwhite: Yes, I think so.

We are trying to report the invalid state, the endpoint it applied to and the entire request. Not guaranteeing that it's really not got some obvious dumb typo that I just cannot see, but it looks right.

Lukasa:
retest this please

Lukasa:
retest this please

Issue #180, opened 2015-02-10
ip route add can fail


plwhite:
In particular, it can fail if the route exists and is being imported from BIRD. We should use "ip route replace" instead.

plwhite:
Fixed in PR #181.

DaveLangridge:
Tested fix. Works fine.

PR #181, opened 2015-02-10
Use ip route replace instead of add

plwhite:
Fixes timing window when route exists during live migration


plwhite:
Neil, are you OK to review this? It's kind of obvious what I'm doing and why.

Fixes bug #180.

Lukasa:
retest this please

Lukasa:
retest this please

PR #183, opened 2015-02-10
Fix delay seen on initial pings

fasaxc:
Fixes https://github.com/Metaswitch/calico/issues/164.  Initial ping was delayed by the default value of the proxy_delay setting, which caused the ARP response from the host to be delayed.  Since the ARPs are point-to-point in our implementation, I can't see any reason that delay would be helpful.  (The manpage says the default value of "80 jiffys" prevents a flood in certain circumstances.)

Retested using calico-docker revision ad6ec3bdb23d08faf15050b3e3467a494208c8db with felix container replaced by the code from this branch.  Seeing initial pings in range of 1-2ms.

### DIFF
--- a/calico/felix/devices.py
+++ b/calico/felix/devices.py
@@ -89,6 +89,9 @@ def configure_tap(tap):
     with open("/proc/sys/net/ipv4/conf/%s/proxy_arp" % tap, 'wb') as f:
         f.write('1')
 
+    with open("/proc/sys/net/ipv4/neigh/%s/proxy_delay" % tap, 'wb') as f:
+        f.write('0')
+
 
 def add_route(type, ip, tap, mac):
     """
--- a/calico/felix/test/test_devices.py
+++ b/calico/felix/test/test_devices.py
@@ -21,7 +21,6 @@ Test the device handling code.
 import logging
 import mock
 import os
-import sys
 import unittest
 import uuid
 
@@ -31,6 +30,11 @@ import calico.felix.test.stub_utils as stub_utils
 
 # Logger
 log = logging.getLogger(__name__)
+
+# Canned mock calls representing clean entry to/exit from a context manager.
+M_ENTER = mock.call().__enter__()
+M_CLEAN_EXIT = mock.call().__exit__(None, None, None)
+
 
 class TestDevices(unittest.TestCase):
     def setUp(self):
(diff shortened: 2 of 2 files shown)


Lukasa:
ok to test

Lukasa:
This looks good to me, but would like @plwhite to verify as well.

Lukasa:
Ah, also, @fasaxc, can you fix up the history? You seem to have three merge commits and nothing else in the history. We generally like to keep the history as linear as possible,.

plwhite:
Looks good to me - happy to merge once the history is sorted (this should be a single simple commit).

Lukasa:
Uh, I actually feel nervous about this.

In particular, I think that this causes problems if a guest bridges together multiple internal interfaces *and* their external interface. We then run the risk of proxy ARPing for their internal interfaces and that will confuse them no end.

Is that a real concern, or am I being needlessly paranoid?

plwhite:
To me that feels a bit paranoid, but I'm no expert in this area; it might be something people do all the time that I'm just not aware of.

fasaxc:
@Lukasa ooh, hadn't thought of a virtual network inside a VM.  How do we configure the IP/subnet of the VM interface these days; what do we advertise over DHCP?  (Do we send them a //32 or, a subnet and a gareway, or a /0?)

I'm thinking, if we send them a subnet matching our IP pool then they shouldn't ARP for anything outside that and, later, we could drop any ARPs outside that range.  I don't have a VM-based system running here to experiment with and the Docker prototype doesnt' use DHCP so it can configure a default static connected route.

I'll clean up the history.  I got derailed by line endings and had to do an unusual stash merge to fix.

Lukasa:
We give them a /32 and a gateway.

Agreed that they shouldn't ARP for anything outside it, but that doesn't prevent them doing it, and AFAICT you can't configure proxy arp in that granularity in the kernel. It would be a bit sad to have to write a userspace proxy-arp daemon...

fasaxc:
What about arptables?

Lukasa:
It's a possibility. I don't know how it plays with proxy arp though.

Lukasa:
I suppose we could drop outgoing arps if we had to.

Ugh, feels like a really unpleasant solution. May be no other options though. =(

fasaxc:
Doesn't seem crazy to police what ARPs guests are allowed to send, especially if we only ever expect them to ARP the gateway.  Might be a security hole to allow guests to send spoofed ARP packets anyway; might allow them to poison the host's ARP cache, for example.

I've rebased the pull request.

Lukasa:
We should definitely have an ARP discussion at some point to really nail this down.

neiljerram:
Note that I don't think it requires active 'spoofing' for a guest to send an ARP request for something other than the gateway address. It only needs a non-default route in the routing table. 

For example imagine a tenant VM 10.65.0.2 providing some bump in the wire service, and another VM wanting to route via that for some traffic: ip route add <prefix> via 10.65.0.2. Then that VM will ARP for 10.65.0.2 instead of for 10.65.0.1.

neiljerram:
Just discussed with @Lukasa and we decided to merge this so that it makes today's release.  It's certainly helpful for what we imagine to be most Calico deployments, and the scenario where it may be problematic is a relatively esoteric one.  If that esoteric scenario (or any bad consequence) pans out in practice, we'll revisit this change.

Issue #184, opened 2015-02-11
Felix does not proactively withdraw routes.


Lukasa:
When Felix gets an ENDPOINTDESTROYED message it does not actively remove the routes from the kernel. Instead, it waits for the kernel to do it when the tap interface is destroyed.

If the tap interface never gets destroyed, this won't happen, and the route will continue to slosh about the calico network forever.

PR #189, opened 2015-02-12
Delete routes when endpoint destroyed

plwhite:
Fixes issue #184.

### DIFF
--- a/calico/felix/endpoint.py
+++ b/calico/felix/endpoint.py
@@ -92,8 +92,23 @@ class Endpoint(object):
         self.state          = Endpoint.STATE_DISABLED
 
     def remove(self):
-        # Delete a programmed endpoint. Remove the rules only, since the routes
-        # will vanish in due course when the tap interface goes.
+        """
+        Delete a programmed endpoint. Remove the routes, then the rules.
+        """
+        if devices.tap_exists(self.tap):
+            for type in (futils.IPV4, futils.IPV6):
+                try:
+                    ips = devices.list_tap_ips(type, self.tap)
+                    for ip in ips:
+                        devices.del_route(type, ip, self.tap)
+                except futils.FailedSystemCall:
+                    # There is a window where the tap interface gets deleted
+                    # under our feet. If the tap interface has gone now, ignore
+                    # the error, otherwise rethrow it.
+                    if devices.tap_exists(self.tap):
+                        raise
+                    break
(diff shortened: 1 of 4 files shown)


calico/felix/endpoint.py
```diff
@@ -92,8 +92,23 @@ def __init__(self, uuid, mac):
...
+        """
+        Delete a programmed endpoint. Remove the routes, then the rules.
+        """
+        for type in (futils.IPV4, futils.IPV6):
+            try:
+                if devices.tap_exists(self.tap):
+                    ips = devices.list_tap_ips(type, self.tap)
+                    for ip in ips:
+                        devices.del_route(type, ip, self.tap)
+            except:
```
Lukasa: Let's avoid the naked `except`.

Lukasa: This is also overbroad. We check if the tap exists in both branches, so we should be able to reduce the scope of this try...except block.

Lukasa: Additionally, I think if the tap existed when we started and doesn't exist when we're done we don't need to rethrow. This logic is definitely included in the current code and we want it, but I think we can still get it with just one code block.

Lukasa:
One note here. =)

plwhite:
What we have now is this.

    for type in (futils.IPV4, futils.IPV6):
        try:
            if devices.tap_exists(self.tap):
                ips = devices.list_tap_ips(type, self.tap)
                for ip in ips:
                    devices.del_route(type, ip, self.tap)
        except:
            if devices.tap_exists(self.tap):
                raise
            break

I think you are proposing this :

    if devices.tap_exists(self.tap):
        for type in (futils.IPV4, futils.IPV6):
            try:
                ips = devices.list_tap_ips(type, self.tap)
                for ip in ips:
                    devices.del_route(type, ip, self.tap)
            except (futils.WhateverTheExceptionIs):
                if devices.tap_exists(self.tap):
                    raise
                break


Is that right? No objection in principle to that restructuring.

Lukasa:
@plwhite Exactly right. I'm a big believer in small try...except blocks. =D

plwhite:
Back to you. Note that I did a `push -f` so that the intermediate version would not appear in the history, but small enough that should not matter much.

Lukasa:
Cool, LGTM. :cake:

PR #191, opened 2015-02-12
Neutron driver support for live migration

neiljerram:


Lukasa:
This in principle looks good to me. I'm going to sit on it until after we cut our new release to get as much time as possible to soak in our automated and manual testing.

neiljerram:
Cool, thanks for the review.

PR #193, opened 2015-02-13
[WIP] Remove python-iptables

Lukasa:
This pull request has primarily been opened for code review purposes. This is not to be merged yet.


calico/felix/fiptables.py
```diff
@@ -1,214 +1,451 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+felix.fiptables
+~~~~~~~~~~~~
+
+IP tables management functions. This is a wrapper round python-iptables that
```
Lukasa: Not any more it isn't!

calico/felix/fiptables.py
```diff
@@ -1,214 +1,451 @@
...
+class Chain(object):
+    """
+    Representation of an iptables chain.
+    """
+    def __init__(self, table, name):
+        self.type = table.type
+        self.table = table
+        self.name = name
+        self.rules = []
+        self.table.chains[name] = self
```
Lukasa: I've never been a fan of this pattern. I don't like having objects that are conceptually owned or contained within other objects add themselves to their parent (a table is made up of chains, so why do chains add themselves to tables?). Really, I'd rather this last step be an extra line in `Table.get_chain` if possible.

calico/felix/fiptables.py
```diff
@@ -1,214 +1,451 @@
...
+        self.table.ops.append([IPTABLES_CMD[self.type],
+                               "-w",
+                               "-t",
+                               self.table.name,
+                               "-F",
+                               self.name])
+        del self.rules[:]
+
+    def delete_rule(self, rule):
+        # The rule must exist or it is an error.
```
Lukasa: [nitpicking] In today's episode of "Cory is a stickler for grammar", the pronoun 'it' here seems to be incorrect. The rule isn't an error, surely?

calico/felix/fiptables.py
```diff
@@ -1,214 +1,451 @@
...
+        self.rules.remove(rule)
+        args = [IPTABLES_CMD[self.type],
+                "-w",
+                "-t",
+                self.table.name,
+                "-D",
+                self.name]
+        args.extend(rule.generate_fields())
+        self.table.ops.append(args)
+        log.debug("Removing rule : %s", args)
```
Lukasa: Can this log come higher up the function? I think it gets logged too late to be helpful in debugging in many cases.

calico/felix/fiptables.py
```diff
@@ -1,214 +1,451 @@
...
+
+        If force_position is True, then the rule is added at the specified point
+        unless it already exists there. If force_position is False, then the
+        rule is added only if it does not exist anywhere in the list of rules.
+
+        Position starts at offset 0, even though iptables wants it to start at
+        offset 1.
+        """
+        if position == RULE_POSN_LAST:
+            position = len(self.rules)
```
Lukasa: This logic plays poorly with the iptables lock that we wait for. It's possible (albeit unlikely) that the table changes under our feet and the rule no longer goes in at the end. Ideally we'd have this logic switch from 'insert' to 'append'.

calico/felix/fiptables.py
```diff
@@ -1,214 +1,451 @@
...
+                "-w",
+                "-t",
+                self.table.name,
+                "-I",
+                self.name,
+                str(position + 1)]
+        args.extend(rule.generate_fields())
+        self.table.ops.append(args)
+        self.rules.insert(position, rule)
+        log.debug("Creating rule : %s", args)
```
Lukasa: Similarly, this log is a bit late.

calico/felix/fiptables.py
```diff
@@ -1,214 +1,451 @@
...
+
+    def __str__(self):
+        output = "  Chain %s\n" % self.name
```
Lukasa: Why is there a space here?

calico/felix/fiptables.py
```diff
@@ -1,214 +1,451 @@
...
+        for rule in self.rules:
+            output += "    %s\n" % rule
+
```
Lukasa: Building a string like this is really very slow. Consider instead doing this:

```python
'\n'.join(['Chain %s' % self.name].extend('  %s' % rule for rule in self.rules)
```

(I haven't run that code, so it may need a small tweak, but the general logic is sound.

calico/felix/fiptables.py
```diff
@@ -1,214 +1,451 @@
...
+    Representation of an iptables table.
+    """
+    def __init__(self, type, name):
+        self.type = type
+        self.name = name
+        self.chains = {}
+        self.ops = []
+
+    def is_chain(self, name):
+        return (name in self.chains)
```
Lukasa: No need for the parentheses here.

calico/felix/fiptables.py
```diff
@@ -1,214 +1,451 @@
...
+                      self.type)
+            chain = Chain(self, name)
+            self.ops.append([IPTABLES_CMD[self.type],
+                              "-w",
+                              "-t",
+                              self.name,
+                              "-N",
+                              name])
+
+        return chain
```
Lukasa: This feedback is take-it-or-leave-it, but arguably it would be better style (and slightly more performant) to do a try...except KeyError block here. This would fit the 'Easier to Ask Forgiveness than Permission' style commin in Python code. I don't care very much, but thought I would note it.

calico/felix/fiptables.py
```diff
@@ -1,214 +1,451 @@
...
+                        "-X",
+                        chain_name])
+        del self.chains[chain_name]
+
+    def __str__(self):
+        output = "TABLE %s (%s)\n" % (self.name, self.type)
+        for chain_name in sorted(self.chains.keys()):
+            output += str(self.chains[chain_name])
+            output += "\n"
+        return output
```
Lukasa: Similar note about the expense of string building here. Also the expense of sorting: is it really worth it?

calico/felix/fiptables.py
```diff
@@ -1,214 +1,451 @@
...
+        parse_fields.
+        """
+        fields = []
+
+        for field in Rule.FIELD_TO_FLAG:
+            value = getattr(self, field)
+            if value is not None and value[0] == "!":
+                fields.extend(["!", Rule.FIELD_TO_FLAG[field], value[1:]])
+            elif value is not None:
+                fields.extend([Rule.FIELD_TO_FLAG[field], value])
```
Lukasa: Again, I think I'd rather use try...except than getattr here.

Lukasa:
So far looks pretty good. It's a big diff so hard to review effectively, but what I can see looks totally reasonable.

You've got some notes from me and some of your own TODOs which are all good ideas, so I hope to see those changes coming in. =)

calico/felix/fiptables.py
fasaxc: Also looks like there's a reference cycle here, which is best avoided in Python since it negates Python's reference counting and relies on Python's (relatively weak, stop-the-world) garbage collector.  Should Chain hold a weak ref to Table?

calico/felix/fiptables.py
```diff
@@ -1,214 +1,451 @@
...
+
+        return chain
+
+    def delete_chain(self, chain_name):
+        log.debug("Delete chain %s from table %s (%s)",
+                  chain_name,
+                  self.name,
+                  self.type)
+        self.ops.append([IPTABLES_CMD[self.type],
+                        "-w",
```
fasaxc: Minor niggle from having to maintain code like this in the past, can we use long-form '--foo-bar' args or comment what the args are?  Avoids an awkward trip to the manpage and trying to grep for "-w" or "-t"

calico/felix/fiptables.py
Lukasa: Probably a good idea. This is a fairly trivial reference cycle (three objects) and none of the objects involved have finalizers so in practice I don't think CPython will care much, but a weakref certainly can't hurt. Docs are [here](https://docs.python.org/2/library/weakref.html).

fasaxc: Yep, not saying we'll leak, just that we move from fast and precise ref counting to GC, which I think is best avoided in Python.

calico/felix/fiptables.py
plwhite: I think you are incorrectly assuming that the pronoun "it" here applies to the rule. When I say "it is cold today", I am not referring to an object, merely making a point about the current state.

However, I'll change "it" to "there" and we can all be happy.

calico/felix/fiptables.py
plwhite: The logic behind the space is that when you turn a table into a string, each chain is indented by the right number of characters for it all to look right.

However, I can leave it more naturally formatted and add the spaces when we are building up the table, which I think is actually better.

calico/felix/fiptables.py
plwhite: I am OK with the weak reference thing - but I am nervous about changing the code so that the adding of the table link is done externally. The problem is that the Chain object can be created either because we have decided to explicitly create it (it did not exist before) or because we have just read it. Those code paths are actually in different places, and having identical code in two places does not smell right, to put it mildly. I actually fixed a bug in an earlier iteration that did it this way.

How strongly do you feel about this?

plwhite: In the end I just did as you suggested - decided I didn't really care either way.

calico/felix/fiptables.py
plwhite: It is slow, but on the other hand there are not many chains or rules. I ended up cutting dramatically the number of such string concatenations, but not all (because readability got too low). We use `join` for all the big lists, but there are a few concatenations as well.

calico/felix/fiptables.py
plwhite: Sorting is useful because we want the output to be deterministic (test outputs tend to rely on it being consistent within a run), and the same as what you get from iptables itself. I would expect the sorting effort (sorting on some short strings) to be very minor, especially for code that isn't called in the mainline.

Agreed that cutting down the concatenations is worthwhile; done that.

calico/felix/fiptables.py
plwhite: There are no long format arguments for iptables. I'll add a comment at the top of the file of what arguments we use.

calico/felix/fiptables.py
plwhite: We have to wait for the lock because somebody (such as the diags script) might be reading the data under our feet. If somebody is modifying any of our chains under our feet, we are pretty much sunk.

The only case we are left with is where we are appending a rule to a chain that we do not own, where it only really matters that it gets in there, not where, so I think I agree that we might as well use append in that case.

calico/felix/fiptables.py
fasaxc: Maybe we're talking about different versions of iptables, but on Ubuntu 14.04:

$ man iptables | grep -Ee '-.,.+--'
       -t, --table table
       -A, --append chain rule-specification
       -C, --check chain rule-specification
       -D, --delete chain rule-specification
       -D, --delete chain rulenum
       -I, --insert chain [rulenum] rule-specification
       -R, --replace chain rulenum rule-specification
       -L, --list [chain]
       -S, --list-rules [chain]
       -F, --flush [chain]
       -Z, --zero [chain [rulenum]]
       -N, --new-chain chain
       -X, --delete-chain [chain]
       -P, --policy chain target
       -E, --rename-chain old-chain new-chain
       -4, --ipv4
       -6, --ipv6
       [!] -p, --protocol protocol
       [!] -s, --source address[/mask][,...]
       [!] -d, --destination address[/mask][,...]
       -m, --match match
       -j, --jump target
       -g, --goto chain
       [!] -i, --in-interface name
       [!] -o, --out-interface name
       [!] -f, --fragment
       -c, --set-counters packets bytes
       -v, --verbose
       -w, --wait
       -n, --numeric
       -x, --exact

plwhite: You're right. I was being dim (I checked the man page, but clearly incompetently since it says what you say, not what I remembered).

I'll raise an issue and fix it up.

PR #194, opened 2015-02-13
Fix #182: Change Windows-style line endings to Unix-style

### LINKED ISSUE #182: Calico repo uses mixed line endings
Some are Windows CRLF, some are Unix-style LF-only.  Given that Calico is a native Linux project, I suggest we standardise on Unix-style and Windows users use the recommended "check out Windows-style, commit Unix-style" option.

Here's github's take on it: https://help.github.com/articles/dealing-with-line-endings/  Looks like we can check in a .gitattributes file that will configure git correctly for whatever we decide.

neiljerram:


fasaxc:
:+1:

PR #195, opened 2015-02-13
Send ENDPOINTDESTROYED rsp even when endpoint is unknown (fixes #192)

### LINKED ISSUE #192: No Felix response for ENDPOINTDESTROYED for absent endpoint
If Felix gets an ENDPOINTDESTROYED request for an endpoint that it doesn't recognize, it doesn't send any response back.

The code in handle_endpointdestroyed just logs and returns, in that case.  It should be more similar to what handle_endpointupdated does in the absent endpoint case.

neiljerram:


calico/felix/felix.py
```diff
@@ -392,26 +390,35 @@ def handle_endpointdestroyed(self, message, sock):
...
             endpoint = self.endpoints.pop(delete_id)
+
+            # Unsubscribe from ACL information for this endpoint.
+            self.sockets[Socket.TYPE_ACL_SUB].unsubscribe(
+                delete_id.encode('utf-8')
+            )
+
+            # Remove programming for this endpoint.
+            endpoint.remove()
+
```
Lukasa: This change leads to the `except` block being overbroad. In particular, we'll hit the except branch if `self.sockets` does not contain a `Socket.TYPE_ACL_SUB` socket entry.

You should consider using Python's try..except..else` syntax here, making this block:

```python
        try:
            # Remove this endpoint from Felix's list of managed
            # endpoints.
            endpoint = self.endpoints.pop(delete_id)
        except KeyError:
            log.error("Received destroy for absent endpoint %s", delete_id)

            fields = {
                "rc": RC_NOTEXIST,
                "message": "Endpoint %s does not exist" % delete_id,
            }
        else:
            # Unsubscribe from ACL information for this endpoint.
            self.sockets[Socket.TYPE_ACL_SUB].unsubscribe(
                delete_id.encode('utf-8')
            )

            # Remove programming for this endpoint.
            endpoint.remove()
```

Lukasa:
One note.

Lukasa:
LGTM. :cake:

Issue #61, opened 2014-12-09
Felix assumes endpoint interface names are OpenStack format


plwhite:
Felix assumes that each endpoint has an interface name which starts with `tap` and then the first 11 characters of the endpoint ID. That assumption is correct for OpenStack, and good enough for the Docker prototypes, but it would not be good enough for production systems where the interface name is more likely to start with `veth` (since it is not a tap interface), and might not follow those naming conventions.

The right solution is to : 

* add a new optional parameter "interface" on the `ENDPOINTCREATED` and `ENDPOINTUPDATED` messages;
* change Felix to use that field, defaulting if not present to current behaviour.

plwhite:
Worth saying that while having an odd hard-coded name for the interface is a bit weird outside OpenStack, if we do not set up the interface ourselves (which might well be the case in at least some integrations), then Calico fundamentally won't work without this fix.

plwhite:
Fixed PR #201

Issue #175, opened 2015-02-09
error.value unresolved according to PyCharm


fasaxc:
Loading up the felix code in PyCharm, I see a bunch of uses of <some exception>.value highlighted as unresolved by the semantic checker, are those supposed to be error.args[0] or am I missing something?

        except InvalidRequest as error:
            # Invalid request fields. Return an error.
            fields = {
                "rc": RC_INVALID,
                "message": error.value,
            }

plwhite:
Fixed PR #201

PR #201, opened 2015-02-18
Make interface name configurable

plwhite:
Fixes issues #61 and #175.

As a result of changing some firewall code, I had to work on some previously untested code - and so this checkin enormously improves test coverage of invalid requests and responses in Felix, as well as tidying up that code.


matthewdupre:
ok to test

matthewdupre:
Two minor comments on the config file inline and one wider comment about specification of interface name format below.  The code itself looks fine.

Did you consider making the plugin entirely responsible for specifying interface naming?

Pros:
- Interface naming is orchestrator specific, and orchestrator specific code belongs in the plugin.
- No need for additional config options (since the correct settings will just be a lookup based on the plugin anyway).
- API is simplified by making the interface field compulsory.

Cons:
- We also use the prefix on its own for creating the intermediate chains.  This could also be passed on the API, but that would be a fairly unpleasant field; or refactored out of the rules, but that has implications of its own that I haven't fully thought through.
- Adds code to the plugins, but it's very simple.  They'd (other than whichever was lucky enough to be the default) all need equivalent instructions anyway.

We should talk about this and decide whether to make such a change.

plwhite:
More changes.
- Interface prefix is specified on RESYNCSTATE response now.
- We make interface_name mandatory on the API.
- Get the suffix by taking the interface name and removing the prefix from it.
Back for review.

calico/felix/config.py
```diff
@@ -217,4 +218,3 @@ def validate_addr(self, name, addr):
             raise ConfigException("Invalid or unresolvable %s value : %s" %
                                   (name, addr),
                                   self._config_path)
```
matthewdupre: Please remove this file from the change - we don't want the double blank line or to remove the blank line at end of file.

calico/felix/felix.py
```diff
@@ -612,7 +686,14 @@ def run(self):
         endpoint_resync_needed = False
         acl_resync_needed = False
 
-        active_sockets = fsocket.poll(self.sockets.values(),
+        if self.iface_prefix:
+            poll_list = self.sockets.values()
+        else:
+            # Note got an first resync response (as no interface prefix), so
```
matthewdupre: Note -> not.

calico/felix/test/data/felix_veth.cfg
```diff
@@ -0,0 +1,12 @@
+[global]
+FelixHostname = test_hostname
+# Plugin and ACL manager addresses
+PluginAddress = 127.0.0.1
+ACLAddress    = 127.0.0.1
+InterfacePrefix = veth
+
```
matthewdupre: InterfacePrefix isn't needed any more.

matthewdupre:
I've made a few small comments above - also please update the API document and fix the UT failures above.

plwhite:
Retest this

Lukasa:
@neiljerram This looks like your tests are failing: care to take a look?

neiljerram:
The UT code needs to be updated to expect the new interface_name key/value pair.

Lukasa:
Sadly this pull request now requires rebasing. =(

Issue #206, opened 2015-02-19
Felix has an undocumented upper bound on the number of rules in a security group.


Lukasa:
When Felix creates an ipset it has a maximum size of 65536 rules. If that size is exceeded further elements do not get added to the ipset. This leads to transparent failure to correctly program security group configuration. This is exacerbated by #205.

fasaxc:
Reading the AWS docs the other day, I think they limit security groups to 50 or 100 rules.  (Of course, that's not much help if one rule is expanded to 65000 ipset entries ;) )

Lukasa:
Agreed. I don't really care what we do here, but if we have an upper bound it certainly shouldn't be *undocumented*. =)

plwhite:
I think this issue goes away with PR #307, so closing it down. If we find that there are still problems of this nature, we'll open another issue.

Issue #205, opened 2015-02-19
Saying 'all UDP traffic' leads to terrible ipsets.


Lukasa:
If you add a security group rule that says 'any UDP port' or 'any TCP port' in OpenStack, that translates into a port range of 1-65535. That causes Felix to create an ipset with 65535 entries. In addition to being a gratuitous waste of memory (1.5 MB for those entries alone) it causes Felix to stop inserting new rules, as per issue #206.

plwhite:
Duplicate of issue #197

PR #208, opened 2015-02-19
[DO NOT MERGE] Increase set size to 6553500 elements

Lukasa:
This represents one very stupid 'fix' to #206 and ameliorates #205. We **absolutely should not merge this**, because it increases the resting size of an ipset, and doesn't deal with most of the core problems. This PR is open only to provide a clue to a temporary patch for anyone who encounters these problems.


plwhite:
I think this PR can be closed down - we are never going to do this, and it's not worth keeping the PR open. @Lukasa do you agree? It might be that I'm missing a customer short term need for it.

PR #209, opened 2015-02-19
Remove python 2.7 only features so we work on python 2.6 again.

matthewdupre:
Red Hat / CentOS 6.5 only come with python 2.6, so we need to rip out all python 2.7 features to support them.

### DIFF
--- a/calico/felix/fiptables.py
+++ b/calico/felix/fiptables.py
@@ -267,7 +267,7 @@ class Rule(object):
 
     FLAG_TO_FIELD = dict(tuples)
     KNOWN_FIELDS = [ a[1] for a in tuples ]
-    FIELD_TO_FLAG = {field: flag for flag, field in tuples}
+    FIELD_TO_FLAG = dict((field, flag) for flag, field in tuples)
 
     def __init__(self, type, target=None):
         self.type = type
--- a/calico/felix/test/test_endpoint.py
+++ b/calico/felix/test/test_endpoint.py
@@ -22,6 +22,7 @@ import mock
 import sys
 import unittest
 import uuid
+from contextlib import nested
 
 import calico.felix.devices as devices
 import calico.felix.endpoint as endpoint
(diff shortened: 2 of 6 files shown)


calico/felix/fiptables.py
```diff
@@ -267,7 +267,7 @@ class Rule(object):
 
     FLAG_TO_FIELD = dict(tuples)
     KNOWN_FIELDS = [ a[1] for a in tuples ]
-    FIELD_TO_FLAG = {field: flag for flag, field in tuples}
+    FIELD_TO_FLAG = dict((field: flag) for flag, field in tuples)
```
Lukasa: This won't work.

calico/felix/test/test_endpoint.py
```diff
@@ -53,18 +53,21 @@ def test_remove_deleted(self):
...
+                with mock.patch(
+                    'calico.felix.devices.del_route',
+                    side_effect=futils.FailedSystemCall("blah",
+                                                        ["dummy", "args"],
+                                                        1,
+                                                        "",
+                                                        "")) as mock_del_route:
+                    with mock.patch(
+                        'calico.felix.frules.del_rules') as mock_del_rules:
+                        ep.remove(iptables_state)
```
Lukasa: MY EYES!

That said, I don't know what you could do about it.

fasaxc: `contextlib.nested`?

Lukasa: Oh yeah! I knew that was somewhere, but I could only see contextlib.ExitStack in recent docs so I got tunnel vision.

plwhite: Is this a python 2.6 oddity that prevents you just having multiple "with" statements (which works for 2.7 so long as you use enough backslashes on the end of lines)? If so, then presumably there are a host of other places where this problem occurs, since I use that logic all the time in tests.

fasaxc: Yep, the multiple-with statement was adding in 2.7:  `with foo as far, bax as biff:` isn't valid in 2.6.  `with nested(foo, bax) as (far, biff):` is roughly equivalent in 2.6 (it's not quite as good due to some problematic corner cases that I'm not well versed on).

PR #211, opened 2015-02-21
Plugin connectivity improvements

neiljerram:
This PR also adds unit testing of the Neutron driver that raises its code coverage to 57%.

### DIFF
--- a/calico/openstack/mech_calico.py
+++ b/calico/openstack/mech_calico.py
@@ -38,6 +38,7 @@ import json
 import time
 from neutron import context as ctx
 from neutron import manager
+from zmq.error import Again
 
 LOG = log.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


Lukasa:
LGTM. :cake:

neiljerram:
Could I merge this now, or are there considerations against doing that?

Lukasa:
It's not clear to me that I should be code reviewing this. =) Recommend that @plwhite takes a look too.

neiljerram:
Thanks @plwhite !

calico/openstack/mech_calico.py
```diff
@@ -607,6 +621,11 @@ def felix_heartbeat_thread(self, hostname):
                 else:
                     LOG.error("Unexpected response from Felix on %s: %s"
                               % (hostname, rsp))
+            except Again:
+                LOG.error("No response from Felix within allowed time (%sms)" %
+                          HEARTBEAT_RESPONSE_TIMEOUT)
```
plwhite: Shouldn't you throw a FelixUnavailable exception here, as in the comment above?

neiljerram: No, because there is no caller who'd be interested in that - we're on a dedicated heartbeat thread here.

plwhite: The comment not so many lines above says that you do raise the exception, and I think should be tidied up (presumably a bad block copy). Then I'm happy to merge.

neiljerram: I think the comment is meaning that the poll call will raise an exception - but I'm happy to clarify that, if that will address your concern here.

neiljerram: Sorry, I mean the recv_json call.

Issue #168, opened 2015-02-09
Felix is failing to recognise certain IP v6 addresses


plwhite:
Here's a log showing the issue.

    2015-02-04 18:02:45,465 [WARNING] calico.felix.devices 72: No IP address found in line fe80::/64  proto kernel  metric 256  for tap47568f4fc56

So far benign, but we should tidy up the scary log (which I think just means that we assume that all IP v6 addresses must have more non-zero fields than that).

Lukasa:
I think this is the point where I re-iterate some code review feedback I gave a while back: we should parse IP addresses properly, acknowledging that they're structured data. Regular expressions are a bad tool for this. =)

fasaxc:
The netaddr library is the best I know of for parsing all flavors IP addresses in Python.

plwhite:
I have a fix for this, but blocked behind review of PR #201

plwhite:
Fixed in PR #219

PR #218, opened 2015-02-24
Improve handling of invalid IP addresses

plwhite:
Fixes bug #168


plwhite:
Ditching because of merge history problems.

PR #219, opened 2015-02-24
Improve handling of invalid IP addresses

plwhite:
Fixes bug #168

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,8 +1,9 @@
 [run]
 include =
-    *felix*
-    *acl_manager*
-    *openstack*
+    calico/felix/*
+    calico/acl_manager/*
+    calico/openstack/*
+    calico/common.py
 omit =
     calico/felix/test/*
     calico/acl_manager/test/*
--- a/calico/common.py
+++ b/calico/common.py
@@ -23,16 +23,11 @@ Calico common utilities.
 """
 import logging
 import logging.handlers
+import netaddr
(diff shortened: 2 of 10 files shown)


matthewdupre:
Code looks OK to me - couple of small comments:

- we only ever call 'validate_ipv4... or validate_ipv6...', so there's little reason to have both implemented separately (especially when they only differ by one digit in implementation).  A single validate address function would serve nicely (it could take the same optional parameter as netaddr.IPAddress), or even just call into netaddr directly.
- netaddr doesn't seem to be installed by default on windows (it is on all the Linux boxes I could find).  @Lukasa any idea why this is?

Lukasa:
@matthewdupre `netaddr` is a third-party Python library. This means it's a third-party dependency for us, and our dependency resolution stuff needs to be updated appropriately. That means updating `felix_requirements.txt` to mention `netaddr`, and then to update the debian and RPM packaging logic to add the dependency for `calico-felix` (probably called python-netaddr).

matthewdupre:
So my take away from that is that there isn't really such a thing as a 'base python install', because such things come with various selections of non-standard libraries.

Pity.

Lukasa:
Nope. There is absolutely a base python install.

However, netaddr is sufficiently popular as to ensure that it'll almost certainly be installed on any Linux system very nearly straight away. For example, apt-get quite probably uses it.

Issue #207, opened 2015-02-19
Felix incorrectly assumes that ipsets treat port 0 as 'all ports'.


Lukasa:
Per the comment on line 623 of `frules.py`, Felix believes that port 0 means 'all ports' in ipsets. Based on my testing this is untrue, and I can find no support for this view in the man page.

plwhite:
Turns out that you can create such rules quite successfully, but they don't work, at least not on the version I was running with. The correct way to specify them is to give them as port range 1 to 65535, which then hits issue #197.

Lukasa:
Resolved in #223.

PR #223, opened 2015-02-25
Do not use port 0 to mean all ports

plwhite:
Actually, this is valid and required for some protocols, but not tcp, sctp,
udp or udplite where the only way to specify a port of "any" is to give an
explicit port range.

Fixes bug #207.


calico/common.py
```diff
@@ -67,6 +67,19 @@ def validate_ip_addr(addr, version):
...
 
 
+def validate_cidr(cidr, version):
+    """
+    Validates that a CIDR is valid. Returns true if valid, false if
+    not. Version can be "4", "6", None for "IPv4", "IPv6", or "either"
+    respectively.
+    """
+    try:
+        ip = netaddr.IPNetwork(cidr, version=version)
```
matthewdupre: When we come back to this code it would probably be nicer to just pass the `IPNetwork`s (and `IPAddress`es) around and convert them to strings immediately before programming.

calico/felix/frules.py
```diff
@@ -574,12 +574,18 @@ def update_ipsets(type,
...
     for rule in rule_list:
         if rule.get('cidr') is None:
-            log.error("Invalid %s rule without cidr for %s : %s" %
-                      (descr, suffix, rule))
+            log.error("Invalid %s rule without cidr for %s : %s",
+                      descr, suffix, rule)
+            continue
+
+        if (type == IPV4 and not common.validate_cidr(rule['cidr'], 4) or
+            type == IPV6 and not common.validate_cidr(rule['cidr'], 6)    ):
```
Lukasa: Might be wise to bracket each of these conditional sections. Suspect most people don't know operator precedence order well enough to be confident of what will happen here.

calico/common.py
plwhite: Agreed - don't propose doing that now, but not too far away.

calico/felix/frules.py
plwhite: OK - will do that.

plwhite:
@matthewdupre, are you OK to merge this now? No huge hurry to be honest, just keen to get it off the list.

PR #228, opened 2015-03-02
Use default metadata port 8775.

Lukasa:
This switches the default metadata port from 9697 (which is for neutron's metadata proxy) to 8775 (which is for nova-api, the service that actually provides metadata in Calico).

### DIFF
--- a/calico/felix/config.py
+++ b/calico/felix/config.py
@@ -72,7 +72,7 @@ class Config(object):
                                                   "127.0.0.1")
         self.METADATA_PORT   = self.get_cfg_entry("global",
                                                   "MetadataPort",
-                                                  "9697")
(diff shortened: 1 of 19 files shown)


Lukasa:
Some docs pages have been updated: [the Calico Configuration page](https://github.com/Metaswitch/calico-docs/wiki/Calico-Configuration) had the default value of metadata port changed to 8775, the [Red Hat Packaged Install](https://github.com/Metaswitch/calico-docs/wiki/Red-Hat-Packaged-Install) page had the change of the Metadata IP setting removed, as did the [Ubuntu Packaged Install](https://github.com/Metaswitch/calico-docs/wiki/Ubuntu-Packaged-Install) page.

PR #230, opened 2015-03-02
Felix programs proxy NDP

spikecurtis:
Felix enables proxy NDP on endpoint interfaces and programs the addresses of the gateways configured.

- Modifies endpoint.Address to track the gateway for each `addr` passed on the API.
- Address is now backed by netaddr.IPAddress, rather than strings for the IP and gateway.
- Refactor code in felix.py that updates Endpoint objects to be a method on the Endpoint.
- Improved UT of endpoint updates.

### DIFF
--- a/calico/felix/endpoint.py
+++ b/calico/felix/endpoint.py
@@ -24,6 +24,8 @@ from calico import common
 from calico.felix import devices
 from calico.felix import frules
 from calico.felix import futils
+from calico.felix.exceptions import InvalidRequest
+from netaddr import IPAddress
 import logging
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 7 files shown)


Lukasa:
ok to test

calico/felix/endpoint.py
```diff
@@ -36,26 +38,49 @@ class Address(object):
...
+                    "IP %s is not the same IP version as gateway %s" %
+                    (self.ip, self.gateway))
+
+
+class InconsistentIPVersion(Exception):
+    """
+    Tried to create an Address with inconsistent IP versions between
+    properties, e.g. IP and gateway.
+    """
+    pass
```
Lukasa: Can you move this into the exceptions module?

Lukasa: Oof, nevermind, no such module exists. I'd like to have one with `InvalidAddress` in it, but if you are opposed to the idea I don't consider it blocking for this PR.

calico/felix/endpoint.py
```diff
@@ -179,13 +204,19 @@ def remove_routes(routes, type):
...
-                                 if addr.type is futils.IPV4])
-            ipv6_intended = set([addr.ip.encode('ascii')
-                                 for addr in self.addresses
-                                 if addr.type is futils.IPV6])
+            ipv4_intended = set([str(addr.ip)
+                                 for addr in self.addresses.values()
+                                 if addr.version is 4])
+            ipv6_intended = set([str(addr.ip)
+                                 for addr in self.addresses.values()
+                                 if addr.version is futils.IPV6])
```
Lukasa: You might be able to save yourself some overhead and typing by using

```python
[ip for ip, addr in self.addresses.items() if addr.version == 4]
```

Lukasa: Separately, why are we using the literal `4` here?

calico/felix/endpoint.py
Lukasa: Oh hey, you added such a module!

calico/felix/endpoint.py
```diff
@@ -270,3 +301,65 @@ def update_acls(self):
...
+
+        :param fields: dictionary of Endpoint Update API fields.
+        :return: None.
+        :throws: InvalidRequest if unsuccessful.
+        """
+
+        try:
+            mac = fields['mac']
+        except KeyError:
+            raise InvalidRequest("Missing \"mac\" field", fields)
```
Lukasa: Probably just as easy to use single-quotes rather than escape the doubles.

Lukasa:
Some minor notes, but otherwise I'm happy. =)

Lukasa:
ok to build

matthewdupre:
retest this please

calico/felix/endpoint.py
spikecurtis: We're using the literal 4 because that's the netaddr IPAddress API.

spikecurtis:
retest this please

matthewdupre:
retest this please

Lukasa:
The ocelot unit test failed because this PR needs rebasing on master. I'll just do it so that @spikecurtis doesn't hate us. =)

Lukasa:
Ah, actually I can't, because it's in @spikecurtis's repository which I don't have permissions to. For now, we'll take the pass on Bobcat UT as evidence of success.

Lukasa:
Regardless, this looks entirely plausible: has anyone done some live test of this change?

spikecurtis:
I'm doing live testing now.  Hit one bug which is fixed in latest commit.

Lukasa:
Thanks @spikecurtis! :cake: :star:

Issue #237, opened 2015-03-04
Felix cyclically restarts when previous state is in iptables6


spikecurtis:
I'm seeing an issue where old iptables6 state appears to be causing felix to cyclically restart.  In the traceback below `tape25102c4c2c` is not an active endpoint from the EP API.  It must be finding out about the (now nonexistent) interface via state leftover in iptables6 or ipsets from before a hard restart of felix.  

```
2015-03-04 22:57:46,451 [ERROR] calico.felix.felix 807: Felix exiting after uncaught exception
Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/calico/felix/felix.py", line 798, in main
    agent.run()
  File "/usr/lib/python2.7/dist-packages/calico/felix/felix.py", line 725, in run
    self.complete_endpoint_resync(False)
  File "/usr/lib/python2.7/dist-packages/calico/felix/felix.py", line 263, in complete_endpoint_resync
    frules.del_rules(self.iptables_state, found_suffix, type)
  File "/usr/lib/python2.7/dist-packages/calico/felix/frules.py", line 496, in del_rules
    iptables_state.apply()
  File "/usr/lib/python2.7/dist-packages/calico/felix/fiptables.py", line 440, in apply
    futils.multi_call(ops)
  File "/usr/lib/python2.7/dist-packages/calico/felix/futils.py", line 112, in multi_call
    check_call(["bash", name])
  File "/usr/lib/python2.7/dist-packages/calico/felix/futils.py", line 90, in check_call
    raise FailedSystemCall("Failed system call", args, retcode, stdout, stderr)
FailedSystemCall: Failed system call (retcode : 1, args : ('bash', '/tmp/tmp3PgfsM'))
  stdout  : Executing : ip6tables --wait --table filter --delete felix-TO-ENDPOINT -o tape25102c4c2c -j felix-to-e25102c4c2c

  stderr  : ip6tables: No chain/target/match by that name.

```

spikecurtis:
Full logs from one felix cycle are at http://pastebin.com/pdaSkwnF

Lukasa:
Felix seems to get all in a tizzy here.

It's going to remove state for a VM that it no longer owns, which includes some stuff in iptables. When it decides to remove it, it emits the following logs:

```
2015-03-04 22:57:46,975 [WARNING] calico.felix.felix 262: Removing IPv6 rules for removed object e25102c4c2c
[...]

PR #239, opened 2015-03-05
Improve a few logs to include IPv4/6 type

plwhite:


plwhite:
Haven't diagnosed bug #237, so at least fixed up some logs.

Lukasa:
@plwhite I've been seeing that test failure intermittently for a little while now, might be good to fix it if we can.

Lukasa:
Nevermind, that's an ACL manager test.

Lukasa:
At this stage I'm torn, because this is an obvious merge with no real cost, but I'm nervous about the intermittent test failure cascading over to the master branch.

Lukasa:
Ah screw it, when it fails there I can beat @matthewdupre up until he fixes it. =)

PR #240, opened 2015-03-05
Handle flaky ACL tests by overriding assertDictEqual.

Lukasa:
This deals with the intermittent ACL manager unit test case failure that I've seen over the last few days by forcing unordered comparison of lists nested inside more complex dictionary structures.

### DIFF
--- a/calico/acl_manager/test/test_processor.py
+++ b/calico/acl_manager/test/test_processor.py
@@ -60,6 +60,7 @@ class TestProcessor(unittest.TestCase):
         be ordered for dict comparison. It's an unfortunate hack, but c'est la
         vie.
         """
+        self.assertEqual(len(d1), len(d2))
         for k,v1 in d1.iteritems():
             self.assertIn(k, d2, msg)
             v2 = d2[k]
@@ -68,7 +69,6 @@ class TestProcessor(unittest.TestCase):
                 self.assertItemsEqual(v1, v2, msg)
             else:
                 self.assertEqual(v1, v2, msg)
-        return True
 
     def test_case1(self):
         """
--- a/calico/acl_manager/test/test_processor.py
+++ b/calico/acl_manager/test/test_processor.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import collections
 import sys
 from copy import deepcopy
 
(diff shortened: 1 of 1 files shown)


calico/acl_manager/test/test_processor.py
```diff
@@ -53,6 +54,22 @@ def tearDown(self):
                          'outbound': [],
                          'outbound_default': 'deny'}}
 
+    def assertDictEqual(self, d1, d2, msg=None):
+        """
+        This override for assertDictEqual ensures that lists do not need to
+        be ordered for dict comparison. It's an unfortunate hack, but c'est la
+        vie.
+        """
+        for k,v1 in d1.iteritems():
```
matthewdupre: This only tests if d1 is a subdictionary of d2.  Please add the reverse comparison (nicest to just loop this?).

calico/acl_manager/test/test_processor.py
```diff
@@ -53,6 +54,22 @@ def tearDown(self):
...
+            self.assertIn(k, d2, msg)
+            v2 = d2[k]
+            if (isinstance(v1, collections.Iterable) and
+                not isinstance(v1, basestring)):
+                self.assertItemsEqual(v1, v2, msg)
+            else:
+                self.assertEqual(v1, v2, msg)
+        return True
```
matthewdupre: assertDictEqual shouldn't really return anything - it's probably harmless enough, but may as well remove so it matches the standard assertDictEqual.

Lukasa:
@matthewdupre Done. =)

Lukasa:
retest this please

Stupid Jenkins

Lukasa:
retest this please

Lukasa:
@plwhite These failures *are* Felix failures, which I have also seen intermittently fail. It'd be nice if you could take a look.

Lukasa:
@plwhite Sadly, the problem seems to be an ordering one:

```
@@ -53,7 +53,7 @@

     -j felix-PREROUTING
   
   Chain felix-PREROUTING
-    -d 169.254.169.254/32 -p tcp -m tcp -j DNAT --dport 80 --to-destination 127.0.0.1:8775
+    -d 169.254.169.254/32 -p tcp -m tcp -j DNAT --to-destination 127.0.0.1:8775 --dport 80
   
 TABLE filter (IPv4)
   Chain FORWARD
@@ -110,6 +110,6 @@

     -j felix-PREROUTING
   
   Chain felix-PREROUTING
-    -d 169.254.169.254/32 -p tcp -m tcp -j DNAT --dport 80 --to-destination 127.0.0.1:8775
+    -d 169.254.169.254/32 -p tcp -m tcp -j DNAT --to-destination 127.0.0.1:8775 --dport 80
```

Lukasa:
These failures are inevitable because `Rule.parameters` is a `dict`, and therefore has no defined ordering.

Lukasa:
@plwhite I'm in a bit of a tough place here. It'd be nice to use `collections.OrderedDict`, but frankly the ordering is really only relevant to the tests, and otherwise we don't care. For that reason it feels stupid to bring in ordered dict and pay the price for maintaining that ordering in all cases.

It really just feels like that test is ill-considered: it cares about ordering in a situation where we really don't.

PR #241, opened 2015-03-05
Fix unit tests on Python 2.6

Lukasa:
These failures were introduced by #230.


Lukasa:
Ok, the remaining Ocelot failure is the one that should be fixed by #240. @matthewdupre, if you're convinced feel free to merge (or not!).

PR #243, opened 2015-03-05
Convert Rule parameters to a list of tuples.

### LINKED ISSUE #242: Felix stores iptables parameters in a dictionary, which it shouldn't.
The `fiptables.Rule` class stores rule parameters in a dictionary. This is problematic for a number of reasons:

1. It makes testing tricky.
2. It prevents us having multiple instances of a parameter with the same name (most notably `-m` which brings in modules and you could easily want more than one of).
3. It loses ordering, which *can* matter (in cases like the above!).
4. It leads to subtle bugs, like you can see in the following code's comment, which clearly expects a behaviour that is not present:

    ```python
        def create_target(self, name, parameters=None):
            self.target = name
            # Note that this adds parameters, rather than replacing existing ones.
            for key in parameters:
                self.parameters[key] = parameters[key]
    ```

All those reasons would be acceptable or work-around-able if we were using the number one reason to use a dictionary, which is O(1) lookup. But we aren't. We never do direct elementwise lookup, we only ever do iteration over the whole collection.

This property could easily be turned into a list of tuples. This would fix all the above problems immediately at no extra cost.

Lukasa:
Resolves #242.

### DIFF
--- a/calico/felix/fiptables.py
+++ b/calico/felix/fiptables.py
@@ -262,7 +262,7 @@ class Rule(object):
         self.match = None
 
         # We bundle rule and match parameters together for easier coding.
-        self.parameters = {}
+        self.parameters = []
 
     def parse_fields(self, line, fields):
         """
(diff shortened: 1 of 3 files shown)


Lukasa:
This is blocked behind the merge of #241.

Lukasa:
Now unblocked.

Lukasa:
@plwhite Ok, we need a solution to #241 by midday tomorrow (to unblock a release). Feel free to use this one, or one of your own.

Lukasa:
@plwhite is not here today, so he lost his veto. @matthewdupre, you're up.

matthewdupre:
Looks fine.

PR #247, opened 2015-03-06
Update diags script

matthewdupre:


debian/rules
```diff
@@ -15,4 +15,5 @@ override_dh_install:
 	install etc/bird/*.template debian/tmp/usr/share/calico/bird
 	install -d debian/tmp/usr/bin
 	install -m 755 etc/*.sh debian/tmp/usr/bin
+    install -m 755 utils/diags.sh debian/tmp/usr/bin/calico-diags
```
Lukasa: Indentation is tricky.

utils/diags.sh
```diff
@@ -7,15 +7,17 @@ then
...
 NEUTRON_DIR=/var/log/neutron
+NOVA_DIR=/var/log/nova
 date=`date +"%F_%H-%M-%S"`
-diags_dir=/tmp/"$date"
+diags_dir=./"$date"
 system=`hostname`
-echo "  created dir $diags_dir"
+echo "  creating dir $diags_dir"
 mkdir "$diags_dir"
 pushd "$diags_dir" > /dev/null
```
Lukasa: I think we care greatly if the mkdir or pushd commands fail: recommend bailing out in those cases.

PR #251, opened 2015-03-09
Refactor Neutron driver in preparation for etcd transport implementation

neiljerram:
This pull request refactors the Neutron driver in preparation for implementing the new etcd-based transport.  To do that, it:
- creates an internal interface to separate the OpenStack logic from the transport (0MQ or etcd)
- separates out the parts of the new internal architecture into their own files.


neiljerram:
Would anyone like to review, once (and assuming that) the automated builds have passed?

calico/openstack/t_etcd.py
```diff
@@ -0,0 +1,49 @@
...
+#    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#    License for the specific language governing permissions and limitations
+#    under the License.
+
+# Etcd-based transport for the Calico/OpenStack Plugin.
+
+# Calico imports.
+from calico.openstack.transport import CalicoTransport
+
+LOG = None
```
Lukasa: Any reason to change the naming convention here?

Lukasa:
As always, thousand-line diffs are tricky to review, but the general shape of this looks good to me.

PR #256, opened 2015-03-10
Updates to documentation

Lukasa:
This pull requests adds some new documentation and cleans up some old stuff, in the pursuit of having a slightly better documentation story.

### DIFF
--- a/docs/source/configuring-calico.rst
+++ b/docs/source/configuring-calico.rst
@@ -1,0 +1,12 @@
+Configuring Calico
+=====================
+
+Contents:
+
+.. toctree::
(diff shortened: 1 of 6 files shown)


Lukasa:
@neiljerram Initial markups made, let me know what you think.

neiljerram:
Agree about not adding a security example - I wasn't sure so thought I'd just raise it.
Did we not agree to leave IPv6 where it was? - I think that change is still in the set.
Otherwise just a couple textual comments.

Lukasa:
Made your textual comments, we'll discuss IPv6 today. =)

PR #257, opened 2015-03-10
Don't finish mechanism driver / ACL Manager audit until the initial GROUPUPDATEs have been sent

matthewdupre:
ACL Manager needs to be able to detect audit failures and retry, which it can't do if the driver returns the GETGROUPS response immediately.

This is an fix to https://github.com/Metaswitch/calico/issues/255


Lukasa:
We're going to not merge this until Nick gives it a spin, because if it's insufficient to fix his problem it might not be worthwhile.

matthewdupre:
Looks like we can merge this now.

Lukasa:
This had to be merged by hand because it played poorly with @neiljerram's refactor for transport-agnostic code.

PR #260, opened 2015-03-10
Use config where appropriate, and use suffix of interface for chains.

plwhite:
This has had virtually no testing (though I have got it through "hello world"); you can merge it if you like, or just see how I think we should handle the chain issue (that I did to get familiar with the way the code works).

### DIFF
--- a/calico/felix/fiptables.py
+++ b/calico/felix/fiptables.py
@@ -36,16 +36,17 @@ _log = logging.getLogger(__name__)
 
 class DispatchChains(Actor):
     """
-    Actor that owns the felix-TO/FROM-ENDPOINT chains, which we use to 
+    Actor that owns the felix-TO/FROM-ENDPOINT chains, which we use to
     dispatch to endpoint-specific chains.
 
     LocalEndpoint Actors give us kicks as they come and go so we can
     add/remove them from the chains.
     """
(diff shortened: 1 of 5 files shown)


calico/felix/fiptables.py
```diff
@@ -57,25 +59,26 @@ def on_endpoint_chains_ready(self, iface_name, endpoint_id):
         :param iface_name: name of the linux interface.
         :param endpoint_id: ID of the endpoint, used to form the chain names.
         """
-        if self.iface_to_ep_id_suffix.get(iface_name) != endpoint_id:
-            self.iface_to_ep_id_suffix[iface_name] = endpoint_id
+        if self.iface_to_ep_id.get(iface_name) != endpoint_id:
```
fasaxc: I think this hasn't been renamed in the __init__.  If the chain names are going to be based on the iface name only then maybe we just need a set of those rather than a map to ep id?

plwhite: Yes - good spot. It was that kind of thing that meant that I didn't want to just commit it! I just fixed that.

It's quite possible that we don't need the map. I wasn't all that sure what it was actually used for, though I was wondering if we would want to know about it, if only because of logging the endpoint id, so I decided to just leave it in for now.

PR #264, opened 2015-03-12
Fix broken link to architecture doc

ed-harrison:
The link to the architecture doc is broken (it is pointing to a now deleted file).


ed-harrison:
Note - I've not actually worked out how to test that this change will work on ReadTheDocs.

ed-harrison:
I've also updated the Installation instructions as these seemed to be a bit broken.

PR #284, opened 2015-03-19
Bring forward changes from calico-docs.wiki

Lukasa:
This should bring forward all changes made to the Calico docs wiki since the last time we synchronized them.


matthewdupre:
Sorry Neil - you can't use service for upstart jobs on EL6, and you can't get the respawn behavior from a sysV init script.

neiljerram:
Ah, OK.  initctl is just fine then.

PR #285, opened 2015-03-19
Implementation and UT for etcd transport for plugin (aka Neutron driver)

neiljerram:
There's still some more UT to do here, but I think it's worth getting review eyes on this and possibly banking it in the master branch.

Note that:
- this PR still leaves 0MQ as the default transport - because that's what we still want in master
- enhanced UT covers both the 0MQ and etcd transports, as well as the common part of the plugin code
- the UT passes.

### DIFF
--- a/calico/openstack/t_etcd.py
+++ b/calico/openstack/t_etcd.py
@@ -225,7 +225,7 @@ class CalicoTransportEtcd(CalicoTransport):
 
                 # Map the protocol field from Neutron to etcd format.
                 if rule['protocol'] is None:
-                    raise UnsupportedError('protocol None')
+                    pass
                 elif rule['protocol'] == 'icmp':
                     etcd_rule['protocol'] = {'IPv4': 'icmp',
                                              'IPv6': 'icmpv6'}[ethertype]
--- a/calico/openstack/t_etcd.py
+++ b/calico/openstack/t_etcd.py
@@ -29,6 +29,8 @@ from calico.openstack.transport import CalicoTransport
 
 LOG = None
 
+PERIODIC_RESYNC_INTERVAL_SECS = 30
+
 
 class CalicoTransportEtcd(CalicoTransport):
     """Calico transport implementation based on etcd."""
--- a/calico/openstack/t_etcd.py
+++ b/calico/openstack/t_etcd.py
@@ -81,42 +81,32 @@ class CalicoTransportEtcd(CalicoTransport):
                 # We have a key/value pair for an OpenStack endpoint.  Extract
                 # the endpoint ID and hostname from the key, and read the JSON
                 # data as a dict.
-                id = m.group("endpoint_id")
+                endpoint_id = m.group("endpoint_id")
                 hostname = m.group("hostname")
                 data = json_decoder.decode(child.value)
 
-                if id in ports:
-                    # OpenStack still has an endpoint with this ID.
-                    new_data = self.port_etcd_value(ports[id])
-
-                    if ports[id]['binding:host_id'] == hostname:
-                        # The endpoint's current host matches the etcd key.
-                        if data == new_data:
-                            # The endpoint data is unchanged.  No change
-                            # needed.
-                            pass
-                        else:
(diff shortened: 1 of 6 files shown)


neiljerram:
@Lukasa, I've speculatively assigned this to you for review - is that OK?

Lukasa:
Fine by me, though I don't think I'll get to it until tomorrow.

calico/openstack/t_etcd.py
```diff
@@ -17,10 +17,25 @@
 
 # Etcd-based transport for the Calico/OpenStack Plugin.
 
+# Standard Python library imports.
```
Lukasa: This comment is weird, probably because it's really close to saying "Python standard library imports", which many of these aren't.

calico/openstack/t_etcd.py
```diff
@@ -17,10 +17,25 @@
...
 # Calico imports.
 from calico.openstack.transport import CalicoTransport
 
 LOG = None
+OPENSTACK_ENDPOINT_RE = re.compile(
+    r'^/calico/host/(?P<hostname>[^/]+)/.*openstack.*/endpoint/(?P<endpoint_id>[^/]+)')
+OPENSTACK_POLICY_RE = re.compile(
+    r'^/calico/policy/profile/(?P<profile_id>[^/]+)/tags')
+
+json_decoder = json.JSONDecoder()
```
Lukasa: Why is this a module-scoped object?

calico/openstack/t_etcd.py
neiljerram: Can you recommend something more appropriate?

calico/openstack/t_etcd.py
neiljerram: No good reason, I think.  What would you recommend instead?

calico/openstack/t_etcd.py
alexwlchan: Imports are supposed to be split into three groups &ndash; standard library, related third-party, and local application &ndash; with a newline between each. Since the `etcd` and `eventlet` libraries aren't part of the standard library, they should be split into their own import group. More like:

```python
# Standard library imports
import json
import re
import time

# Related third-party imports
import etcd
import eventlet

# Calico imports
```

There's more detail about how import statements should be formatted [in PEP 8](https://www.python.org/dev/peps/pep-0008/#imports).

calico/openstack/t_etcd.py
```diff
@@ -34,16 +49,283 @@ def __init__(self, driver, logger):
...
+
+    def port_profile_id(self, port):
+        return '_'.join(port['security_groups'])
+
+    def resync_security_groups(self):
+        # Get all current security groups from the OpenStack database and key
+        # them on security group ID.
+        self.sgs = {}
+        for sg in self.driver.get_security_groups():
+            self.sgs[sg['id']] = sg
```
Lukasa: Is this a good idea? Storing the full set of security groups on the box in memory is likely to be a scale bottleneck.

calico/openstack/t_etcd.py
```diff
@@ -34,16 +49,283 @@ def __init__(self, driver, logger):
...
+                    etcd_rule['dst_tag'] = rule['remote_group_id']
+                    etcd_rule['dst_net'] = net
+                    etcd_rule['dst_ports'] = port_spec
+                    outbound.append(etcd_rule)
+                    LOG.info("=> Outbound Calico rule %s" % etcd_rule)
+
+        return {'inbound_rules': inbound, 'outbound_rules': outbound}
+
+    def profile_tags(self, profile_id):
+        return profile_id.split('_')
```
Lukasa: Methods like this (that don't use `self` anywhere in their body should probably be moved to being functions instead.

calico/openstack/t_etcd.py
Lukasa: Instance scoped (`self.json_decoder`).

Lukasa:
Closing this, as the work has been merged into #307.

PR #288, opened 2015-03-20
Fix incorrect line break in faq.rst

ed-harrison:
I think the intent was there should be a line break after the link to the "Why Calico" blog, but this wasn't coming through.  I thought it might be better to refer to the blog at the end of the answer, anyway, so have moved the reference.


lwr20:
Makes sense.  LGTM.

PR #291, opened 2015-03-23
Add list of third-party attributions

alexwlchan:
This is a straight copy of the file README.txt from the calico-docs repo, copied here so it doesn't get lost.when the calico-docs repo is retired.

Related: https://github.com/Metaswitch/calico-docs/issues/3

### DIFF
--- a/ATTRIBUTIONS.txt
+++ b/ATTRIBUTIONS.txt
@@ -1,0 +1,62 @@
+A	Third Party Software Attributions
+
+The following third party components are included with Project Calico, and are made 
+available by third parties under separate license terms.  Applicable license terms and 
+information regarding the authors and copyright holders are provided along with the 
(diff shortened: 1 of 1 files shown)


Lukasa:
@neiljerram and I were quite unsure that we should reproduce that notice in this repository, given that this repository does not distribute or contain any code from any of those projects in it. My alternative is in #292.

That said, someone who isn't me or neil should decide which of these approaches to take. @ed-harrison seems like a good start!

Lukasa:
ok to test

ed-harrison:
I'm happy to agree with Cory's plan (sorry Alex!).  I'll merge Cory's suggestion and then just run it by the legal department to confirm they are happy.

alexwlchan:
Seems sensible. I'll clean up after myself.

PR #292, opened 2015-03-23
Add third party licenses to docs.

Lukasa:
Rather than leave them in the repository, given that it has nothing to do with the projects mentioned here, move these license notices to the docs instead.


ed-harrison:
LGTM

PR #293, opened 2015-03-24
Neutron mechanism driver: always get Neutron DB handle before using it in theoretical start-of-day window

matthewdupre:
add_port_gateways() requires self.db to have been set up, but potentially _get_db() hasn't been called yet.

This window was observed using a dummy Neutron for testing - it could only be hit in a live deployment if create_port_postcommit() can be called before any Felixes have connected - I'm not sure if this is possible.


Lukasa:
`create_port_postcommit` can absolutely be called before any Felices have connected. :+1:

PR #295, opened 2015-03-26
Initial etcd documentation.

Lukasa:
This is a first pass at the etcd documentation. Feedback would be appreciated!


docs/source/etcd-architecture.rst
```diff
@@ -0,0 +1,239 @@
+Calico etcd-based architecture
+==============================
+
+The current version of Calico is built around `etcd`_. etcd is a distributed,
+consistent key value store for shared configuration and service discovery with
+a focus on being simple, secure, fast, and reliable.
+
+In Calico, etcd is used as the data store and communication mechanism for all
+the Calico components. This data store contains all the information the various
+Calico components to set up the Calico network.
```
alexwlchan: This sentence seems to be missing a word. Possibly "all the information the [...] components *need* to set up the Calico network"?

docs/source/etcd-architecture.rst
```diff
@@ -0,0 +1,239 @@
...
+
+.. _calico-felix-component:
+
+Felix
+-----
+
+Felix is the most important component in the Calico network: without it, no
+network programming can be achieved. It is a daemon that runs on every machine
+that provides endpoints: that means on nodes that host containers or VMs in
+most cases.
```
alexwlchan: I would rearrange the final clause to put "in most cases" first. I think it makes the sentence easier to read.

docs/source/etcd-architecture.rst
```diff
@@ -0,0 +1,239 @@
...
+
+Depending on the specific orchestrator environment, Felix is responsible for
+some or all of the following tasks:
+
+Interface Creation and Management
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+In some environments, particularly containerized ones, Felix is responsible
+for creating and managing network interfaces to its endpoints. This usually
+involves creating and managing 'veth pairs': a pair of virtual ethernet
```
alexwlchan: Ethernet should be capitalised.

docs/source/etcd-architecture.rst
```diff
@@ -0,0 +1,239 @@
...
+
+Felix is responsible for programming routes to the endpoints on its host into
+the Linux kernel FIB. This ensures that packets destined for those endpoints
+that arrive on at the host are forwarded accordingly.
+
+ACL Programming
+~~~~~~~~~~~~~~~
+
+Felix is also responsible for programming ACLs into the Linux kernel. These
+ACLs are used to ensure that only valid traffic can be send between
```
alexwlchan: "can be sen*t*"

docs/source/etcd-architecture.rst
```diff
@@ -0,0 +1,239 @@
...
+order to render a single complex operation (e.g. live migration) into the
+series of simpler operations the rest of the Calico network expects.
+
+Feedback
+~~~~~~~~
+
+If necessary, the orchestrator plugin will provide feedback from the Calico
+network into the orchestrator. Examples include: providing information about
+Felix liveness; marking certain endpoints as failed if network setup failed;
+and so on.
```
alexwlchan: Drop the "and so on"; it's redundant.

docs/source/etcd-architecture.rst
```diff
@@ -0,0 +1,239 @@
...
+into the network.
+
+
+.. _calico-bgp-component:
+
+BGP Client (BIRD)
+-----------------
+
+Calico deploys a BGP client on every node that also hosts a
+:ref:`calico-felix-component`. The role of the BGP client is read routing state
```
alexwlchan: I think this should be "The role [...] is *to* read routing state"

docs/source/etcd-architecture.rst
```diff
@@ -0,0 +1,239 @@
...
+become tricky to maintain.
+
+For that reason, in larger deployments Calico will deploy a BGP route
+reflector. This component, commonly used in the Internet, acts as a central
+point to which the BGP clients connect, preventing them from needing to talk to
+every single BGP client in the cluster.
+
+For redundancy, multiple BGP route reflectors can be deployed seamlessly.
+
+In calico, this BGP component is also most commonly `BIRD`_, configured as a
```
alexwlchan: Calico should be capitalised.

docs/source/etcd-data-model.rst
```diff
@@ -0,0 +1,243 @@
...
+
+``hostname``
+  the hostname of the compute server
+
+``orchestrator_id``
+  the name of the orchestrator that owns the endpoint, e.g. ``"docker"`` or
+  ``"openstack"``.
+
+``workload_id``
+  a identifier provided by the orchestrator to relate multiple endpoints that
```
alexwlchan: Should be "an identifier"

alexwlchan:
I've made a few comments inline. Mostly typos and the like.

More generally, for small changes like that, is the preferred method:

1. Making inline comments, as above, or;
2. Making a new commit to the branch that fixes those issues.

I don't mind which, but the latter feels a bit more intrusive.

Lukasa:
Thanks @alexwlchan! I've pushed markups. =)

As to your question, I generally prefer the approach you've taken here. It's definitely a bit aggressive to just push up changes, especially as it can be hard to tell the line between a stylistic decision and an error. Often better to let the original committer make the decision, unless for some reason they can't.

docs/source/api-proposal.rst
```diff
@@ -1,1000 +0,0 @@
-Calico API Design
```
neiljerram: Should we not leave this in place until we actually switch to etcd?

Perhaps retitle it as something like 0MQ API, and add a link upfront to the etcd doc, with a note that we're in the process of transitioning.

Lukasa: Not sure: @ed-harrison?

docs/source/etcd-architecture.rst
```diff
@@ -0,0 +1,238 @@
+Calico etcd-based architecture
+==============================
+
+The current version of Calico is built around `etcd`_. etcd is a distributed,
+consistent key value store for shared configuration and service discovery with
+a focus on being simple, secure, fast, and reliable.
+
+In Calico, etcd is used as the data store and communication mechanism for all
```
neiljerram: "Calico uses etcd as ...."

docs/source/etcd-architecture.rst
```diff
@@ -0,0 +1,238 @@
+Calico etcd-based architecture
```
neiljerram: I'm not sure, but I think this might be topsy-turvy.  The architecture, surely, is a design for components that fulfil particular roles, and the use of etcd is an internal detail.  So with that view I'd rather title this "Calico Architecture"...

Except maybe not, if we're really making a big deal of the idea that 3rd party components can orchestrate by writing directly into etcd....  In that view I guess a key property of the architecture would be having etcd as an external interface.

I'm not sure - but I think perhaps a bit more thought is needed on what we want the message to be here.  (And to take care that we're not writing in terms that reflect history (i.e. 0MQ) that will soon be forgotten.)

docs/source/etcd-architecture.rst
```diff
@@ -0,0 +1,238 @@
...
+----------
+
+Calico is made up of the following interdependent components:
+
+- :ref:`calico-felix-component`, the primary Calico agent that runs on each
+  machine that hosts endpoints.
+- :ref:`calico-orchestrator-plugin`, orchestrator-specific code that tightly
+  integrates Calico into that orchestrator.
+- :ref:`calico-etcd-component`, running unmodified and used by the other
+  components as a data store.
```
neiljerram: "running unmodified" feels like unwarranted detail here.  I'd prefer just "the data store".

docs/source/etcd-architecture.rst
```diff
@@ -0,0 +1,238 @@
...
+
+.. _calico-felix-component:
+
+Felix
+-----
+
+Felix is the most important component in the Calico network: without it, no
+network programming can be achieved. It is a daemon that runs on every machine
+that provides endpoints: in most cases that means on nodes that host containers
+or VMs.
```
neiljerram: This sounds silly to me, a whole para that basically says nothing.  Can you instead say a summary of what it does, like "programs routes and iptables, and other things in the host's Linux kernel, so as to provide the desired connectivity for endpoints on that host".

docs/source/api-proposal.rst
ed-harrison: Yes, I think we should leave the 0MQ API description in place until we switch over.  Assuming we decide to actually publish the etcd stuff now (i.e. publish what Cory's done this week rather than leave it until we're ready to transition) then I agree with Neil that having a pointer in the 0MQ API doc to the new datastore description would be sensible.

docs/source/etcd-architecture.rst
Lukasa: I think that's extreme. The second sentence provides meaningful information. I propose to keep that but add the summary.

docs/source/etcd-architecture.rst
```diff
@@ -0,0 +1,238 @@
...
+some or all of the following tasks:
+
+Interface Creation and Management
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+In some environments, particularly containerized ones, Felix is responsible
+for creating and managing network interfaces to its endpoints. This usually
+involves creating and managing 'veth pairs': a pair of virtual Ethernet
+interfaces that behave as though they're connected together with a single
+Ethernet cable.
```
neiljerram: This isn't currently true.  Instead, the calico-docker orchestrator does this, and I'm not aware of any plan to change that.

neiljerram: On the other hand, you could say here that it monitors for when the relevant interfaces appear and become programmable.

docs/source/etcd-architecture.rst
```diff
@@ -0,0 +1,238 @@
...
+endpoints, and ensure that endpoints are not capable of circumventing
+Calico's security measures. For more on this, see :doc:`security-model`.
+
+State Reporting
+~~~~~~~~~~~~~~~
+
+Felix is responsible for providing data about the health of the network. In
+particular, it reports errors and problems with configuring its host. This data
+is written into etcd, to make it visible to other components and operators of
+the network.
```
neiljerram: Really?  I don't think any of this is true yet, so would delete this subsection.

Lukasa: Very little of this document is true yet. I'd argue if we ship without this functionality we're being extremely reckless, so it doesn't feel aggressive to have this section in place.

docs/source/etcd-architecture.rst
```diff
@@ -0,0 +1,238 @@
...
+separate plugins for each major cloud orchestration platform (e.g. OpenStack,
+Kubernetes). The purpose of these plugins is to bind Calico more tightly into
+the orchestrator, allowing users to manage the Calico network just as they'd
+manage network tools that were built in to the orchestrator.
+
+A good example of an orchestrator plugin is the Calico Neutron ML2 mechanism
+driver. This component integrates with Neutron's ML2 plugin, and allows users
+to configure the Calico network by making Neutron API calls. This provides
+seamless integration with Neutron, and lowers the barrier of entry to using
+Calico.
```
neiljerram: Delete "and lowers .... Calico".  The fact that we've written Calico lowers the barrier ...  The fact that we've written this doc lowers the barrier ...  In other words, we could say that after almost every sentence in our documentation - but it doesn't really add anything.

docs/source/etcd-architecture.rst
```diff
@@ -0,0 +1,238 @@
...
+network into the orchestrator. Examples include: providing information about
+Felix liveness; marking certain endpoints as failed if network setup failed.
+
+
+.. _calico-etcd-component:
+
+etcd
+----
+
+Calico uses etcd as its backing data store. etcd is a distributed key-value
```
neiljerram: Delete backing.

docs/source/etcd-architecture.rst
ed-harrison: Yes, we ought to decide on the overall messaging approach.
If we look at this document in six month's time, it will just describe the Calico Architecture.  The fact that we moved from 0MQ to etcd will be lost int he mists of time.  Consequently, I think I'd rather see this just called "Calico Architecture".
However, in introducing the "new" architecture, we'll probably want to make it clear that we're moving to use etcd.  That may mean that we:

-  use the blog and mailing list to advertise our "upcoming architecture changes" with a heavy reference to etcd
-  at that point have this document sit in the "Future" section of the docs with a clear "Coming soon" banner and have it referenced from the current 0MQ architecture and API docs
-  once we actually release the etcd code into master, we move these new docs to replace the current 0MQ ones.

Alternatively, we could leave these docs in a branch (still, potentially, blog their existence) and then just merge them all into master when the code is ready.

I'm interested in opinions here, but we'll probably want to make a proposal to (and get sign off from) the CCDT.

docs/source/etcd-architecture.rst
```diff
@@ -0,0 +1,238 @@
...
+store that has a focus on consistency, ensuring that Calico can always build an
```
neiljerram: I'm not sure what the real point is here - but I suspect too technical for the current context.  As a DB-mostly-ignoramus, I'd interpret "consistent" as "gives me back the same value as I just wrote into it", and hence my reaction to seeing that written here is a bit like seeing a car described as "able to drive along a road".

There may well be a real more substantive point here - but if so it needs to be brought out.

Lukasa: Mmm, it's a tricky one. 'Consistent' in this context is a term-of-art, and the sentence "gives me back the same value as I just wrote into it" is too vague to correctly correspond to what 'consistent' means.

Basically, I was trying to straddle the line between saying nothing about what etcd is and dropping into a sermon on the CAP theorem and consensus protocols.

I'm not sure what I should be shooting for here, so some guidance would be appreciated.

docs/source/etcd-architecture.rst
```diff
@@ -0,0 +1,238 @@
...
+The etcd component is distributed across the entire deployment. It is divided
+into two groups of machines: the core cluster, and the proxies.
+
+In Calico, we deploy an etcd cluster. In small deployments this can be an etcd
+cluster of one, which provides no redundancy but is simple and low-overhead.
+In larger deployments we scale this up, as per the `etcd admin guide`_.
+
+Additionally, on each machine that hosts either a
+:ref:`calico-felix-component` or a :ref:`calico-orchestrator-plugin`, we run
+an etcd proxy. This is an attempt to reduce load on the core cluster and to
```
neiljerram: s/is an attempt to reduce/reduces/

docs/source/etcd-architecture.rst
```diff
@@ -0,0 +1,238 @@
...
+ensure that nodes are shielded from the specifics of the etcd cluster.
```
neiljerram: s/to ensure/ensures/

neiljerram: actually, even nicer, I think: "and shields nodes from the specifics ..."

docs/source/etcd-architecture.rst
```diff
@@ -0,0 +1,238 @@
...
+components to read from the database (which is their most common operation), as
+they can distribute their reads around the cluster.
+
+Communication
+~~~~~~~~~~~~~
+
+etcd is also used as a communication bus between components through the
+mechanism of having various components watch keys for changes. This allows the
+act of committing state to the database to cause that state to be programmed
+into the network.
```
neiljerram: You've said this already, so delete this subsection.

Lukasa: Wait, where did I say it? Because it'd be better said here in a section that is easy to find, than said somewhere else scattered in the prose.

docs/source/etcd-architecture.rst
```diff
@@ -0,0 +1,238 @@
...
+
+.. _calico-bgp-rr-component:
+
+BGP Route Reflector (BIRD)
+--------------------------
+
+For larger deployments, simple BGP can become a limiting factor because it
+requires every BGP client to be connected to every other BGP client in a mesh
+topology. This requires an ever increasing number of connections that rapidly
+become tricky to maintain.
```
neiljerram: Well, so does scaling in general...  I think you need to be more precise here, something about an N^2 configuration requirement and N^2 live BGP peering sessions.

docs/source/etcd-architecture.rst
```diff
@@ -0,0 +1,238 @@
...
+route reflector rather than as a standard BGP client.
+
+The BGP route reflector is responsible for the following tasks:
+
+Centralised Route Distribution
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+When :ref:`calico-bgp-component` advertises routes from its FIB to the route
+reflector, the route reflector advertises those routes out to the other nodes
+in the deployment.
```
neiljerram: Possibly worth mentioning here - even though it's a well known RR property - that it does this in a way that does not put itself into the routing data path.

docs/source/etcd-data-model.rst
```diff
@@ -0,0 +1,243 @@
...
+
+In Calico, etcd is used as the data store and communication mechanism for all
+the Calico components. This data store contains all the information the various
+Calico components to set up the Calico network.
+
+This document discusses the way Calico stores its data in etcd. This data store
+and structure acts as Calico's primary external and internal API, granting
+developers exceptional control over what Calico does. This document does not
+discuss the way this data gets turned into actual network topologies by
+individual components: for more on that, see :doc:`etcd-architecture`.
```
neiljerram: I'm not sure `etcd-architecture` discusses that either....  Perhaps a better wording for what you mean would be "This document does not describe the components that read and write this data so as to provide the connectivity that endpoints in a Calico network want: for more on that ...."

docs/source/etcd-data-model.rst
```diff
@@ -0,0 +1,243 @@
...
+
+.. _etcd: https://github.com/coreos/etcd
+
+Objects
+-------
+
+Calico focuses on the following major object types, stored in etcd:
+
```
neiljerram: Really good to have this all here.  When this is merged, please delete (or obsolete/archive?) the similar google doc, to guard against anyone making further changes to that version.

docs/source/etcd-architecture.rst
neiljerram: Line 141

docs/source/etcd-architecture.rst
neiljerram: Agreed, and sorry. Your proposal sounds good to me.

docs/source/etcd-architecture.rst
```diff
@@ -0,0 +1,238 @@
...
+
+The current version of Calico is built around `etcd`_. etcd is a distributed,
+consistent key value store for shared configuration and service discovery with
+a focus on being simple, secure, fast, and reliable.
+
+In Calico, etcd is used as the data store and communication mechanism for all
+the Calico components. This data store contains all the information the various
+Calico components need to set up the Calico network.
+
+This document discusses the various pieces of the Calico etcd-based
```
ed-harrison: Given the positioning as "Calico Architecture", I think you can potentially remove the paragraph above this one and just start with "This document discusses..."

docs/source/etcd-architecture.rst
ed-harrison: Neil raises an interesting point about where the Calico Docker Orchstrator fits in the architecture.  I guess in the context of this doc, the LWO actually acts as the plug-in and orchestrator together?

docs/source/etcd-architecture.rst
ed-harrison: Agreed - providing Peter/Shaun agree that this is going to be in the code at the point we merge into master (or, if not, we move the "Coming soon" banner from the top of this article to this section).

docs/source/etcd-architecture.rst
ed-harrison: Agreed...  I'd also rather not acknowledge there is a barrier to entry for using Calico at all (lowered or not).

docs/source/api-proposal.rst
Lukasa: Do we plan to do that?

docs/source/etcd-architecture.rst
Lukasa: Mentioned it in the section above.

Lukasa:
Ok, addressed a substantial quantity of @neiljerram and @ed-harrison's feedback.

docs/source/etcd-architecture.rst
neiljerram: Are you thinking just of the Felix liveness reporting here?  If so, I'd be happy with describing that more specifically.  I'm not aware of any plan for more general 'state reporting' than that - whereas we do have code nearly ready to go for everything else described in this doc.

docs/source/etcd-architecture.rst
```diff
@@ -0,0 +1,238 @@
...
+
+etcd
+----
+
+Calico uses etcd as its backing data store. etcd is a distributed key-value
+store that has a focus on consistency, ensuring that Calico can always build an
+accurate network.
+
+In addition to its role as Calico's primary data store, etcd also acts as a
+communication mechanism between the various components. We do this by having
```
ed-harrison: Replace "We do" with "Calico does"...

docs/source/etcd-architecture.rst
Lukasa: I was also thinking failure detection (e.g. "I couldn't add this route because reasons").

neiljerram: That would indeed be very nice to have!

neiljerram: (Assuming it doesn't just say "because reasons" :-) )

Lukasa: Heh, agreed. ;) @fasaxc @plwhite are we expecting to add this for MVP?

docs/source/etcd-architecture.rst
```diff
@@ -0,0 +1,238 @@
...
+the non-etcd components watch certain points in the keyspace to ensure that
+they see any changes that have been made, allowing them to respond to those
+changes in a timely manner.
+
+The etcd component is distributed across the entire deployment. It is divided
+into two groups of machines: the core cluster, and the proxies.
+
+In Calico, we deploy an etcd cluster. In small deployments this can be an etcd
```
ed-harrison: Delete the first sentence and then say something like "For small Calico deployments, the core cluster can be an etcd cluster of one node (which would typically be co-located with the plug-in).  This deployment model is simple, but provides no etcd redundancy - in the case of an etcd failure, the plug-in will have to rebuild the database.  In larger deployments, the core cluster can be scaled up, as per..."

docs/source/etcd-architecture.rst
ed-harrison: OK - my suggested text above doesn't make sense if the plug-in has a proxy as well even in the simple case...

docs/source/etcd-architecture.rst
```diff
@@ -0,0 +1,238 @@
...
+
+etcd is responsible for performing all of the following tasks:
+
+.. _etcd admin guide: https://github.com/coreos/etcd/blob/master/Documentation/admin_guide.md#optimal-cluster-size
+
+Data Storage
+~~~~~~~~~~~~
+
+etcd stores the data for the Calico network in a distributed, consistent,
+fault-tolerant manner. This set of properties ensures that the Calico network
```
ed-harrison: Is it only fault tolerant if the cluster size is > 1?

docs/source/etcd-data-model.rst
```diff
@@ -0,0 +1,243 @@
+Calico etcd Data Model
+======================
+
+The current version of Calico is built around `etcd`_. etcd is a distributed,
+consistent key value store for shared configuration and service discovery with
+a focus on being simple, secure, fast, and reliable.
+
```
ed-harrison: As with the previous doc, I'd delete the paragraph above.

docs/source/understanding.rst
```diff
@@ -10,4 +10,5 @@ also of considerations for deploying Calico at large scale.
    l2-interconnectFabric
    arch-felix-and-acl
    calico-neutron-api
-   api-proposal
+   etcd-architecture
```
ed-harrison: As above, at this point we might want to put the etcd stuff in a "Futures" section, then move it here when we actually merge etcd into master.

Lukasa: That's a question, I think: do we plan to merge this immediately and move it later, or just hold off until the switch?

fasaxc:
Had a skim while I was checking what we could send out as a pre-release.  It looks like all the OpenStack setup stuff still includes ACL manager nad doesn't say how to configure the plugin, if it needs any config.

Lukasa:
I honestly don't know anything about the plugin configuration: care to enlighten me?

fasaxc:
Hehe, I don't know either, that's why I was looking in the docs ;-)

neiljerram:
Reading this thread now...                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                        From: Shaun CramptonSent: Monday, 13 April 2015 18:56To: Metaswitch/calicoReply To: Metaswitch/calicoCc: Neil JerramSubject: Re: [calico] Initial etcd documentation. (#295)Hehe, I don't know either, that's why I was looking in the docs ;-)

—Reply to this email directly or view it on GitHub.

neiljerram:
No, the plugin doesn't currently have any config.

It could do, e.g. interval for resync against neutron DB (currently hardcoded); and might need more in future, e.g. related to election between multiple plugin's. But currently, nothing.

fasaxc:
Please make sure that the docs say to use a unique --initial-cluster-token each time you crate an etcd cluster.  Without that, it defaults the cluster ID to a static value, which means we can't spot if the cluster ID changes.

docs/source/etcd-architecture.rst
ed-harrison: Where are we on this with what is planned for 0.16?  i.e. can we make this doc match what will be present in 0.16?

docs/source/etcd-architecture.rst
```diff
@@ -0,0 +1,230 @@
+Calico Architecture
```
ed-harrison: Given where we are now (i.e. we're publishing this change as part of the first etcd release) I think this document should be named "architecture.rst".

docs/source/understanding.rst
ed-harrison: As above, I think this should be called "architecture.rst".

docs/source/understanding.rst
```diff
@@ -10,4 +10,5 @@ also of considerations for deploying Calico at large scale.
    l2-interconnectFabric
    arch-felix-and-acl
```
ed-harrison: I think we should now remove arch-felix-and-acl.

Lukasa:
Alright, that was extraordinarily painful (particularly rebasing back on to master), but I believe we're finally close enough to merge this in.

Lukasa:
By the way, this is an impossible diff to review, so I think it should just be directly merged, and then review should be done on the rendered output on RTD.

Lukasa:
@ed-harrison, the privilege of hitting the big green merge button is yours. :shipit:

ed-harrison:
For the record, we'll merge this shortly before we actually push the 0.16 release button.

ed-harrison:
For the record, the previous merge from Alex Chan is the docs change associated with #408.

PR #296, opened 2015-03-26
Update openstack.rst

ed-harrison:
Rogue full-stop causing DCL bullet style catastrophe.


Lukasa:
OH NO! Don't let Ian see it! :fearful: :runner: :boom:

PR #300, opened 2015-03-28
L3 fabric

liljenstolpe:
l3 document done


liljenstolpe:
Looks clean - merging

PR #302, opened 2015-03-31
NJ review comments on L3 interconnect doc

neiljerram:


liljenstolpe:
The overlong hyphens should render as en-dashes, which are, I believe, grammatically correct.  However, its no big deal.  Merged.

neiljerram:
Thanks for the merge.  On the dash/hyphen point I'm not sure you're right; I'm pretty sure most dashes in words are hyphens and not en-dashes.  E.g. see http://english.stackexchange.com/questions/2116/when-should-i-use-an-em-dash-an-en-dash-and-a-hyphen.

Also, as a practical matter, it would be just weird if RST (or equally TeX) forced us to write the most common dash usage in a less intuitive way.  In my TeX past I've always used - in words, -- for ranges (en dash) and --- for punctuation (em dash).

PR #304, opened 2015-04-01
Add Google Analytics to our docs

Lukasa:
This change brings in a local copy of the ReadTheDocs theme, and then adds a Google Analytics code to the bottom of every page.

Rather than review all of this, it's probably most important to review the changes to `conf.py` in the first commit, and then all of the second commit.


Lukasa:
Responded to all your feedback @neiljerram. =)

ed-harrison:
This all looks reasonable to me, but think the best way to actually see if it works is to try it.  Is there any way to test it without just merging it live (where my main concern is that if we merge it, we'll break the docs completely - if we merge and find the docs are still OK, but the analytics don't work, then at least we'll know the general theme stuff is OK and can fix the analytics in isolation).

Lukasa:
We won't break the docs completely, I don't think: the local build works fine.

ed-harrison:
Let's give it a go then!

alexwlchan:
I don't know if it was intentional, but the main font for me has changed from Lato to Arial (Chrome and Firefox, Windows 7; Safari, iOS 8.2). My local build is still showing Lato.

ed-harrison:
Interesting.  I thought something was possibly slightly different, but couldn't put my finger on it.  FWIW, I don't think using Arial is a bad thing (and I assume we were using Lato as that was just the default rather than through any deliberate choice) - but others may disagree.

neiljerram:
FWIW I prefer the look now on docs.projectcalico.org - which I think means with the Arial.

In particular, the lowercase a looks a bit odd, in my local build.

PR #305, opened 2015-04-01
Improve search results in docs with short search tokens

alexwlchan:
> Search results with tokens with length <= 2 would fail silently, and not give
the user any explanation as to why they weren't seeing any results. This fix
removes those tokens from the search, and prints a warning to the user that
said tokens have been removed.

As discussed previously.

My JavaScript isn't great, so feedback would be appreciated!


alexwlchan:
The latest commit isn't strictly related to the search tokens, but is instead fixing a broken link on the search page that I noticed while I was poking around.

Lukasa:
For the latest commit @alexwlchan, you may want to offer a pull request upstream, [here](https://github.com/snide/sphinx_rtd_theme).

Lukasa:
@alexwlchan You seem to have changed the mode on all these files: do you mind fixing that up?

alexwlchan:
@Lukasa On the modes: oops, fixed.

Lukasa:
This looks entirely plausible to me. :cake: :sparkles:

Lukasa:
@alexwlchan I forgot to test this locally, so I tested it live: it seems that if you specify only one token that is itself too short (e.g. just 'ip') we don't print the warning.

alexwlchan:
@lukasa Hmm. I'm sure I tested that case. Possible I broke something when I tidied it up; I’ll take a look in a few hours.

alexwlchan:
@Lukasa I can’t reproduce this. I’m getting the warning if I use single tokens which are too short. e.g. this search for “ip”:

screenshot_2015-04-01_21-49-25

These are the test cases I’ve been trying: ip, ip address, ip at address, ip at os address, ip at, ip at os, and the same cases with single letter tokens. All behave exactly was I expect (i.e. printing the warnings as expected).

I’m not a JS expert, so it‘s possible this is a browser compatibility bug. I’ve tested in Chrome and FF on Windows 7, and Safari in OS X. If you’re still able to repro, could you tell me what browser you’re using, and I’ll look into this in the morning.
.

Lukasa:
I'll take a look at it tomorrow, I was getting it on Chrome on Win7.

ed-harrison:
It's working OK for me on Chrome, Win 8.1.

PR #313, opened 2015-04-02
Plugin/etcd: Provide config that Felices need, in etcd data

neiljerram:


Lukasa:
Assuming Ocelot tests pass, this looks good to me.

PR #314, opened 2015-04-02
Purge all mentions of the calico-docs repository

alexwlchan:
Per [pull request #4](https://github.com/Metaswitch/calico-docs/pull/4) on the calico-docs repo, this commit purges all mentions of the calico-docs repository from this repo.

A grep for the phrase “calico-docs” in my local copy now shows no mentions of the phrase except in the Debian and RPM changelogs announcing the move.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,8 +37,7 @@ To get started, you first need a working installation of
 stable build of Calico following the instructions
 [here](http://www.projectcalico.org/download/).
 
-Technical documentation is
-[on ReadTheDocs](http://docs.projectcalico.org/);
+Technical documentation is [here](http://docs.projectcalico.org/);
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -38,7 +38,7 @@ stable build of Calico following the instructions
 [here](http://www.projectcalico.org/download/).
 
 Technical documentation is
-[here](https://github.com/Metaswitch/calico-docs/wiki);
+[on ReadTheDocs](http://docs.projectcalico.org/);
```
ed-harrison: I'm not sure we need to advertise that we use RTD, so I think this can still say [here].

alexwlchan: Okay, I've reverted it to [here].

ed-harrison:
LGTM

PR #318, opened 2015-04-07
Felix start-of-day cleanup and multiple other fixes

### LINKED ISSUE #309: Felix dies if /calico key is missing from etcd.
Felix should presumably be somewhat resilient to etcd existing but being uninitialized. As it stands, right now Felix will launch and crash repeatedly, eventually causing the init system to leave it dead. That's less than idea.

Felix should be able to tolerate the absence of this key until such time as it appears.

The traceback:

```
  File "/usr/local/lib/python2.7/dist-packages/calico-0.14-py2.7.egg/calico/felix/felix.py", line 156, in main
    config = Config("/etc/calico/felix.cfg")
  File "/usr/local/lib/python2.7/dist-packages/calico-0.14-py2.7.egg/calico/felix/config.py", line 79, in __init__
    self.ETCD_PORT)
  File "/usr/local/lib/python2.7/dist-packages/calico-0.14-py2.7.egg/calico/felix/fetcd.py", line 451, in load_config
    global_cfg = client.read("/calico/config/")
  File "/usr/local/lib/python2.7/dist-packages/python_etcd-0.3.2-py2.7.egg/etcd/client.py", line 344, in read
    self.key_endpoint + key, self._MGET, params=params, timeout=timeout)
  File "/usr/local/lib/python2.7/dist-packages/python_etcd-0.3.2-py2.7.egg/etcd/client.py", line 582, in api_execute
    return self._handle_server_response(response)
  File "/usr/local/lib/python2.7/dist-packages/python_etcd-0.3.2-py2.7.egg/etcd/client.py", line 598, in _handle_server_response
    etcd.EtcdError.handle(**r)
  File "/usr/local/lib/python2.7/dist-packages/python_etcd-0.3.2-py2.7.egg/etcd/__init__.py", line 142, in handle
    raise exc(msg)
KeyError: 'Key not found : /calico'

```

### LINKED ISSUE #312: Security groups fail in etcd.
On my test install, everything goes wrong. iptables makes it clear that parsing the rules hasn't gone well:

```
Chain felix-p-_1f35a710c69e75a-i (1 references)
 pkts bytes target     prot opt in     out     source               destination 
    0     0 DROP       all  --  any    any     anywhere             anywhere             /* ERROR failed to parse rules DROP: */

Chain felix-p-_1f35a710c69e75a-o (1 references)
 pkts bytes target     prot opt in     out     source               destination 
    0     0 DROP       all  --  any    any     anywhere             anywhere             /* ERROR failed to parse rules DROP: */
```

With lots of this in the logs:

```
2015-04-02 16:08:13,030 [ERROR][25] calico.felix.frules 158: Failed to convert rules to fragments: [{'ip_version': 4, 'dst_net': u'0.0.0.0/0', 'dst_tag': None, 'dst_ports': None}, {'ip_version': 6, 'dst_net': u'::/0', 'dst_tag': None, 'dst_ports': None}].  Will DROP!
  File "/usr/local/lib/python2.7/dist-packages/calico-0.14-py2.7.egg/calico/felix/frules.py", line 152, in rules_to_chain_rewrite_lines
    on_deny=on_deny))
  File "/usr/local/lib/python2.7/dist-packages/calico-0.14-py2.7.egg/calico/felix/frules.py", line 187, in rule_to_iptables_fragment
    assert not unknown_keys, "Unknown keys: %s" % ", ".join(unknown_keys)
AssertionError: Unknown keys: ip_version
```

Here's an example of some rules:

```
ubuntu@calico-st02:~/calico$ etcdctl get /calico/policy/profile/20834313-58bc-4867-80c9-876c4c0f8ffa_7184c0df-294c-4932-8504-e5e1dd82b0cc/rules | python -m json.tool
{
    "inbound_rules": [
        {
            "icmp_type": null,
            "ip_version": 4,
            "protocol": "icmp",
            "src_net": null,
            "src_ports": null,
            "src_tag": "20834313-58bc-4867-80c9-876c4c0f8ffa"
        },
        {
            "ip_version": 4,
            "protocol": "tcp",
            "src_net": "172.18.203.223/32",
            "src_ports": [
                22
            ],
            "src_tag": null
        },
        {
            "ip_version": 6,
            "src_net": null,
            "src_ports": null,
            "src_tag": "7184c0df-294c-4932-8504-e5e1dd82b0cc"
        },
        {
            "ip_version": 4,
            "src_net": null,
            "src_ports": null,
            "src_tag": "7184c0df-294c-4932-8504-e5e1dd82b0cc"
        },
        {
            "ip_version": 4,
            "protocol": "tcp",
            "src_net": "172.18.203.223/32",
            "src_ports": [
                22
            ],
            "src_tag": null
        }
    ],
    "outbound_rules": [
        {
            "dst_net": "0.0.0.0/0",
            "dst_ports": null,
            "dst_tag": null,
            "ip_version": 4
        },
        {
            "dst_net": "::/0",
            "dst_ports": null,
            "dst_tag": null,
            "ip_version": 6
        },
        {
            "dst_net": "::/0",
            "dst_ports": null,
            "dst_tag": null,
            "ip_version": 6
        },
        {
            "dst_net": "0.0.0.0/0",
            "dst_ports": null,
            "dst_tag": null,
            "ip_version": 4
        }
    ]
}

```

### LINKED ISSUE #317: Minor: plugin shouldn't write null for unused fields in rules
It's allowed by felix but the intention was to only supply the things that we want to match on.  If nothing else it keeps the data small.

{"ip_version": 4, "dst_net": "0.0.0.0/0", "dst_tag": null, "dst_ports": null}

fasaxc:
This revision of etcd felix does start-of-day cleanup and is the first that passes the FV tests.  It's a collection of a few major chunks and lots of minor fixes that I don't have time to unpick before I leave.  The history is a bit messy due to all the different pieces.  I think it'd be hard to unpick all the minor bits and bobs so it might be best to squash merge once we get to that point rather than worrying too much about it.  Thoughts?

* implements start-of-day processing: waits for a /calico/config/ready flag to read "true" before doing its initial sync.  Deletes unknown IP chains and sets.  (It does this lazily on a 30s timer to give the initial snapshot some time to be applied; I think that's good enough for now since the snapshot should be fast and we probably want to bound how long we have old config in place anyway even during a graceful restart.)  General approach is not to enter the etcd polling loop until all the config values are present.
* fixes #317: removes nulls from our JSON
* fixes #312: failure to parse rules with ip_version specified
* fixes #309: felix dies if /calico key is missing.  It now stays dormant until its required keys are present.

I'll drop some signposts for the minor changes into the review as comments.

In any case, I think we want to base future changes on top of these or they're likely to conflict!


calico/common.py
```diff
@@ -192,16 +193,22 @@ def complete_logging(logfile=None,
...
-
-        formatter = logging.Formatter(FORMAT_STRING)
-        file_handler = logging.handlers.TimedRotatingFileHandler(
-            logfile, when="D", backupCount=10
-        )
-        file_handler.setLevel(file_level)
-        file_handler.setFormatter(formatter)
-        file_handler.addFilter(GreenletFilter())
-        root_logger.addHandler(file_handler)
+        if not file_handler:
```
fasaxc: This change makes the logging config idempotent so we can call it whenever the config changes.

calico/felix/actor.py
```diff
@@ -17,7 +17,85 @@
 felix.actor
 ~~~~~~~~~~~
 
-Actor infrastructure used in Felix.
+A queue-based Actor framework that supports efficient handling of
```
fasaxc: All this commenting seemed to get lost in the big gevent-proto->etcd merge but I think that was all that was lost.

calico/felix/actor.py
```diff
@@ -37,184 +115,17 @@
...
 
 
-DEFAULT_QUEUE_SIZE = 10
-
 ResultOrExc = collections.namedtuple("ResultOrExc", ("result", "exception"))
 
 # Local storage to allow diagnostics.
 actor_storage = gevent.local.local()
 
-_refs = {}
```
fasaxc: Moved these less-important classes to the bottom of the file.

calico/felix/actor.py
```diff
@@ -224,9 +135,8 @@ class Actor(object):
     max_ops_before_yield = 10000
     """Number of calls to self._maybe_yield before it yields"""
 
-    def __init__(self, queue_size=None, qualifier=None):
-        queue_size = queue_size or self.queue_size
-        self._event_queue = Queue(maxsize=queue_size)
+    def __init__(self, qualifier=None):
+        self._event_queue = Queue()
```
fasaxc: @plwhite I've just made the queues unbounded.  It's simpler than the spawn approach, which was causing problems anyway.

calico/felix/config.py
```diff
@@ -75,9 +75,23 @@ def __init__(self, config_path):
                                            "FelixHostname",
                                            socket.gethostname())
 
-        cfg_dict = fetcd.load_config(self.HOSTNAME,
-                                     self.ETCD_PORT)
+        self.STARTUP_CLEANUP_DELAY = 30
```
fasaxc: Reworked this so that fetcd can call update_config once the config is ready rather than needing it to be ready beforehand.

calico/felix/endpoint.py
```diff
@@ -51,11 +51,14 @@ def __init__(self, config, ip_type,
...
 
-        # Endpoints that are in use; LocalEndpoint actors referenced by
-        # endpoint_id.
+        # All endpoint dicts that we know about.
         self.endpoints_by_id = {}
         self.endpoint_id_by_iface_name = {}
 
+        # Set of endpoints that are live on this host.  I.e. ones that we've
+        # increffed.
+        self.local_endpoint_ids = set()
```
fasaxc: Tracking this explicitly to make it much simpler to get the code correct (there was a bug before)

calico/felix/fiptables.py
```diff
@@ -290,31 +299,83 @@ def delete_chains(self, table_name, chain_names, callback=None):
...
-
+        modify_succeeded = False
         try:
             self._calculate_index_changes()
-            input_lines = self._calculate_ipt_input()
+            input_lines = self._calculate_ipt_modify_input()
             self._execute_iptables(input_lines)
+            modify_succeeded = True
+            try:
+                input_lines = self._calculate_ipt_delete_input()
```
fasaxc: I split the iptables calls into two batches, one that modifies the chains (and replaces any chains that we're deleting with a drop rule).  And then one that actually does the delete.  This two-pass approach means that:

* we can better handle failure to delete chains that are referenced by orphaned chains that we haven't cleaned up yet
* when we're deleting orphan chains, we decouple them all so they don't reference each other before we actually do the delete
* we fail safe, dropping traffic rather than failing to delete a chain.

calico/felix/frules.py
```diff
@@ -145,11 +151,11 @@ def rules_to_chain_rewrite_lines(chain_name, rules, ip_version, tag_to_ipset,
         for r in rules:
             rule_version = r.get('ip_version')
             if rule_version is None or rule_version == ip_version:
-                fragments.append(rule_to_iptables_fragment(chain_name, r,
-                                                           ip_version,
-                                                           tag_to_ipset,
-                                                           on_allow=on_allow,
-                                                           on_deny=on_deny))
+                fragments.extend(rule_to_iptables_fragments(chain_name, r,
```
fasaxc: This removes the limit on the number of ports in a `src/dst_ports` match

calico/openstack/t_etcd.py
```diff
@@ -292,15 +292,21 @@ def profile_rules(self, profile_id):
...
                 if rule['direction'] == 'ingress':
-                    etcd_rule['src_tag'] = rule['remote_group_id']
-                    etcd_rule['src_net'] = net
-                    etcd_rule['src_ports'] = port_spec
+                    if rule['remote_group_id'] is not None:
+                        etcd_rule['src_tag'] = rule['remote_group_id']
+                    if net is not None:
+                        etcd_rule['src_net'] = net
+                    if port_spec is not None:
+                        etcd_rule['dst_ports'] = port_spec
```
fasaxc: This line is where the actual fix is.  Ingress rules should match on dest port, not source port.

calico/openstack/t_etcd.py
```diff
@@ -352,3 +358,5 @@ def provide_felix_config(self):
         # different things.  Here we provide the prefix that Neutron
         # uses.
         self.client.write('/calico/config/InterfacePrefix', 'tap')
+        # TODO Set this flag only once we're really ready!
+        self.client.write("/calico/config/Ready", 'true')
```
fasaxc: @neiljerram We need to code to write this flag after the etcd database is properly in-sync but I wasn't sure where to put it int he plugin code so I wrote it by default.

calico/openstack/t_etcd.py
```diff
@@ -292,15 +292,21 @@ def profile_rules(self, profile_id):
                 # Put it all together and add to either the inbound or the
                 # outbound list.
                 if rule['direction'] == 'ingress':
-                    etcd_rule['src_tag'] = rule['remote_group_id']
-                    etcd_rule['src_net'] = net
-                    etcd_rule['src_ports'] = port_spec
+                    if rule['remote_group_id'] is not None:
```
fasaxc: These None checks just clean up the JSON but there's also a fix below.

neiljerram:
Hi @fasaxc - some high level comments from me...

- I would like this to be split up and resubmitted as a set of smaller, independent pull requests.  (As I've also written about in email.)

- I'm concerned about the 'ready' logic.  I haven't seen a design for this, and don't understand why it's believed to be needed.  I'll also follow up about that in email.

- In all the places where you've provided explanatory notes above: please instead add your explanations as comments in the code.

Thanks!

calico/felix/actor.py
plwhite: Yes, I'm much happier about that approach!

Issue #316, opened 2015-04-02
interface_up() failed with IOError


fasaxc:
@robbrockbank Raising this to track your exception hit in the FV tests.

2015-04-02 23:05:35,388 [ERROR][33] calico.felix.actor 322: Exception processing 813b5eb3df76 (on_interface_update)
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/calico-0.14-py2.7.egg/calico/felix/actor.py", line 320, in _step
    result = msg.method()
  File "/usr/local/lib/python2.7/dist-packages/calico-0.14-py2.7.egg/calico/felix/endpoint.py", line 229, in on_interface_update
    self._configure_interface()
  File "/usr/local/lib/python2.7/dist-packages/calico-0.14-py2.7.egg/calico/felix/endpoint.py", line 334, in _configure_interface
    elif not devices.interface_up(self._iface_name):
  File "/usr/local/lib/python2.7/dist-packages/calico-0.14-py2.7.egg/calico/felix/devices.py", line 194, in interface_up
    flags = f.read().strip()
IOError: [Errno 22] Invalid argument

fasaxc:
@plwhite I think this is one for you to take a look at.  Looks like we hit a race where we check that the interface is still alive but it's gone before/during the read to get it's current state.

fasaxc:
I think this is fixed.

PR #319, opened 2015-04-07
Do not terminate if interface deleted under our feet.

plwhite:
Also minor comment tidy ups after fixing commenting of that routine.
Fixes issue #316.


plwhite:
Matt - assigned you because you are in the same time zone; relatively trivial fix, but let me know if you need more background. Thanks.

matthewdupre:
Looks fine.

PR #321, opened 2015-04-07
Config tests

plwhite:
Get config code UTed and ready for review.

### DIFF
--- a/calico/felix/test/test_config.py
+++ b/calico/felix/test/test_config.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# Copyright 2015 Metaswitch Networks
+# Copyright 2014 Metaswitch Networks
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


calico/felix/test/test_config.py
```diff
@@ -0,0 +1,214 @@
...
+real_client = etcd.client
+
+# Logger
+log = logging.getLogger(__name__)
+
+
+# Set of results from etcd.
+etcd_results = {}
+
+from collections import namedtuple
```
alexwlchan: I’d make a few changes to these module imports:

* There are module imports mixed with program flow – I’d consolidate them all at the top of the script.
* I’d remove `import mock`, since it doesn’t seem to be used.
* Maybe this is just personal preference, but I’d use the `try ... except ImportError` idiom for unittest. I don’t like that we have to import the sys module just to import another module.

calico/felix/test/test_config.py
```diff
@@ -0,0 +1,214 @@
...
+        """
+        Test various ways of defaulting config.
+        """
+        files = [ "felix_missing.cfg", # does not exist
+                  "felix_empty.cfg", # empty file
+                  "felix_empty_section.cfg", # file with empty section
+                  "felix_extra.cfg", # extra config is just logged
+                  ]
+
+        for file in files:
```
alexwlchan: I would replace this with `for filename in files` to avoid overriding the builtin file.

I don’t think it would cause any problems here, but it can’t hurt to be safe.

calico/felix/test/test_config.py
Lukasa: The `try...except ImportError` idiom doesn't work here because python 2.6 does have a unittest module, just not one with the features we need.

alexwlchan: @Lukasa I was actually thinking of doing

    try:
        import unittest2 as unittest
    except ImportError:
        import unittest

so we’d only hit the except block in versions of Python that didn’t have the unittest2 module; i.e. Python 2.7 or higher. We’d get the unittest2 module in Python 2.6. Or have I missed something?

plwhite: I've moved the imports to the top of the file and made them alphabetical for tidiness.

calico/felix/test/test_config.py
plwhite: OK, will do.

neiljerram:
In case this conversation might precipitate more non-trivial work in the config area - please see also my comments at https://github.com/Metaswitch/calico/issues/326.

matthewdupre:
This isn't currently passing UTs, but the failures are relatively insignificant, so I'm going to merge it now so we can start merge to master and start testing sooner.  I'll fix these failures up in parallel.

PR #322, opened 2015-04-07
Add script to validate iptables assumptions.

fasaxc:
This manual script replicates the manual testing I've done of interleaved iptables operations.  It validates various assumptions such as iptables-restore failing at COMMIT time if the table has changed.

Since it has to be run as root, I just made it a vanilla script rather than a test case.


plwhite:
I think this is useful stuff. I'd like to see some more comments in it, but I'm not sure it's worth doing that right now.

PR #327, opened 2015-04-07
Let environment override global config file for docker support.

### LINKED ISSUE #326: Add ability to override felix.cfg with environment variables
This makes it much easier to configure felix when it's running under a docker container.

fasaxc:
- felix now allows GLOBAL_ETCDHOST and GLOBAL_ETCDPORT to be
  specified.
- fixes #326.

### DIFF
--- a/calico/felix/config.py
+++ b/calico/felix/config.py
@@ -24,12 +24,13 @@ On instantiation, this module automatically parses the configuration file and
 builds a singleton configuration object. Other modules should just import the
 Config object and use the fields within it.
 """
+import os
+
 import ConfigParser
 import logging
 import socket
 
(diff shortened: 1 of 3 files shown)


Lukasa:
LGTM.

fasaxc:
@spikecurtis under feedback from Neil, I've changed this to use FELIX as the prefix for the variables and combined the ETCD variables into one: FELIX_ETCDADDR=localhost:1234 should do the trick.

spikecurtis:
I tested this out, and it allows me to override the etcd address Felix uses.  Nice!

plwhite:
Again, merging since I think it is good enough for later full review.

PR #331, opened 2015-04-08
Differentiate between different etcd errors.

### LINKED ISSUE #330: etcd thread dies if etcd connection fails.
Two related issues: etcd thread wasn't handling poll failures gracefully and retrying.  When etcd thread died, it didn't take the process with it.

fasaxc:
- Retry connection failures.
- Detect sync errors and trigger resync.
- Let all other errors trigger resync.
- Use a urllib3 Timeout object to set connect vs read timeout;
  for polling, we want a short connect timeout but a long read
  timeout.

Fixes #330.

### DIFF
--- a/calico/felix/felix.py
+++ b/calico/felix/felix.py
@@ -95,7 +95,7 @@ def _main_greenlet(config):
         iface_watcher.start()
         etcd_watcher.start()
 
-        greenlets = [
+        monitored_items = [
             v4_updater.greenlet,
             v4_ipset_mgr.greenlet,
             v4_rules_manager.greenlet,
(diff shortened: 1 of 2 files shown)


calico/felix/felix.py
```diff
@@ -95,7 +95,7 @@ def _main_greenlet(config):
         iface_watcher.start()
         etcd_watcher.start()
 
-        greenlets = [
+        monitored_items = [
```
fasaxc: This startup code is in flux, I plan to rework it so it's not worth reviewing.

calico/felix/fetcd.py
```diff
@@ -181,28 +182,44 @@ def watch_etcd(self):
...
-                               "on: %s < %s", response.modifiedIndex,
-                               next_etcd_index)
-                next_etcd_index = max(response.modifiedIndex,
-                                      next_etcd_index + 1)
+                except EtcdException as e:
+                    # Sadly, python-etcd doesn't have a clean exception
+                    # hierarchy; look at the message.  We only log the stack
+                    # trace for errors we're not expecting to avoid copious
+                    # log spam.
+                    if e.message == "No more machines in the cluster":
```
Lukasa: Is it a good idea to test that the message is exactly this? Might be better to have something like `if 'no more machines' in e.message.lower()`.

Lukasa:
Nothing very exciting here, one small note.

calico/felix/fetcd.py
fasaxc: Yeh, sounds good.

plwhite:
I think this code is good enough for merging - note that this has not been reviewed so that means "does not seem likely to make things worse" rather than that I have reviewed very hard (that review is pending).

PR #332, opened 2015-04-08
Rework startup to avoid concurrent access to config object

fasaxc:
I've done the simplest thing; we no longer try to update the config on each pass around the resync loop.  I'm interested in @plwhite's opinion on whether we should kill felix if we do get a config update that we don't know how to handle (which is all but the Ready flag right now).

Due to conflicts between this and my pull req to handle etcd errors correctly, I've rebased this pull req on top of that one.


plwhite:
ok to test

PR #334, opened 2015-04-08
Rework IptablesUpdater to simplify logic

fasaxc:
- Move DispatchChains to own file.
- Let each IptablesUpdater handle only one table.  This massively simplifies the book-keeping.
- Extract out a class to track the state of the current update batch.
- Force cleanup and ensure_rule_inserted run in their own batch in order to simplify the book-keeping.  (This required a change to the actor_event decorator so I took the opportunity to rename it to actor_message, which, I think is a better name.)
- Only try to delete unreferenced chains in cleanup so that we're less likely to hit issues.


ed-harrison:
ok to test

plwhite:
I'm nervous about this commit. While the code mostly looks sensible, you have introduced two iptables updaters for IPV4 - despite the comment in fiptables that says that you should only ever have one because they can conflict. That worries me a lot (though I realise that the NAT table does not get changed very often). I think there probably needs to be an answer to that before merging.

fasaxc:
Ah, it turns out you *can* make concurrent changes to different tables.  The granularity of the read/modify/write is at the table level so the old code was actually wrong in assuming that a commit would be atomic across multiple tables.  I'll fix up the commenting!

fasaxc:
Fixed up the commenting and rolled best-effort deletion into this patch (closing out some more of the fixmes).  That makes cleanup thorough; it should always delete all our unreferenced chains.  It uses a loop  that repeatedly deletes only chains that it explicitly knows are unreferenced according to the dataplane.  That should cover deleting orphaned endpoint chains and then profile chains in turn.

The low-level delete function does a best-effort delete.  Doing a binary chop on batches of chains to delete.

matthewdupre:
Looks like the issue above has been cleaned up, so I'm going to merge this.  It'll receive a full review then.

matthewdupre:
I've rebased this onto the etcd branch.

matthewdupre:
I can't merge this - there's a logical conflict between commit c6ec52ce, which involves adding a parameter to EtcdWatcher::watch_etcd and a refactor made in the 'Rename actor_event decorator to actor_message' commit, which stops parameters being passable to actor_messages.

Shaun - please fix this up so I can merge.

fasaxc:
Rebased onto etcd branch, should now be good to go.

PR #336, opened 2015-04-09
Fix failure to parse unicode data when validating ports.

plwhite:
Turns out that isinstance reckons that u"blah" is not a string. So made more robust.

### DIFF
--- a/calico/felix/fetcd.py
+++ b/calico/felix/fetcd.py
@@ -466,18 +466,25 @@ def validate_rule_port(port):
             return "integer out of range"
         return None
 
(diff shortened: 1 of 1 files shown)


plwhite:
Can I get this merged ASAP? It's trivial, and it is what was blocking Dave's testing.

matthewdupre:
Merged this already, although I have some questions about restricting ports to 0<port<65536.  In particular will we ever be passed a port range starting from 0, and will ICMP codes use the port field?

This can wait until we do the main code review of this section though.

PR #343, opened 2015-04-13
Compute and Control install on Ubuntu changes

lwr20:


docs/source/ubuntu-opens-install.rst
```diff
@@ -42,44 +42,21 @@ Icehouse or Juno.
...
-if later versions of those packages are released by Ubuntu.
+Edit ``/etc/apt/preferences`` to add the following lines, whose effect
+is to prefer Calico-provided packages for Nova and Neutron even if later
+versions of those packages are released by Ubuntu.
 
 ::
 
     Package: *
-    Pin: origin binaries.projectcalico.org
+    Pin: origin launchpad.net/project-calico
```
matthewdupre: This won't work - copy the pin line from Juno instead (it's magic...).

docs/source/ubuntu-opens-install.rst
```diff
@@ -292,14 +328,10 @@ And/or for IPv6 connectivity between compute hosts:
    configuration appropriately. You should consult the relevant
    documentation for your chosen BGP stack.
 
-9.  Create the ``/etc/calico/felix.cfg`` file by taking a copy of the
-    supplied sample config at ``/etc/calico/felix.cfg.example``. Then,
-    in ``/etc/calico/felix.cfg``:
-
-    -  Change both the ``PluginAddress`` and ``ACLAddress`` settings to
-       the host name or IP address of the controller node.
+14.  Create the ``/etc/calico/felix.cfg`` file by taking a copy of the
```
matthewdupre: I think it's worth mentioning that you can specify the FelixHostname used, which must match nova's hostname for that box.

Lukasa: I don't think that's worth mentioning, really. In 90% of cases people won't mess with this, in which case both nova and we will use `socket.gethostname()` and be done with it. In the remaining 10% people will presumably go looking for config when they find it doesn't work.

PR #347, opened 2015-04-13
Add python-gevent as a dependency for the debs.

fasaxc:


fasaxc:
@robbrockbank tested and this seems to do the trick but please can someone check that I've added the deps to the correct debs?

matthewdupre:
Looks fine.

PR #352, opened 2015-04-14
Be resilient to missing per-host config.

fasaxc:


calico/felix/fetcd.py
```diff
@@ -290,10 +290,16 @@ def _load_config_dict(self):
...
+        try:
+            host_cfg = self.client.read("/calico/host/%s/config/" %
+                                        self.config.HOSTNAME)
+        except KeyError:
+            _log.info("No configuration overrides for this node")
+        else:
+            configs.append(host_cfg)
 
-        for child in itertools.chain(global_cfg.children, host_cfg.children):
+        for child in itertools.chain(*[cfg.children for cfg in configs]):
```
Lukasa: This feels like overkill, given that we don't ever expect to have more configs. Why not just set `host_cfg` to the empty list before the try...except, and then keep the rest of the code the same?

Lukasa:
One note from me.

calico/felix/fetcd.py
fasaxc: Agree, can probably do this a bit cleaner.  That said, host_cfg isn't a list, it's an etcd node representing the directory.  cfg.children returns an iterator.

matthewdupre:
I'll merge this now so we can test with it today - @fasaxc could you please make any markups required and submit them separately (rolled into some other PR is fine since they'll only be small).

Thanks.

PR #354, opened 2015-04-14
Copyright 2015

neiljerram:


matthewdupre:
There are more places we need to update - for example debian/copyright, setup.py, calico/common.py.

I suggest grepping the whole codebase for '2014'; and perhaps 'copyright' as well.

neiljerram:
Sorry.  Think I've got them all now.

matthewdupre:
Looks good

PR #356, opened 2015-04-14
Reinstate modifiedIndex tracking

fasaxc:
Needed to avoid lots of spurious polling in Docker land.


fasaxc:
Merging this as an urgent fix for the Docker team.  It reinstates some previous function and the Docker team have tested it today.

PR #358, opened 2015-04-14
Workaround for etcd polling issue

fasaxc:
Trigger a full reconnect whenever we hit a timeout to avoid incorrect connection reuse.


fasaxc:
Merging this ahead of review because we think this fixes #359, which is a nasty bug for people to run into on day one.  Docker team have given it some testing and it seems to resolve the issue.

PR #362, opened 2015-04-15
Partial support for anycast, cleanup/comment ipsets code

fasaxc:
Previously, anycast didn't work because the ipsets management code assumed only a single endpoint could have one IP (so it would remove the IP from the ipset when _any_ endpoint with that IP went away).

I reworked the code to do all the index updates in some utility methods while at the same time changing the index structure to maintain a set of endpoints that share each IP.

This is an essential part of the anycast puzzle but we're not yet sure if it does everything that we need.  In particular, we're not sure if we correctly handle multiple routing rules for the same address in all cases.

### DIFF
--- a/calico/felix/ipsets.py
+++ b/calico/felix/ipsets.py
@@ -251,26 +251,6 @@ class IpsetManager(ReferenceManager):
 
         _log.info("Endpoint update complete")
 
-    def _remove_mapping(self, tag_id, endpoint_id, ip_address):
-        """
-        Removes the tag->endpoint->IP mapping from indexes and updates
-        any ActiveIpset if the IP is no longer present in the tag.
-
-        :return: True if the update resulted in removing that IP from the tag.
-        """
-        ep_ids = self.ip_owners_by_tag[tag_id][ip_address]
-        ep_ids.discard(endpoint_id)
-        ip_removed = False
-        if not ep_ids and ip_address in self.ips_in_tag:
-            self.ips_in_tag[tag_id].discard(ip_address)
-            del self.ip_owners_by_tag[tag_id][ip_address]
-            ip_removed = True
-        if ip_removed and self._is_starting_or_live(tag_id):
-            _log.debug("Removing %s from active tag %s", ip_address, tag_id)
-            self.objects_by_id[endpoint_id].remove_member(ip_address,
-                                                          async=True)
-        return ip_removed
-
     def _add_mapping(self, tag_id, endpoint_id, ip_address):
         """
         Adds the given tag->endpoint->IP mapping to the index and updates
@@ -287,6 +267,34 @@ class IpsetManager(ReferenceManager):
             self.objects_by_id[endpoint_id].add_member(ip_address, async=True)
         return ip_added
 
+    def _remove_mapping(self, tag_id, endpoint_id, ip_address):
+        """
+        Removes the tag->endpoint->IP mapping from indexes and updates
+        any ActiveIpset if the IP is no longer present in the tag.
+
+        :return: True if the update resulted in removing that IP from the tag.
+        """
+        ep_ids = self.ip_owners_by_tag[tag_id][ip_address]
+        ep_ids.discard(endpoint_id)
+        ip_removed = False
+        if not ep_ids and ip_address in self.ips_in_tag:
+            ips_in_tag = self.ips_in_tag[tag_id]
+            ips_in_tag.discard(ip_address)
+            if not ips_in_tag:
+                del self.ips_in_tag[tag_id]
+            del self.ip_owners_by_tag[tag_id][ip_address]
+            ip_removed = True
+        if ip_removed and self._is_starting_or_live(tag_id):
+            _log.debug("Removing %s from active tag %s", ip_address, tag_id)
+            self.objects_by_id[endpoint_id].remove_member(ip_address,
+                                                          async=True)
+        return ip_removed
+
+    def _add_profile_index(self, prof_id, endpoint_id):
+        if prof_id is None:
+            return
+        self.endpoint_ids_by_profile_id[prof_id].add(endpoint_id)
+
     def _remove_profile_index(self, prof_id, endpoint_id):
         if prof_id is None:
             return
@@ -296,10 +304,6 @@ class IpsetManager(ReferenceManager):
             _log.debug("No more endpoints use profile %s", prof_id)
             del self.endpoint_ids_by_profile_id[prof_id]
 
-    def _add_profile_index(self, prof_id, endpoint_id):
-        if prof_id is None:
-            return
-        self.endpoint_ids_by_profile_id[prof_id].add(endpoint_id)
 
 class ActiveIpset(RefCountedActor):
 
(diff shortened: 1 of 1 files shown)


Lukasa:
@fasaxc Who are you expecting to review this patch?

fasaxc:
Key motivation was to get anycast ready for the Maine Coons.  They can build from an integration branch so it's not blocking them.  Apart from that, it cleans up (and hopefully simplifies) that code a lot, making it ready for at least a first pass code review.  Not sure if Ocelots want to review it before merge or merge and then review later?

fasaxc:
@plwhite Please can you review this when you get back?

tomdee:
The Maine Coon's would really like this in 0.16. It would mean that the next release we do can be against a "proper" calico release rather than from a branch. But, it's still only a WIBNI so not worth delaying 0.16 for.

Symmetric:
I tested this patch yesterday, and I can verify that it fixes the ipsets for duplicate IPs (for both cases: endpoints on the same host, and endpoints on different hosts).

calico/felix/ipsets.py
```diff
@@ -159,103 +176,137 @@ def on_tags_update(self, profile_id, tags):
...
+        if endpoint is None:
+            return set()
+        return set(map(futils.net_to_ip,
+                       endpoint.get(self.nets_key, [])))
+
 
     @actor_message()
     def on_endpoint_update(self, endpoint_id, endpoint):
+        """
+        Update tag memberships and indexes with the new endpoint dict.
```
plwhite: Perhaps a bit pernickety, but "indexes" -> "indices" throughout.

"Indexes" appears to be correct in the context of the index of a book, but not in the context of indexed data.

calico/felix/ipsets.py
```diff
@@ -159,103 +176,137 @@ def on_tags_update(self, profile_id, tags):
...
+
+        :return: True if the IP wasn't already in that tag.
+        """
+        ep_ids = self.ip_owners_by_tag[tag_id][ip_address]
+        ip_added = not bool(ep_ids)
+        ep_ids.add(endpoint_id)
+        self.ips_in_tag[tag_id].add(ip_address)
+        if ip_added and self._is_starting_or_live(tag_id):
+            _log.debug("Adding %s to active tag %s", ip_address, tag_id)
+            self.objects_by_id[tag_id].add_member(ip_address, async=True)
```
plwhite: I disagree with using add_member here. It seems to me that we could be making 1000 add_member calls because somebody adds a single tag to a profile. Better (more efficient and simpler) to make a single call to replace_member once we have got our data structures correct (which in fact might mean that add_member and remove_member can be retired, since I don't think they would then ever be called).

calico/felix/ipsets.py
```diff
@@ -64,8 +63,11 @@ def __init__(self, ip_type):
         self.tags_by_prof_id = {}
         self.endpoints_by_ep_id = {}
 
-        # Indexes.
-        self.endpoint_ids_by_tag = defaultdict(set)
+        # Main index self.ip_owners_by_tag[tag][ip] == set([endpoint_id])
```
plwhite: I found this comment pretty confusing. The point of this ip_owners_by_tag thing is that when you remove an endpoint, you want to remove its IP addresses from ips_in_tag if another endpoint had the same ID (which is possible, thanks to anycast). I think you need a comment (just a line or two) making this clear.

calico/felix/ipsets.py
```diff
@@ -149,9 +145,30 @@ def on_tags_update(self, profile_id, tags):
             None if deleted.
         """
         _log.info("Tags for profile %s updated", profile_id)
-        old_tags = self.tags_by_prof_id.get(profile_id, [])
-        new_tags = tags or []
-        self._process_tag_updates(profile_id, set(old_tags), set(new_tags))
+
+        # General approach is to default to the empty list if the new/old
```
plwhite: I think the top level comment on this routine should say what the routine does and why - i.e. that it is called when the tags in a profile have changed, and hence the content of the ipset must change.

That sounds obvious, but it took me a moment to figure out and an extra one line comment in each method of this type would be helpful here.

plwhite:
Several review comments there - largely comment changes, but would also like to see `add_member` / `remove_member` usage sorted out before checkin. Once those are done, happy to merge.

plwhite:
@fasaxc - this is still with you to apply markups.

fasaxc:
Thanks, sorry, didn't spot that you'd reviewed it!

fasaxc:
@neiljerram I'd like to get this into 0.17 if I can since it helps the Maine coons and I don't want this patch to get stale.  I'll move it to 0.17 for now and then triage it back out again tomorrow if I just can't get to it.

calico/felix/ipsets.py
fasaxc: You've won me over; especially now that we're bank-switching the ipsets, there's no advantage to incremental updates.

fasaxc:
Applied code review markups but I think this needs another round of auto-FV and manual testing before we can merge it.

fasaxc:
@neiljerram @plwhite  I think that means we have to slip it out to 0.18.

lxpollitt:
Out of interest, was the add_member / remove_member an problem introduced by the anycast changes, or was this something that existed before and we would have wanted to tidy up anyway?  If the latter then I would have preferred it was raised and handled as a separate issue rather than as part of code review of the anycast changes.  We have a customer who plans to use anycast as part of their install w/c May 4th.  It would be very good if this was on a versioned release rather than some integration branch, and until it is the calico-docker releases will continue to fudge their releases rather than using the "real" released Felix.

neiljerram:
It seems this now just needs "another round of [...] manual testing" (per @fasaxc above).  So hopefully that can happen for 0.18.

Can anyone say more about what that manual testing is, and if they have it in plan for this week?

fasaxc:
@lxpollitt we were using add/remove before.

calico/felix/ipsets.py
fasaxc: Added.

fasaxc:
I've given this an FV run and had a poke int he logs and it seems to be working as intended.  @robbrockbank can I pass to you/team Maine Coon to do some anycast tire kicking and then merge?

robbrockbank:
LGTM

Created containers with identical IPs on different hosts and on the same host.  Tested for both IPv6 and IPv6.

Main use case of identical IPs on different host is good.  I see "ip route" to local container change to be a remote route when the local container is removed.  Bird config correctly has local and remote routes when present and correctly when container is removed.

Multiple containers on the same host - only seeing one local route in the bird config and in "ip route".    Removing a container maintains the local route until the last local container is removed.

lslatr:
I would like to get an anycast configuration working as described here... is there an example configuration somewhere that might point me in the right direction?  I've done a good bit of looking around and haven't found anything.

matthewdupre:
I don't think there's much in the way of examples.  Essentially it would involve giving the same IP to multiple endpoints; modifying the BIRD configuration to handle that (add path/multi path) and making sure everything else can deal.

PR #365, opened 2015-04-15
Comment out options in example felix.cfg so Felix can use its defaults

matthewdupre:
Felix picks socket.gethostname by default, which will work reasonably often and is therefore a better default than 'hostname'.  Just comment out the example.


Lukasa:
:cake: :sparkles: :cookie:

PR #366, opened 2015-04-15
[RFC] Felix: no longer read a config file (felix.cfg)

neiljerram:


Lukasa:
As discussed verbally, at this point in time I'm strongly against this change. =)

neiljerram:
I'm not strongly _pro_ this change in particular.  But I am bothered by the current complexity of the ways that Felix can be configured, and would prefer to streamline that somehow.

fasaxc:
Agree that config is a rough edge at the moment, this was just plonked in from the zmq felix with minimum rework.  I do think we need somewhere to put the etcd cluster details though.  The default of accessing a proxy on localhost won't be good enough for some use cases.

nbartos:
Somehow configuration changes (etcd host/port, hostname) must be able to be made.  I don't have a strong argument for a config file vs command line arguments, but using environment variables seems icky, and reminds me of other things I don't like.

fasaxc:
I think we've tied don't what we're going to do here and we're resigned to keeping the config file so let's close this RFC.  I do like the idea of not _requiring_ a config file, though, and defaulting to connecting to a local etcd proxy.

Issue #371, opened 2015-04-15
Apparent concurrent runs of felix cleanup


fasaxc:
I had expected this to be a non-issue since it should all get serialized appropriately but, if nothing else, it's very hard to follow.

There seems to be an underlying issue that felix is seeing an unexpected event (repeated sets of the Ready flag) that's triggering a resync much more often than expected.  I plan to clean up both that and the fact that we can trigger multiple resyncs.

fasaxc:
OK, we've flushed out a few bugs in diagnosing this.  I'll split these out into pull reqs (first one already done):

* Incorrect processing of needs_own_batch messages: #372 
* Incorrect calculation of affected chains in the iptables index.
* Felix being too cautious and resyncing when the plugin rewrites the config to the same values.
* Multiple cleanups could get scheduled, making the log output confusing.
* iptables wasn't waiting for the lock when we use it for reading the current state.

fasaxc:
All those other issues were split out into their own pull reqs.  This one can now be closed.

PR #372, opened 2015-04-15
Fix needs_own_batch processing and add UT.

fasaxc:
needs_own_batch wasn't being properly honoured.  The message was batched with following messages.  Fix the logic and add UT to verify it.


fasaxc:
I believe this fixes #371

robbrockbank:
This looks good to me.

fasaxc:
OK, this doesn't fix #371 but it this does fix a real issue and add UT.

PR #375, opened 2015-04-16
Fix endpoint updating

fasaxc:
Mark endpoint as dirty if its data changes (before we only dealt with transitioning from ready to non-ready).

Also fix a lookup that may fail to safely default.

### DIFF
--- a/calico/felix/endpoint.py
+++ b/calico/felix/endpoint.py
@@ -175,6 +175,8 @@ class LocalEndpoint(RefCountedActor):
         # Track whether the last attempt to program the dataplane succeeded.
         # We'll force a reprogram next time we get a kick.
         self._failed = False
+        # And whether we've received an update since last time we programmed.
+        self._dirty = False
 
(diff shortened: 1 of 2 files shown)


calico/felix/endpoint.py
```diff
@@ -264,7 +271,7 @@ def _maybe_update(self, was_ready):
         is_ready = self._ready
         if not is_ready:
             _log.debug("%s not ready, waiting on %s", self, self._missing_deps)
-        if self._failed or is_ready != was_ready:
+        if self._failed or self._dirty or is_ready != was_ready:
             ifce_name = self._iface_name
             if is_ready:
                 # We've got all the info and everything is active.
```
robbrockbank: Do you not need to set self._dirty to False at some point?

fasaxc: Yes!

robbrockbank:
looks good.

PR #376, opened 2015-04-16
Avoid concurrent cleanups

fasaxc:
Promote UpdateSplitter back to being an actor so it can manage concurrency.  Track a flag to avoid doing concurrent cleanups.

Concurrent cleanups _should_ be fine because the actors that we're calling into serialize the requests but they were producing a lot of log spam and slowing things down a lot, making it hard to track down the real issue that we were hitting today.

### DIFF
--- a/calico/felix/devices.py
+++ b/calico/felix/devices.py
@@ -141,8 +141,8 @@ def add_route(ip_type, ip, interface, mac):
     :param str mac: MAC address. May not be none unless ips is empty.
     :raises FailedSystemCall
     """
-    if mac is None and ips:
-        raise ValueError("mac must be supplied if ips is not empty")
+    if mac is None and ip:
+        raise ValueError("mac must be supplied if ip is provided")
 
     if ip_type == futils.IPV4:
         futils.check_call(['arp', '-s', ip, mac, '-i', interface])
--- a/calico/felix/devices.py
+++ b/calico/felix/devices.py
@@ -302,6 +302,7 @@ class InterfaceWatcher(Actor):
                     rta_data = rta_data[:-1]
                     if msg_type == RTM_NEWLINK:
                         _log.debug("Detected new network interface : %s", rta_data)
-                        self.update_splitter.on_interface_update(rta_data)
+                        self.update_splitter.on_interface_update(rta_data,
+                                                                 async=True)
                     else:
                         _log.debug("Network interface has gone away : %s", rta_data)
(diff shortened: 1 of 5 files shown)


calico/felix/devices.py
```diff
@@ -141,8 +141,8 @@ def add_route(ip_type, ip, interface, mac):
     :param str mac: MAC address. May not be none unless ips is empty.
```
Lukasa: This docstring line seems wrong (refers to `ips`, which doesn't exist), mind updating it?

calico/felix/splitter.py
```diff
@@ -66,14 +70,20 @@ def apply_snapshot(self, rules_by_prof_id, tags_by_prof_id,
...
+                                                 async=True))
+            self._cleanup_scheduled = True
 
+    @actor_message()
     def trigger_cleanup(self):
         """
         Called from a separate greenlet, asks the managers to clean up
         unused ipsets and iptables.
         """
+        self._cleanup_scheduled = False
```
Lukasa: I think there's still a risk of concurrently cleaning up here.

In particular, the way this flag is used allows multiple greenlets to execute this function at once. This could lead to the problematic logging situation you have mentioned already. Is it worth additionally adding a semaphore that is used only by this function to ensure that multiple cleanups can't actually execute at the same time?

fasaxc: trigger_cleanup() is now an actor_message so it'll get queued up on the UpdateSplitter's queue, that means it can't run concurrently with apply_snapshot() so this is safe.

fasaxc:
@robbrockbank Please can you review this, I'd like to get it in today.  If you can't review then I'll rebase #377 onto master because that one _is_ definitely required for release.

calico/felix/fetcd.py
```diff
@@ -242,18 +243,21 @@ def watch_etcd(self, update_splitter):
                 profile_id, rules = parse_if_rules(response)
                 if profile_id:
                     _log.info("Scheduling profile update %s", profile_id)
-                    update_splitter.on_rules_update(profile_id, rules)
+                    update_splitter.on_rules_update(profile_id, rules,
+                                                    async=False)
```
robbrockbank: I assume async=False sticks to existing behavior?  Is this as you'd expect though?

robbrockbank:
Looks good.  Discussed with Shaun the async=False comment above - understand that moving these to async=True may be useful ... but needs further thought.

Issue #374, opened 2015-04-15
Openstack plugin keeps setting config values that are already set


fasaxc:
Since felix watches /calico/, this causes events to go to every felix even though the value doesn't change.  Presently, this also triggers felix to do a resync, which s bad because it seems to happen after every update so we're turning every update into a resync.

I'll put up a pull req to prevent felix from being so dumb but I don't think we should be generating all these events at the plugin side.

fasaxc:
@neiljerram Please can you look at he plugin side of this tomorrow?

PR #377, opened 2015-04-16
Don't resync when config changes

fasaxc:
Conflicted with pull req 376 so I rebased on top of that.  376 needs to be merged to master before this can be reviewed.

Plugin is rewriting config with the same values, triggering a felix resync.  Disable the resync for now.

### DIFF
--- a/calico/felix/fetcd.py
+++ b/calico/felix/fetcd.py
@@ -257,7 +257,7 @@ class EtcdWatcher(Actor):
                     continue
                 profile_id, tags = parse_if_tags(response)
                 if profile_id:
-                    _log.info("Scheduling profile update %s", profile_id)
+                    _log.info("Scheduling tags update %s", profile_id)
                     update_splitter.on_tags_update(profile_id, tags,
                                                    async=False)
                     continue
@@ -288,12 +288,10 @@ class EtcdWatcher(Actor):
                     _log.warning("Global config changed but we don't "
                                  "yet support dynamic config response: %s",
                                  response)
-                    continue_polling = False
                 if response.key.startswith(self.my_config_prefix):
                     _log.warning("Config for this felix changed but we don't "
                                  "yet support dynamic config response: %s",
                                  response)
-                    continue_polling = False
 
     def _load_config_dict(self):
         """


Lukasa:
@fasaxc Looks like your commits are wandering around multiple branches: d1cee6a9cba1dd3e4812b55ba859967c677e18b2 was also present in and merged in #369. It should never have been in #369 of course (I thought it was weird when I did a post-hoc review of it), but now that it has been merged there do you mind dropping it from this PR and rebasing?

Lukasa:
In fact, the only new commit in this PR is a1036ec502f403b362ad7e5e9be10855a8c202ce, which doesn't conflict with anything else, so it would be best to just cherry-pick that change in. I'll go ahead and do that.

Lukasa:
Argh, ok, cherry-picking won't work well here either. Blarg. Let's sit on this until I can serialize some of these pull requests.

ed-harrison:
Am I right in thinking this addresses the Felix side of #374?

fasaxc:
@Lukasa Did you see my comment: "Conflicted with pull req 376 so I rebased on top of that. 376 needs to be merged to master before this can be reviewed.".  I figured that, once 376 was merged, then this pull req would only show the deltas on top of that.

Is there a better way to handle such a case?

fasaxc:
@ed-harrison Yes, but I spotted another case where we'll trigger a resync.  I'll raise a separate issue for that.

Lukasa:
@fasaxc I did see that, but the problem was that part of these changes also got merged as part of #369, which means that this PR has a duplicate commit, and merging it will also bring in that duplicate commit. It was *that* that I was trying to avoid, and ideally would like you to fix. =)

fasaxc:
OK, I'll rebase this onto master once #376 is merged.  This one is required for the release.  If we can't get #376 merged then we should manually take the resync changes from this branch and apply to master.

robbrockbank:
This looks good.

fasaxc:
Rebased onto master so it should be only a one tiny commit now.

fasaxc:
@Lukasa please can you pick this up?

Lukasa:
\o/ This looks far nicer, thanks @fasaxc!

PR #378, opened 2015-04-16
Fix calculation of batch-affected chains and deletable chains.

fasaxc:


fasaxc:
@robbrockbank Probably best for me to talk you through this one tomorrow.

fasaxc:
@Lukasa , @robbrockbank didn't get to this one and I think it is needed.  

The problem was that I was trying to be too clever to avoid some work.  The UpdateBatch class calculates the index updates for a particular sequence of iptables updates.  We need indexes to track chain dependencies so that we don't delete chains that are still referenced.

Part of its job is to keep track of all the chains that we're about to execute operations on (`self.affected_chains`).  We then output a "create or flush" line to the iptables-restore command for each of those.

I was trying to avoid outputting a "create-or-flush" line for chains that we were only deleting.  But the logic was wrong and too hard to fix in the start-of-day case where chains may be referenced by "orphan" chains.

The chains_to_delete logic also had a bug, it wasn't whitelisting all the referenced chains that we will be stubbing out.

Lukasa:
This looks plausible, but the reality is that the logic is sufficiently complicated that I'd need to spend a few hours ramping up on it to really effectively review. I'll try to find that time if possible, but otherwise @matthewdupre is free to merge if he feels comfortable doing it.

PR #380, opened 2015-04-16
Improve logging around IDs, refcounting and iptables programming.

fasaxc:
Just a bunch of logging improvements and a few asserts added while tracking down today's bugs.

### DIFF
--- a/calico/felix/endpoint.py
+++ b/calico/felix/endpoint.py
@@ -111,6 +111,7 @@ class EndpointManager(ReferenceManager):
         """
         if self._is_starting_or_live(endpoint_id):
             # Local endpoint thread is running; tell it of the change.
+            _log.info("Update for live endpoint %s", endpoint_id)
             self.objects_by_id[endpoint_id].on_endpoint_update(endpoint,
                                                                async=True)
         if endpoint is None:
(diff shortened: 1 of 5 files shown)


Lukasa:
Lots of these logging improvements consist simply of elevating debug logs to info level. What's the reasoning behind that?

fasaxc:
@Lukasa I think it was a mistake to have the "active" actors (i.e. the ones that handle the things this host actually cares about) log their state changes at debug.  There should be only a few of them (<500?) so, while these changes might have an impact, they shouldn't be _too_ chatty.  Without logs for their updates and life-cycle changes, the diags we were getting from the FV weren't enough to go on.  

The only one I'm a little scared about is tracing out the raw iptables updates and deps.  Those are very valuable but they could be large.

Lukasa:
It's still not clear to me why we're deciding to impact *all* users by logging this stuff at debug level, rather than just solving the problem for the FV case by turning the logging level up for the FV.

I'm worried about us generating a truly unholy amount of logs.

apollitt:
We need to get our logs to a level where we can diagnose problems in the field.  If based on our FV debugging experience we think we need to have better logging for that, then I suspect the same applies to our field logging.  Note I said "better logging".  In some cases this may mean "more logs", but not necessarily.

fasaxc:
I think this tunes it to a better balance for diagnosing general problems.  We don't log every update, only changes that impact this node.

PR #381, opened 2015-04-16
Plugin/etcd: Only write config values when not already correct

neiljerram:


neiljerram:
Addresses https://github.com/Metaswitch/calico/issues/374

ed-harrison:
LGTM

calico/openstack/t_etcd.py
```diff
@@ -352,11 +352,25 @@ def security_group_updated(self, sg):
...
+        """
+        # First read the config values, so as to avoid unnecessary
+        # writes.
+        prefix = None
+        ready = None
+        try:
+            prefix = self.client.read('/calico/config/InterfacePrefix').value
+            ready = self.client.read('/calico/config/Ready').value
+        except Exception:
+            pass
```
Lukasa: At the very least you want to log here, so that we can work out that exceptions are occurring if we hit them live.

Lukasa:
One note from me.

neiljerram:
Delta pushed.

Issue #340, opened 2015-04-09
Ensure we detect complete replacement of etcd cluster


fasaxc:
If the etcd server is replaced wholesale with another one running on the same IP address, felix could issue a poll to the new server for an etcd_index that is far in the future as far as that new server is concerned.

This would be hard to hit in a traditional mainline etcd deployment but some people may run etcd on a RAM disk as an ephemeral store with a floating IP, making it much easier to hit this window.

etcd clusters do have UUIDs so if there's some way we can check the UUID hasn't changed since our last poll, that'd do the trick.

fasaxc:
Looks like there's at least a partial solution.  When etcd responds, the response includes the cluster ID.  There's just the slight wrinkle that, if a poll fails, we don't get a response.

fasaxc:
I asked the etcd devs if there's a way to close that window already: https://github.com/coreos/etcd/issues/2649

fasaxc:
Turns out that etcd writes its headers at the start of the long poll, including the one we need.  We'll need to get access to the header but it's already doing what we need.

They also seem to support streaming events via an option: stream=true.  I couldn't find reference to that in the docs but, if we could use it then that'd prevent us from needing to reconnect every poll as we do now.

fasaxc:
This urllib3 incantation does the trick to load from their streaming model:
```
>>> import urllib3
>>> http = urllib3.HTTPConnectionPool("localhost",4001)
>>> r = http.request_encode_url("GET","/v2/keys/calico/",{"wait":"true", "stream":"true", "recursive":"true"}, preload_content=False)
>>> r.getheaders()
HTTPHeaderDict({'transfer-encoding': 'chunked', 'x-raft-index': '661276', 'x-raft-term': '7', 'x-etcd-index': '116', 'date': 'Fri, 10 Apr 2015 05:10:23 GMT', 'cont
ent-type': 'application/json', 'x-etcd-cluster-id': '7e27652122e8b2ae'})
>>> r.readline()
'{"action":"set","node":{"key":"/calico/foo","value":"baz","modifiedIndex":117,"createdIndex":117},"prevNode":{"key":"/calico/foo","value":"baz","modifiedIndex":11
6,"createdIndex":116}}\n'
```
I'm wondering if we should just drop python-etcd; it's a long way from supporting this.

Lukasa:
python-etcd is such a tiny wrapper it's really hard to see what it gets us. The other option, though, is to fork it and add the feature we need ourselves.

[...]

PR #383, opened 2015-04-16
Felix resyncs when the etcd cluster restarts while it's watching

matthewdupre:
This is the Felix part of a fix to issue #340 - it also requires an appropriately patched version of python-etcd (https://github.com/Metaswitch/python-etcd/pull/2)

### DIFF
--- a/calico/felix/fetcd.py
+++ b/calico/felix/fetcd.py
@@ -18,7 +18,7 @@ felix.fetcd
 
 Etcd polling functions.
 """
-from etcd import EtcdException
+from etcd import EtcdException, EtcdClusterIdChanged
 import etcd
 import itertools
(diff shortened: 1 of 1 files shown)


matthewdupre:
@fasaxc you might want to take a look too

Lukasa:
This looks plausible, though I'd like to know why the UT failed.

matthewdupre:
The UT is pretty unexciting - it didn't have the right version of python-etcd, so couldn't import the exception.

fasaxc:
I have a pull req open for the same feature, which includes adjusting the python-etcd version and updating tox config: https://github.com/Metaswitch/calico/pull/384 so I'm going to close this one.

PR #384, opened 2015-04-16
Reconnect to etcd if cluster ID changes.

fasaxc:
* Peg python-etcd to our forked version.
* Catch exception and resync in fetcd.py.


fasaxc:
@robbrockbank I think we need to get this in.  Let's discuss because @matthewdupre had some concerns about the version number.

robbrockbank:
This looks good.  I'll merge.

Discussion with Shaun -- there may be a follow up action with minor modifications to the version format - but that should not stop this going in.

PR #387, opened 2015-04-16
Synchronise data model with v1 specification.

fasaxc:
* Move data model definitions to shared file.
* Add v1 prefix to keys.
* Add missing workload_id to endpoint keys WARNING: this is just using the endpoint ID because I didn't know what it should be!

### DIFF
--- a/calico/felix/fetcd.py
+++ b/calico/felix/fetcd.py
@@ -49,7 +49,7 @@ class EtcdWatcher(Actor):
         super(EtcdWatcher, self).__init__()
         self.config = config
         self.client = None
-        self.my_config_dir = None
+        self.my_config_dir = dir_for_per_host_config(self.config.HOSTNAME)
 
     @actor_message()
     def load_config(self):
@@ -71,7 +71,6 @@ class EtcdWatcher(Actor):
                                     self.config.LOGLEVSYS,
                                     self.config.LOGLEVSCR)
             configured = True
-        self.my_config_dir = dir_for_per_host_config(self.config.HOSTNAME)
 
     @actor_message()
     def wait_for_ready(self):
--- a/calico/felix/fetcd.py
+++ b/calico/felix/fetcd.py
@@ -31,7 +31,7 @@ from urllib3.exceptions import ReadTimeoutError
 from calico import common
 from calico.datamodel_v1 import (VERSION_DIR, READY_KEY, CONFIG_DIR,
                                  RULES_KEY_RE, TAGS_KEY_RE, ENDPOINT_KEY_RE,
-                                 per_host_config_dir)
+                                 dir_for_per_host_config)
 from calico.felix.actor import Actor, actor_message
 
 _log = logging.getLogger(__name__)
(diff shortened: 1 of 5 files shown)


fasaxc:
This needs to go into the release.  I think the most important gate for merging this in is running the FVs against it.

fasaxc:
Discussed with the Maine Coon team about calico-docker.  Turns out that the changes to python-etcd that went in earlier make updating calico-docker much more painful than I'd thought.  They've taken on the action to do that work next time they rev the version of felix they're pulling into their build.  calico-docker compatibility shouldn't block this PR from going in.

fasaxc:
@neiljerram I think this needs to go into the release but I didn't manage to get a clean FV run against it tonight.  When I run a single FV, it fails very quickly with a 500 Internal Server Error (presumably from the OpenStack node) but I couldn't find anything in neutorn/server.log.

You might want to debug on the calicobobcat-ci01 control server, which is patched with the bobcat-integration branch (which has this change and my other pending changes merged in).  I run the test via this jenkins job: http://calicobobcat-jenkins.gulfstream:8080/job/test-class/ but the diags seem to be pretty poor.  It fails almost instantly.

calico/felix/fetcd.py
```diff
@@ -284,11 +274,11 @@ def watch_etcd(self, update_splitter):
...
                                  "resync.", response.action, response.key)
                     continue_polling = False
-                if response.key.startswith("/calico/config"):
+                if response.key.startswith(CONFIG_DIR):
                     _log.warning("Global config changed but we don't "
-                                 "yet support dynamic config response: %s",
-                                 response)
-                if response.key.startswith(self.my_config_prefix):
+                                 "yet support dynamic config key=%s",
+                                 response.key)
```
neiljerram: WHy the change here from response to response.key ?

calico/openstack/t_etcd.py
```diff
@@ -22,16 +22,19 @@
...
+                                 key_for_config)
 from calico.openstack.transport import CalicoTransport
 
 LOG = None
 OPENSTACK_ENDPOINT_RE = re.compile(
-    r'^/calico/host/(?P<hostname>[^/]+)/.*openstack.*/endpoint/(?P<endpoint_id>[^/]+)')
-OPENSTACK_POLICY_RE = re.compile(
-    r'^/calico/policy/profile/(?P<profile_id>[^/]+)/tags')
+    r'^' + VERSION_DIR +
+    r'/host/(?P<hostname>[^/]+)/.*openstack.*/endpoint/(?P<endpoint_id>[^/]+)')
```
neiljerram: Can we use HOST_DIR here instead of VERSION_DIR + /host ?

neiljerram:
Couple of comments above.  I also have a tiny delta to push shortly, with changes that I made myself.

calico/felix/fetcd.py
fasaxc: I think that might be a merge conflict resolution mistake.  I changed it to response so that it dumps the whole response including its key, action and data.  response is correct.

calico/openstack/t_etcd.py
fasaxc: Yes, I'll add that in.

fasaxc:
@neiljerram signed this off for me to self-merge after I made those markups.

Issue #385, opened 2015-04-16
Add configuration to the plugin to set etcd server to connect to


fasaxc:
Presently, the plugin expects to find an etcd proxy on localhost.  I think we need to be flexible here and allow that to be overridden to connect to an arbitrary address.

This is blocking a customer so we need to do this ASAP.

PR #389, opened 2015-04-16
[RFC] Plugin/etcd: allow configuration of etcd host and port

neiljerram:
This PR has my best guess at how to allow configuration of the etcd node/proxy that the Neutron plugin connects to.  If I've got it right, the configuration should be specified in the neutron.conf file (/etc/neutron/neutron.conf) in a [calico] section, like this:

    ...
    [calico]
    etcd_host = 172.18.203.27
    # etcd_port = 4001
    ...

Unfortunately I can't currently system test this, as I don't have access to an OpenStack/Calico cluster from home.  If anyone else would like to, please go ahead - you should be able to drop the modified t_etcd.py in place on the controller, then 'service neutron-server restart'.

### DIFF
--- a/calico/openstack/t_etcd.py
+++ b/calico/openstack/t_etcd.py
@@ -24,8 +24,20 @@ import json
 import re
 import time
 
+# OpenStack imports.
+from oslo.config import cfg
+
 # Calico imports.
 from calico.openstack.transport import CalicoTransport
+
+# Register Calico-specific options.
+calico_opts = [
+    cfg.StrOpt('etcd_host', default='localhost',
+               help="The hostname or IP of the etcd node/proxy"),
+    cfg.IntOpt('etcd_port', default=4001,
+               help="The port to use for the etcd node/proxy"),
+]
+cfg.CONF.register_opts(calico_opts, 'calico')
 
 LOG = None
 OPENSTACK_ENDPOINT_RE = re.compile(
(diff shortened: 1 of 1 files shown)


neiljerram:
Would also appreciate comments on whether the config model looks OK.  Structurally I haven't thought of any alternatives to using neutron.conf - but reviewers might prefer different names.

fasaxc:
In felix, we specify the "authority", i.e. hostname:port.  Probably best to be consistent with the felix config file?

fasaxc:
The key is called "EtcdAddr" (I didn't pick the naming scheme)

fasaxc:
I'll merge this into bobcat-integration so that @robbrockbank can run an FV run against it.

neiljerram:
Doing host:port was my initial preference too.  However, the OpenStack config infrastructure doesn't appear to allow this kind of value.  So I retreated to specifying host and port independently.

apollitt:
I think that doing it the OpenStack way is a sensible enough position.  It is after all an OpenStack config file.

ed-harrison:
Addresses #385

fasaxc:
Agree with @apollitt, using OpenStack's typed config variables and naming scheme is the way to go.

neiljerram:
@fasaxc You mean as the PR is currently?

ed-harrison:
I'm assuming @fasaxc did mean "as it is now".  LGTM, so merging.

PR #391, opened 2015-04-16
Plugin etcd disappearance

neiljerram:
I've had a little time this evening to investigate the resilience of the plugin to etcd resets, and this PR shows the result of that so far.  I've added UT to reset the etcd data at various possible points, then looked at the effect of that on the plugin code - which revealed KeyError exceptions at lines 218 and 220 where the plugin tries to read ..../rules and ..../tags for a profile.  The second commit here covers that.

I don't think this is a complete review, yet, I'm afraid, but I would hope that, with this, the plugin will have enough resilience to get through some initial testing of a moving etcd node.

### DIFF
--- a/calico/openstack/t_etcd.py
+++ b/calico/openstack/t_etcd.py
@@ -151,7 +151,11 @@ class CalicoTransportEtcd(CalicoTransport):
                     # cases the etcd key is no longer valid and should be
                     # deleted.  In the migration case, data will be written
                     # below to an etcd key that incorporates the new hostname.
-                    self.client.delete(child.key)
+                    try:
+                        self.client.delete(child.key)
+                    except etcd.EtcdKeyNotFound:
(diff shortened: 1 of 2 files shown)


neiljerram:
Hi @fasaxc  - There's also this PR that I believe is good to go now, and likely beneficial for our customers, even if it isn't yet a complete treatment of resilience w.r.t. etcd moving.

I'm afraid it will need some de-conflicting, though, in order to merge _after_ the versioning change.  Do you mind taking that on?

fasaxc:
Yep, I'll take it.

fasaxc:
@robbrockbank I've rebased this, please can you review?

robbrockbank:
Need to handle no key at line 154?

fasaxc:
I think so, I'll add that in.

robbrockbank:
Looks good

PR #396, opened 2015-04-17
Plugin/etcd: expect EtcdKeyNotFound instead of KeyError

neiljerram:


neiljerram:
Anyone around who can quickly review this?  It's blocking progress on fixing our FV rig.

PR #400, opened 2015-04-17
Doc: update for reduced set of configurable felix.cfg settings

neiljerram:


ed-harrison:
This change, in itself, is OK, but it does mean that we're no longer documenting the configuration which has been moved to etcd - see #402.

Lukasa:
:sparkles: :cake: :cookie:

Issue #382, opened 2015-04-16
Felix triggers full resync when directory in etcd is deleted (e.g. a whole profile)


fasaxc:
We get the event for the directory and don't spot that we can handle that case efficiently by generating deletes for its tags and rules so we think we need a full resync.

PR #404, opened 2015-04-17
Handle deletion of profiles without a resync.

fasaxc:


fasaxc:
Haha, you've been waiting for this one, eh? @Lukasa

fasaxc:
We sanity checked this on the FV rig against one or two tests and it passed, I think ti's good to merge.

calico/datamodel_v1.py
```diff
@@ -57,8 +57,12 @@
     r'/(?P<hostname>[^/]+)/.+/endpoint/(?P<endpoint_id>[^/]+)')
 
 
+def dir_for_host(hostname):
+    return VERSION_DIR + "/host/%s" % hostname
```
robbrockbank: better as HOST_DIR + "/%s" %  hostname
?

calico/datamodel_v1.py
```diff
@@ -57,8 +57,12 @@
...
+
+
 def dir_for_per_host_config(hostname):
-    return HOST_DIR + "/%s/config/" % hostname
+    return HOST_DIR + "/%s/config" % hostname
```
robbrockbank: and maybe dir_for_host(hostname) + "/config"

calico/datamodel_v1.py
```diff
@@ -80,3 +84,15 @@ def key_for_profile_tags(profile_id):
 
 def key_for_config(config_name):
     return CONFIG_DIR + "/%s" % config_name
+
+
+def get_profile_id_for_profile_dir(key):
+    """
+    :param str key: etcd key.
+    :returns The profile ID if this is a profile dir or None if not.
```
robbrockbank: or False if the key isn't valid ?

robbrockbank: or should that case also return None?

calico/felix/fetcd.py
```diff
@@ -267,12 +288,13 @@ def watch_etcd(self, update_splitter):
...
-                if response.action not in ("set", "create"):
-                    # FIXME: this check is over-broad.
+                if (response.action not in ("set", "create") and
+                        any((response.key.startswith(pfx) for pfx in
+                             PREFIXES_TO_RESYNC_ON_CHANGE))):
                     # It's purpose is to catch deletions of whole directories
                     # or other operations that we're not expecting.
-                    _log.warning("Unexpected action %s to %s; triggering "
-                                 "resync.", response.action, response.key)
+                    _log.warning("Unexpected event: %s; triggering resync.",
```
robbrockbank: should this really be a warning just for non-deletes?

calico/datamodel_v1.py
fasaxc: Yep

calico/datamodel_v1.py
fasaxc: Yep.

calico/datamodel_v1.py
fasaxc: Should return None.  Will fix.

calico/felix/fetcd.py
fasaxc: No, this case actually catches delete for, say a whole workload or a whole host right now.  In those cases, we can't yet handle it gracefully so we have to resync.

robbrockbank:
looks good

fasaxc:
Fixes #382

PR #410, opened 2015-04-20
Documentation for installing etcd on RHEL

Lukasa:
This change removes references to the ACL manager and instead documents how to download and install etcd on RHEL.

The next step of this is to get some tests of these instructions: that's with you @DaveLangridge.

### DIFF
--- a/docs/source/redhat-opens-install.rst
+++ b/docs/source/redhat-opens-install.rst
@@ -133,7 +133,7 @@ On a control node, perform the following steps:
         curl -L  https://github.com/coreos/etcd/releases/download/v2.0.9/etcd-v2.0.9-linux-amd64.tar.gz -o etcd-v2.0.9-linux-amd64.tar.gz
         tar xvf etcd-v2.0.9-linux-amd64.tar.gz
         cd etcd-v2.0.9-linux-amd64
-        mv etcd* /usr/local/bin/etcd
+        mv etcd* /usr/local/bin/
 
    - Create an etcd user::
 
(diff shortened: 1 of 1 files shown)


DaveLangridge:
A few changes suggested. 
1) adduser intruction should be something like:

adduser -s /sbin/nologin -d /var/lib/etcd/ etcd

2) RAM disc stuff. On control, after creating user.(same as ubuntu instructions.)

mount -t tmpfs -o size=512m tmpfs /var/lib/etcd

then add to end of /etc/fstab/

tmpfs /var/lib/etcd tmpfs nodev,nosuid,noexec,nodiratime,size=512M 0 0

3) install of etcd (controll and compute)
mv etcd* /usr/local/bin/    
(to move ectdctl as well as etcd)


4) Need extra installs to build python-etcd (compute node already has the dev tools bit):

yum groupinstall 'Development Tools'
yum install python-devel libffi-devel openssl-devel

5) Joint compute/control node doesn't need etcd stuff again. (Same bug in ubuntu instructions.)

6) Don't need to change PluginAddress in felix.cfg. (step 15 of compute)

7) Also needed yum install python-gevent

neiljerram:
Explicit instructions for python-gevent will not be needed, assuming https://github.com/Metaswitch/calico/pull/442 goes in.

neiljerram:
@Lukasa , @DaveLangridge is planning to test an install on RHEL6.5 tomorrow morning, so I wonder if it would be possible for you to make the remaining doc markups, and for me to merge them, today?  Then @DaveLangridge 's testing can also recheck the complete doc.

Lukasa:
@DaveLangridge Ok, I've applied the markups.

docs/source/redhat-opens-install.rst
```diff
@@ -131,17 +123,126 @@ On a control node, perform the following steps:
...
+     reinstated at boot time::
+
+        tmpfs /var/lib/etcd-rd tmpfs nodev,nosuid,noexec,nodiratime,size=512M 0 0
+
+   - Get etcd running by providing an init file.
+
+     For RHEL 6.5, place the following in ``/etc/init/etcd.conf``, replacing
+     ``<hostname>`` and ``<public_ip>`` with their appropriate values for the
+     machine. For ``<cluster_id>`` it can be any textual string, but make sure
+     to use a new one each time.
```
matthewdupre: Wording is a bit clumsy ('Use any string for <cluster_id>' would be clearer), and 'each time' is a bit ambiguous - it should be unique to the cluster but then fixed I think?

PR #420, opened 2015-04-20
Remove incorrect comment.

nbartos:
It looks like this got added when someone was copying code from a
previous block, but it didn't get corrected.


Lukasa:
Heh, good spot @nbartos, thanks!

Need to consult with some of my colleagues to work out why @calico-jenkins became so chatty.

Lukasa:
:cake: :sparkles:

PR #421, opened 2015-04-20
Fix ICMP rule generation in plugin; add missing support for ICMP code

### LINKED ISSUE #414: Felix fails to program rules for ICMP port
Looks like a mismatch between plugin and felix: the plugin is specifying both icmp_type and ports.  Also looks like felix needs to handle multiple ICMP ports and port ranges, which we may not at present (just spotted that multiport doesn't support icmp, d'oh).

```
Apr 18 19:58:33.553531 10.41.0.18 err calico-felix: Failed to convert rules to fragments: [{'ip_version': 4, 'src_tag': u'ed1c1daf-8cb8-44c8-a2cc-6ca518f31301', 'icmp_type': 255, 'protocol': u'icmp', 'dst_ports': [255]}].  Will DROP!
  File "/usr/lib64/python2.7/site-packages/calico/felix/frules.py", line 153, in rules_to_chain_rewrite_lines
  File "/usr/lib64/python2.7/site-packages/calico/felix/frules.py", line 209, in rule_to_iptables_fragments
  File "/usr/lib64/python2.7/site-packages/calico/felix/frules.py", line 295, in _rule_to_iptables_fragment
AssertionError: Protocol icmp not supported with dst_ports ({'protocol': u'icmp', 'dst_ports': ['255'], 'src_tag': u'ed1c1daf-8cb8-44c8-a2cc-6ca518f31301', 'ip_version': 4, 'icmp_type': 255, 'src_ports': []})
```

fasaxc:
* Plugin was generating rules with source/dest port as well as ICMP type/code, which felix was asserting on.  Fixed by moving the ICMP code into an alternating branch with the source/dest port code.
* Felix was missing support for the ICMP code.  I've added this in but there's a caveat: felix can only match on icmp type or icmp type and code; it can't match on code alone, which does seem to be allowed in OpenStack.  Need to discuss the correct approach here.  (Expanding to 255 rules seems overkill?)
* Added in missing validation for ICMP type/code.
* Added plugin UTs.

Not to reviewer: the first commit in the pull req makes the change to the plugin.  "Add UTs" commit refactors the plugin code to make it easier to test the rule creation in isolation.

Fixes #414.

### DIFF
--- a/calico/openstack/test/test_plugin_etcd.py
+++ b/calico/openstack/test/test_plugin_etcd.py
@@ -22,6 +22,7 @@ import eventlet
 import json
 import mock
 import unittest
+from calico.felix import fetcd
 
 import calico.openstack.test.lib as lib
 import calico.openstack.mech_calico as mech_calico
@@ -547,6 +548,15 @@ class TestPluginEtcd(lib.Lib, unittest.TestCase):
     def assertNeutronToEtcd(self, neutron_rule, exp_etcd_rule):
         etcd_rule = t_etcd._neutron_rule_to_etcd_rule(neutron_rule)
         self.assertEqual(etcd_rule, exp_etcd_rule)
+
+        # Check felix is happy with generated rule.
+        if neutron_rule["direction"] == "ingress":
+            rules = {"inbound_rules": [neutron_rule],
+                     "outbound_rules": []}
+        else:
+            rules = {"outbound_rules": [neutron_rule],
+                     "inbound_rules": []}
+        fetcd.validate_rules(rules)
 
 
 def neutron_rule(overrides):
--- a/calico/openstack/test/test_plugin_etcd.py
+++ b/calico/openstack/test/test_plugin_etcd.py
@@ -495,3 +495,69 @@ class TestPluginEtcd(lib.Lib, unittest.TestCase):
             {mech_calico.api.NETWORK_TYPE: 'vlan'},
             mech_calico.constants.AGENT_TYPE_DHCP
         ))
+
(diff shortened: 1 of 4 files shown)


calico/openstack/test/test_plugin_etcd.py
```diff
@@ -495,3 +496,78 @@ def test_check_segment_for_agent(self):
...
+        "ethertype": "IPv4",
+        "protocol": None,
+        "remote_ip_prefix": None,
+        "remote_group_id": None,
+        "direction": "ingress",
+        "port_range_min": None,
+        "port_range_max": None,
+    }
+    rule.update(overrides)
+    return rule
```
robbrockbank: missing newline at end of file

calico/openstack/test/test_plugin_etcd.py
```diff
@@ -495,3 +496,78 @@ def test_check_segment_for_agent(self):
...
+            "port_range_max": 100,
+        }), {
+            'ip_version': 6,
+            'protocol': 'icmpv6',
+            'src_net': '::/0',
+            'icmp_type': 123,
+            'icmp_code': 100,
+        })
+
+    def assertNeutronToEtcd(self, neutron_rule, exp_etcd_rule):
```
robbrockbank: minor point: found it a little confusing at first that neutron_rule is trumping the function with the same name.

fasaxc: Good spot, I'll rename to avoid confusion.

fasaxc:
FV run was successful, merging on @robbrockbank's say-so.

PR #426, opened 2015-04-21
When reconnecting, copy old cluster ID from previous etcd Client

fasaxc:
IMPORTANT: relies on new python-etcd version https://github.com/Metaswitch/python-etcd/pull/3.  that needs to be reviewed/merged/packaged first.

If we reconnect without doing a resync then we were clobbering the cluster ID that our version of python-etcd maintains.  This fix copies the cluster ID to the newly created client except when we resync.

### DIFF
--- a/calico/felix/fetcd.py
+++ b/calico/felix/fetcd.py
@@ -214,7 +214,7 @@ class EtcdWatcher(Actor):
                 except (ReadTimeoutError, SocketTimeout) as e:
                     # This is expected when we're doing a poll and nothing
                     # happened. socket timeout doesn't seem to be caught by
-                    # urllib3 0.7.1.  Simply reconnect.
+                    # urllib3 1.7.1.  Simply reconnect.
                     _log.debug("Read from etcd timed out (%r), retrying.", e)
                     # Force a reconnect to ensure urllib3 doesn't recycle the
                     # connection.  (We were seeing this with urllib3 1.7.1.)
--- a/calico/felix/fetcd.py
+++ b/calico/felix/fetcd.py
@@ -18,12 +18,11 @@ felix.fetcd
 
 Etcd polling functions.
 """
-import socket
+from socket import timeout as SocketTimeout
 from etcd import (EtcdException, EtcdClusterIdChanged, EtcdKeyNotFound,
                   EtcdEventIndexCleared)
 import etcd
 import httplib
-import itertools
 import json
 import logging
 import gevent
@@ -109,7 +108,7 @@ class EtcdWatcher(Actor):
                 gevent.sleep(RETRY_DELAY)
                 continue
 
-    def _reconnect(self):
+    def _reconnect(self, copy_cluster_id=True):
         _log.info("(Re)connecting to etcd...")
         etcd_addr = self.config.ETCD_ADDR
         if ":" in etcd_addr:
@@ -118,7 +117,13 @@ class EtcdWatcher(Actor):
         else:
             host = etcd_addr
             port = 4001
-        self.client = etcd.Client(host=host, port=port)
+        if self.client and copy_cluster_id:
+            old_cluster_id = self.client.expected_cluster_id
+            _log.info("Old etcd cluster ID was %s.", old_cluster_id)
+        else:
+            old_cluster_id = None
+        self.client = etcd.Client(host=host, port=port,
+                                  expected_cluster_id=old_cluster_id)
 
     @actor_message()
     def watch_etcd(self, update_splitter):
@@ -130,7 +135,7 @@ class EtcdWatcher(Actor):
         """
         while True:
             _log.info("Reconnecting and loading snapshot from etcd...")
-            self._reconnect()
+            self._reconnect(copy_cluster_id=False)
             self.wait_for_ready()
 
             # Load initial dump from etcd.  First just get all the endpoints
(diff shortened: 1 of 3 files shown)


tox.ini
```diff
@@ -7,5 +7,5 @@ deps =
     mock
     coverage>=4.0a5
     unittest2
-    git+http://github.com/Metaswitch/python-etcd.git@d5e8dc849f200ecfcb207086ce0d44f238c82aa5#egg=python-etcd
```
fasaxc: I pushed a tag to python-etcd for the new version.

Lukasa:
@fasaxc Can we get a milestone for this?

tox.ini
Lukasa: The tag is great, but doesn't solve the packaged side of the problem. Gotta bump the version number too. ;)

felix_requirements.txt
```diff
@@ -1,4 +1,4 @@
 gevent
 greenlet
 netaddr
-python-etcd==0.3.3-calico-1
+python-etcd==0.3.3-calico-2
```
Lukasa: Why does this not point to the GitHub repository?

neiljerram: I wonder if there's a case for doing that for gevent too, for Debian compatibility. We don't have Debian as a target platform for any of our packaged install methods; but we might want to help Debianites trying out Calico from source.

Although - does that mean we'd also have to add cython here?

Lukasa:
Ok, I've pushed the equivalent code to the icehouse-testing and juno-testing repositories as `python-etcd_0.3.3-1calico0.16~rc1`. This patch needs to be updated to tweak our dependencies in the debian file as well.

Lukasa:
(It's extremely annoying that the version number has diverged from the Python version number, so we really need to aim to upstream our patches if at all possible, or at least fix up the Ubuntu version number.)

Lukasa:
Some notes before this gets merged. =)

felix_requirements.txt
Lukasa: Also, if we're not pointing at the github repository, we should be using >=, not ==.

Lukasa: For examples of a problem caused by this already, see #431.

fasaxc: I had two thoughts in mind:
* putting the github link in here didn't seem to work when I tried it
* seemed wrong for our package itself to insist on getting the code from git rather than requiring a specific version
* I pegged it at that version since a 0.3.4 release from upstream would trump our fork's version.

fasaxc:
@Lukasa I feel like I'm missing a piece of the puzzle wrt to package versioning.  @matthewdupre said he was building the package on launchpad and it auto-assigned a number?

Shouldn't the deb be 0.3.3-2calico0.17-rc1 (2 coming from the calico-2 version suffix I added?)

fasaxc:
(Agree, would be good to upstream this.)

fasaxc:
Taking this back until I get to the bottom of the issue that Nick is hitting.

calico/felix/fetcd.py
```diff
@@ -203,44 +208,25 @@ def watch_etcd(self, update_splitter):
                                                 recursive=True,
                                                 timeout=Timeout(connect=10,
                                                                 read=90),
-                                                check_cluster_id=True)
+                                                check_cluster_uuid=True)
```
fasaxc: This was the bug: just a typo in the name :-(

fasaxc:
I think I've found the bug that Nick was hitting: it was a typo in the argument name that was swallowed by the `**kwdargs` so we didn't spot it.  With me to retest.

calico/felix/fetcd.py
```diff
@@ -203,44 +208,25 @@ def watch_etcd(self, update_splitter):
...
-                except ReadTimeoutError:
+                except (ReadTimeoutError, SocketTimeout) as e:
                     # This is expected when we're doing a poll and nothing
-                    # happened.
-                    _log.debug("Read from etcd timed out, retrying.")
-                    self._reconnect()
-                except socket.timeout:
-                    # That this leaks out appears to be an artifact of running
-                    # urllib3 on top of gevent.  Be defensive and reconnect.
-                    _log.warning("Raw socket.timeout leaked out of "
```
fasaxc: This is actually the mainline, I think.  My local testing wasn't hitting it because my virtualenv was poisoned by urllib3 1.10.  I've combined it into the ReadTimeout and demoted back to debug level.

fasaxc:
I've retested with e3f5b43 on the FV rig and verified by hand that that we now detect etcd cluster rebuild.  I did hit an apparent etcd bug, which may be fixed by a version bump of etcd: #444.

On the versioning, I'm going round in circles a bit.  pip is complaining that that version isn't PEP 440 compliant: https://www.python.org/dev/peps/pep-0440/#local-version-identifiers so maybe we should change it.  "0.3.3+calico-2" seems to be the recommended one.  WDYT?

Agree we should upstream the python-etcd changes and I've created a branch to do so.

calico/felix/fetcd.py
```diff
@@ -203,44 +209,25 @@ def watch_etcd(self, update_splitter):
...
-                                 "python-etcd.  Reconnecting...",
-                                 exc_info=True)
+                    # happened. socket timeout doesn't seem to be caught by
+                    # urllib3 0.7.1.  Simply reconnect.
```
Lukasa: This comment almost certainly should read 'urllib3 1.7.1'.

felix_requirements.txt
```diff
@@ -1,4 +1,4 @@
 gevent
 greenlet
 netaddr
-python-etcd>=0.3.3-calico-1
+python-etcd==0.3.3-calico-2
```
Lukasa: Please skip the hard pin, it just causes us problems if we have to version rev again.

Lukasa:
I think we don't really care that pip is complaining, given that we want to upstream this anyway.

Lukasa:
I've applied my markups myself because they're so small: @neiljerram has done a quick over-the-shoulder code review of my delta, so if/when these tests pass I'll merge this myself.

PR #430, opened 2015-04-21
Fix up a range of commnents.

plwhite:
While code-reading for the design doc, I found some poor comments. Fixed these up, both clarifications and tidying up TODO markers (so they are all TODO, not a mix of TODO and FIXME).

### DIFF
--- a/calico/felix/fetcd.py
+++ b/calico/felix/fetcd.py
@@ -319,8 +319,8 @@ class EtcdWatcher(Actor):
                 if (response.action not in ("set", "create") and
                         any((response.key.startswith(pfx) for pfx in
                              PREFIXES_TO_RESYNC_ON_CHANGE))):
-                    # It's purpose is to catch deletions of whole directories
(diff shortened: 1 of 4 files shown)


calico/felix/fetcd.py
```diff
@@ -319,8 +319,8 @@ def watch_etcd(self, update_splitter):
                 if (response.action not in ("set", "create") and
                         any((response.key.startswith(pfx) for pfx in
                              PREFIXES_TO_RESYNC_ON_CHANGE))):
-                    # It's purpose is to catch deletions of whole directories
-                    # or other operations that we're not expecting.
+                    # The purpose of this code is to catch deletions of whole
+                    # directories or other operations that we're not expecting.
```
matthewdupre: Or just 'Catch deletions of ...'?

plwhite: Yes, that is better. I'll fix it that way.

Issue #436, opened 2015-04-21
ActiveIpset failed because ipset already existed


fasaxc:
@nbartos hit this during his FV testing:

```
21491-Apr 21 16:15:30.593386 10.41.0.18 err calico-felix: _finish_msg_batch failed.
21492-Traceback (most recent call last):
21493-  File "/usr/lib64/python2.7/site-packages/calico/felix/actor.py", line 248, in _step
21494-  File "/usr/lib64/python2.7/site-packages/calico/felix/ipsets.py", line 345, in _finish_msg_batch
21495-  File "/usr/lib64/python2.7/site-packages/calico/felix/ipsets.py", line 388, in _sync_to_ipset
21496-  File "/usr/lib64/python2.7/site-packages/calico/felix/futils.py", line 99, in check_call
21497-FailedSystemCall: Failed system call (retcode : 1, args : ('ipset', 'restore', '-file', '/tmp/tmpEx166Z'))
21498-  stdout  : 
21499:  stderr  : ipset v6.21.1: Error in line 1: Set cannot be created: set with the same name already exists
```
Since we check whether the ipset exists in advance, perhaps there's a bug in the checking code that makes it assume there's no ipset when in fact there is.

Is there some way to make the ipset operations idempotent?  How about passing "-exist" on the create and always flushing after that?

nbartos:
+1 for idempotent

plwhite:
Can you provide a reference to that atomicity? I believe that ipset restore is NOT atomic, and so this fix is not correct.

plwhite:
I have reproed this, and will put together a fix. It's caused by "set_exists" being calculated before the previous actor exits, in a window where the ipset is used, goes out of scope, and then gets recreated.

plwhite:
We've had some going round in circles on this one. PR #443 fixed it, but so does the latest iteration of PR #437. I've closed the first, and the second is pending review markups and merge.

fasaxc:
Fixed by #437.

PR #437, opened 2015-04-21
Make ipset sync idempotent.

fasaxc:
* Avoids reliance on checking whether set exists.
* Removes different cases.

### DIFF
--- a/calico/felix/futils.py
+++ b/calico/felix/futils.py
@@ -102,7 +102,7 @@ def check_call(args, input_str=None):
     retcode = proc.returncode
     if retcode:
         raise FailedSystemCall("Failed system call",
-                               args, retcode, stdout, stderr)
+                               args, retcode, stdout, stderr, input=input_str)
 
     return CommandOutput(stdout, stderr)
 
--- a/calico/felix/futils.py
+++ b/calico/felix/futils.py
@@ -46,19 +46,22 @@ SHORTENED_PREFIX = "_"
 
 
 class FailedSystemCall(Exception):
-    def __init__(self, message, args, retcode, stdout, stderr):
+    def __init__(self, message, args, retcode, stdout, stderr, input=None):
         super(FailedSystemCall, self).__init__(message)
         self.message = message
         self.args = args
         self.retcode = retcode
         self.stdout = stdout
         self.stderr = stderr
+        self.input = input
 
     def __str__(self):
         return ("%s (retcode : %s, args : %s)\n"
                 "  stdout  : %s\n"
-                "  stderr  : %s\n" %
-                (self.message, self.retcode, self.args, self.stdout, self.stderr))
+                "  stderr  : %s\n"
+                "  input  : %s\n" %
+                (self.message, self.retcode, self.args,
+                 self.stdout, self.stderr, self.input))
 
 
 def call_silent(args):
@@ -90,7 +93,9 @@ def check_call(args, input_str=None):
     """
     log.debug("Calling out to system : %s" % args)
 
+    stdin = subprocess.PIPE if input_str is not None else None
     proc = subprocess.Popen(args,
+                            stdin=stdin,
                             stdout=subprocess.PIPE,
                             stderr=subprocess.PIPE)
     stdout, stderr = proc.communicate(input=input_str)
--- a/calico/felix/futils.py
+++ b/calico/felix/futils.py
@@ -74,7 +74,8 @@ def call_silent(args):
     except FailedSystemCall as e:
         return e.retcode
 
-def check_call(args):
+
+def check_call(args, input_str=None):
     """
     Substitute for the subprocess.check_call function. It has the following
     useful characteristics.
@@ -92,7 +93,7 @@ def check_call(args):
     proc = subprocess.Popen(args,
                             stdout=subprocess.PIPE,
                             stderr=subprocess.PIPE)
-    stdout, stderr = proc.communicate()
+    stdout, stderr = proc.communicate(input=input_str)
     retcode = proc.returncode
     if retcode:
         raise FailedSystemCall("Failed system call",
--- a/calico/felix/ipsets.py
+++ b/calico/felix/ipsets.py
@@ -370,6 +370,16 @@ class ActiveIpset(RefCountedActor):
         # We have got the set into the correct state.
         self.programmed_members = self.members.copy()
 
+    def __str__(self):
+        return self.__class__.__name__ + "<queue_len=%s,live=%s,msg=%s," \
+                                         "name=%s,id=%s>" % (
+            self._event_queue.qsize(),
+            bool(self.greenlet),
+            self._current_msg,
+            self.name,
+            self._id,
+        )
+
 
 def tag_to_ipset_name(ip_type, tag, tmp=False):
     """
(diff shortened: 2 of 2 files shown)


fasaxc:
This is an attempt to fix #436

fasaxc:
@plwhite please can you review?  I'm not entirely sure why you added the double buffer but I _think_ `ipset restore` is atomic so this _should_ work.  OTOH, you might have had a different reason that I wasn't aware of!

calico/felix/ipsets.py
```diff
@@ -127,11 +122,10 @@ def cleanup(self):
...
         # only clean up our own rubbish.
         pfx = IPSET_PREFIX[self.ip_type]
-        tmppfx = IPSET_TMP_PREFIX[self.ip_type]
-        felix_ipsets = set([n for n in all_ipsets if n.startswith(pfx) or
-                                                     n.startswith(tmppfx)])
+        felix_ipsets = set([n for n in all_ipsets if n.startswith(pfx)])
         whitelist = set()
-        for ipset in self.objects_by_id.values():
+        for ipset in (self.objects_by_id.values() +
+                      self.stopping_objects_by_id.values()):
```
fasaxc: If there's a stopping ActiveIpset, it will be trying to do its own cleanup.  I thought this might have been the bug that Nick was hitting but that wasn't borne out by his logs.

plwhite:
Can you provide a reference to that atomicity? I believe that ipset restore is NOT atomic, and so this fix is not correct.

fasaxc:
I was basing it on code reading in the ipset source code; it buffers up all the commands and then calls a function called ipset_commit() that actually applies the changes.  I wasn't able to follow it further than that, though, so you're probably right that we shouldn't rely on it.

I still think there's something in the approach that avoids needing to check if anything exists.  Using the -exist flag, we could do this:

* always create both sets with -exist
* always flush the tmp set
* write the data to the tmp set
* swap
* always delete the tmp set

No need to check for existence and should work from any state.

fasaxc:
@plwhite Updated with the idea that I was talking about; I like having just one code path that should handle all cases.  What do you think?

calico/felix/ipsets.py
```diff
@@ -131,7 +120,8 @@ def cleanup(self):
         felix_ipsets = set([n for n in all_ipsets if n.startswith(pfx) or
                                                      n.startswith(tmppfx)])
         whitelist = set()
-        for ipset in self.objects_by_id.values():
+        for ipset in (self.objects_by_id.values() +
+                      self.stopping_objects_by_id.values()):
```
plwhite: Is this "stopping_objects_by_id" the same fix or something else?

calico/felix/ipsets.py
```diff
@@ -328,11 +314,18 @@ def remove_member(self, member):
 
     @actor_message()
     def on_unreferenced(self):
+        # Temporary set should never be left over; just in case, try to
+        # delete it.  Silence errors, we expect this to fail unless we're
+        # cleaning up after an error.
+        futils.call_silent(["ipset", "destroy", self.tmpname])
         try:
-            if self.set_exists:
+            if self.programmed_members is not None:
```
plwhite: Why not delete the set unconditionally? It seems to me that you could either get rid of the test and assume it exists (since it is benign to fail) or (better) test if the set exists by just calling ipset_exists. That feels more robust than figuring out whether the set exists based on whether you have written to it.

calico/felix/ipsets.py
```diff
@@ -351,61 +344,57 @@ def _finish_msg_batch(self, batch, results):
...
+        # The only operation that we're sure is atomic is swapping two ipsets
+        # so we build up the complete set of members in a temporary ipset,
+        # swap it into place and then delete the old ipset.
+        create_cmd = "create %s hash:ip family %s --exist"
+        input_lines = [
+            # Ensure both the main set and the temporary set exist.
+            create_cmd % (self.name, self.family),
+            create_cmd % (self.tmpname, self.family),
+
+            # Flush the temporary set, this is a no-op unless we had a
```
plwhite: Nit-picking, but since I've got other feedback! You have two sentences separated by a comma. Should be ".. set. This is a ..."

calico/felix/ipsets.py
```diff
@@ -351,61 +344,57 @@ def _finish_msg_batch(self, batch, results):
...
 
         # We have got the set into the correct state.
         self.programmed_members = self.members.copy()
 
 
-def ipset_exists(name):
+def tag_to_ipset_name(ip_type, tag, tmp=False):
     """
-    Check if a set of the correct name exists.
+    Turn a tag ID in all its glory into an ipset name.
```
plwhite: I think this comment dates from the days when the tag ID passed around was not truncated - worth just removing the "in all its glory" bit (which I agree that I wrote, so mea culpa!).

calico/felix/ipsets.py
```diff
@@ -351,61 +344,57 @@ def _finish_msg_batch(self, batch, results):
...
+        # Then, atomically swap the temporary set into place.
+        input_lines.append("swap %s %s" % (self.name, self.tmpname))
+        # Finally, delete the temporary set (which was the old active set).
+        input_lines.append("destroy %s" % self.tmpname)
+        # COMMIT tells ipset restore to actually execute the changes.
+        input_lines.append("COMMIT")
+
+        input_str = "\n".join(input_lines) + "\n"
+        cmd = ["ipset", "restore"]
+        process = subprocess.Popen(cmd,
```
plwhite: I don't like calling subprocess directly. We deliberately always do subprocess commands going through futils, and I think you should use futils.check_call (which does involve giving it a default stdin argument).

plwhite:
I've raised a number of comments, but none of them are particularly substantive - happy to merge once those are addressed.

calico/felix/ipsets.py
fasaxc: I think this was another potential race between the start-of-day cleanup code and the stopping ipset cleanup code.  (I believe it would be benign since one of the cleanups would win and we won't start another ipset until both cleanups finish but better safe than sorry.)

calico/felix/ipsets.py
fasaxc: Sounds good, I'll take that snipped from your patch.

calico/felix/ipsets.py
fasaxc: Sounds good.

fasaxc:
Looks like there's a bug in the latest commit, FV tests are not passing on it.

calico/felix/futils.py
```diff
@@ -89,10 +93,12 @@ def check_call(args):
...
+    stdin = subprocess.PIPE if input_str is not None else None
     proc = subprocess.Popen(args,
+                            stdin=stdin,
                             stdout=subprocess.PIPE,
                             stderr=subprocess.PIPE)
-    stdout, stderr = proc.communicate()
+    stdout, stderr = proc.communicate(input=input_str)
     retcode = proc.returncode
     if retcode:
         raise FailedSystemCall("Failed system call",
```
robbrockbank: Shouldn't input_str be passed into FailedSystemCall arguments?

robbrockbank:
I've looked at the last three commits since the original review.

I've one minor comment about passing in the input_str to the exception

calico/felix/futils.py
fasaxc: Yep, will fix.

PR #441, opened 2015-04-22
Beef up syslog format, add a couple of additional logs.

fasaxc:
Having the PID in the syslog is invaluable and adding the file and line number seemed like it'd be useful too.


fasaxc:
@robbrockbank could you take a look at this?  I think you might have more interest in logging than @Lukasa now I think about it.

robbrockbank:
looks fine.

PR #443, opened 2015-04-22
Fix problem where we check whether ipset exists before another actor deletes it.

plwhite:
Fixes bug #436.

### DIFF
--- a/calico/felix/ipsets.py
+++ b/calico/felix/ipsets.py
@@ -289,10 +289,6 @@ class ActiveIpset(RefCountedActor):
         # Members which really are in the ipset.
         self.programmed_members = None
 
-        # Do the sets exist?
-        self.set_exists = ipset_exists(self.name)
-        self.tmpset_exists = ipset_exists(self.tmpname)
-
         # Notified ready?
         self.notified_ready = False
 
@@ -329,10 +325,11 @@ class ActiveIpset(RefCountedActor):
     @actor_message()
     def on_unreferenced(self):
         try:
-            if self.set_exists:
-                futils.check_call(["ipset", "destroy", self.name])
-            if self.tmpset_exists:
-                futils.check_call(["ipset", "destroy", self.tmpname])
+            # Destroy the ipsets - ignoring any errors.
+            _log.debug("Delete ipsets %s and %s if they exist",
+                       self.name, self.tmpname)
+            futils.call_silent(["ipset", "destroy", self.name])
+            futils.call_silent(["ipset", "destroy", self.tmpname])
         finally:
             self._notify_cleanup_complete()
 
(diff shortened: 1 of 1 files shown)


plwhite:
Cory, can you please review for style, and if you are comfortable then I'll get Shaun to review for correct function. Thanks.

calico/felix/ipsets.py
```diff
@@ -289,13 +289,30 @@ def __init__(self, tag, ip_type):
...
+        """
+        if real:
+            if self.set_exists is None:
+                self.set_exists = ipset_exists(self.name)
+            return self.set_exists
+        else:
+            if self.tmpset_exists is None:
+                self.tmpset_exists = ipset_exists(self.tmpname)
+            return self.tmpset_exists
+
```
Lukasa: There are two identical branches of code here, which seem entirely redundant to me, along with a boolean parameter that you pass into the function undecorated (which looks really weird).

The correct refactor here seems to be modifying `ipset_exists` to check if `None` is passed to it, and return `False` in that case. Then, all the calls to `self._set_exists()` turn into calls to `ipset_exists` with parameters on self. Probably saves about 15 lines of diff.

plwhite:
@fasaxc This is a fix for bug #436 that I have reproed here (or rather Dave did) - I'm getting Cory to review for style if he has time, but otherwise it's worth you taking a look.

calico/felix/ipsets.py
plwhite: Why would "None" ever be passed to ipset_exists? I probably need to talk to you because I am missing something.

fasaxc: Think I'd prefer passing in the name of the set to check for and caching in a dict to avoid that duplication.

fasaxc:
Good catch!  I was scratching  my head looking for another case where we'd try to clean up while creating a new one.  I think I'd been working on he assumption that we only manipulate things after starting the Actor.

I'd like to rejig the ref-counting logic to avoid creating the new actor until the old one is gone.  I think that'd make it simpler and avoid the potential for such bugs.  I'll add that to the design thoughts doc.

plwhite:
Having chatted to Cory, he persuaded me that caching the result of the ipset_exists call was overkill - so I have removed that, which makes the code a fair bit cleaner and simpler. Ready for merge (manual tests and UT are OK, FV is pending). Shaun, can you confirm and merge if happy.

fasaxc:
This looks good to me, I'm happy for it to be merged.

However, I'm also keen on the idea of making this code idempotent and branch free, if we can.  I've reworked my proposal to rely only on swap for the atomicity: #437. What do you think to that approach?

If we do merge this, we'll still want the other fix from my #437 pull req, which prevents cleanup from trying to cleanup chains that are stopping.

fasaxc:
(Sorry, would have liked to discuss this morning but we ran out of overlap and I didn't spot you'd updated this until after I tweaked my pull req.)

plwhite:
@fasaxc wanted to refactor this code in PR #437. Hence closing this PR down.

PR #447, opened 2015-04-23
Update bundle for etcd architecture

Lukasa:
This updates our sample bundle to use etcd, not calico-acl-manager.


ed-harrison:
LGTM (!)

PR #448, opened 2015-04-23
Initial code review markups for iptables updater.

plwhite:
I have made some comment changes directly in line - other comments marked with "PLW".

This is a first pass, having found nothing very exciting but a few cases where the logic was not clear.

### DIFF
--- a/calico/felix/fiptables.py
+++ b/calico/felix/fiptables.py
@@ -56,6 +56,7 @@ class IptablesUpdater(Actor):
     Concurrent access to the same table is not allowed by the
     underlying iptables architecture so there should be one instance of
     this class for each table.
+    PLW: per-table, or IP version?
 
     However, this class tries to be robust against concurrent access
     from outside the process by detecting and retrying such errors.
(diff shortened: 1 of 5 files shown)


calico/felix/fiptables.py
```diff
@@ -288,8 +306,11 @@ def _finish_msg_batch(self, batch, results):
 
     def _delete_best_effort(self, chains):
         """
-        Try to delete all the chains in the input list.
+        Try to delete all the chains in the input list. Any errors are silently
+        swallowed.
         """
+        # PLW: Do we really need the retry logic? It feels to me that we could
```
fasaxc: It allows us to keep making progress before we've completed our first cleanup (in case some chains are keeping something referenced that we're not expecting) or if someone makes a manual change that keeps one of our chains referenced by mistake.

I see it as analogous to using call_silent to delete ipsets except with the optimisation of deleting as many as possible in batches.

calico/felix/fiptables.py
```diff
@@ -383,6 +406,7 @@ def _calculate_ipt_delete_input(self, chains):
         if found_delete:
             return input_lines
         else:
+            #PLW: Nothing seems to catch this exception (unlike from modify)
```
fasaxc: I think we'll never get here due to the way batches are calculate.  That said, no harm in adding the except: block in the right place.

calico/felix/fiptables.py
```diff
@@ -608,8 +636,15 @@ def _stub_drop_rules(chain):
 def extract_unreffed_chains(raw_save_output):
     """
     Parses the output from iptables-save to extract the set of
-    unreferenced chains, which should be safe to delete.
+    unreferenced chains which are safe to delete.
     """
+    # PLW: This method is only called by load_unreferenced_chains.
```
fasaxc: Idea was to make this dead-easy to UT by making it a pure function.  That's the only reason it's split out.

fasaxc:
@plwhite I think I've addressed all your comments, back to you for another look.  You might want to look at the commit on its own to see your in-code comments side-by-side with my updates.

fasaxc:
Oh, the other reason for pulling out the UpdateBatch (now renamed _Transaction, which I think conveys its purpose much better) was to make it easier to UT.  It's not exactly a pure function but it's much easier to prod and poke than the IptablesUpdater itself, which has a lot of side-effects.

plwhite:
OK - your markups look good to me, so going to merge.

plwhite:
@fasaxc I had to revert this fix, as it broke the FVs. There was a typo in member names of the transaction (missing underscores for the updates field), and I decided to hold off to get some other fixes wrapped up today. Can you sanity check that this is all that was wrong and get it checked in?

PR #450, opened 2015-04-23
Use commit id instead of tag in tox dependency

matthewdupre:
For some reason using the tag seems to cause (intermittent?) failures in our
UTs.  It seems to be resolved by reverting to the commit id, so I'm just going
to do that rather wasting time investigating.


fasaxc:
Sounds fine.

Issue #451, opened 2015-04-23
Linux kernel treats ICMP type/code 255 as a wildcard


fasaxc:
Felix programs a rule like this, expecting it to match packets of type 255 and code 255 only:

```
-A felix-p-_5f77875ac4e22aa-i -p icmp -m icmp --icmp-type 255/255 -j RETURN
```

But iptables-save returns it as this:

```
-A felix-p-_5f77875ac4e22aa-i -p icmp -m icmp --icmp-type any -j RETURN
```

Assuming it is correctly reporting what the kernel is doing, this is broken.

I think that means we either need to whitelist only the valid ICMP types or switch to using a different approach as outlined in #424.

fasaxc:
@neiljerram  Bit annoying that Openstack allows arbitrary matching here but that's awkward to support in the kernel.  Is there any way to tell Openstack we can't support that combination?  If not, I guess we just have to figure out another way to implement this in felix (ideas in #424).

Lukasa:
@fasaxc As far as I know there's no way to report the limitation to OpenStack.

Given that it can be programmed elsewhere, I'd be interested to see what that security group rule turns into using a different plugin. I may be able to bully someone else in the org into testing for me.

fasaxc:
I'm erring towards the u32 approach described in #424.  It feels wrong but I think we should be able to isolate it and build a good FV for it.  It is one of the explicit examples for the u32 match extension.

neiljerram:
Is there any urgency here?  Note that ICMP type 255 isn't even a defined value yet (except as 'reserved').  I'll move this out to 1.0 for now; happy to move back if that's the wrong thing.

fasaxc:
I can't see this ever bubbling up to the top of the list give that ICMP 255 isn't a valid value.

PR #452, opened 2015-04-24
Add "icmp_code" to whitelist of allowed rule fields.

fasaxc:
Also beef up UTs and validation to check whitelist against rules.

This fixes the issue at hand but I'm not sure we can merge without addressing #451 in some way.

### DIFF
--- a/calico/felix/frules.py
+++ b/calico/felix/frules.py
@@ -53,6 +53,7 @@ KNOWN_RULE_KEYS = set([
     "dst_tag",
     "dst_ports",
     "icmp_type",
+    "icmp_code",
     "ip_version",
 ])
 
--- a/calico/felix/frules.py
+++ b/calico/felix/frules.py
@@ -302,6 +302,15 @@ def _rule_to_iptables_fragment(chain_name, rule, ip_version, tag_to_ipset,
 
     if rule.get("icmp_type") is not None:
         icmp_type = rule["icmp_type"]
+        if icmp_type == 255:
+            # Temporary work-around for this issue:
+            # https://github.com/Metaswitch/calico/issues/451
+            # This exception will be caught by the caller, which will replace
+            # this rule with a DROP rule.  That's arguably better than
+            # forbidding this case in the validation routine, which would
+            # replace the whole chain with a DROP.
+            _log.error("Kernel doesn't support matching on ICMP type 255.")
+            raise UnsupportedICMPType()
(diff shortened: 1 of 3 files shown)


Lukasa:
This is not assigned to a person or a release, which means it'll languish.

fasaxc:
Hmm, it was meant to be with me or Rob, I think.  Github was broken last night; losing comments and updates.  (I was also tired so may have messed up.)

fasaxc:
@robbrockbank found in manual testing that the ICMP type/code match of 255/255 is interpreted by the kernel as a wildcard so I've added a workaround to this fix.  (It's not safe to merge the fix without the workaround because it would open up a security hole.)

Resolving the kernel issue is tracked under #451.

fasaxc:
(Workaround is to log an error and generate an unconditional DROP rule if someone tries to use 255/255.)

calico/felix/frules.py
```diff
@@ -53,6 +53,7 @@
     "dst_tag",
     "dst_ports",
     "icmp_type",
+    "icmp_code",
```
fasaxc: Actual bug fix here.

calico/openstack/test/test_plugin_etcd.py
```diff
@@ -551,10 +551,10 @@ def assertNeutronToEtcd(self, neutron_rule, exp_etcd_rule):
 
         # Check felix is happy with generated rule.
         if neutron_rule["direction"] == "ingress":
-            rules = {"inbound_rules": [neutron_rule],
+            rules = {"inbound_rules": [etcd_rule],
```
fasaxc: UT bug: was validating wrong dict!

calico/felix/frules.py
```diff
@@ -199,16 +194,22 @@ def rule_to_iptables_fragments(chain_name, rule, ip_version, tag_to_ipset,
...
+            rule_copy["src_ports"] = src_ports
+            rule_copy["dst_ports"] = dst_ports
+            frag = _rule_to_iptables_fragment(chain_name, rule_copy, ip_version,
+                                              tag_to_ipset,  on_allow=on_allow,
+                                              on_deny=on_deny)
+            fragments.append(frag)
+        return fragments
+    except Exception:
+        # Defensive: isolate failures to parse the rule (which has already
+        # passed validation byt his point) to this chain.
```
robbrockbank: typo:
  byt his

robbrockbank:
LGTM

PR #457, opened 2015-04-24
Add missing jump target to ICMPv6 from endpoint rule.

fasaxc:
@neiljerram this should fix the ICMP issue you were seeing but I can't test it locally without an otherwise working IPv6 setup.  Do you have time to give it a spin?


neiljerram:
Yes, that's exactly what I thought too.  I'll merge as soon as the automatic tests pass.

PR #469, opened 2015-04-28
Tidy up line wrapping in RHEL install docs

alexwlchan:
Tiny fix: just some line wrapping and replacing a pair of hyphens and en dashes.

@ed-harrison: assigning to you since you wrote this warning originally.


neiljerram:
ok to test

PR #471, opened 2015-04-28
Dispatch chains review and test.

Lukasa:
This is the result of my code review and unit test of the `dispatch.py` module.

This is divided into a number of commits that are well isolated from each other, I recommend reviewing them individually rather than focussing on the diff as a whole.

The first commit pulls some utility functions out of `endpoints.py` into `frules.py`. This prevents an import cycle where `dispatch.py` needed to import `endpoints.py`, which needs to import `dispatch.py`: now both simply import `frules.py`. The functions that got moved were logically part of `frules.py` anyway.

The second commit tidies up `dispatch.py` a little bit: it's small and can easily be verified to have no functional effect. It takes advantage of the first commit to move an inline import up to the top of the module.

The third commit refactors the implementation and interface of `DispatchChains`. This class used a dictionary in its implementation, but never once accessed the values of that dictionary in order to do anything with them. As a result, I pulled that out and replaced it with a set. This obviously changed the interface as well, which no longer wants to deal with dictionaries or dict values, so I had to touch the bits of the code that touched the `DispatchChains` class.

Finally, I wrote UT for the end result.

### DIFF
--- a/calico/felix/dispatch.py
+++ b/calico/felix/dispatch.py
@@ -44,28 +44,26 @@ class DispatchChains(Actor):
         self.config = config
         self.ip_version = ip_version
         self.iptables_updater = iptables_updater
-        self.iface_to_ep_id = {}
+        self.ifaces = set()
         self._dirty = False
 
     @actor_message()
-    def apply_snapshot(self, iface_to_ep_id):
+    def apply_snapshot(self, ifaces):
         """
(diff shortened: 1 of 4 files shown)


plwhite:
Two comments added - one is really just a comment that you should consider and feel free to ignore if you like, and one needs addressing.

Issue #407, opened 2015-04-17
Make configuration composable from multiple sources


fasaxc:
Felix now expects the metadata address to be in etcd but the plugin isn't writing it there.

fasaxc:
A customer hit this and it's blocking them; I've offered them a patch that reinstates the old behaviour (loading from the config file) while we get this sorted.

fasaxc:
The customer says that, in their infrastructure, the metadata IP will be different on each node and it's not known by neutron.  I'm wondering whether we need to be flexible about where we get our configuration from in order to support such cases.  How about we use the same key names in the config file, etcd (global and per-host scopes), and our environment variables (for docker support) and inherit missing config from further up the hierarchy:

* environment variables trump all (only used in docker or manual execution so feels like these should win)
* config file trumps etcd config
* per-host etcd config trumps global etcd config
* global etcd config has lowest priority.

Lukasa:
I can get behind this principle, except that the relationship between per-host ETCD config and the config file seems wooly. There's no obvious stable ordering there. Any reason we shouldn't use something like confd to ensure that the two data sources are forcefully kept in sync?

neiljerram:
Some details that I don't understand here...

- Firstly, what precisely is the Metadata configuration requirement? In the absence of any configuration, I believe the default is localhost:8775. Is the customer's requirement here to have the metadata API running on just one node, and hence to require a specific IP instead of localhost? Does that IP need to change dynamically during the life of the cluster? 

- Secondly, I believe the architecture now is (a) for such configuration to be provided, where necessary, via the etcd data model (instead of felix.cfg); (b) to regard the etcd data model as a public API that Calico customers / integrators can write to, in addition to (say) the OpenStack plugin writing to it.

Hence, if a particular deployment requires metadata <addr> and <port>, I thought the architecture for that would be for that deployment to do

    etcdctl set /calico/v1/config/MetadataAddr <addr>
    etcdctl set /calico/v1/config/MetadataPort <port>

Is there something about that picture that is wrong, or that doesn’t meet the customer requirement? 

Certainly we could also add a plugin setting, to get it to set these values, but I don't think that's necessary. And I don't understand at all why it's being discussed to reintroduce felix.cfg settings.

neiljerram:
[...]

PR #472, opened 2015-04-28
Allow for config to be read from config files.

plwhite:
Logic is to take config from environment variables, config files, etcd per-host config, etcd global config in that order.

Resolves issue #407.

### DIFF
--- a/calico/felix/config.py
+++ b/calico/felix/config.py
@@ -77,7 +77,7 @@ class ConfigParameter(object):
     - Where the value was read from
     """
     def __init__(self, name, description, default,
-                 sources=DEFAULT_SOURCES, int=False):
+                 sources=DEFAULT_SOURCES, value_is_int=False):
         """
         Create a configuration parameter.
         :param str description: Description for logging
@@ -90,7 +90,7 @@ class ConfigParameter(object):
         self.sources = sources
         self.value = default
         self.active_source = None
-        self.value_is_int = int
+        self.value_is_int = value_is_int
 
     def set(self, value, source):
         """
@@ -108,6 +108,8 @@ class ConfigParameter(object):
             self.active_source = source
 
             if self.value_is_int:
+                # Set value before the call to int, so the ConfigException has
+                # the right value if / when it goes wrong.
                 self.value = value
                 try:
                     self.value = int(value)
@@ -134,6 +136,13 @@ class Config(object):
         - Configuration file
         - per-host etcd (/calico/vX/config)
         - global etcd (/calico/vX/host/<host>/config)
+
+        After object creation, the environment variables and config file have
+        been read, and the variables ETCD_ADDR and HOSTNAME have been set and
+        validated. The caller is then responsible for reading the remaining
+        config from etcd and calling report_etcd_config with the returned
+        values before the rest of the config structure can be used.
+
         :raises EtcdException
         """
         self.parameters = {}
@@ -144,10 +153,11 @@ class Config(object):
                            socket.gethostname(), sources = [ENV, FILE])
 
         self.add_parameter("StartupCleanupDelay", "Delay before cleanup starts",
-                           30, int = True)
+                           30, value_is_int=True)
         self.add_parameter("MetadataAddr", "Metadata IP address or hostname",
                            "127.0.0.1")
-        self.add_parameter("MetadataPort", "Metadata Port", 8775, int=True)
+        self.add_parameter("MetadataPort", "Metadata Port",
+                           8775, value_is_int=True)
         self.add_parameter("InterfacePrefix", "Interface name prefix", None)
         self.add_parameter("LogFilePath",
                            "Path to log file", "/var/log/calico/felix.log")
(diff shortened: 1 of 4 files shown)


calico/felix/config.py
```diff
@@ -46,166 +46,299 @@
...
+        :param list sources: List of valid sources to try
+        :param str default: Default value
+        :param bool int: Integer value?
+        """
+        self.description = description
+        self.name = name
+        self.sources = sources
+        self.value = default
+        self.active_source = None
+        self.value_is_int = int
```
Lukasa: Consider using a different name than `int`: it's generally considered bad form to shadow builtins, partly because it confuses syntax highlighters!

calico/felix/config.py
```diff
@@ -46,166 +46,299 @@
...
+            log.debug("Read value %r for %s (%s) from %r",
+                      value,
+                      self.name,
+                      self.description,
+                      source)
+
+            self.active_source = source
+
+            if self.value_is_int:
+                self.value = value
```
Lukasa: This double assignment looks weird, but reading the code makes it apparent that it's for the benefit of the `ConfigException`. A comment might be a good idea here.

calico/felix/config.py
```diff
@@ -46,166 +46,299 @@
...
+        :param host_dict: Dictionary of etcd parameters
+        :param global_dict: Dictionary of global parameters
+        :raises ConfigException
+        """
+        log.debug("Configuration reported from etcd")
+        for source in (LOCAL_ETCD, GLOBAL_ETCD):
+            if source == LOCAL_ETCD:
+                cfg_dict = host_dict
+            else:
+                cfg_dict = global_dict
```
Lukasa: This is a nasty antipattern. Instead:

`for source, cfg_dict in ((LOCAL_ETCD, host_dict), (GLOBAL_ETCD, global_dict)):`

calico/felix/config.py
```diff
@@ -46,166 +46,299 @@
...
+            for name, parameter in self.parameters.iteritems():
+                if source in parameter.sources and name in cfg_dict:
+                    parameter.set(cfg_dict.pop(name), source)
 
-        self._config_path = config_path
-        self._items = {}
+            self._warn_unused_cfg(cfg_dict, source)
 
-        self.read_cfg_file(config_path)
+        self._finish_update(final=True)
```
Lukasa: This function has a bad name and a bad docstring. 

- "Report back configuration parameters read from etcd": report them back to where?
- Can I run this function multiple times? It calls something called `_finish_update`: what does that have to do with anything?

calico/felix/config.py
```diff
@@ -46,166 +46,299 @@
...
 
-        self.ETCD_ADDR = self.get_cfg_entry("global", "EtcdAddr",
-                                            "localhost:4001")
+    def _validate_cfg(self, final=True):
+        """
+        Firewall that the config is not invalid.
+        :param stage: Which source we just loaded.
```
Lukasa: This parameter does not exist.

calico/felix/test/test_config.py
```diff
@@ -20,60 +20,26 @@
...
     def setUp(self):
-        # Stub out etcd.
-        etcd.Client = StubEtcd
-        etcd_results.clear()
+        pass
 
     def tearDown(self):
-        # Unstub etcd.
-        etcd.Client = real_client
+        pass
```
Lukasa: If these aren't doing anything, please remove them. =)

Lukasa:
retest this please

calico/felix/config.py
plwhite: Done

calico/felix/config.py
plwhite: Done

calico/felix/config.py
plwhite: Done

calico/felix/config.py
plwhite: Clarified.

calico/felix/config.py
plwhite: Doh.

plwhite:
This fix is ready to go, and has been tested. We're still waiting for a full FV run before merging (and that is blocked on an intermittent failure fixed in another PR, so may not be ready until tomorrow morning), but this fix is hopefully final.

calico/felix/config.py
```diff
@@ -46,166 +46,311 @@
...
+        self.parameters = {}
+
+        self.add_parameter("EtcdAddr", "Address and port for etcd",
+                           "localhost:4001", sources = [ENV, FILE])
+        self.add_parameter("FelixHostname", "Felix compute host hostname",
+                           socket.gethostname(), sources = [ENV, FILE])
+
+        self.add_parameter("StartupCleanupDelay", "Delay before cleanup starts",
+                           30, value_is_int=True)
+        self.add_parameter("MetadataAddr", "Metadata IP address or hostname",
```
fasaxc: Should we drop this default?  It's not correct for anything apart from OpenStack.  I think I added it as a workaround while we didn't have config at all.

fasaxc: Happy for that to be handled in a separate issue rather than blocking this work from going in.

calico/felix/config.py
```diff
@@ -46,166 +46,311 @@
...
+        if not final:
+            # Do not check that unset parameters are defaulted; we have more
+            # config to read.
+            return
+
+        for name, parameter in self.parameters.iteritems():
+            if parameter.value is None:
+                # No value, not even a default
+                raise ConfigException("Missing undefaulted value",
+                                      self.parameters["InterfacePrefix"])
```
fasaxc: Looks like the error is specific to InterfacePrefix even though it's in a generic loop?

calico/felix/config.py
```diff
@@ -46,166 +46,311 @@
...
+        - storing the parameters in the relevant fields in the structure
+        - validating the configuration is valid (for this stage in the process)
+        - updating logging parameters (if this is the final call)
+        :param final: Have we completed (rather than just read env and config file)
+        """
+        self.ETCD_ADDR = self.parameters["EtcdAddr"].value
+        self.HOSTNAME = self.parameters["FelixHostname"].value
+        self.STARTUP_CLEANUP_DELAY = self.parameters["StartupCleanupDelay"].value
+        self.METADATA_IP = self.parameters["MetadataAddr"].value
+        self.METADATA_PORT = self.parameters["MetadataPort"].value
```
fasaxc: Should this class handle converting `str` to and `int`?  Seems a bit odd to validate that, say the port looks like an int, but not to go the whole hog and store it as one.

calico/felix/config.py
plwhite: In practice yes, although we might well add more mandatory attributes later.

calico/felix/config.py
plwhite: It does - for integer values, the value is stored as an int (see the "set" method).

fasaxc:
FVs passed against this branch (apart from unrelated plugin exception).  Merging...

PR #473, opened 2015-04-28
Move common validation code to single place.

plwhite:
This hopefully stops the crazy etcdclusteridchanged import errors in openstack test code.


Lukasa:
LGTM.

Issue #329, opened 2015-04-08
Reinstate felix's ability to specify config file path


fasaxc:
We used to use argparse to accept a --config-file option but we lost that when we moved to python 2.6 (at least on the etcd branch).  We should reinstate that by using optparse (or decide it's not needed).

Lukasa:
This needs an assignee.

Lukasa:
This has been languishing, but is reasonably important, so I'm moving to 0.18.

neiljerram:
@plwhite Would you like to include this in your config work too?  It's not strictly the same, so I'd equally be happy not to do that (and then probably leave it until after 0.18).  What do you think?

plwhite:
Fix with Cory to review, so passing ownership to him.

PR #484, opened 2015-04-29
Reinstate ability to take file path as command line parameter.

plwhite:
Fixes bug #329.

### DIFF
--- a/calico/felix/felix.py
+++ b/calico/felix/felix.py
@@ -25,6 +25,7 @@ from gevent import monkey
 monkey.patch_all()
 
 import logging
+import optparse
 import os
 
 import gevent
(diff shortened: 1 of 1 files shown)


fasaxc:
I gave this a quick spin locally, only issue I could see was that --help gives an exception:
```
Usage: calico-felix [options]

Options:
  -h, --help            show this help message and exit
  -c CONFIG_FILE, --config-file=CONFIG_FILE
                        configuration file to use
2015-04-29 12:01:54,671 [ERROR][63369/1] calico.felix.felix 194: Felix exiting due to exception
  File "/home/gulfstream/git/calico/calico/felix/felix.py", line 182, in main
    logging.DEBUG)
  File "/home/gulfstream/git/calico/calico/common.py", line 227, in complete_logging
    logfile, when="D", backupCount=10
  File "/usr/lib/python2.6/logging/handlers.py", line 164, in __init__
    BaseRotatingHandler.__init__(self, filename, 'a', encoding, delay)
  File "/usr/lib/python2.6/logging/handlers.py", line 64, in __init__
    logging.FileHandler.__init__(self, filename, mode, encoding, delay)
  File "/usr/lib/python2.6/logging/__init__.py", line 827, in __init__
    StreamHandler.__init__(self, self._open())
  File "/usr/lib/python2.6/logging/__init__.py", line 846, in _open
    stream = open(self.baseFilename, self.mode)
IOError: [Errno 13] Permission denied: '/var/log/calico/felix.log'

```

Lukasa:
Shaun's problem is the perfect example of why you should never use a plain `except` statement.

Normally optparse would exit on `--help` or any other error. However, to do that is uses sys.exit, which raises a Python exception. If the prior author had used `except Exception` that would not have been caught, and instead you'd have dropped down to the `BaseException` block, which would have printed a thoroughly ugly traceback.

At the very least, remove the inner try..except block. Config files are non-optional, and their absence should not pass unremarked.

I also think this whole section needs re-evaluating now. Are we actually using real threads, such that sys.exit() won't work?

plwhite:
Config files are optional, but bad parameters should lead to exiting. And this is all before we have started any greenlets at all.

Hence I think the right thing to do is just to make "except" be "except Exception". Does that sound right to you? I obviously need to actually test that!

calico/felix/felix.py
```diff
@@ -159,29 +160,36 @@ def _main_greenlet(config):
...
-                pass
+            common.complete_logging("/var/log/calico/felix.log",
+                                    logging.DEBUG,
+                                    logging.DEBUG,
+                                    logging.DEBUG)
+        except Exception:
+            pass
 
-            raise
+        raise
```
Lukasa: What is this raise statement for? Raising the exception here will, in practice, kill the process, so having gone to all the effort of completing the logging setup is a waste of time, *especially* as the exception isn't logged out explicitly anyway.

calico/felix/felix.py
```diff
@@ -159,29 +160,36 @@ def _main_greenlet(config):
...
 
-        # Create configuration, reading defaults from file if it exists.
+    # Create configuration, reading defaults from file if it exists.
+    try:
+        parser = optparse.OptionParser()
+        parser.add_option('-c', '--config-file', dest='config_file',
+                          help="configuration file to use",
+                          default="/etc/calico/felix.cfg")
+        options, args = parser.parse_args()
+        config = Config(options.config_file)
```
Lukasa: I still think this `try` block is overbroad. Errors from optparse should not be caught, regardless of what they are, so let's restrict the `try` block to only the `config = Config(options.config_file)` line.

Lukasa:
Some more notes inline.

calico/felix/felix.py
plwhite: The idea is that the calling function which catches the exception logs it. Except that the calling function got refactored out of existence at some point. I'll add a log before the raise (that will indeed take the process down).

calico/felix/felix.py
plwhite: I think this is reasonable, because we want to catch the exception and log it. But as you pointed out above, that change got broken at some stage.

plwhite: Fair enough. I think it is reasonable that we just terminate with some exception on screen for bad command arguments (vs. bad config file, that will be logged).

Lukasa:
@plwhite You've assigned this back to me but haven't addressed any of the feedback (despite saying you were going to in at least one case). What's the story here?

plwhite:
@Lukasa I pushed some changes but failed to notice that the push failed. Repushed. Sorry about that.

Issue #422, opened 2015-04-20
Felix logging goes to stdout/err (at least until we load config)


fasaxc:
The default_logging() function installs a global StreamHandler that logs to stdout.  I think this is causing issues for @nbartos at Piston Cloud; he wants to be able to suppress all logging apart from syslog and, at least until we conenct to etcd, we're logging to stdout.

fasaxc:
Nick also found a case where gevent is logging directly to stderr, from it's hub.py.  It simply writes directly to stderr if a greenlet dies with an exception.

Lukasa:
Log levels should be configurable, they always used to be.

fasaxc:
I think that'd be good enough as long as we don't log anything to stdout/err in the mainline case where everything is working.

DaveLangridge:
On RHEL 7.1 all the felix logs are going to messages.

DaveLangridge:
Ignore the above. They're now going to /var/log/calico/felix.log as expected.

neiljerram:
@plwhite Can you look at covering this in your config file work?

plwhite:
My fix just means that if you have a config file turning down log levels, or use the env variable FELIX_LOGSEVERITYSCREEN to critical, then there is no logging to screen.

PR #485, opened 2015-04-29
Allow disabling of screen logging

plwhite:
Ensure that default logging levels are low enough that turning off logging through environment variables leads to no logs to screen unless things go very badly wrong.

Fixes bug #422 

Will need a rebase before merge, as based on issue #407 fix, which is not yet merged.


fasaxc:
FVs have passed against this branch (apart from unrelated plugin error).

Issue #455, opened 2015-04-24
RHEL 6.5 install issues.


DaveLangridge:
suid isn't available on RHEL 6.5. I removed it (do we need to run it as etcd?). Also, the directory give is wrong. My /etc/init/etcd.conf on control node is (also nee similar changes for compute):

```
[root@calico-rh04 ~]# cat /etc/init/etcd.conf
description "etcd"
author "etcd maintainers"

start on stopped rc RUNLEVEL=[2345]
stop on runlevel [!2345]

respawn

script

        env ETCD_DATA_DIR=/var/lib/etcd
        export ETCD_DATA_DIR

        exec /usr/local/bin/etcd --name  calico-rh04                                                        \
                           --advertise-client-urls "http://172.18.203.104:2379,http://172.18.203.104:4001" \
                           --listen-client-urls http://0.0.0.0:2379,http://0.0.0.0:4001              \
                           --listen-peer-urls http://0.0.0.0:2380                                    \
                           --initial-advertise-peer-urls "http://172.18.203.104:2380"                \
                           --initial-cluster-token calicorh6                                         \
                           --initial-cluster calico-rh04=http://172.18.203.104:2380                  \
                           --initial-cluster-state new
end script
```

DaveLangridge:
I can't install libffi-devel - its not available from the repos setup so far. I tried rpmforge ( https://gist.github.com/rderoldan1/5920539 ) with el5 replaced with el6 and it still failed.

Lukasa:
Ugh.

We don't *need* to run etcd as a non-privileged user, but it makes me sad to just run everything as root. At this point though RHEL 6.5 is kinda forcing my hand, so we can remove it.

The `libffi-devel` requirement is a weird one. We need it because python-etcd depends on PyOpenSSL, but we don't actually need PyOpenSSL here (because we aren't using SSL at any point). Which is kinda crappy.

DaveLangridge:
Solution for libffi-devel:
Enable Redhat repository Optional subchannel as follows:
•	Search /etc/yum.repos.d for the string ‘rhel-6-server-optional-rpms’
•	In the ‘rhel-6-server-optional-rpms’ section change the enable = value from 0 to 1
•	Run ‘yum list libffi-devel’ – to verify the package is now seen (investigate if not)
•	Run ‘yum install libffi-devel’

DaveLangridge:
The plugin is hitting exceptions:
[...]

PR #488, opened 2015-04-29
Retire RHEL 6.5 instructions until we can fix them up, or are convinced ...

ed-harrison:
...there is no demand.


ed-harrison:
Currently, we do not support RHEL 6.5 due to the issues highlighted in #455.  Consequently, we should retire the RHEL 6.5 instructions until we can either fix #455 or we decide that there is no demand for support on anything earlier than RHEL 7.

Lukasa:
\o/

The king is dead, long live the king! :sparkles:

Issue #466, opened 2015-04-28
Default BIRD filter is foot-shootingly broad


fasaxc:
Our default BIRD export filter is to export anything but 0.0.0.0/0.  That means that we export, for example, the routes assigned to the ns-* device that dnsmasq uses.

At the customer trial today, we got bitten by this: they were using a shared subnet for the IP pool and for VM addresses.  When we turned on BIRD, it redistributed the ns-* route and broke networking.

It'd be much safer, in OpenStack, to add a condition on interface name.  We used condition `ifname ~ "tap*"` to resolve the problem.

Lukasa:
Based on some email trails I've seen, @liljenstolpe is actually the one planning to take a swing at this.

liljenstolpe:
I will have a simple fix in today.  We need to think about a cleaner fix, but it probably needs new config-file gen script variables.  When do we want to to discuss that?

liljenstolpe:
Pull request issued to close

fasaxc:
@liljenstolpe have you given this any testing?

neiljerram:
Fixed by #491

PR #491, opened 2015-04-29
Issue466filter

liljenstolpe:
We are only export routes that are !default AND originate on one of the Calico virtual interfaces.

### DIFF
--- a/etc/bird/calico-bird.conf.template
+++ b/etc/bird/calico-bird.conf.template
@@ -1,8 +1,6 @@
 
 router id @MY_IP_ADDRESS@;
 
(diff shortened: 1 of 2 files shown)


Lukasa:
For #466.

This is a good start, but before we merge we should confirm that similar fixes have gone in to all our install steps and mechanisms that require them. I'm responsible for getting this into chef and Juju. @emmagordon, you'll want to be aware of this for Fuel.

lxpollitt:
Don't forget Calico Docker support too!

neiljerram:
Fixes #466

PR #493, opened 2015-04-30
Make felix restart graceful

### LINKED ISSUE #339: Don't stub out chains during "graceful restart" window
Currently, if the endpoints, dispatch chains and profile rules don't all feed their config into the IptablesUpdater before it does its first update then it'll stub out chains that may still be valid.  I propose that we add a grace period at start of day where we don't stub out any chain that already exists and instead we wait until the first call to cleanup() to stub out any chains we've not heard of yet.

fasaxc:
This needs the FVs to be run over it as well as some manual testing.

* Track the chains that are really in the dataplane.
* Don't flush or stub out chains that already exist until
  end of GR window.

Fixes #339.

### DIFF
--- a/calico/felix/fiptables.py
+++ b/calico/felix/fiptables.py
@@ -102,14 +102,26 @@ class IptablesUpdater(Actor):
         self._table = table
         if ip_version == 4:
             self._restore_cmd = "iptables-restore"
+            self._save_cmd = "iptables-save"
             self._iptables_cmd = "iptables"
         else:
             assert ip_version == 6
             self._restore_cmd = "ip6tables-restore"
+            self._save_cmd = "ip6tables-save"
             self._iptables_cmd = "ip6tables"
 
+        self._chains_in_dataplane = None
+        """
+        Set of chains that we know are actually in the dataplane.  Loaded
+        at start of day and then kept in sync.
+        """
+        self._grace_period_finished = False
+        """
+        Flag that is set after the graceful restart window is over.
+        """
+
         self._explicitly_prog_chains = set()
-        """Set of chains that have been explicitly programmed."""
+        """Set of chains that we've explicitly programmed."""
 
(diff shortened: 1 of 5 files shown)


ed-harrison:
What's the plan for doing the testing on this?  @fasaxc is this with you, or are you expecting some help from the Ocelot team?

plwhite:
See my comment in the issue - I think this is a workaround for the way the cleanup currently works, and it would be potentially cheaper and simpler to tweak the cleanup code so the issue goes away.

neiljerram:
AFAIK this isn't needed for 0.18, so deferring as we already have too much else to get in today.

fasaxc:
I think we can punt this to 0.19.  It's not as important as the other things that are going on.

fasaxc:
Oops, one of my new UTs depended on the print order of a dict.  Fixing up...

fasaxc:
Should be fixed now.

plwhite:
I hate to go back on my view that we should just accept this, but I've been a bit shaken by just how many lines of code it is for something so marginal (though much of it is tests). That doesn't mean we shouldn't take it, but it does mean that I need to spend more time reviewing it than I had expected, and have not had time yet.

fasaxc:
That'll teach me to add UT and extra comments :-P

ed-harrison:
Assigning this to @Lukasa as this has been languishing for a long time now and has already been deferred from two releases.  As @plwhite is OOO today, it makes sense for someone else to have a look so that we can hopefully get the code in and running through some tests before the 0.20 release.

Lukasa:
This is a complex review job and it's hard to do effectively, but there's nothing *obviously* wrong with this. I'm going to merge for now, but further feedback on this change is welcomed.

Lukasa:
Also merged to feature/neutron-ha

PR #494, opened 2015-04-30
Code review and tests for UpdateSplitter

Lukasa:
This pull request contains minor review markups (docstrings only) and unit tests for the `UpdateSplitter`.

Before merging this, I have more substantive review feedback (/cc @fasaxc) that should be considered.

The `UpdateSplitter` class knows way too much about the objects it contains. At its core, this class is nothing more than a message broker: given a message A, it routes it to all actors that handle message A. The problem is that it doesn't ask actors whether they can handle message A, it knows which actors can do it and what arguments they take.

This limits this class quite severely, and also manages to inflate the number of lines of code in the class. For example, right now there are four `on_*_update` methods on the class. Each method does nothing more than call the equivalently named method on the managers that have that method. There is no earthly reason that all four methods could not be replaced by a little bit of magic:

```python
def __getattr__(self, name):
    def proxy_function(self, *args, **kwargs):
        for manager in self.managers:
            if hasattr(manager, name):
                getattr(manager, name)(*args, **kwargs, async=True)

    setattr(self, name) = proxy_function
```

The only reason I can think of not to do this is because it harms introspection, but that's ok, because you can always replace them with simple wrapper methods that call `self._proxy_function` instead. Hell, you could even use a decorator if you wanted.

There are other patterns you could use: managers could register method names they want to handle with the UpdateSplitter, for example, which is more explicit but slightly more verbose. Or the code that creates the UpdateSplitter could do so. Or we could require that managers have a parameter that names all the messages they want to receive.

Any of these changes make this class substantially more general, which makes it more robust in the face of extension and re-use. Switching to this general architecture makes this class usable in many other roles.

The only hiccup in this plan is `apply_snapshot`. This is because each manager takes *different* arguments to its `apply_snapshot` method. This is nuts: it forces the UpdateSplitter to know about the arguments each actor needs to take, for no good reason. Each manager should be updated to have the exact same message signature for `apply_snapshot`, and it should simply ignore anything it doesn't need.

That leaves us with the other wart: cleanup. There are two problems here. The first is that cleanup, as written, works weirdly (and probably incorrectly). The second is that it's not clear that cleanup has a good design. I'll handle them in that order.

First, cleanup works weirdly. The cleanup process stops for a given type of manager when any manager of that type throws an exception. Why does an exception in the ip4tables manager prevent us cleaning up ip6tables? That feels wrong, and should be changed, unless there's something I'm not seeing.

The follow-on is that the whole architecture is odd. We schedule cleanup for ~30 seconds after a snapshot is applied. This is *obviously* racy: so obvious, in fact, that the comment above this block outright says so, and waves away concerns about dropping packets.

This feels crazy. If we want to cleanup, we should do it properly. Each manager should be capable of taking a cleanup message and acting on it, or it should automatically clean up after itself. Why can managers not clean up automatically after snapshots are applied? Ordering appears to be the concern here (we need to clean up iptables before ipsets), but again it feels like managers should be making repeated attempts to delete anything they fail to delete the first time *anyway*.

I think cleanup needs much more thought applied to it before we call ourselves happy.

For the moment, I see no reason not to merge these tests and markups: they've been written anyway, and they won't slow us down much if we rewrite this module. Nevertheless, we should think hard about the design here.

### DIFF
--- a/calico/felix/splitter.py
+++ b/calico/felix/splitter.py
@@ -28,6 +28,10 @@ _log = logging.getLogger(__name__)
 
 
 class UpdateSplitter(Actor):
+    """
+    Actor that takes the role of message broker, farming updates out to IPv4
+    and IPv6-specific actors.
+    """
     def __init__(self, config, ipsets_mgrs, rules_managers, endpoint_managers,
                  iptables_updaters):
         super(UpdateSplitter, self).__init__()
@@ -44,6 +48,12 @@ class UpdateSplitter(Actor):
         """
         Replaces the whole cache state with the input.  Applies deltas vs the
         current active state.
+
+        :param rules_by_prof_id: A mapping of security profile ID to profile
+            rules.
+        :param tags_by_prof_id: A mapping of security profile ID to profile
+            tags.
+        :param endpoints_by_id: A mapping of endpoint ID to endpoint data.
         """
         # Step 1: fire in data update events to the profile and tag managers
         # so they can build their indexes before we activate anything.
@@ -123,6 +133,9 @@ class UpdateSplitter(Actor):
 
     @actor_message()
     def on_interface_update(self, name):
+        """
+        Called when an interface state has changed.
+        """
         _log.info("Interface %s state changed", name)
         for endpoint_mgr in self.endpoint_mgrs:
             endpoint_mgr.on_interface_update(name, async=True)
--- a/calico/felix/splitter.py
+++ b/calico/felix/splitter.py
@@ -49,11 +49,11 @@ class UpdateSplitter(Actor):
         Replaces the whole cache state with the input.  Applies deltas vs the
         current active state.
 
-        :param rules_by_prof_id: A mapping of security profile ID to profile
-            rules.
-        :param tags_by_prof_id: A mapping of security profile ID to profile
-            tags.
-        :param endpoints_by_id: A mapping of endpoint ID to endpoint data.
+        :param rules_by_prof_id: A dict mapping security profile ID to a list
+            of profile rules, each of which is a dict.
+        :param tags_by_prof_id: A dict mapping security profile ID to a list of
+            profile tags.
+        :param endpoints_by_id: A dict mapping endpoint ID to endpoint data.
         """
         # Step 1: fire in data update events to the profile and tag managers
         # so they can build their indexes before we activate anything.
--- a/calico/felix/test/test_splitter.py
+++ b/calico/felix/test/test_splitter.py
@@ -1,0 +1,258 @@
+# -*- coding: utf-8 -*-
+# Copyright 2014, 2015 Metaswitch Networks
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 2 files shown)


calico/felix/splitter.py
```diff
@@ -44,6 +48,12 @@ def apply_snapshot(self, rules_by_prof_id, tags_by_prof_id,
         """
         Replaces the whole cache state with the input.  Applies deltas vs the
         current active state.
```
plwhite: Well done documenting what these are, but I'd like more.

* Is it conventional that mapping => dict? If so, fine. If not, please replace mapping with dict. 
* We should probably specify what a rule is here.

I'm thinking something like :
``:param rules_by_prof_id: A dict mapping security profile ID to a list of rules, each of which is a dict``

plwhite:
Happy to merge, but would like a trivial wording markup first.

On the more general point : 

* Agree that this has too much carnal knowledge, though I am less unhappy about that.
* I agree that cleanup feels wrong. I'm going to think more about that one, because that is what worries me most.

Lukasa:
@plwhite Addressed.

fasaxc:
It's not too obvious from the code (on my list of feedback from Matt to apply already!) but there's reasoning behind the ordering in apply_snapshot.  We do send the snapshots on asynchronously but sending them in the order that we do ensures that the new data arrives in the tag and rules managers before any increfs from the endpoints that are modified/created by the snapshot.  That isn't important for eventual correctness(!) but it does mean that we'll program down the right data first time rather than programming a endpoint chains with stubbed-out profile chains and then filling them in later.

I have nothing against being generic in principle but maintaining that ordering is sensible. I'm a tad wary of having a manager accept information it doesn't need on its apply_snapshot() call.  It makes it harder to reason about who needs what and who might have taken a copy of what.  Something to balance with the desire to be super generic?

In my GR pull req, I've removed the exception handling in cleanup.  The lower-layer managers now catch reasonable exceptions and any that get relayed back here should probably be fatal.

The 30 second timer: first lets define what we mean by a race here!  

* I think felix is eventually consistent; it will get the dataplane into the right state as per the snapshot, eventually.
* The race is between programming the dataplane from that snapshot and issuing the cleanup call.  If the dataplane hasn't been fully updated by the time we pull the cleanup lever we might clean up something that really does still exist in the snapshot but which hasn't yet been programmed into the dataplane.  Then, after the cleanup, we will program that something into the dataplane correctly.

The reason I don't think that's too crazy is that you can cast it a different way: felix has a 30 second graceful restart window.  If it manages to complete the resync in 30s then it won't drop packets.  If it doesn't then it will assume that the prior dataplane config is stale, remove it, and then continue programming the dataplane to get it in sync.  It wasn't a totally pragmatic decision to do this on a timer; we'd been talking to a customer the day before who was advocating for a hard limit on how stale config could be.

I don't think it'd too crazy to have a limit on how stale we let the dataplane programming get.  If we do have such a limit, we can hang the cleanup off it...

lxpollitt:
I suggest this one is best resolved with a live conversation rather than lots more comments back and forth here?  Remember the standing guidance that if the proposed change will not get us to 1.0 faster (and not doing the change now will not cause huge pain later in 1.x) then we shouldn't be doing it.  I don't know how that guidance evaluates in relation to the issues being discussed here, but please don't forget it.  :-)

plwhite:
I am going to merge this pull request. We still need to discuss the cleanup stuff a bit more, though I agree that would be better done in real time rather than through the medium of GitHub. My action to make that conversation happen.

PR #497, opened 2015-04-30
Felix Review comments and some UT

matthewdupre:
I've done some review and UT, principally on the actor and refcount classes.

My general feeling was that while the code looks pretty good, there's worryingly much of it - it wasn't always obvious to me that it was justified and not fixing a problem we don't have (yet).

### DIFF
--- a/calico/felix/actor.py
+++ b/calico/felix/actor.py
@@ -338,7 +338,6 @@ class Actor(object):
         pass
 
     def _maybe_yield(self):
-    # MD4: Premature optimization or really worth including?
         """
         With some probability, yields processing to another greenlet.
         (Utility method to be called from the actor's greenlet during
@@ -509,8 +508,6 @@ def actor_message(needs_own_batch=False):
             if (not on_same_greenlet and
                     not async and
                     _log.isEnabledFor(logging.DEBUG)):
-                    # MD4: I don't understand what the point of this
-                    # isEnabledFor check is.  Please remove or explain.
                 _log.debug("BLOCKING CALL: %s", calling_path)
 
             # OK, so build the message and put it on the queue.
--- a/calico/felix/actor.py
+++ b/calico/felix/actor.py
@@ -20,7 +20,7 @@ felix.actor
 A queue-based Actor framework that supports efficient handling of
 batches of messages.  Each Actor instance has its own greenlet
 and a queue of pending messages.  Messages are sent by making calls
-to a method decorated by the @actor_message decorator.
+to methods decorated by the @actor_message decorator.
 
 When an actor_message-decorated method is called from another greenlet
 the method call is wrapped up as a Message object and put on the
@@ -338,6 +338,7 @@ class Actor(object):
         pass
 
     def _maybe_yield(self):
+    # MD4: Premature optimization or really worth including?
         """
         With some probability, yields processing to another greenlet.
         (Utility method to be called from the actor's greenlet during
@@ -508,6 +509,8 @@ def actor_message(needs_own_batch=False):
             if (not on_same_greenlet and
                     not async and
                     _log.isEnabledFor(logging.DEBUG)):
+                    # MD4: I don't understand what the point of this
+                    # isEnabledFor check is.  Please remove or explain.
                 _log.debug("BLOCKING CALL: %s", calling_path)
 
             # OK, so build the message and put it on the queue.
--- a/calico/felix/endpoint.py
+++ b/calico/felix/endpoint.py
@@ -53,8 +53,6 @@ class EndpointManager(ReferenceManager):
 
         # All endpoint dicts that we know about.
         self.endpoints_by_id = {}
-        # MD4: Why do we need this?  There's already an objects_by_id (and it's
-        # even used in this file...)
         self.endpoint_id_by_iface_name = {}
 
         # Set of endpoints that are live on this host.  I.e. ones that we've
@@ -132,9 +130,6 @@ class EndpointManager(ReferenceManager):
             self.endpoint_id_by_iface_name[endpoint["name"]] = endpoint_id
 
         if endpoint and endpoint["host"] == self.config.HOSTNAME:
-        # MD4: Do we really need to do all the above work for non-local
-        # endpoints?  Since the config for hostname is always available, can't
-        # we move this check to the top?
             _log.debug("Endpoint is local, ensuring it is active.")
             if endpoint_id not in self.local_endpoint_ids:
                 # This will trigger _on_object_activated to pass the endpoint
--- a/calico/felix/endpoint.py
+++ b/calico/felix/endpoint.py
@@ -53,6 +53,8 @@ class EndpointManager(ReferenceManager):
 
         # All endpoint dicts that we know about.
         self.endpoints_by_id = {}
+        # MD4: Why do we need this?  There's already an objects_by_id (and it's
+        # even used in this file...)
         self.endpoint_id_by_iface_name = {}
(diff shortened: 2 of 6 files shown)


calico/felix/actor.py
```diff
@@ -338,6 +338,7 @@ def _finish_msg_batch(self, batch, results):
         pass
 
     def _maybe_yield(self):
+    # MD4: Premature optimization or really worth including?
```
fasaxc: I think adding this is just being a good citizen.  E.g. if we're doing a big apply_snapshot, we should yield the thread from time to time to allow etcd polling and iptables commands to run.

calico/felix/actor.py
```diff
@@ -508,6 +509,8 @@ def queue_fn(self, *args, **kwargs):
             if (not on_same_greenlet and
                     not async and
                     _log.isEnabledFor(logging.DEBUG)):
+                    # MD4: I don't understand what the point of this
```
fasaxc: Ah, this has been refactored (maybe by Peter?).  I think the calling path lookup used to be done after this guard but I see we're not using that for something else now so it's no longer needed.

calico/felix/endpoint.py
```diff
@@ -53,6 +53,8 @@ def __init__(self, config, ip_type,
 
         # All endpoint dicts that we know about.
         self.endpoints_by_id = {}
+        # MD4: Why do we need this?  There's already an objects_by_id (and it's
```
fasaxc: endpoints_by_id is full of endpoint dicts.  objects_by_id contains the active LocalEndpoint objects that have been ref-counted into existence.

calico/felix/endpoint.py
```diff
@@ -130,6 +132,9 @@ def on_endpoint_update(self, endpoint_id, endpoint):
             self.endpoint_id_by_iface_name[endpoint["name"]] = endpoint_id
 
         if endpoint and endpoint["host"] == self.config.HOSTNAME:
+        # MD4: Do we really need to do all the above work for non-local
+        # endpoints?  Since the config for hostname is always available, can't
+        # we move this check to the top?
```
fasaxc: We do need to store non-local endpoints and we need to handle endpoint == None (hence no host).  We should just pass the host in, I think we need to do that anyway for other reasons (it helps us handle any issues with duplicate endpoint IDs during a migration)

fasaxc: Actually, I'm wrong on that.  We don't need to store non-local endpoints any more.  We do need to do that in the tag management code but I can't see any reason to do it here.

calico/felix/fiptables.py
```diff
@@ -271,6 +271,8 @@ def _finish_msg_batch(self, batch, results):
                 _log.error("Non-retryable error from a combined batch, "
                            "splitting the batch to narrow down culprit.")
                 raise SplitBatchAndRetry()
+                # MD4: This introduces code to the actor and makes failure
```
fasaxc: Yes, I think it is if we want to combine multiple iptables requests into one batch (which gives us a massive perf improvement) and, at the same time, isolate iptables failures to the particular request that caused them.

I think the latter is important in order to isolate failures to a particular rule chain or endpoint.  I think that makes felix a lot more robust in the face of mistakes in our iptables rule generation and the like.

calico/felix/refcount.py
```diff
@@ -108,6 +116,9 @@ def decref(self, object_id):
         Return a reference and garbage-collect the backing actor if it is no
         longer referenced elsewhere.
         """
+        # MD4 RefHelper implies that this method can't be called while the object is
+        # starting - if this is true, it needs to be in this docstring, if not then
+        # RefHelper can be simplified a bunch.
```
fasaxc: It is valid to decref while an object is starting.  The reason that RefHelper doesn't do that is to simplify the code in RefHelper.  Without that, RefHelper needs to account for potential in-flight callbacks from  objects that it started but then stopped again before receiving the callbacks.

I wrote the code that way at first but it was ugly and Peter suggested the approach in RefHelper.

calico/felix/refcount.py
```diff
@@ -326,6 +342,7 @@ def _notify_ready(self):
         self._manager.on_object_startup_complete(self._id, self, async=True)
 
     def _notify_cleanup_complete(self):
+        # MD4 Please move this after on_unreferenced, to better reflect the lifecycle of a RefCountedActor
```
fasaxc: +1

calico/felix/refcount.py
```diff
@@ -203,6 +214,10 @@ def _is_starting_or_live(self, obj_id):
 class RefHelper(object):
     """
     Helper class for a client of a ReferenceManager.  Manages the
+    # MD4 I think the first sentence here is confusing - this is not a helper class for most
+    # clients of this class - only those that want to wait for creation of multiple
+    # RefCountedActors.  I'm also not entirely sold on its existence, since it only has one
+    # user.  I'd be tempted to move it into profilerules.py at least.
```
fasaxc: I had intended to use this in several places but we ended up with Endpoints no needing to actually get the reference to their profile.  I still think it's important and that it belongs here:

* It's important because this code is hard to get right.
* It belongs here because it knows about actor/ref count internals so it's more tightly coupled with those.  It's also completely generic and it's not that unlikely that LocalEndpoints will need to use one at some point.

calico/felix/refcount.py
```diff
@@ -97,6 +97,14 @@ def on_object_startup_complete(self, object_id, obj):
...
                       "in state %s", obj.ref_mgmt_state)
             return
+            # MD4 It's not obvious to me why this branch is needed - I suspect it's because of asynchronous completion
+            # (thanks to actors) of startup while the manager thinks it's being (e.g.) destroyed, but this definitely
+            # needs commenting at least.
+
+            # MD4 On a related note, it wasn't clear to me what the intended way of
+            # covering these sort of things is.  The use of call_via_cb in the test code is
+            # very opaque - is there a reason you used these rather than something equivalent
+            # to the N-BASE's TB_SCHEDULER / SCHEDULE_ONE()?
```
fasaxc: The first iteration of the code used AsyncResult as its primary means of dealing with result passing.  I think it's an artifact of that.  I don't think I can justify rewriting the UTs to use a different technique but thanks for the feedback nonetheless.

fasaxc: I think the ref counting stuff needed this more because the RefMgr actually starts other actors and I wanted to get some FV of that start process rather than fully dummy it out.  Elsewhere, I've used single-stepping.  You can create an actor and then run actor._step(), IIRC.

calico/felix/test/test_refcount.py
```diff
@@ -120,6 +193,10 @@ def on_object_cleanup_complete(self, *args, **kwargs):
         super(RefMgrForTesting, self).on_object_cleanup_complete(*args,
                                                                  **kwargs)
 
+    @actor_message()
```
fasaxc: This doesn't need to be an actor_message.  I've tidied up the docs under #500 to reflect that.

(It's safe to make it one but it'll get called from the Actor's greenlet so it'll bypass the queue anyway.)

fasaxc:
Addressed your comments inline and in #499 and #500.  UTs LGTM (I only had a quick skim though).

matthewdupre:
OK, that all sounds fine - I'll pull the comments out and merge the UTs in.

For the future, I think it might have been easier to re-review if you'd made the markups in this PR where possible (deleting the comments as they're made).  No need to change it now, just seems nicer that way.

PR #499, opened 2015-04-30
NEEDS FV Replace endpoint ID with tuple that includes host and workload too.

### LINKED ISSUE #439: Handle migrating endpoint keeping same UUID
felix assumes that an endpoint's ID belongs to a particular host.   However, during a migration, endpoints keep their ID while moving between hosts.

As long as the plugin deletes the old etcd endpoint before creating the new, this should work fine.  However, if the ordering gets inverted, felix may update its indexes incorrectly.  In particular, if felix sees:

* add host-A/endpoint-1
* add host-B/endpoint-1
* delete host-A/endpoint-1

then it will delete host-B/endpoint-1 from its index by mistake.

Suggest we fix up this case in felix, even if the plugin is behaving correctly.

fasaxc:
@matthewdupre this addresses the code review comment about not needing to track non-local endpoints in the EndpointManager.  (I needed to pass in the hostname to handle deletes but doing it this way fixes #439 too.)

Removes assumption that endpoint_id is globally unique, which doesn't hold during a VM migration.

Lays the groundwork for handling host and workload deletion without a resync.  (This needs further work.)

### DIFF
--- a/calico/datamodel_v1.py
+++ b/calico/datamodel_v1.py
@@ -24,6 +24,7 @@ made in a new copy of the file with revved version suffix.  That allows
 us to maintain multiple copies of the data model in parallel during
 migrations.
 """
+from collections import namedtuple
 import logging
 import re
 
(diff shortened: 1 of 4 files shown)


fasaxc:
Passed a kick-the-tires mainline-tests FV.

matthewdupre:
I'm going to merge this now; it'll run through the FV then.

Issue #489, opened 2015-04-29
KeyError in plugin while trying to create VMs during FV run


fasaxc:
The bobcat's FV rig reported this error:

```
2015-04-29 18:55:02,549 [ERROR][22365] calicotest.common.testutils 87: Hit unexpected exception in function - not retrying.
Traceback (most recent call last):
  File "/var/lib/jenkins/workspace/fv-tests/calicotest/common/testutils.py", line 84, in retry_until_success
    result = function()
  File "/var/lib/jenkins/workspace/fv-tests/calicotest/openstack/vm.py", line 259, in raise_unless_state
    raise VmErrorState("VM is in state 'ERROR' and must be "
VmErrorState: VM is in state 'ERROR' and must be deleted.
ERROR
2015-04-29 18:55:39,038 [INFO][22365] calicotest.openstack.topology 321: Running TestBasicNetworks.setUpClass
```

which coincided with this stack trace in the neutron server.log (note: the logs are in different timezones):

```
2015-04-29 10:55:00.954 1018 WARNING urllib3.connectionpool [-] HttpConnectionPool is full, discarding connection: localhost
2015-04-29 10:55:00.955 1018 ERROR neutron.plugins.ml2.managers [req-211ac354-831f-45cc-8b49-ef32c847a422 None] Mechanism driver 'calico' failed in create_port_postcommit
2015-04-29 10:55:00.955 1018 TRACE neutron.plugins.ml2.managers Traceback (most recent call last):
2015-04-29 10:55:00.955 1018 TRACE neutron.plugins.ml2.managers   File "/usr/lib/python2.7/dist-packages/neutron/plugins/ml2/managers.py", line 168, in _call_on_drivers
2015-04-29 10:55:00.955 1018 TRACE neutron.plugins.ml2.managers     getattr(driver.obj, method_name)(context)
2015-04-29 10:55:00.955 1018 TRACE neutron.plugins.ml2.managers   File "/usr/lib/python2.7/dist-packages/calico/openstack/mech_calico.py", line 139, in create_port_postcommit
2015-04-29 10:55:00.955 1018 TRACE neutron.plugins.ml2.managers     self.transport.endpoint_created(port)
2015-04-29 10:55:00.955 1018 TRACE neutron.plugins.ml2.managers   File "/usr/lib/python2.7/dist-packages/calico/openstack/t_etcd.py", line 294, in endpoint_created
[...]

Issue #465, opened 2015-04-28
Unable to delete a security group from an OpenStack guest VM


robbrockbank:
I'm performing some fast "toggling" of a security group from on a VM - whereby I'm quickly adding and removing a security group from a VM.  I'm using the NovaClient to do this.

At some point I hit an exception in the NovaClient attempting to remove the security group from the VM.  This error in the client coincides with exceptions raised in the neutron server logs in the calico plugin code.

Details of the exception raised in my test harness, and stack traces from the control nodes are shown below.

I can provide full trace of this - but it's quite large.  Just let me know.

2015-04-27 04:48:06,420 [ERROR][15539] calicotest.common.testutils 30: Test calicotest.openstack.test_security_groups.TestBasicSecurityGroups.test_add_remove_group terminated with exception
Traceback (most recent call last):
  File "/var/lib/jenkins/workspace/test-class/calicotest/common/testutils.py", line 22, in call_test_hooks
    test_method(self)
  File "/var/lib/jenkins/workspace/test-class/calicotest/openstack/test_security_groups.py", line 190, in test_add_remove_group
    self._add_remove_group(False)
  File "/var/lib/jenkins/workspace/test-class/calicotest/openstack/test_security_groups.py", line 272, in _add_remove_group
    remove_security_group()
  File "/var/lib/jenkins/workspace/test-class/calicotest/openstack/test_security_groups.py", line 224, in remove_security_group
    test_vm.remove_security_group(self.security_group.name)
  File "/var/lib/jenkins/workspace/test-class/calicotest/openstack/vm.py", line 194, in remove_security_group
    self.vm.remove_security_group(groupname)
  File "/var/lib/jenkins/workspace/test-class/env/local/lib/python2.7/site-packages/novaclient/v1_1/servers.py", line 368, in remove_security_group
    self.manager.remove_security_group(self, security_group)
  File "/var/lib/jenkins/workspace/test-class/env/local/lib/python2.7/site-packages/novaclient/v1_1/servers.py", line 1160, in remove_security_group
    self._action('removeSecurityGroup', server, {'name': security_group})
  File "/var/lib/jenkins/workspace/test-class/env/local/lib/python2.7/site-packages/novaclient/v1_1/servers.py", line 1240, in _action
    return self.api.client.post(url, body=body)
  File "/var/lib/jenkins/workspace/test-class/env/local/lib/python2.7/site-packages/novaclient/client.py", line 490, in post
[...]

PR #502, opened 2015-05-01
Retry if security groups are missing

Lukasa:
If for any reason we can't find a security group, we'll try up to five times across the course of one second to read the data. This is to make allowances for short data races where the security group information *is* coming, it just hasn't arrived yet.

This is another of the series of band-aid fixes that is part of #489, mostly aimed at #465.

### DIFF
--- a/calico/openstack/t_etcd.py
+++ b/calico/openstack/t_etcd.py
@@ -268,7 +268,27 @@ class CalicoTransportEtcd(CalicoTransport):
         inbound = []
         outbound = []
         for sgid in self.profile_tags(profile_id):
-            for rule in self.sgs[sgid]['security_group_rules']:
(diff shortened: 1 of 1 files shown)


fasaxc:
If it helps, felix is robust to rules, tags and endpoints being written in any order.

Lukasa:
> but does the scheduling allow self.sgs to be updated in that scenario

Yes, because this fix causes us to yield the flow of control. I believe we are not in possession of any relevant database transactions at this point in the execution, so I believe it's safe to yield here and the security groups can be updated.

It may be intelligent, if we time out, to schedule a resync in that scenario: that's a reasonable enhancement here.

neiljerram:
Changes here were merged in https://github.com/Metaswitch/calico/pull/508

PR #505, opened 2015-05-01
Plugin: provide correct workload ID - fixes #445

### LINKED ISSUE #445: Enhance plugin to set the etcd workload ID to the instance ID, not the endpoint ID
Following on from #386, we'd like to round out the plugin so that it writes the instance ID as the etcd workload ID. Currently it writes the endpoint ID, which leads to some weird redundancy in the key.

neiljerram:


ed-harrison:
Should there be some test code to go with this?

neiljerram:
Added proper UT now - please could you look again, and merge if OK?

ed-harrison:
LGTM... but, I'd like @fasaxc to confirm that he doesn't see any unintended consequences of this.

fasaxc:
Since this is a breaking change, I'm waiting for a response from a customer on whether it will impact their upgrade process.

fasaxc:
Customer got back to me and is happy for this to go in.

PR #506, opened 2015-05-01
doc: explain why someone would want to switch to an IP fabric

vincentbernat:
The current recommandation is to use an Ethernet fabric. The
documentation explains what are the limitations of an IP
fabric (commodity switches have to learn many routes and they may not be
able too) but doesn't highlight the limitations of an Ethernet fabric.

This commit adds a paragraph explaining why someone would want to use an
IP fabric. At some point, the vrouters may not scale. I am unsure if
Linux would be able to handle thousands of routes in the kernel
efficiently, so I don't stress too much on this but I mention it. Each
time you scale your network, all vrouters are also using more
CPU/memory. So, the main advantage of an IP fabric is that you can buy a
couple of high end routers as a spine and let the vRouters and the ToR
switches only learn local IP addresses. I am also adding a sentence
about anycasting which is a cool feature coming with an IP fabric. I've
stayed clear of Ethernet downsides as they are solved in the Ethernet
fabric by using seperate planes.

### DIFF
--- a/docs/source/l3-interconnectFabric.rst
+++ b/docs/source/l3-interconnectFabric.rst
@@ -519,7 +519,16 @@ installation that uses an IP fabric will be limited to the routing table
 size of its constituent network hardware, with a reasonable upper limit
 today of 128,000 endpoints.
 
+On the other hand, with an Ethernet frabric, all your vRouters have to
+handle the endpoints for the whole network (both in BGP and in the
+kernel). Switching to an IP fabric allows you to shift the burden to
+the North: if your spine is part of the IP fabric, each vRouter may
+just learn the routes to other endpoints in the rack. Only the spine
+has to know all endpoints and, as there are fewer of them, it may be
+less expensive to scale them appropriately.
 
+An IP fabric also enables some scenarios difficult to implement with
+an Ethernet fabric like anycast IP.
 
 .. [#endpoints]
    In Calico's terminology, an endpoint is an IP address and interface.


liljenstolpe:
Greetings Vincent,

One of the reasons that we support the Ethernet fabric as well as the IP fabric is because of route table growth.  As a general rule, a Linux host has MUCH more space for routing tables (basically infinite) when compared to a ToR.  Most ToR switches today support, at most 128k v4 routes, and 64k v6 routes.  There are LOTS of Linux boxes out there in the world that support multiple full internet routing tables (so Nx500K routes).  It's not uncommon to see servers with +1M routes on them.  The whole reason behind Calico is to put the state into more scalable platforms.  We are MUCH less concerned about blowing out the routing table on a Linux server than on a ToR (or spine for that matter).  Performance-wise, there doesn't seem to be operational impact, and the amount of memory consumed is small in comparison to the memory available on these servers.

As far as Anycast, we agree, but it's not just limited to the IP fabric.  We can do Anycast on the Ethernet fabric as well (remember, the Calico nodes are routers).  Anycast is a core feature of Calico, and isn't fabric dependent.

vincentbernat:
Didn't test, but to be sure, +1M routes in the routing table, not only in BIRD, right? Linux servers with a lot of routes are usually route servers and therefore, the routes never hit the kernel and the never see any bytes of traffic. I wouldn't expect Linux to perform enough to route 10G-like traffic without a huge performance impact due to route lookups, even with the "new" LC-trie algorithm. But I may be plain wrong as I've never tested that.

As for anycast routing, I didn't think it through and you are right.

vincentbernat:
Well, I should have tested a bit, the LC-trie compression is quite efficient if IPs are "dense". For example, with 65000 different IPv4 routes with only the last 2 bytes changing, my LC-trie is like this:

```
Basic info: size of leaf: 40 bytes, size of tnode: 40 bytes.
Main:
        Aver depth:     1.99
        Max depth:      2
        Leaves:         64519
        Prefixes:       64519
        Internal nodes: 2
          3: 1  16: 1
        Pointers: 65544
Null ptrs: 1024
Total size: 6561  kB
```

I don't know exactly how the lookup is working but the depth being minimal and the lookup at the next hop being likely to be done in an array, I think that the lookup should be fast.

If you have real tests about that, I would love to see them.

vincentbernat:
I close the PR since its content is quite incorrect.

PR #508, opened 2015-05-04
Fix plugin threading

fasaxc:
* Cory's retry fix in case we race between SG arrival and endpoint arrival.  (I think this may have been a red herring, I can't seem to hit that case now I have the other fixes in place.)
* The periodic resync thread was clobbering the `needed_profiles` and `sgs` fields on the transport.  I believe the assumption was that this was safe due to eventlet's co-op threading model but that's not true when we're doing IO to access etcd.  We should rework more substantially but I think this fix is correct (if not particularly architectural!)
* There was a race between loading the list of ports in an SG and a port being deleted, catch the PortNotFound exception and treat as if that port wasn't in the SG.

### DIFF
--- a/calico/openstack/t_etcd.py
+++ b/calico/openstack/t_etcd.py
@@ -283,7 +283,27 @@ class CalicoTransportEtcd(CalicoTransport):
         inbound = []
         outbound = []
         for sgid in self.profile_tags(profile_id):
-            for rule in self.sgs[sgid]['security_group_rules']:
+            # Be tolerant of a security group not being here. Allow up to 5
+            # attempts to get it, waiting a few hundred ms in between: we might
+            # just be racing slightly ahead of a security group update.
+            rules = None
+            retries = 5
+            while rules is None:
+                try:
+                    rules = self.sgs[sgid]['security_group_rules']
+                except KeyError:
+                    LOG.warning("Missing info for SG %s: waiting.", sgid)
+                    retries -= 1
+
+                    if not retries:
+                        LOG.error("Gave up waiting for SG %s", sgid)
+                        raise
+
+                    # Wait for 200ms
+                    eventlet.sleep(0.2)
+
+
+            for rule in rules:
                 LOG.info("Neutron rule  %s : %s", profile_id, rule)
                 etcd_rule = _neutron_rule_to_etcd_rule(rule)
                 if rule['direction'] == 'ingress':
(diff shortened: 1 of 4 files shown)


fasaxc:
@neiljerram or @Lukasa please can you pick up this review tomorrow?  I've had this running in my churn test for about an hour with no issues (old code lasted 30s).  It hasn't had a full, "normal", FV run against it.

To be clear, I think we should follow up with a more architectural fix but this was the most straightforward way I could think to get the code correct (at least in the non-HA case).

calico/openstack/t_etcd.py
```diff
@@ -67,19 +69,25 @@ def initialize(self):
...
-        # Spawn a green thread for periodically resynchronizing etcd against
-        # the OpenStack database.
-        eventlet.spawn(self.periodic_resync_thread)
-
         # We will remember the OpenStack security group data, between resyncs,
         # so that we can generate profiles when needed for new or updated
         # endpoints.  Start off with an empty set.
         self.sgs = {}
+        # Semaphore to protect all access to the sgs dict.
+        self._sgs_semaphore = Semaphore(1)
```
Lukasa: This semaphore is unnecessary.

eventlet is like gevent, it's a co-operative concurrency framework with implicit yielding. This means that a semaphore is only useful if it is unacceptable to see partial state of an object. That does not apply in this case.

There are only three accesses of the `sgs` dict in this code. One is a write into the dict in `security_group_updated`: this is a single write, so the flow of control will not be yielded and therefore it cannot conflict with anything.

Another is a read from the dict in `profile_rules`. This is protected by a retry, so it tolerates a partially complete dictionary, and therefore does not require the semaphore (it will yield if the dictionary element it's looking for is not present).

The last is the delete-and-write in `resync_security_groups`. The only reason to have a semaphore cover this operation is to avoid anyone seeing the partially populated (or empty) dictionary while we make DB queries, but given that neither of the other entry points is going to be perturbed by this it's hard to see why the semaphore is necessary here, also.

The only reason I can see for the semaphore is if we believe that a security group resynchronization is going to take more than 20 * 200ms (4 seconds). Do we believe that's a likely scenario?

Lukasa: Ah screw it, while it's not necessary it's not like it hurts more than the alternative. Let's be defensive and worry about it later.

ed-harrison:
Just for the sake of the cross reference going the other way, we believe this PR now closes #489.

PR #511, opened 2015-05-05
Neutron HA feature branch

Lukasa:
This branch acts as a target for all pull requests for the Neutron HA work. It also acts as a source for running unit tests and package builds. Finally, it will allow us to merge this extremely substantial change in one fell swoop once it's ready.

I will be regularly rebasing this branch on top of master.


lxpollitt:
I think the reference counting needs more discussion/thought.  The intention with the current data model is that profiles can exist independently as valid objects even if no workloads are currently referencing them.  Mandating they are automatically deleted when no workloads reference them might work for OpenStack scenario (where OpenStack is the master source of truth) but I don't think it works for some other orchestration scenarios (including some of the container orchestration scenarios) where etcd is the master data store.

Symmetric:
I think there's an atomicity issue here that needs to be fleshed out: we can atomically CAS the refcount up and down, but the profile delete isn't going to be atomic with the CAS to 0. What if another component tries to create a new reference to a profile with the same name before the previous delete has occurred?

I think we really want [multi-key transactions](https://github.com/coreos/etcd/issues/860) (etcd v3.0) for this. If we need to do this now, could consider a flattened JSON blob for profiles or an etcd lock on profile add/delete operations to serialize access.

As currently proposed this wouldn't negatively impact the Kubernetes plugin though, since there's one profile per pod, and only one process (kubelet) managing each pod. Probably need @spikecurtis to comment on how this would interact with Mesos integration.

Bottom line, since you create an empty profile on the command line in the vanilla docker demo, I think we need empty profiles to stick around. At the very least, can we make this profile GC controlled by a config flag?

Lukasa:
> Mandating they are automatically deleted when no workloads reference them might work for OpenStack scenario (where OpenStack is the master source of truth) but I don't think it works for some other orchestration scenarios (including some of the container orchestration scenarios) where etcd is the master data store.

It's a good thing this document doesn't mandate that behaviour then! :wink: There is nothing in this document that says that profiles need to be deleted when the profile's reference count drops to zero. In the OpenStack case such a behaviour would be ideal, though as @Symmetric points out we might have some trouble here.

> If we need to do this now

We can buy time by not deleting profiles at all, but at some point we do need to tackle this.

> What if another component tries to create a new reference to a profile with the same name before the previous delete has occurred?

Good spot, this won't work.

> etcd lock on profile add/delete operations to serialize access.

This is probably the direction I'd want to go at this stage. I'm desperately trying to keep the data model back-compatible with the current one, so switching the profile representation feels deeply unpleasant.

Of course, this is a problem in the currently existing data model: there's no safe way to delete security profiles at this moment.

Arguably a similar problem exists with endpoints, though their smaller valuespace means that this problem can be avoided by simply never creating an endpoint with the same workload and endpoint id as a previous one.

lxpollitt:
@lukasa - Perhaps I am misunderstand this PR then.  What is meant by "If a
+control-plane component compares-and-swaps this key down to ``0``, it must then
+issue an etcd delete of the profile directory"?  I had interpreted that as mandatory deletion of a profile when no workloads are referencing it.

Lukasa:
>  I had interpreted that as mandatory deletion of a profile when no workloads are referencing it.

Arg crap, I have a commit in my work machine that removes that clause, which I guess I didn't push. Now I look like an idiot. =D

Lukasa:
Ok folks, I've updated this to talk about our list of security profiles instead. The refcounting change is staying in place because the `neutron-changes` branch will remove it.

fasaxc:
I merged master in and fixed the UT conflict.  UTs now pass.  I didn't want to rebase and break dependent branches.

Lukasa:
Alright, the main chunk of neutron changes have been merged to this branch.

Lukasa:
I think our package build script doesn't like the name of this branch, I'll try to fix that up.

Lukasa:
retest this please

calico/openstack/mech_calico.py
```diff
@@ -120,146 +173,437 @@ def update_subnet_postcommit(self, context):
...
+        This method acts as a the periodic resynchronization logic for the
+        Calico mechanism driver.
+
+        On a fixed interval, it spins over the entire database and reconciles
+        it with etcd, ensuring that the etcd database and Neutron are in
+        synchronization with each other.
+        """
+        # The very first thing we do is sleep for our startup interval before
+        # initializing our state.
+        eventlet.sleep(STARTUP_DELAY_SECS)
```
fasaxc: @Lukasa @neiljerram I gave this feedback already on Neil's version of this patch but I don't think that sleeping is the answer.  IMO, We should either plumb through a kick from Neutron or do this lazily on the first call.

My reasoning is simple: sleeping is *never* the answer.  It always bites eventually.

Here, there's another issue that I'm worried about: having live threads when you fork is generally a bad idea because the threading libraries don't always handle forks properly.  IIRC, gevent does need to be re-initialised after a fork and eventlet is based on the same library underneath.  That wouldn't surprise me because libevent is based on polling on file handles, which are awkwardly shared over a fork.

calico/openstack/t_etcd.py
```diff
@@ -41,379 +48,99 @@
...
-        return {'inbound_rules': inbound, 'outbound_rules': outbound}
-
-    def profile_tags(self, profile_id):
-        return profile_id.split('_')
+        # Elector, for performing leader election.
+        # TODO: This doesn't handle forking yet, we need to mix the PID in here
+        # somewhere.
+        self.elector = Elector(
+            client=self.client,
+            server_id=cfg.CONF.calico.elector_name,
```
fasaxc: Surely this doesn't work if you have multiple API workers since they'll share config?  It needs to include the PID, at least (or just make it a UUID).

calico/datamodel_v1.py
```diff
@@ -44,6 +44,9 @@
 POLICY_DIR = VERSION_DIR + '/policy'
 PROFILE_DIR = POLICY_DIR + "/profile"
 
+# Key used for leader election by Neutron mechanism drivers.
+NEUTRON_ELECTION_KEY = VERSION_DIR + '/neutron_election'
```
fasaxc: Not sure this should go in the versioned directory, it's plugin-specific and felix won't be expecting the compare-and-set operations.  How about putting it under `/calico/openstack/...`

calico/openstack/t_etcd.py
fasaxc: Oh, there's a TODO to that effect just above.

calico/openstack/mech_calico.py
Lukasa: So the sleep logic here comes from the VMWare Neutron plugin, which does exactly the same thing: we even use the same duration as they do.

We discussed doing it lazily on the first call (and as you can see from the rest of the diff, it's easy to do), but we rejected it because we want to make sure that start-of-day sync happens even if no-one is actively triggering Neutron events. If we think that it's not important that that occurs, then we can remove the sleep and go back to lazy initialization.

Plumbing through a hook from Neutron is something I'd rather not do if we can avoid it, because making patches to Neutron ends up being a colossal pain in the ass from a long-term maintenance perspective. I do not really want to maintain a patched version of Neutron for all time if I can possibly avoid it.

calico/openstack/t_etcd.py
Lukasa: I need to remove the TODO, the election code actually does handle the PID already. =)

calico/datamodel_v1.py
Lukasa: Agree, and agree. I'll move it for now but leave it lower-case, I'll let you come up with some naming conventions! =D

Lukasa: Moved in 5bd530597a4ae7b6b8f622ffe7876232e1bcfc86

calico/openstack/t_etcd.py
Lukasa: Comment removed in bc45315d6ce62263999d996f3e97bc18cc758512.

BTW, the relevant line of code in the election code is [this one](https://github.com/Metaswitch/calico/blob/feature/neutron-ha/calico/election.py#L170-L172).

calico/openstack/mech_calico.py
Lukasa: I wasn't clear enough here: I'd like your input on whether we can abandon start of day sync. For the moment I've concluded we can't, so this is the approach that will stay in for the rest of today. If you think we can give up on start of day, let's re-evaluate. Feel free to call me later today to talk it over.

fasaxc: Ugh, you're right, we probably can't get away without start-of-day sync.  The ideal would probably be to combine our approaches.  The lazy one had some of benefits:

* if we get an API call, we initialise immediately
* even if it makes a mistake and initialises too soon, it detects a PID change later
* it made sure to tear down the DB and etcd connections after the fork

You could make sure my approach does start of day sync by doing this:

```
eventlet.spawn_later(30, self._maybe_init)
```
No big deal though, as long as you guys have thought this through and as long as it works!  I was more worried that no-one rebutted my feedback.

Lukasa:
@plwhite Your merge broke the UT in this branch, can you please fix it up?

Lukasa:
\o/ Merged! :sparkles: :cake: :sparkles: :cookie: :shipit: :fist: :+1:

lxpollitt:
\o/ indeed!

Issue #446, opened 2015-04-23
Plugin does not clear up etcd keys


Lukasa:
Sadly, the plugin appears to leave empty etcd keys around. The deployment below has only ever had 4 VMs on it, but it has a ton of redundant key information:

```
ubuntu@calico-vm15:~$ etcdctl ls /calico --recursive
/calico/v1
/calico/v1/host
/calico/v1/host/calico-vm13
/calico/v1/host/calico-vm13/workload
/calico/v1/host/calico-vm13/workload/openstack
/calico/v1/host/calico-vm13/workload/openstack/0a58d2ad-b356-49e2-85d0-6f68a60ebac5
/calico/v1/host/calico-vm13/workload/openstack/0a58d2ad-b356-49e2-85d0-6f68a60ebac5/endpoint
/calico/v1/host/calico-vm13/workload/openstack/0a58d2ad-b356-49e2-85d0-6f68a60ebac5/endpoint/0a58d2ad-b356-49e2-85d0-6f68a60ebac5
/calico/v1/host/calico-vm13/workload/openstack/137bb978-3bd1-449b-9b8f-9d2a5a1608e1
/calico/v1/host/calico-vm13/workload/openstack/137bb978-3bd1-449b-9b8f-9d2a5a1608e1/endpoint
/calico/v1/host/calico-vm13/workload/openstack/137bb978-3bd1-449b-9b8f-9d2a5a1608e1/endpoint/137bb978-3bd1-449b-9b8f-9d2a5a1608e1
/calico/v1/host/calico-vm13/workload/openstack/2072ec1f-3c2b-4ff2-ba92-cae054c626ac
/calico/v1/host/calico-vm13/workload/openstack/2072ec1f-3c2b-4ff2-ba92-cae054c626ac/endpoint
/calico/v1/host/calico-vm13/workload/openstack/aed8d0f5-07a1-4f8a-944a-dcd13ae1b838
/calico/v1/host/calico-vm13/workload/openstack/aed8d0f5-07a1-4f8a-944a-dcd13ae1b838/endpoint
/calico/v1/host/calico-vm13/workload/openstack/aed8d0f5-07a1-4f8a-944a-dcd13ae1b838/endpoint/aed8d0f5-07a1-4f8a-944a-dcd13ae1b838
/calico/v1/host/calico-vm15
/calico/v1/host/calico-vm15/workload
/calico/v1/host/calico-vm15/workload/openstack
[...]

PR #513, opened 2015-05-06
Clean up empty etcd directories, handle that in felix

fasaxc:
* In the plugin, attempt to delete parent directories using etcd parameter dir=true.  That does a non-recursive delete, only if the directory is empty.
* In felix, cache the endpoint IDs in the etcd layer so we can generate delete events for the children of a deleted directory.  I did it this way because it avoided polluting the rest of the code with knowledge of the etcd directory structure.  (It doesn't mean much to the endpoint/tag manager that the orchestrator dir was deleted.)
* The old way of parsing was getting untenable so I devised a tree structure to hang the handlers from.
* Bit annoying that I've still got the parse_if... functions hanging around.  I'd like to clean those up but this fix is too important to block behind that.

### DIFF
--- a/calico/felix/fetcd.py
+++ b/calico/felix/fetcd.py
@@ -475,8 +475,8 @@ class EtcdWatcher(Actor):
 
     def on_workload_delete(self, response, hostname, orchestrator,
                            workload_id):
-        _log.info("Workload dir %s/%s/%s deleted, removing contained hosts",
-                  hostname, orchestrator, workload_id)
+        _log.debug("Workload dir %s/%s/%s deleted, removing endpoints",
+                   hostname, orchestrator, workload_id)
         for endpoint_id in list(self.endpoint_ids_per_host[hostname]):
             if (endpoint_id.orchestrator == orchestrator and
                     endpoint_id.workload == workload_id):
--- a/calico/felix/fetcd.py
+++ b/calico/felix/fetcd.py
@@ -300,7 +300,7 @@ class EtcdWatcher(Actor):
         self.splitter.apply_snapshot(rules_by_id,
                                      tags_by_id,
                                      endpoints_by_id,
-                                     async=False)
+                                     async=True)
         # The etcd_index is the high-water-mark for the snapshot, record that
         # we want to poll starting at the next index.
         self.next_etcd_index = initial_dump.etcd_index + 1
@@ -420,7 +420,7 @@ class EtcdWatcher(Actor):
         _log.debug("Endpoint %s updated", combined_id)
         self.endpoint_ids_per_host[combined_id.host].add(combined_id)
         endpoint = parse_endpoint(self.config, endpoint_id, response.value)
-        self.splitter.on_endpoint_update(combined_id, endpoint)
+        self.splitter.on_endpoint_update(combined_id, endpoint, async=True)
 
     def on_endpoint_delete(self, response, hostname, orchestrator,
                            workload_id, endpoint_id):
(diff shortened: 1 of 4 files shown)


fasaxc:
Fixes #446.

fasaxc:
Passed the mainline FVs (calicotest.openstack.test_mainline_connectivity).  Please can someone (for example @plwhite) review this on Wednesday; it's required by a customer.

fasaxc:
I left this full FV run gong against this and my other patches from today: http://calicobobcat-jenkins.gulfstream:8080/job/fv-tests/773/

calico/felix/fetcd.py
```diff
@@ -60,6 +77,74 @@ def __init__(self, config):
...
+                                        "set": self.on_endpoint_set
+                                    })
+                                }
+                            })
+                        })
+                    }
+                })
+            }
+        }
+
```
Lukasa: Is Felix written in Javascript now?

All joking aside, this is an excellent example of where a dictionary is a *terrible* design and interface choice, not least because populating this dict is such an obvious exercise in madness. Additionally, this data structure is presented with exactly zero comments explaining how it's supposed to be structured, which means I have got genuinely no idea what this is, what it's for, and why it looks the way it looks.

Pull this whole mess out (along with `handle_update`), split it into a class that allows handler registration, and then give a vaguely sensible API for handling that kind of registration. Something like:

```python
self.dispatcher = EventDispatcher()
self.dispatcher.register('/Ready', set=self.on_ready, delete=self.force_resync)
self.dispatcher.register('/policy', delete=self.force_resync)
self.dispatcher.register('/policy/profile', delete=self.force_resync)
self.dispatcher.register('/policy/profile/<profile_id>/', delete=self.on_profile_delete)
self.dispatcher.register(/policy/profile/<profile_id>/tags', set=self.on_tags_set, delete=self.on_tags_delete)

self.dispatcher.dispatch(...)
```

This is definitely going to be more work than this dictionary, but it's a small price to pay for code that is debuggable.

Lukasa: Oh, this is required by a customer.

:disappointed:

Fine, I suppose we have to merge this as-is. Doing this better really needs to be a top priority though.

plwhite:
I'm afraid I couldn't understand the logic behind this code. From the point of view of Felix, we have several interesting deletions.

* Profiles, rules and endpoints were already handled in the code as is, as were rules and endpoints.
* We need to spot the ready flag being removed.
* Anything higher in the hierarchy can be ignored, because notifications arrive for every level in the hierarchy during a deletion (I think). That amounts to just removing the catch all "do a total resync when any higher level directory gets removed".

There's probably some subtlety I'm missing, but I can't see what. If it's just that we may only be told of top level changes then surely what we need to do is almost as trivial.

calico/felix/fetcd.py
fasaxc: Yeh, that's cleaner.  Maybe I was trying too hard to avoid duplicating the paths over and over.

fasaxc:
@plwhite Sadly etcd doesn't distinguish between a recursive and non-recursive delete in an event.  It's true that the plugin won't do recursive deletes (except for profiles) right now but felix needs to be defensive and assume the worst.

fasaxc:
Just pushed a work in progress so marking as DO NOT MERGE.

fasaxc:
I've manually tested this, deleting each dir in turn so we could merge it but it'd be good to add some UT of the fairly-complex Dispatcher code.

fasaxc:
Worth another review anyway.

neiljerram:
@plwhite Can I have your thoughts on this?  Is it close enough to mergeability to get into 0.19 today?

fasaxc:
I'm working on that UT.

fasaxc:
Added UTs for the PathDispatcher and EtcdWatcher event handlers.

Lukasa:
FYI, the cleaning up empty etcd directories fix is also present in the `neutron-changes` branch, in a different form.

plwhite:
I have put together another pull request, PR #537, which does much the same thing in a different way.

fasaxc:
Closing in favour of PR based on neutron-ha branch.

PR #514, opened 2015-05-06
Add script that automates the merging required for a Debian/Ubuntu package

neiljerram:


matthewdupre:
We're going to add this to the Neutron repository instead.

matthewdupre:
No obvious branch to use, just putting it here.

PR #518, opened 2015-05-08
Add support for list of profiles on endpoints.

fasaxc:
* Profiles are executed in order.  First accept or deny wins.
* The default if there is no match is to drop the packet.
* Single-profile case passed some basic FVs.

Caveats:

* Since we need to return from a chain rather than accept,
  and we need to distinguish explicit return the implicit
  end-of-chain return.  I have used the MARK action to
  distinguish the cases.
* The combination of anti-spoof and profile dispatch in the
  from-endpoint chains is now suboptimal (3nm rules, where
  we could do n+3m instead).
* The tag index is now 4 levels deep which would be a problem
  if we had an endpoint with many IPs, many profiles and
  many tags per profile.  This could be decomposed into
  multiple smaller indices.

### DIFF
--- a/calico/common.py
+++ b/calico/common.py
@@ -261,12 +261,25 @@ def validate_endpoint(config, endpoint):
     elif endpoint["state"] not in ("active", "inactive"):
         issues.append("Expected 'state' to be one of active/inactive.")
 
-    for field in ["name", "mac", "profile_id"]:
+    for field in ["name", "mac"]:
         if field not in endpoint:
             issues.append("Missing '%s' field." % field)
         elif not isinstance(endpoint[field], StringTypes):
             issues.append("Expected '%s' to be a string; got %r." %
                           (field, endpoint[field]))
+
+    if "profile_id" in endpoint:
+        if "profile_ids" not in endpoint:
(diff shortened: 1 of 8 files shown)


fasaxc:
@Lukasa I think you wanted this to test against tomorrow.  I haven't tested multiple profiles yet but WCGW?

fasaxc:
@Lukasa I've sweetened the deal with some UTs for IpsetManager.  Note: this PR is against your integration branch, not master so I'm expecting we'll give this a good FV before we merge the whole thing together onto master.

Lukasa:
This is with you for review this week @plwhite.

fasaxc:
@plwhite I've been ruminating on this over the weekend and I think it would be cleaner (and better performing) to reify the profiles so that the RulesManager's IDs are tuples containing the sequence of profile IDs.  However, I think that's time that would be better spent elsewhere so I can't justify making that change.

plwhite:
This hit dull merge errors. I therefore have manually merged it (cherry-pick) into feature/neutron-ha, and this can be closed.

Issue #520, opened 2015-05-08
Memory allocation exception gets swallowed.


DaveLangridge:
I was running mismatched versions of plugin and felix (accidentally). This caused a memory leak. Eventually memory allocation exceptions were raised, but handled. These continued. Probably should do something better here, like restart felix.

2015-05-07 21:34:55,701 [ERROR][3523/7399] calico.felix.actor 227: Exception processing 3f1bd9f99fef (on_endpoint_update)
Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/calico/felix/actor.py", line 225, in _step
    result = msg.method()
  File "/usr/lib/python2.7/site-packages/calico/felix/endpoint.py", line 238, in on_endpoint_update
    self._deconfigure_interface()
  File "/usr/lib/python2.7/site-packages/calico/felix/endpoint.py", line 376, in _deconfigure_interface
    devices.set_routes(self.ip_type, set(), self._iface_name, None)
  File "/usr/lib/python2.7/site-packages/calico/felix/devices.py", line 182, in set_routes
    current_ips = list_interface_ips(ip_type, interface)
  File "/usr/lib/python2.7/site-packages/calico/felix/devices.py", line 59, in list_interface_ips
    ["ip", "route", "list", "dev", interface]).stdout
  File "/usr/lib/python2.7/site-packages/calico/felix/futils.py", line 100, in check_call
    stderr=subprocess.PIPE)
  File "/usr/lib64/python2.7/site-packages/gevent/subprocess.py", line 238, in __init__
    errread, errwrite)
  File "/usr/lib64/python2.7/site-packages/gevent/subprocess.py", line 646, in _execute_child
    self.pid = fork()
OSError: [Errno 12] Cannot allocate memory

fasaxc:
That's odd, the exception should get put into the result object, which should either get handled or cause Felix to get killed later on if it doesn't get handled.

Lukasa:
It's coming from gevent on an OOM condition: I think at this stage you can assume that nothing sensible happens.

fasaxc:
Annoyingly, it's not even a BaseException because it's wrapping an errno.  Either way, Felix should die if such an exception is not handled; I'll have to dig into why that's not happening.

fasaxc:
I believe swallowing the error was caused by the same issue that was leaking memory in the first place: #523

neiljerram:
[...]

PR #523, opened 2015-05-08
Fix reference leak in felix caused by reference cycle.

fasaxc:
Stop attaching the message to the ExceptionTrackingRef beause it indirectly holds a reference to the TrackedAsyncResult.

### DIFF
--- a/calico/felix/actor.py
+++ b/calico/felix/actor.py
@@ -363,10 +363,6 @@ def wait_and_check(async_results):
         r.get()
 
 
-_refs = {}
-_ref_idx = 0
-
-
 class Message(object):
     """
     Message passed to an actor.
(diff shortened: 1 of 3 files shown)


fasaxc:
I believe this fixes #520.  I think it was both the reason we were leaking memory as well as the reason we didn't die as a result.

calico/felix/actor.py
```diff
@@ -505,10 +516,9 @@ def queue_fn(self, *args, **kwargs):
 
             # OK, so build the message and put it on the queue.
             partial = functools.partial(fn, self, *args, **kwargs)
-            result = TrackedAsyncResult(method_name)
+            result = TrackedAsyncResult((caller, self.name))
```
robbrockbank: method_name is no longer referenced ... can remove

calico/felix/actor.py
```diff
@@ -450,11 +455,17 @@ class TrackedAsyncResult(AsyncResult):
     """
     def __init__(self, tag):
         super(TrackedAsyncResult, self).__init__()
-        self.__ref = ExceptionTrackingRef(self, _reap_ref)
-        self.__ref.tag = tag
-
-    def set_msg(self, msg):
-        self.__ref.msg = msg
+        # Avoid keeping a reference to the week ref directly; look it up
```
robbrockbank: week -> weak

calico/felix/actor.py
```diff
@@ -450,11 +455,17 @@ class TrackedAsyncResult(AsyncResult):
...
+        # when needed.  Also, be careful not to attach any debugging
+        # information to the ref that could produce a reference cycle.  The
+        # tag should be something simple like a string or tuple.
+        tr = ExceptionTrackingRef(self, _reap_ref)
+        tr.tag = tag
+        self.__ref_idx = tr.idx
+
+    @property
+    def __ref(self):
+        return _refs[self.__ref_idx]
```
robbrockbank: Can this return None?

fasaxc: No, it can't, the ref is removed from the dict when the TrackedAsyncResult is GCed so there's no way to hit this line after the ref is removed from the dict.

Basically, this code is intended to call _reap_ref when the TrackedAsyncResult gets GCed.  It's almost equivalent to just using a __del__ method on the TrackedAsyncResult but, because it uses a weakref to get the callback, it doesn't mess up garbage collection.

The annoying part is that the weakref callback doesn't get called if the weakref dies before the TrackedAsyncResult does so we have to stash it in a strongly-referenced dict and then remove it after the TrackedAsyncResult gets reaped.  (We can't stash the weak ref on the TrackedAsyncResult because an objects fields get GCed before it does!)

calico/felix/actor.py
fasaxc: Hmm, think I'll add that into the tag.

calico/felix/actor.py
fasaxc: I'll add an explanation to the code.

robbrockbank:
LGTM

PR #524, opened 2015-05-11
Mock out gethostbyname so we don't have to wait for timeouts during UTs.

fasaxc:
The felix UTs used to take 60s to run, turned out it was really calling gethostbyname(), which has a long timeout.  Mock that out.


robbrockbank:
LGTM

PR #526, opened 2015-05-11
Add UT for iptables rules generation

fasaxc:
Also fixes port list splitting.  Previously, was splitting on 15 ports but asserting on <15 in the list.  Now synced up.


plwhite:
@fasaxc I'm concerned that we have an assert about number of ports in this code - that should be firewalled for when we first load the rule, not there (though no harm in firewalling it there to).

Otherwise, I'm happy just to merge.

plwhite:
I made a daft comment, then retracted it. Think we should just merge this now.

PR #529, opened 2015-05-12
Canonicalise IPs and MAC addresses at validation time.

### LINKED ISSUE #357: Felix should canonicalise IP addresses
I didn't use netaddr in the felix code for fear of exploding our occupancy.  However, that leaves a gap, particularly for IPv6, where we might treat different spellings of the same address as a different address.

We could do this somewhere between the etcd layer and the update splitter.

fasaxc:
* Add some basic UTs for the mainline validation cases too.

Fixes #357

### DIFF
--- a/calico/common.py
+++ b/calico/common.py
@@ -32,6 +32,7 @@ import netaddr.core
 import os
 import sys
 from types import StringTypes
+from netaddr.strategy import eui48
 
 _log = logging.getLogger(__name__)
 
@@ -106,6 +107,13 @@ def validate_ip_addr(addr, version):
         return False
 
 
+def canonicalise_ip(addr, version):
+    if addr is None:
+        return None
+    ip = netaddr.IPAddress(addr, version=version)
+    return intern(str(ip))
+
+
 def validate_cidr(cidr, version):
     """
     Validates that a CIDR is valid. Returns true if valid, false if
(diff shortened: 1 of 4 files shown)


fasaxc:
We may still want to pass IPAddress objects around but lets decouple that decision and get this fixed now.

calico/common.py
```diff
@@ -119,6 +127,29 @@ def validate_cidr(cidr, version):
...
+    return intern(str(nw))
+
+
+def validate_mac(mac):
+    try:
+        netaddr.EUI(mac, dialect=eui48.mac_unix)
+    except (netaddr.core.AddrFormatError, ValueError, TypeError):
+        return False
+    else:
+        return True
```
alexwlchan: Is there a reason we validate MAC address this way, rather than calling <code>netaddr.valid_mac(<em>mac</em>)</code>?

calico/datamodel_v1.py
```diff
@@ -44,22 +44,30 @@
...
+# endpoint ID.
+_ID_RE_FRAGMENT = r'[0-9a-zA-Z-_\.]+'
+# Validation regexes for profiles and endpoint IDs.
+GENERIC_ID_RE = re.compile(r'^' + _ID_RE_FRAGMENT + r'$')
+PROFILE_ID_RE = GENERIC_ID_RE
+ENDPOINT_ID_RE = GENERIC_ID_RE
+
 # Regex to match profile rules, capturing the profile ID in capture group
 # "profile_id".
 RULES_KEY_RE = re.compile(
```
plwhite: This means that if you have a tag that has an invalid character, instead of canonicalising it we'll just ignore it. That seems wrong to me - am I missing something?

plwhite:
I think we should be taking odd characters in the names and replacing them, rather than just ignoring them (which from a quick scan seems to be what happens here). Can you comment @fasaxc?

calico/common.py
fasaxc: Didn't know about that function.  Sounds ideal.

fasaxc:
The datamodel says which characters are valid in the etcd paths.  Agree it might bite us to just drop anything else on the floor though.  To be honest, I was trying to avoid needing to sanitise IDs in several places.

We need to make sure that any substitution we do avoids aliasing.  For example, if we have two profiles "a+b" and "a:b" and we sanitise them both to "a_b" then we'll alias.  

I'll have a think if there's a cleaner way.

fasaxc:
I think this is going to conflict with WiP on the HA branch.  I'll rebase it on top of that on Monday once it's been merged to master.

fasaxc:
I've put up #568 with the canonicalisation part of this and dropped the ID validation.  I think @plwhite has another PR in flight for that, which doesn't have the flaw of ignoring the bad data.

PR #530, opened 2015-05-12
Implement tree-based dispatch chains.

fasaxc:
Only support a root chain and a single layer below that based on the first character of the endpoint's suffix.

Low urgency but I'd like to have it in the back pocket in case our users start to but up against the perf issues that Dave was seeing.


fasaxc:
@DaveLangridge maybe you could give this branch a spin on your dataplane perf test.  It should reduce the impact of having lots of endpoints on one host.

plwhite:
@fasaxc : One bug. `calculate_updates` takes a set, but the tests provide it with a list to provide regressible output. I'd be inclined to fix that with a comment tweak, but it's so minor (and is effectively only in test code) that I don't think I care. Certainly doesn't block merging, so going to merge now.

PR #531, opened 2015-05-13
Clean up ProfileRules class, optimise and add UTs

fasaxc:
Customer was seeing churn on the profiles, it probably wasn't a felix issue but I thought I'd clean up and test the ProfileRules class anyway:

* Moved all the function to the finish_msg_batch function so that we can coalesce multiple updates in a batch.
* Track whether the profile is dirty.
* Add UTs that cover the ProfileRules class thoroughly.

### DIFF
--- a/calico/felix/actor.py
+++ b/calico/felix/actor.py
@@ -367,9 +367,9 @@ class Message(object):
     """
     Message passed to an actor.
     """
-    def __init__(self, method, results, caller_path, recipient,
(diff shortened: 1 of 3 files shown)


fasaxc:
Oh, looks like a bonus debugging change made it in too, reporting the result of any blocking calls at debug level only.

Lukasa:
Please keep remembering that this change should also be merged into `feature/neutron-ha` so that our merge gets easier. =)

plwhite:
LGTM.

Issue #515, opened 2015-05-06
Plugin can get VM config wrong after VM create failure. 


DaveLangridge:
Nova tries to create a VM on compute node A, and it fails (e.g. somone hasn't changed the qemu.conf file correctly at install), then nova tries another node - B. Nova now thinks the VM is on B (though it doesn't remove the tap interface on A), but felix thinks its on A, so routing etc is on A, not B. The etcd config is showing the endpoint on A, so this is probably a plugin problem.

fasaxc:
I think one our customers might be hitting this so I've triaged it into 0.19.

plwhite:
I had a quick think about this, and if the etcd configuration is incorrect (i.e. shows it on the wrong host) then Felix is blameless. Of course, it's possible that the endpoint is in some sense in both hosts, and we probably need to consider how Felix would handle that, but it's not clear that this is a high priority.

fasaxc:
Neutron definitely seems to think there are two copies of the port.  On the customer's system the ports had different types though.  Maybe we're misunderstanding an update from Neutron or maybe it's a Neutron bug.

Lukasa:
I think the mechanism driver logs out the port dict whenever a port update comes in, so it'd be interesting to see if the logs contain the two port dicts. That should help us tell the difference between the two.

neiljerram:
Sounds like more info needed, and I don't see any chance of this making 0.19, so deferring.

fasaxc:
Dave, does it make sense for you to repro this to get the port dict diags?

DaveLangridge:
I've just repro'd this. Logs are quite big. I could upload them, but probably better for you to look at calico-rh01. 
To repro, I edited the libvirt config on rh02 to ensure that libvirt would fail to create the VM. Then added a VM. The attempt to create it on rh02 fails, then it gets created on rh01. Looks like the plugin thinks its on rh02 still. 
ID is 3ee54ff4-a744-469a-8708-eb4b7fbd1ba2
I can find a Created port log at 13:28:36.714 for rh02. 
Then an Updated port at 13:28:36.844. dict still says rh02, and this is followed by a Migration part 1 log. Then another Updated port at 13:28:37.028 with Migration part 2. This still says 'calico-rh02'. So is this something we're interpreting incorrectly, or a neutron bug

Lukasa:
Interesting, the port binding is definitely saying calico-rh02 is the destination of the migration. Weird. I wonder what we missed here.

[...]

PR #536, opened 2015-05-14
Avoid excessive processing during endpoint creation.

Lukasa:
As noticed in #515, during endpoint creation events we can actually end up processing three events and doing three times the work we need to do. The flow is as follows:

1. create_port_postcommit is called, which writes state into etcd. At the end of its processing it updates the port status to ACTIVE, which causes...
2. update_port_postcommit is called, with the port binding VIF type set to unbound. This causes us to (mistakenly) interpret it as the first step of a migration, which causes us to *delete* the state we just wrote to etcd. Subsequently, the binding gets bound, at which point...
3. update_port_postcommit is called *again*, this time with the port bound, which we interpret as the second step of a migration, where we re-create the port.

In practice, the end result of this is actually right, but it causes us to take out two database transactions (including some heavy write locks on MySQL) where we could avoid the majority of them.

This change reduces the work we do, but it's a bit speculative and needs some testing. It's made up of a few observations:

1. If we're asked to create a port, we shouldn't bother with it if the port is unbound, as unbound ports cannot possibly expect connectivity.
2. If we're updating a port, and the only change in it is the port state (ACTIVE->DOWN or DOWN->ACTIVE), we don't want to bother doing anything (nothing we record has actually changed).
3. If we're updating a port we need to check whether the binding state has changed: if it has, we may need to delete it or create it.

I have a follow-on question: with this change, do we still need the non-Icehouse migration steps? I *suspect* that we actually can't hit them now (they'll be two separate `self._update_port` calls), in which case we can actually safely remove them entirely.

### DIFF
--- a/calico/openstack/mech_calico.py
+++ b/calico/openstack/mech_calico.py
@@ -249,7 +249,10 @@ class CalicoMechanismDriver(mech_agent.SimpleAgentMechanismDriverBase):
         # A migration only applies if the binding host ID has changed:
         # otherwise, this is a plain update.
         if original['binding:host_id'] == port['binding:host_id']:
-            return self._update_port(context, port)
+            status_change = port_status_change(port, original)
+            LOG.debug("Status change event: %s", status_change)
+            if not status_change:
+                return self._update_port(context, port)
 
         # This is a migration, handle the appropriate migration step.
         if port['binding:vif_type'] == 'unbound':
@@ -666,3 +669,26 @@ def profile_from_neutron_rules(profile_id, rules):
             outbound_rules.append(rule)
 
     return SecurityProfile(profile_id, inbound_rules, outbound_rules)
+
+
+def port_status_change(port, original):
+    """
+    Checks whether a port update is being called for a port status change
+    event.
+
+    Port activation events are triggered by our own action: if the only change
+    in the port dictionary is activation state, we don't want to do any
+    processing.
+    """
+    # Be defensive here: if Neutron is going to use these port dicts later we
+    # don't want to have taken away data they want. Take copies.
+    port = port.copy()
+    original = original.copy()
+
+    port.pop('status')
+    original.pop('status')
+
+    if port == original:
+        return True
+    else:
+        return False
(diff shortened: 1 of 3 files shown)


Lukasa:
@fasaxc @matthewdupre @neiljerram So, we need to decide whether or not this potentially destabilising change should land tomorrow or not. Your input would be valuable.

fasaxc:
Can we ignore "ACTIVE->DOWN" transitions; couldn't that be a manual "disabling" of the port?

calico/openstack/mech_calico.py
```diff
@@ -240,15 +246,37 @@ def update_port_postcommit(self, context):
...
+        if port_status_change(port, original):
+            LOG.info('Called for port status change, no action.')
+            return
+
+        # Now, re-read the port.
+        with context._plugin_context.session.begin(subtransactions=True):
+            port = self.db.get_port(context._plugin_context, port['id'])
+
+            # Now, fork execution based on the type of update we're performing.
+            # There are a few: first, a port becoming bound (binding vif_type
```
fasaxc: Use bullet points, please

Lukasa:
@fasaxc Pass. We'll test tomorrow: @DaveLangridge can you remember to check this?

fasaxc:
Seems like a pretty mainline case overall so if we can look out for any other ways that an endpoint can be disabled and explicitly test those then we should be able to give this a good workout.

We have migration FV tests too so that should get covered by an FV run.

fasaxc:
@Lukasa I tried to merge this with feature/neutron-ha.  It was a tricky merge and the UTs didn't pass afterwards.

I may have cocked up the merge but it looks like the UT updates the SG bindings and then sends an update_port_postcommit but your new check decides the port hasn't changed so does nothing.  (Maybe that was something you mentioned on the call earlier.)  I wasn't sure if the UT or your change was at fault.

Merge commit was af537fe to my smc-neutron-ha-pre-rel branch.  I was trying to build a branch that I could run the FVs against with  your changes and directory deletion.

Lukasa:
@fasaxc I'm quite confident that the UT here is wrong, if I do this today I'll merge by hand. Thanks for trying it though!

calico/openstack/mech_calico.py
Lukasa: Done, see 96b6744b7c5064138d9e785ca965787dc3b27879

Lukasa:
For the record, 2ede877706c624977c3b3fa776cb46f939b3f7e3 contains the necessary cleanup. The plugin tests are unfortunately prone to re-using some global state, so you kinda need to bend over backwards to avoid busting stuff. Separately, they just weren't emulating what Neutron does very well. Amusingly, like all good test failures this really would only happen in one or the other branch!

PR #537, opened 2015-05-14
Add capability to cope with deletion of any directory

plwhite:
This turned out to be much more complex than I expected, but was my attempt to rework PR#513. Note the following.

* The code uses endpoint_id to mean either the endpoint ID string or an EndpointId structure interchangeably. I fixed this (because I was literally so confused I was unable to make the fix otherwise).

* I have moved all of the understanding of the deletion hierarchy into the datamodel code, where it belongs.

* The deletion code got quite messy, so I pushed it into a separate subroutine in fetcd.

Net of this is that I think this is a good PR, but I now understand why PR#513 got so complex!

### DIFF
--- a/calico/felix/fetcd.py
+++ b/calico/felix/fetcd.py
@@ -18,6 +18,7 @@ felix.fetcd
 
 Etcd polling functions.
 """
+from collections import defaultdict
 from socket import timeout as SocketTimeout
 from etcd import (EtcdException, EtcdClusterIdChanged, EtcdKeyNotFound,
                   EtcdEventIndexCleared)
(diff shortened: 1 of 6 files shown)


plwhite:
Shaun, can you let me know what you think? I naturally prefer my version of this, mostly because I have tried to move the complicated logic into a methon in the datamodel code, but you should definitely have a look. I think that independently of this, fetcd is getting a bit ugly and could use a code tidy (and some tests!), but hopefully it isn't any worse.

Tests are a bit limited still (though I have manually checked that it works, and run the existing and new UTs).

fasaxc:
Either approach works.  I did implement (and UT!) that dispatcher code on the back of first-round code review from you and Cory.  Would have been nice if you'd put your heads together and given me consistent feedback if you thought both my approach and his suggestion was wrong.  (I thought I'd addressed your comment about why it was all needed.)

I was trying to get away from a big list of regexes but I don't feel too strongly about that; it's how Tornado and others dispatch URLs so it must work.

If we're doing it this way, might as well use a loop to do the matches:
```
for name, regex in [(DELETED_ENDPOINT, ENDPOINT_KEY_RE), ...]:
    m = regex.match(path)
    if m:
        captures = m.groupdict()
        captures["type"] = name
        return captures
```

If we had FooInfo named tuples for the other resources then we could do:
```
for info_type, regex in [(EndpointInfo, ENDPOINT_KEY_RE), ...]:
    m = regex.match(path)
    if m:
        return info_type(**m.groupdict())
```
and then fetcd.py wouldn't need to be creating EndpointInfo() objects at all.

I also liked the bit of re-work I did to split watch_etcd() and _wait_for_etcd_event().  Maybe you could reinstate that?  It was on the path to cleaning up and UTing fetcd.py.

calico/felix/fetcd.py
```diff
@@ -44,14 +56,6 @@
 
 RETRY_DELAY = 5
 
-# If we see an unhandled event (e.g. a directory deletion) for keys in any of
-# these prefixes, we'll abort our polling and resync.
-PREFIXES_TO_RESYNC_ON_CHANGE = [
```
fasaxc: There's still a reference to this in the code.

calico/felix/fetcd.py
```diff
@@ -275,17 +291,13 @@ def watch_etcd(self, update_splitter):
                                       response.modifiedIndex) + 1
 
                 if response.action == "delete":
```
fasaxc: My version of this change added support for compareAndDelete, expire and the like.  I wanted to add that in so we can remove the code that does a resync when we hit something unexpected.

fasaxc:
The change I mentioned above:
```
for info_type, regex in [(EndpointInfo, ENDPOINT_KEY_RE), ...]:
    m = regex.match(path)
    if m:
        return info_type(**m.groupdict())
```
would also let us decouple this change from being delete-specific.  We could just ask the datamodel to parse the path and tell us what type of *Info object it corresponds to.

fasaxc:
Cory, Ed and I discussed and decided to go with my more well-tested fix.  I expect the best possible fix is somewhere between our approaches!

PR #538, opened 2015-05-14
Force lookup of security group bindings

Lukasa:
This guarantees that we have up-to-date security group information. For some reason unbeknownst to me, the `get_port` call does not actually return up-to-date information. This change instead forcibly looks up the security group port binding table, which is actually accurate.


Lukasa:
Having had casual code-review from @neiljerram, I'm merging this into the feature branch.

PR #539, opened 2015-05-14
Handle odd timing window bug in etcd.

plwhite:


plwhite:
This can be merged, but we should hold off until we have thought about whether it is a good fix or not.

lxpollitt:
@plwhite - What is the actual etcd bug here?  I would like to understand any etcd issues we hit since etcd quality/readiness is something that comes up in some customer discussions.  We are developing a good relationship with the CoreOS etcd folks, so being on top of the etcd issues we encounter is useful for that too.

fasaxc:
Comment in the code:
```
# We only set _master when we have successfully written and
# rewritten. This is because testing appeared to reveal some
# odd timing windows where two writes with prevExist within 1ms
# can both succeed, so safer to find out right away.
```

Lukasa:
Note that we took this approach because repro is a bit tricky and because we're using a down-level etcd (2.0.0). It's possible this was fixed somewhere in between 2.0.0 and now, so we probably need to start sorting out our processes for updating etcd versions.

fasaxc:
I'll have a bash at reproing this and checking whether it's 2.0.0-specific.

calico/election.py
```diff
@@ -175,7 +175,7 @@ def _become_master(self):
             self._etcd_client.write(self._key,
                                     id_string,
                                     ttl=self._ttl,
-                                    prevExists=False,
+                                    prevExist=False,
```
fasaxc: No harm in being defensive but have you seen any issues with this fix in place?  It's certainly incorrect without this, python-etcd just swallows the incorrect argument and each worker clobbers the value :-(

I wrote a little repro script that hammers on etcd (2.0.0) with 100 worker processes (yay `multiprocessing`) and I can't seem to hit the window you mention.  It was all simple python, no eventlet and no Neutron forking.  If we do hit issues in Neutron but not in my test, it may be an artefact of the fork or the eventlet monkey-patch instead.

fasaxc: Stuck my script in a gist: https://gist.github.com/fasaxc/27f7322e5809b651855a

plwhite:
Agreed that etcd silently swallows arguments, which is very tedious.

prevExist replaced prevExists in a release a little while ago (after a bad checkin to etcd that did the opposite earlier!). I suspect that part of the change might be because I had an odd version of etcd on the test box, but we should use whatever is right for the version of etcd we have.

This is all totally untested by me - it's a speculative fix to show how I think we could fix this if we hit the problem, no more than that.

Lukasa:
I believe this is no longer required, as we addressed this problem at a higher level (and updated etcd).

PR #547, opened 2015-05-15
Handle etcd deletion-like actions in leader election.

fasaxc:
We were polling forever if the key expired.  Just happened to hit that case in my testing.


plwhite:
LGTM

PR #548, opened 2015-05-15
Handle directory deletion in Felix.

fasaxc:


fasaxc:
Merging after discussion with @Lukasa

PR #551, opened 2015-05-15
Start periodic resync thread from initialisation rather than vice versa.

### LINKED ISSUE #550: HA test: Ready flag persistently not set
Don't have much to go on yet apart from felix complaining that the Ready flag is missing.  Actively investigating on my test rig...

fasaxc:
* Resync thread was only running in the parent process.
* But Electors got started lazily in the children.
* If parent didn't win the election, periodic resync didn't
  happen.

This change:

* Starts the periodic resync lazily.
* Defensively checks the PID and re-initialises everything if
  it changes.
* Adds the ability to stop the Elector so we can recreate it.

Fixes #550

### DIFF
--- a/calico/openstack/mech_calico.py
+++ b/calico/openstack/mech_calico.py
@@ -25,6 +25,7 @@
 # TODO: Update reference to new etcd architecture document
 #
 # It is implemented as a Neutron/ML2 mechanism driver.
+import os
 import eventlet
 
 from collections import namedtuple
@@ -51,6 +52,8 @@ AGENT_TYPE_FELIX = 'Felix (Calico agent)'
 # The interval between period resyncs, in seconds.
 # TODO: Increase this to a longer interval for product code.
 RESYNC_INTERVAL_SECS = 60
+# When we're not the master, how often we check if we have become the master.
+MASTER_CHECK_INTERVAL_SECS = 5
 
 # We wait for a short period of time before we initialize our state to avoid
 # problems with Neutron forking.
@@ -74,9 +77,9 @@ def requires_state(f):
     from Neutron.
     """
     @wraps(f)
-    def wrapper(*args, **kwargs):
-        args[0]._init_state()  # args[0] will be 'self'.
-        return f(*args, **kwargs)
+    def wrapper(self, *args, **kwargs):
+        self._init_state()
+        return f(self, *args, **kwargs)
 
     return wrapper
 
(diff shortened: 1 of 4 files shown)


fasaxc:
@Lukasa @plwhite This PR fixes various issues I flushed out on the HA branch today.

To be clear, I've never hit the case where we call `_init_state()` and then get forked, that's just defensive (I cribbed it from my super-defensive tactical fix).  We could probably do without that bit (and remove the `stop()` methods on the transport and Elector that I added).  Then again, it does defend against Neutron changes that break our assumptions.

I wonder if we want to have some sort of liveness check between PR thread and Elector.  E.g. the PR thread could call a method to say "I'm still alive" and then the Elector could use a leaky bucket to decide if the PR thread has died.

We definitely don't want to be master without a healthy PR thread, though.

Probably best if you guys just take this over and run with it on Monday rather than passing it back for markups.

calico/openstack/mech_calico.py
```diff
@@ -100,26 +103,52 @@ def __init__(self):
...
     def _init_state(self):
         """
         Creates the connection state required for talking to the Neutron DB
         and to etcd. This is a no-op if it has been executed before.
         """
-        if self._state_initialized:
+        current_pid = os.getpid()
+        if self._my_pid == current_pid:
+            # We've initialised our PID and it hasn't changed since last time,
+            # nothing to do.
```
Lukasa: Do we really want to do this on *every* call? It's not clear that we do if we can possibly avoid it.

calico/openstack/mech_calico.py
```diff
@@ -468,33 +497,41 @@ def periodic_resync_thread(self):
...
+                    # the master.  Avoids waiting a whole RESYNC_INTERVAL_SECS
+                    # if we just miss the master update.
+                    eventlet.sleep(MASTER_CHECK_INTERVAL_SECS)
+        except:
+            # TODO Should we tear down the process.
+            LOG.exception("Periodic resync thread died!")
+            if self.transport:
+                # Stop the transport so that we give up the mastership.
+                self.transport.stop()
+            raise
```
Lukasa: What is this except block actually trying to catch? Exceptions could be thrown out of `get_admin_context()` I suppose, but it could be brought into the `try..except` above. Otherwise, the only place it could bubble up is from `eventlet.sleep` in the else block, which I think is unlikely to raise exceptions!

Lukasa:
I've got a few small notes here. If you get to them this weekend, great, if not I'll probably just merge this by hand and make the markups myself. =)

calico/openstack/mech_calico.py
fasaxc: Makes sure we try to log any BaseExceptions even if we don't want to catch them.  I've found it useful to do this sort of log around any worker thread in the past.  I think eventlet logs such failures but they go to stdout/err so we don't see them?

calico/openstack/mech_calico.py
fasaxc: It is a bit paranoid but I can't see a way it can go wrong, even if Neutron gets changed quite a bit. Even if our startup timer magically pops before the fork or Neutron completely changes its process model, we'll still catch it.

I only measure a 9ns penalty for the `getpid()` call vs field access; in the noise compared with DB access. 
```
$ python -m timeit -n 1000000 -s 'o = lambda x: x; o.foo=False' -- 'if o.foo: pass'
1000000 loops, best of 3: 0.0497 usec per loop
$ python -m timeit -n 1000000 -s 'from os import getpid' -- 'if getpid() == 1: pass'
1000000 loops, best of 3: 0.0506 usec per loop
```

calico/openstack/mech_calico.py
Lukasa: I suppose I can see the value.

calico/openstack/mech_calico.py
Lukasa: Yeah, the delta over field access isn't bad, but that wasn't really my question: my question was whether we feel the need to do the PID check every time. I guess I can see the value, though we should remember that we're costing ourselves CPU time each time we do it. Yeah, the database is way more expensive, but we have far less control over the DB than we do over our own code.

fasaxc: It is a bit paranoid, the main change here is to initialise the PR thread from the lazy init (which was already being called on every API call) rather than the other way around.  We still don't get a PR thread in every child until it receives an API call but the parent always starts one.  I can't see a way around that unless we can get a post-fork kick.

I think the cost of field access vs getpid is negligible.  If we want to remove the overhead of the lazy-init decorator we'd have to have it patch itself out after it's been called.  I.e. replace the entry in `__dict__` with the rawmethod without decorator.   I expect that's not worth it.

Hmm, maybe there's a generic python/syscall API for registering a post-fork callback?  Then we wouldn't have to get a kick plumbed through Neutron.  Or, we could set up our own IPC channel of some sort before the fork so that parent can prod children.

Lukasa:
Merged manually.

Issue #541, opened 2015-05-15
Felix exception during ipsets cleanup


DaveLangridge:
During churn test on neutron-ha branch

2015-05-14 14:46:01,457 [ERROR][21795/6] calico.felix.splitter 110: ipsets cleanup failed, will retry on resync.
Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/calico/felix/splitter.py", line 108, in trigger_cleanup
    ipset_mgr.cleanup(async=False)
  File "/usr/lib/python2.7/dist-packages/calico/felix/actor.py", line 450, in queue_fn
    return result.get()
  File "/usr/lib/python2.7/dist-packages/calico/felix/actor.py", line 543, in get
    timeout=timeout)
  File "/usr/lib/python2.7/dist-packages/gevent/event.py", line 235, in get
    raise self._exception
AttributeError: 'set' object has no attribute 'owned_ipset_names'

fasaxc:
@DaveLangridge Do you have the whole log for this?  It's probably bretty easy to track down from the owned_ipset_names clue but this is a re-raised error so the stack trace isn't so useful.  there should be another stack trace not far above with same message but with the correct stack.

DaveLangridge:
Is this what you want?

2015-05-14 14:46:01,171 [INFO][21795/878] calico.felix.endpoint 207: Endpoint<IPv6,id=EndpointId<calico-vm15/openstack/4725d075-e899-4232-83b1-636f4a20908f/c110ea3f-3e84-4820-a81a-584402d029ce>,iface=tapc110ea3f-3e> updated: {'ipv4_gateway': u'10.28.0.1', 'name': u'tapc110ea3f-3e', 'state': u'active', 'ipv6_nets': [u'2001:db8:c41:2::1f1c/128'], 'mac': u'fa:16:3e:b0:f2:30', 'ipv4_nets': [u'10.28.31.88/32'], 'ipv6_gateway': u'2001:db8:c41:2::1', 'profile_ids': [u'08c1346b-09d4-468e-a077-0812a7bb8eac']}
2015-05-14 14:46:01,452 [ERROR][21795/10] calico.felix.actor 227: Exception processing 0ef0a097cdb3 (cleanup)
Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/calico/felix/actor.py", line 225, in _step
    result = msg.method()
[...]

PR #552, opened 2015-05-18
Failure caused by use of stopping objects as if it was an object, not…

plwhite:
… a set.

stopping_objects_by_id is a dict whose values are sets, so treat as such.

### DIFF
--- a/calico/felix/ipsets.py
+++ b/calico/felix/ipsets.py
@@ -147,11 +147,15 @@ class IpsetManager(ReferenceManager):
         tmppfx = IPSET_TMP_PREFIX[self.ip_type]
         felix_ipsets = set([n for n in all_ipsets if n.startswith(pfx) or
                                                      n.startswith(tmppfx)])
+
+        # Ask the ipsets for all the names they may use and whitelist.
         whitelist = set()
-        for ipset in (self.objects_by_id.values() +
-                      self.stopping_objects_by_id.values()):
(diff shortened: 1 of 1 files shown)


plwhite:
Simple fix, but really worth holding off until there are some tests for this code (and after feature/neutron-ha goes into master, at which point there may well be tests to extend).

plwhite:
Fixes bug #541.

calico/felix/ipsets.py
```diff
@@ -147,11 +147,15 @@ def cleanup(self):
...
                                                      n.startswith(tmppfx)])
+
+        # Ask the ipsets for all the names they may use and whitelist.
         whitelist = set()
-        for ipset in (self.objects_by_id.values() +
-                      self.stopping_objects_by_id.values()):
-            # Ask the ipset for all the names it may use and whitelist.
+        for ipset in (self.objects_by_id.values()):
+            whitelist.update(ipset.owned_ipset_names())
+        for ipset in [ipsets for ipsets in
```
fasaxc: Isn't `[ii for ii in foo]` just an identity mapping?

Why don't I write some UT for the cleanup code and fix this as part of it?

fasaxc: Never mind, brain fart on my part.  You're using update() on the line below.

fasaxc:
I've put up a version with UT, which was on me anyway: #567

Issue #460, opened 2015-04-25
Felix should validate profile/endpoint/tag IDs


fasaxc:
The data model schema defines the valid range of characters but felix doesn't seem to police them.  It should police that only characters that are valid in the data model are used.

Not catastrophic since it is an internal interface but better safe than sorry.

Assigning to Peter as he did a lot of the validation routines (but anyone could take this on).

plwhite:
Actually, I think the right approach here is that for IDs containing any invalid characters, we should just hash them using the uniquely_shorten function (which would then need to be renamed).

lxpollitt:
I vote for just not allowing invalid characters.  Anything else feels like solving a problem we don't yet have.

PR #553, opened 2015-05-18
Test that characters are valid in tags, rules and enpoints.

plwhite:


plwhite:
No UTs (and I think breaks existing UTs) so absolutely do not merge yet. Tidy up on the way.

plwhite:
Fixes bug #460.

plwhite:
Moving to master.

PR #554, opened 2015-05-18
Initial documentation for 'bird as RR'

Lukasa:
This contains initial documentation for configuring BIRD as a route reflector.


docs/source/bird-rr-config.rst
```diff
@@ -0,0 +1,149 @@
...
+Step 3 (Optional): Set your BIRD IPv6 configuration
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+If you want to use IPv6 connectivity, you'll need to repeat step 2 but using
+``/etc/bird/bird6.conf``. The *only* difference between the two is that, where
+before you set ``<node_ip>`` to the compute node's IPv4 address, this time you
+need to set it to the compute node's IPv6 address.
+
+Note that ``<router_id>`` should still be set to the route reflector's IPv4
+address: you cannot use an IPv6 address in that field.#
```
alexwlchan: Trailing hash should be removed.

docs/source/bird-rr-config.rst
```diff
@@ -0,0 +1,149 @@
...
+Ubuntu 14.04
+^^^^^^^^^^^^
+
+Restart BIRD::
+
+    sudo service bird restart
+
+Optionally, if you configured IPv6 in step 3, also restart BIRD6:
+
+    sudo service bird6 restart
```
alexwlchan: There should be a second colon after BIRD6; right now the `sudo service bird6 restart` line doesn't show up in a code block. (At least, in my local docsbuild.)

docs/source/bird-rr-config.rst
```diff
@@ -0,0 +1,149 @@
...
+
+Restart BIRD::
+
+    systemctl restart bird
+    systemctl enable bird
+
+Optionally, if you configured IPv6 in step 3, also restart BIRD6:
+
+    systemctl restart bird6
+    systemctl enable bird6
```
alexwlchan: Ditto above; missing second colon after BIRD6.

docs/source/bird-rr-config.rst
```diff
@@ -0,0 +1,149 @@
...
+
+RHEL 7
+^^^^^^
+
+First, install EPEL. Depending on your system, the following command may be
+sufficient::
+
+    sudo yum install epel-release
+
+If that fails, try the following instead::
```
ed-harrison: Is there any specific type of error which we expect to hit?

docs/source/bird-rr-config.rst
```diff
@@ -0,0 +1,149 @@
...
+reflector.
+
+These instructions are valid for both Ubuntu 14.04 and RHEL 7.
+
+Prerequisites
+-------------
+
+Before starting this you will need the following:
+
+- A machine running either Ubuntu 14.04 or RHEL 7.
```
ed-harrison: Whilst it should be obvious, I think it would be worth noting this is a "fresh" machine, which will become your route reflector (i.e. not just a random machine you've got lying around that, perhaps, is already acting as a compute node...).

docs/source/bird-rr-config.rst
```diff
@@ -0,0 +1,149 @@
...
+    sudo yum install epel-release
+
+If that fails, try the following instead::
+
+    wget https://dl.fedoraproject.org/pub/epel/7/x86_64/e/epel-release-7-5.noarch.rpm
+    sudo yum install epel-release-7-5.noarch.rpm
+
+With that complete, you can now install BIRD::
+
+    yum install -y bird bird6
```
ed-harrison: Presumably the Ubuntu equivalent installs both BIRD and BIRD6 under the one command?  Maybe worth mentioning IPv6 as well?

docs/source/bird-rr-config.rst
```diff
@@ -0,0 +1,149 @@
...
+
+Step 2: Set your BIRD IPv4 configuration
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Before doing this, you'll need to take note of what BGP AS number you've used
+in your compute node install.
```
ed-harrison: Where would this be found - can we give people a pointer?  [I don't know, for example, whether in the "normal" Calico install we tell people to explicitly configure this, or just assume they'll leave it as a default value.]

docs/source/bird-rr-config.rst
```diff
@@ -0,0 +1,149 @@
...
+
+Open ``/etc/bird/bird.conf`` and initially fill it with the following template,
```
ed-harrison: Open bird.conf "on your route reflector" system.  [i.e. these aren't instructions to do something with the compute node which we mentioned above.]

docs/source/bird-rr-config.rst
```diff
@@ -0,0 +1,149 @@
...
+    debug protocols all;
+
+    # This pseudo-protocol watches all interface up/down events.
+    protocol device {
+      scan time 2;    # Scan interfaces every 10 seconds
+    }
+
+Then, at the end, for each compute node in your deployment add one of the
+following blocks, replacing ``<node_shortname>`` with a short
+purely-alphabetical name for the host, ``<node_ip>`` with the node's IPv4
```
ed-harrison: Is <node_shortname> purely a short name in this file?  i.e. it doesn't have to actually reflect the hostname in any way?

docs/source/bird-rr-config.rst
```diff
@@ -0,0 +1,149 @@
...
+~~~~~~~~~~~~~~~~~~~~
+
+Ubuntu 14.04
+^^^^^^^^^^^^
+
+Restart BIRD::
+
+    sudo service bird restart
+
+Optionally, if you configured IPv6 in step 3, also restart BIRD6:
```
ed-harrison: add another ":" at the end of this line.

docs/source/bird-rr-config.rst
```diff
@@ -0,0 +1,149 @@
...
+
+RHEL 7
+^^^^^^
+
+Restart BIRD::
+
+    systemctl restart bird
+    systemctl enable bird
+
+Optionally, if you configured IPv6 in step 3, also restart BIRD6:
```
ed-harrison: ... and this line.

docs/source/bird-rr-config.rst
```diff
@@ -0,0 +1,149 @@
...
+
+    systemctl restart bird6
+    systemctl enable bird6
+
+Step 5: Reconfigure compute nodes
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+On each of your compute nodes, edit ``/etc/bird/bird.conf`` (and optionally
```
ed-harrison: suggest "and, if you're using IPv6," as opposed to "and, optionally"

docs/source/redhat-opens-install.rst
```diff
@@ -422,10 +422,9 @@ On a compute node, perform the following steps:
     will create BIRD config files for simple topologies -- either a
     peering between a single pair of compute nodes, or to a route
     reflector (to avoid the need for a full BGP mesh in networks with
-    more than two compute nodes). If your topology is more complex, please
-    consult the relevant documentation for your chosen BGP stack or ask
-    the mailing list if you have questions about how BGP relates to
-    Calico.
+    more than two compute nodes). If you decide to do this, instructions for
+    configuring BIRD as a route reflector can be found in
+    :doc:`bird-rr-config`.
```
ed-harrison: So, does the user have to run the calico-gen-bird-conf stuff regardless of whether they are using an RR, or do they just jump to the RR docs here and ignore the rest of this step?  Either way, we need to be clear.

docs/source/bird-rr-config.rst
```diff
@@ -0,0 +1,149 @@
+Configuring BIRD as a BGP Route Reflector
+=========================================
+
+These instructions will take you through installing BIRD as a BGP route
+reflector, and updating your other BIRD instances to speak to your new route
+reflector.
```
emmagordon: Since a few reactions to Calico have been along the lines of 'but BGP is big and scary and hard to configure', is it worth including a link here to some kind of BGP/Route Reflection 101 guide?

docs/source/ubuntu-opens-install.rst
```diff
@@ -360,8 +360,8 @@ On a compute node, perform the following steps:
     configuration is outside the scope of this install document.
 
     If you *are* configuring a full BGP mesh you'll need to handle the BGP
-    configuration appropriately. You should consult the relevant
-    documentation for your chosen BGP stack.
+    configuration appropriately. To configure BIRD as a route reflector,
+    follow the instructions in :doc:`bird-rr-config`.
```
ed-harrison: Ahhh - interestingly the ubuntu instructions seem clearer on the subject of whether this step works whether you have an RR or not.  Can we be consistent between the Ubuntu and RHEL ones?
The implication here, though, is that the calico-gen-bird stuff will create per-host config that works for the RR - so why do the RR instructions say that you need to go in and fix up the bird.conf files on each compute host?

docs/source/bird-rr-config.rst
Lukasa: It was unclear to me when reading, but it's possible the above line only works on CentOS.

docs/source/bird-rr-config.rst
Lukasa: Neither, I think, we just off-handedly use it in some command lines. We can expand on that, if you want.

docs/source/bird-rr-config.rst
Lukasa: Yup

ed-harrison: OK - it probably makes sense to make this clear.  Maybe "replacing <node_shortname> with a purely alphabetical name for the host (this must be unique for each host, but the shortname is only used within this file)."

docs/source/bird-rr-config.rst
Lukasa: Sure: can you point me at one? :smile:

docs/source/ubuntu-opens-install.rst
Lukasa: This is a tricky one. The short argument is that I wanted the RR instructions to work even if the user didn't do this stuff, so I used instructions that will always work. If you think that's unnecessary, I can rewrite.

Lukasa: Ok, added a stanza for this.

docs/source/bird-rr-config.rst
emmagordon: You noticed I didn't include a link then! :p Google search gives loads of these, but not obvious which would be most suitable. A calico blog post on the topic at some point might be the way to go? (with a note to come back and add the link here once it's done)

ed-harrison:
I've made some additional changes so that the RHEL and Ubuntu install docs are consistent about the BRID config.  Back to @Lukasa to confirm he's OK with these changes.

alexwlchan:
While making the latest round of formatting tweaks, I noticed that several of the source files in our docs have recently gained the "executable" flag. Oops.

Lukasa:
Your UT failures are not relevant here, they're an ordering problem in our tests.

ed-harrison:
@alexwlchan In @Lukasa's absence, any chance you could have a look at the latest version of this PR?  I think it is probably good to go, but having both made changes to it, I think Cory and I were probably feeling neither of us could do the merge!

ed-harrison:
Actually @alexwlchan, let me update the PR to resolve conflicts first.

ed-harrison:
I've rebased, so @alexwlchan could you now review this and merge, if you're happy?

alexwlchan:
There was a formatting bug in our Ubuntu install doc that had crept in outside this branch, but otherwise this is fine.

Issue #557, opened 2015-05-18
HA: Check that EtcdExceptions such as EtcdClusterIdChanged handled in mechanism driver


fasaxc:
Any etcd request can fail with a connection problem or EtcdClusterIdChanged.  Need to make sure that the mechanism driver does the right thing.

PR #558, opened 2015-05-18
Various fixes found through manual FV

### LINKED ISSUE #556: HA: Elector doesn't reconnect to etcd after EtcdClusterIdChanged
The EtcdClusterIdChanged exception is persistent.  Once hit, it will be raised on every subsequent watch(); we need to tear down the client and reconnect in that case (unless we change the behaviour in our fork of python-etcd but we've already made moves to upstream that change).

**Edit**: looks like we're going to address this in python-etcd rather than the Elector.

fasaxc:
Started off just fixing #556 and #557 but found a few more issues while I was doing manual FV of those changes:

* Reinit client and elector after EtcdException in mechanism driver (including EtcdClusterIdChanged).
* Have Elector forcefully shut down its greenlet to avoid temporary overlap with dying greenlet.  Use the stop mechanism in the UTs to give it a workout.
* Use decorator to handle exception in each transport API call.
* Simplify exception handling in Elector (seemed odd that we were catching EtcdException and EtcdFooException, which is a subclass -- maybe an artefact of our stub missing this detail?)
* Reduce log verbosity on somewhat-expected failures by feeding exceptions through common logging function (I know at least one customer who might see all the tracebacks as a regression).
* Reduce and randomise the retry timer.  It was taking minutes to even retry in my testing.
* Fix prevExists/prevExist incorrect API call to python-etcd.

Fixes #556 #557.

### DIFF
--- a/calico/election.py
+++ b/calico/election.py
@@ -20,12 +20,12 @@ calico.election
 
 Calico election code.
 """
+from greenlet import GreenletExit
 import etcd
 import eventlet
 from httplib import HTTPException
 import logging
 from socket import timeout as SocketTimeout
-import time
 from urllib3 import Timeout
 from urllib3.exceptions import ReadTimeoutError, ConnectTimeoutError, HTTPError
 
@@ -101,6 +101,8 @@ class Elector(object):
             # Election greenlet failed. Log but reraise.
             _log.exception("Election greenlet exiting")
             raise
+        finally:
+            self._attempt_step_down()
 
     def _vote(self):
         """
@@ -144,17 +146,17 @@ class Elector(object):
                 # Unexpected timeout - reconnect.
                 _log.debug("Read from etcd timed out (%r), retrying.", e)
                 return
+            except etcd.EtcdKeyNotFound:
+                # It should be impossible for somebody to delete the object
+                # without us getting the delete action, but safer to handle it.
+                _log.warning("Implausible vanished key - become master")
+                self._become_master()
             except (etcd.EtcdException, HTTPError, HTTPException,
                     etcd.EtcdClusterIdChanged, etcd.EtcdEventIndexCleared):
                 # Something bad and unexpected. Log and reconnect.
                 _log.warning("Unexpected error waiting for master change",
                              exc_info=True)
                 return
-            except etcd.EtcdKeyNotFound:
-                # It should be impossible for somebody to delete the object
-                # without us getting the delete action, but safer to handle it.
-                _log.warning("Implausible vanished key - become master")
-                self._become_master()
 
             if (response.action in ETCD_DELETE_ACTIONS or
                     response.value is None):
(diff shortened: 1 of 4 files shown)


fasaxc:
There's a change I'm not sure about here: in order to let the Elector step down, I removed the timestamp from the value but I'm not sure if the timestamp was needed for some purpose.

calico/election.py
```diff
@@ -144,17 +146,17 @@ def _vote(self):
                 # Unexpected timeout - reconnect.
                 _log.debug("Read from etcd timed out (%r), retrying.", e)
                 return
+            except etcd.EtcdKeyNotFound:
```
fasaxc: Moved this except block up because it looked unhittable.

fasaxc:
Sorry it's not particularly well split out but I was testing it all together so splitting it out might give a false sense of security!

calico/election.py
```diff
@@ -93,14 +94,21 @@ def _run(self):
...
+                _log.info("Retrying leader election in %.1f seconds",
+                          retry_time)
+                eventlet.sleep(retry_time)
+        except ElectorStopped:
+            _log.info("Elector told to stop.")
+            raise
+        except BaseException as e:
             # Election greenlet failed. Log but reraise.
-            _log.exception("Election greenlet exiting")
+            _log.exception("Election greenlet exiting: %r", e)
```
Lukasa: This change is unneeded: `log.exception` will dump out the entire exception stacktrace.

Lukasa:
Note that this only partially resolves #557. In particular, `EtcdClusterIDChanged` should *not* cause failures of any method in `mech_calico`: they should retry.

plwhite:
The timestamp is there for two reasons.

* Diagnosability. In the field with elections support like to know when an elected instance takes over.
* Reload. Suppose a process has an election, gets and exception and restarts. It avoids the problem where the instance thinks it's elected (although actually that might not matter - the instance never checks the string to see "is that me").

Neither of these is a bomb-proof argument for the timestamp.

calico/election.py
plwhite: I am puzzled by the double sleep. We had a backoff interval of _interval seconds, and you changed it to a random number of seconds without scaling by _interval. Why was that?

calico/election.py
```diff
@@ -173,49 +175,77 @@ def _become_master(self):
...
                                     timeout=self._interval)
-
         except Exception as e:
             # We could be smarter about what exceptions we allow, but any kind
             # of error means we should give up, and safer to have a broad
             # except here.  Since we expect to hit this in the mainline, we
             # don't log out the stack trace.
-            _log.info("Failed to become elected master - key %s: %r",
-                      self._key, e)
+            self._log_exception("become elected master", e)
```
plwhite: Actually, I think _log_exception sometimes does log out the stack trace, doesn't it? Then the comment isn't quite correct.

plwhite:
Right, so I am going to fix the comment independently, and the rest looks fine to me (we could have a debate about the timestamp, but not worth it), so no action.

PR #560, opened 2015-05-19
Tweak comment about logging exception.

plwhite:
This comment claimed not to log the exception stack trace, but the code did in at least some cases. So correct it.


Lukasa:
@matthewdupre This is only a comment change.

PR #564, opened 2015-05-19
Improved etcd exception handling

Lukasa:
This change consists of two parts:

1. Reduce the scope of the `except` block for CAS actions to avoid being aggressive on capturing the exceptions.
2. Retry if the cluster ID has changed. API actions shouldn't fail just because etcd got moved: we should simply try again. This change implements a decorator that ensures that they do.

### DIFF
--- a/calico/openstack/mech_calico.py
+++ b/calico/openstack/mech_calico.py
@@ -591,8 +591,7 @@ class CalicoMechanismDriver(mech_agent.SimpleAgentMechanismDriverBase):
         for endpoint in eps_to_delete:
             try:
                 self.transport.atomic_delete_endpoint(endpoint)
-            except Exception:
-                # TODO: Be more specific.
+            except ValueError:
                 # If the atomic CAD doesn't successfully delete, that's ok, it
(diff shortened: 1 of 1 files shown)


fasaxc:
Only thought is should the retry be pushed down into t_etcd instead?  Feels odd that mech_calico knows about etcd-specific exceptions (but then I guess that split has already eroded a bit!)

Lukasa:
I thought about that (my original implementation actually did it, I still have the commit lying around), but the problem is that it's actually too low a level to handle it. Some API actions involve multiple etcd operations, and if the cluster is moved during one of the later ones we need to assume that we should replay the whole transaction. That's the only way to avoid writing partial data.

A good compromise (that I thought about but skipped for simplicity) is to wrap the cluster ID exception in a more generic one that the t_etcd layer exposes, and then catch that. I'd be amenable to that change if you want to make it.

fasaxc:
Should be a fairly easy change to make later if there's ever another transport, let's not clutter the code with that right now.

PR #568, opened 2015-05-19
Validate and canonicalise IPs and MACs at validation time.

fasaxc:
Add UT for validation, fix various minor issues found while
adding UT.

### DIFF
--- a/calico/common.py
+++ b/calico/common.py
@@ -32,6 +32,7 @@ import netaddr.core
 import os
 import sys
 from types import StringTypes
+from netaddr.strategy import eui48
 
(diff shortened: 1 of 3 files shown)


fasaxc:
@plwhite I stripped out the canonicalisation part of my #529 PR and beefed up the UTs to test all the failure cases.  That helped me find some genuine bugs (usually where we validate x is not None but then access x anyway further down to validate its value).

I think you've got another PR in play to do the ID validation?

plwhite:
Clashes with PR#563; need to do some interesting merging to sort that all out. Will try and get to that tomorrow. Code however looks fine.

PR #570, opened 2015-05-19
Handle EtcdKeyNotFound from atomic delete methods.

fasaxc:
Gets raised instead of ValueError if the key is not present at all.


robbrockbank:
LGTM

Issue #566, opened 2015-05-19
Felix never tells you if the ready flag is missing.


plwhite:
On startup, Felix refuses to load configuration until the "ready" flag is present. This leads to the following comedy chain of events.
* Felix does not make any logs until it has read config (because the severity is too low, and we do not know the log file for sure until after the etcd config is read).
* Felix cannot read config until the ready flag is set.

As a result, if the ready flag is missing, Felix not only never starts, but never makes any logs either, which is not at all easy to debug!

We could partially fix by logging the missing ready flag at huge severity, so at least it goes into syslog and to stdout. That might be the most pragmatic fix.

lxpollitt:
I don't think I fully get the "we do not know the log file for sure until after the etcd config is read" comment.  I thought one of the key motivations for having a local config file was so that you can get logging at the level you desire even if you can't connect to etcd?

plwhite:
Yes, agreed. The logic here in Felix is wrong, since one of the refactors has moved initialising of logging after etcd reading. That's clearly not right and needs sorting out properly.

PR #574, opened 2015-05-21
Fix failure to log when Ready flag missing.

plwhite:
Fixes issue #566


Lukasa:
Let's be clear, we *did* log. There was information in /var/log/upstart/calico-felix.log.

Issue #212, opened 2015-02-23
Security groups that limit outbound traffic can break metadata.


Lukasa:
The metadata service should be explicitly excluded from outbound packet limitations.

fasaxc:
Feels like a hole we should address fairly soon.

plwhite:
I'm not sure I buy this. If you have metadata, the first rules you set up allow it, overriding whatever you do in your security group. What am I missing?

Lukasa:
@plwhite A clearer explanation from me, I think.

See the following OpenStack system, with two VMs on it. Both are in a security group that has *no* outbound rules in it at all (that is, it DENYs all outbound traffic). Below are the iptables for that system:

```
Chain INPUT (policy ACCEPT 7 packets, 674 bytes)
 pkts bytes target     prot opt in     out     source               destination
4925K 1300M felix-INPUT  all  --  any    any     anywhere             anywhere
6866K 1782M nova-api-metadat-INPUT  all  --  any    any     anywhere             anywhere

Chain FORWARD (policy ACCEPT 0 packets, 0 bytes)
 pkts bytes target     prot opt in     out     source               destination
 3247  273K felix-FORWARD  all  --  any    any     anywhere             anywhere
    0     0 nova-filter-top  all  --  any    any     anywhere             anywhere
    0     0 nova-api-metadat-FORWARD  all  --  any    any     anywhere             anywhere

Chain OUTPUT (policy ACCEPT 5 packets, 544 bytes)
 pkts bytes target     prot opt in     out     source               destination
7342K 1507M nova-filter-top  all  --  any    any     anywhere             anywhere
7342K 1507M nova-api-metadat-OUTPUT  all  --  any    any     anywhere             anywhere

Chain felix-FORWARD (1 references)
 pkts bytes target     prot opt in     out     source               destination
 1624  136K felix-FROM-ENDPOINT  all  --  tap+   any     anywhere             anywhere
 1623  136K felix-TO-ENDPOINT  all  --  any    tap+    anywhere             anywhere
 1624  136K ACCEPT     all  --  tap+   any     anywhere             anywhere
 1623  136K ACCEPT     all  --  any    tap+    anywhere             anywhere

Chain felix-FROM-ENDPOINT (2 references)
 pkts bytes target     prot opt in     out     source               destination
   99  8600 felix-from-0d9e6b4b-21  all  --  tap0d9e6b4b-21 any     anywhere             anywhere            [goto]
[...]

PR #575, opened 2015-05-21
Ensure that metadata is not blocked by over-restrictive rules on outb…

plwhite:
…ound traffic.

Fixes issue #212.


plwhite:
This was a simple enough fix that I decided to make it. Since my test box has no functional metadata, I haven't been able to test the function, but it creates the rule OK, and the tests cover it, so close enough for review.

fasaxc:
LGTM.

PR #576, opened 2015-05-21
Initial upgrade procedure documentation

Lukasa:
This is my first draft of documentation for our upgrade procedure.

### DIFF
--- a/docs/source/opens-upgrade.rst
+++ b/docs/source/opens-upgrade.rst
@@ -1,0 +1,133 @@
+Upgrade Procedure (OpenStack)
+=============================
+
+This document details the procedure for upgrading a Calico-based OpenStack
+system. It contains the full steps for upgrading all components and the order
+in which that upgrade should be performed. Most releases do not concurrently
+upgrade all of these components: if a release does not upgrade a given
+component, you may skip those steps.
+
+.. warning:: While the upgrade procedure is very safe, you will be unable to
+             issue API requests to your OpenStack system during the procedure.
+             Please plan your upgrade window accordingly, and see the
+             :ref:`service_impact` section for more details.
+
+
+.. _service_impact:
+
+Service Impact
+--------------
+
+During the upgrade, **all VMs will continue to function normally**: there
+should be no impact on the data plane. However, control plane traffic may fail
+at different points throughout the upgrade.
+
+Generally, users should be prevented from creating or updating virtual machines
+during this procedure, as these actions will fail. VM deletion *may* succeed,
+but will likely be delayed until the end of the upgrade.
+
+For this reason, we highly recommend planning a maintenance window for the
+upgrade. During this window, you should disable all user API access to your
+OpenStack deployment.
+
+Upgrade Procedure
+-----------------
+
+Upgrade is performed in the following stages.
+
+1: Upgrade etcd
+~~~~~~~~~~~~~~~
+
+This step should be run on every machine in your deployment that runs any
+Calico code, and also on the machine running the etcd cluster.
+
+Ubuntu 14.04
+^^^^^^^^^^^^
+
+On the machine running etcd, simply use apt-get to obtain the more recent
+version::
+
+    apt-get update
+    apt-get install etcd
+
+Red Hat 7
+^^^^^^^^^
+
+On the machine running etcd, begin by stopping the etcd process::
+
+    systemctl stop etcd
+
+Then, download the latest binary and install it::
+
+    curl -L  https://github.com/coreos/etcd/releases/download/v2.0.11/etcd-v2.0.11-linux-amd64.tar.gz -o etcd-v2.0.11-linux-amd64.tar.gz
+    tar xvf etcd-v2.0.11-linux-amd64.tar.gz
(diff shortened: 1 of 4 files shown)


docs/source/opens-upgrade.rst
```diff
@@ -0,0 +1,133 @@
...
+
+Ubuntu 14.04
+^^^^^^^^^^^^
+
+First, upgrade packaged components::
+
+    apt-get update
+    apt-get install dnsmasq-base nova-api-metadata neutron-dhcp-agent neutron-common python-etcd calico-compute
+
+Finally, restart Felix to ensure that it picks up any changes::
```
alexwlchan: Nitpick:having both "first ... finally" feels weird to me. "finally" sounds as if we're at the end of a much longer list; I'd just drop it. "Then restart Felix..."

docs/source/opens-upgrade.rst
```diff
@@ -0,0 +1,133 @@
...
+^^^^^^^^^^^^
+
+First, update packaged components::
+
+    apt-get update
+    apt-get install python-etcd calico-control
+
+Finally, restart Neutron to ensure that it picks up any changes::
+
+    service neutron-server restart
```
alexwlchan: Nitpick: ditto above.

docs/source/opens-upgrade.rst
```diff
@@ -0,0 +1,133 @@
...
+but will likely be delayed until the end of the upgrade.
+
+For this reason, we highly recommend planning a maintenance window for the
+upgrade. During this window, you should disable all user API access to your
+OpenStack deployment.
+
+Upgrade Procedure
+-----------------
+
+Upgrade is performed in the following stages.
```
ed-harrison: "stages, which must be performed in the order shown."  [Worth hammering home this point, I think.]

docs/source/opens-upgrade.rst
```diff
@@ -0,0 +1,133 @@
...
+1: Upgrade etcd
+~~~~~~~~~~~~~~~
+
+This step should be run on every machine in your deployment that runs any
+Calico code, and also on the machine running the etcd cluster.
+
+Ubuntu 14.04
+^^^^^^^^^^^^
+
+On the machine running etcd, simply use apt-get to obtain the more recent
```
ed-harrison: Personally, I'd remove "simply" here.

docs/source/opens-upgrade.rst
```diff
@@ -0,0 +1,133 @@
...
+    apt-get install etcd
+
+Red Hat 7
+^^^^^^^^^
+
+On the machine running etcd, begin by stopping the etcd process::
+
+    systemctl stop etcd
+
+Then, download the latest binary and install it::
```
ed-harrison: Here you say download the "latest" binary, but then give the instruction to download 2.0.11.  If we want to pin ourselves to 2.0.11, then say that in the line above (e.g. "Download the correct etcd binary - Calico is currently tested to work with etcd v2.0.11.").

docs/source/opens-upgrade.rst
```diff
@@ -0,0 +1,133 @@
...
+
+First, upgrade python-etcd::
+
+    curl -L https://github.com/Metaswitch/python-etcd/archive/master.tar.gz -o python-etcd.tar.gz
+    tar xvf python-etcd.tar.gz
+    cd python-etcd-master
+    python setup.py install
+
+Then, update the relevant components::
+
```
ed-harrison: Will yum update update other things beyond Calico?  Just thinking that in the Ubuntu instructions we're explicitly upgrading only the Calico stuff, but if yum update could upgrade other packages, that could have unintended consequences.  Is there a way to just update the Calico stuff?

docs/source/opens-upgrade.rst
```diff
@@ -0,0 +1,133 @@
...
+    yum update
```
ed-harrison: ditto above

ed-harrison:
I think you should have a pointer to the upgrade instructions at the top of the relevant install instructions.  e.g. "These instructions will take you through first time installation of Calico...  If you want to upgrade an existing version of Calico, then see..."

Do you think we need to say anything about the upgrade instructions only working to upgrade versions after 0.xx (whatever xx is, in practice)?

docs/source/opens-upgrade.rst
alexwlchan: @ed-harrison Yes and yes. From the [Red Hat docs](https://access.redhat.com/documentation/en-US/Red_Hat_Network_Satellite/5.5/html/Reference_Guide/sect-Reference_Guide-Package_Updater-Updating_Packages_from_the_Command_Line_with_yum.html):

> If no packages are specified, then yum will attempt to update all installed packages.

Fix is to do <code>yum update <em>calico-package-name</em></code> for appropriate package names.

Lukasa:
Thanks @ed-harrison @alexwlchan, those markups have been made.

@ed-harrison, I don't think we need that tag in the upgrade instructions at this time. I think we consider the pre-etcd days to be the dark past of this project, and so we just don't really want to worry about it.

ed-harrison:
LGTM

PR #582, opened 2015-05-22
Fix incorrect underline build warning in docs

ed-harrison:
Incorrect length underline is causing a build warning in our docs.


alexwlchan:
LGTM

PR #586, opened 2015-05-22
Neutron driver: specify fixed MAC address for Calico TAP interfaces

neiljerram:


neiljerram:
Hi @Lukasa - I'm assigning this to you because it looks (from Git) like you're handling the task to rebase our Nova and Neutron work to Kilo.

I think it would be good if our Kilo-based Nova code used the same patch as I am currently proposing upstream at https://review.openstack.org/#/c/146914/, and I have prepared and pushed a branch at https://github.com/Metaswitch/calico-nova/tree/nj-calico_2015.1.0 that does the rebase in this form - i.e. if it works, nj-calico_2015.1.0 should replace the calico_2015.1.0 that you just prepared.

To go with the Nova change as in nj-calico_2015.1.0, there is a corresponding mechanism driver change needed, as in this PR.

Therefore, ideally:

- please review and merge this PR

- do a Kilo FV run using nj-calico_2015.1.0 instead of calico_2015.1.0

- assuming that works, discard the existing calico_2015.1.0 and then rename nj-calico_2015.1.0 to calico_2015.1.0.

Thanks!

neiljerram:
If and when you do this, can you check in particular that it still sets the fixed MAC address that we want on all Calico TAP interfaces?

Lukasa:
Can we safely merge this? It seems like we need to do some conditional work to make sure we only do this on Kilo or higher platforms.

neiljerram:
It bears testing, of course, but I'm pretty sure pre-Kilo platforms will just ignore the added mac_address key.

Lukasa:
Merged manually in 04d92eaf4380b0349941706d2590c38cea7ee1bd.

Issue #534, opened 2015-05-13
Felix runs out of file descriptors at around 160 instances on one host


matthewdupre:
I'm running with the default ulimit of 1024, and after creating 160 VMs it stopped setting up new VMs and cyclically restarted.  We should consider reducing the number of file handles used and (if we can't remove the relation to number of instances) document that the limit may need increasing.
```
2015-05-13 17:47:25,134 [ERROR][12169/352] calico.felix.actor 227: Exception processing 597b3eb38cf7 (on_endpoint_update)
Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/calico/felix/actor.py", line 225, in _step
    result = msg.method()
  File "/usr/lib/python2.7/dist-packages/calico/felix/endpoint.py", line 235, in on_endpoint_update
    self._configure_interface()
  File "/usr/lib/python2.7/dist-packages/calico/felix/endpoint.py", line 360, in _configure_interface
    self.endpoint["mac"])
  File "/usr/lib/python2.7/dist-packages/calico/felix/devices.py", line 182, in set_routes
  File "/usr/lib/python2.7/dist-packages/calico/felix/devices.py", line 62, in list_interface_ips
  File "/usr/lib/python2.7/dist-packages/calico/felix/futils.py", line 100, in check_call
  File "/usr/lib/python2.7/dist-packages/gevent/subprocess.py", line 238, in __init__
  File "/usr/lib/python2.7/dist-packages/gevent/subprocess.py", line 638, in _execute_child
  File "/usr/lib/python2.7/dist-packages/gevent/subprocess.py", line 594, in pipe_cloexec
OSError: [Errno 24] Too many open files
2015-05-13 17:47:25,140 [CRITICAL][12169/352] calico.felix.actor 511: TrackedAsyncResult ('endpoint.py:82:_on_object_started', 'EndpointManager(IPv6) (processing 27dd27bbd479)', u'LocalEndpoint(e3257adf-2ba3-4011-a339-2ceb2d0886c4(IPv6))', 'on_endpoint_update') was leaked with exception OSError(24, 'Too many open files').  Dying.
```

fasaxc:
That's odd.  I'm not sure what we'd be using a proportional number of FDs for.

fasaxc:
Some ideas:
* we're probably using a few hundred for normal things
* could be gevent using FDs for something underneath us
[...]

PR #590, opened 2015-05-27
Increase number of file descriptors in calico-felix.conf

alexwlchan:
This increases the number of file descriptors available to Felix, which should mitigate issue #534 (Felix gets sad when it can't create any more VMs/tap interfaces).

I picked 32K just because it was the first number suggested in that issue. We don't have any documentation for editing calico-felix.conf, so I assume it gets put in place by our installers. This is the only felix.conf file in the /calico repo, so I'm guessing this is it.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,7 @@
   interfaces.
 - Prevent the possibility of gevent context-switching during garbage collection
   in Felix
+- Increase the number of file descriptors available to Felix
 
 ## 0.21
 
--- a/debian/calico-felix.upstart
+++ b/debian/calico-felix.upstart
@@ -5,6 +5,8 @@ author "Project Calico Maintainers <maintainers@projectcalico.org>"
 start on stopped rc RUNLEVEL=[2345]
 stop on runlevel [!2345]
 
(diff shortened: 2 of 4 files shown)


Lukasa:
Do we not need to make a similar change on Ubuntu?

plwhite:
I think this isn't quite right. That file is only used for RPMs, not for the normal upstart job (I think). That means you should also change debian/calico-felix.upstart

Net of this is that 

* You should change that file too.
* Run the changes past @matthewdupre since he's more expert in this area.
* Add a changelog entry for this change.

plwhite:
Comments crossed - @Lukasa and I spotted the same thing, but he didn't get distracted by a mail in the middle!

alexwlchan:
Added the Debian file and changelog entry.

matthewdupre:
Debian and RHEL 6.5 upstart job changes are fine, but you also need to update the RHEL 7 systemd service file (rpm/calico-felix.service).

Specifically LimitNOFILE=32000 in the [service] section - see http://man7.org/linux/man-pages/man5/systemd.exec.5.html.

PR #591, opened 2015-05-27
Migrate to oslo logging when needed.

Lukasa:
Seems that Neutron migrated to Oslo logging during the Kilo cycle, which means we need to change our imports.


Lukasa:
This change actually also contains method overrides for some methods added in Kilo. I'm not 100% confident on those, tbh, but for now they appear to shut Kilo up. We need to make sure @neiljerram casts his eye over those.

Issue #542, opened 2015-05-15
Felix exception during delete port


DaveLangridge:
Found during churn test on neutron-ha branch.

2015-05-14 14:14:38,361 [ERROR][21795/166] calico.felix.endpoint 388: Cannot delete rules for interface tapfaab2e73-0e for EndpointId<calico-vm15/openstack/2712722d-e654-43b0-aebc-7a310468d40a/faab2e73-0e79-4487-b939-7501f9e39456>
Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/calico/felix/endpoint.py", line 379, in _deconfigure_interface
    devices.set_routes(self.ip_type, set(), self._iface_name, None)
  File "/usr/lib/python2.7/dist-packages/calico/felix/devices.py", line 182, in set_routes
    current_ips = list_interface_ips(ip_type, interface)
  File "/usr/lib/python2.7/dist-packages/calico/felix/devices.py", line 62, in list_interface_ips
    ["ip", "-6", "route", "list", "dev", interface]).stdout
  File "/usr/lib/python2.7/dist-packages/calico/felix/futils.py", line 105, in check_call
    args, retcode, stdout, stderr, input=input_str)
FailedSystemCall: Failed system call (retcode : 1, args : ('ip', '-6', 'route', 'list', 'dev', u'tapfaab2e73-0e'))
  stdout  :
  stderr  : Cannot find device "tapfaab2e73-0e"
 
  input  : None

fasaxc:
I *think* this is benign.  It simply means the tap device has been removed under us, which we handle.  We should clean up the log spam, if possible, though.

plwhite:
Yes, this is definitely benign. I'm going to handle it better, but not critically urgent (so fix won't be for a few days).

plwhite:
Looks like interface_exists is failing - presumably because the place it checks (/sys/class/net/<iface>) exists, but the interface is going down at the time so the ip commands think it has gone.

matthewdupre:
Merged a (slightly speculative) fix for this, so I'm closing this for now.  If you hit this again, please reopen the issue.

PR #593, opened 2015-05-28
Suppress exception triggered when removing routes from device being r…

plwhite:
…emoved.

Fixes issue #542.


plwhite:
This is a slightly speculative fix - I know that interface_exists was returning that the interface existed when other ip calls were reporting that it did not, so it feels right though and is pretty safe.

matthewdupre:
Looks fine.

PR #594, opened 2015-05-28
Fix bad assert left in after fix to issue 460

plwhite:
Also, add changelog entry for issue 460 (missed in initial fix).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,7 @@
 # Changelog
 
+- Firewall input characters in profiles and tags.
+
 ## 0.22-dev
 
 - Updated Mechanism driver to specify fixed MAC address for Calico tap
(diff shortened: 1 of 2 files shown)


Lukasa:
@plwhite Changes should go in CHANGES.md, not in debian/changelog.

matthewdupre:
Cory's right in that you should remove the debian changelog entry.  I can add to that that the package build failure is because it wasn't formatted properly either (it's very picky about the signature line).

matthewdupre:
If we can cheaply reintroduce the assert that would be nice, otherwise just need the changelog -> CHANGES.md fix before I can merge.

matthewdupre:
@plwhite Strictly the changelog entry should go under #0.22-dev (or whatever the next release is), but we'll just fix that when we make the release.

PR #597, opened 2015-05-29
Changes to ubuntu install instructions to cover kilo

DaveLangridge:


docs/source/ubuntu-opens-install.rst
```diff
@@ -53,17 +53,16 @@ Configuring the APT software sources
...
 
 Add the Calico PPA.
 
 ::
 
-    sudo apt-add-repository ppa:project-calico/icehouse
+    sudo apt-add-repository ppa:project-calico/<release>
+
 
+Where <release> is icehouse, juno or kilo
```
alexwlchan: Note: to get \<release> to show up in a monospaced font, add backticks, like so: <code>\`\`\<release>``</code>. Also, this sentence needs a full stop.

alexwlchan:
Do we need to add a new link to install instructions for Kilo at the top of the document, or are we assuming that everybody will have Icehouse or Juno already?

docs/source/ubuntu-opens-install.rst
```diff
@@ -17,7 +17,7 @@ Ubuntu 14.04 Packaged Install Instructions
 
 These instructions will take you through a first-time install of Calico using
 the latest packages on a system running Ubuntu 14.04 with either OpenStack
-Icehouse or Juno. If you are upgrading an existing system, please see the
+Icehouse, Juno or kilo. If you are upgrading an existing system, please see the
```
alexwlchan: “Kilo” should be capitalised.

ed-harrison:
@DaveLangridge I've made a few more suggested changes.  One thing which I think would still be worth adding is a note on how to tell whether you've got the correct libvirt or not.  Could you add that and then pass this to @matthewdupre to merge (which we should do once the Kilo packages are actually ready)?

DaveLangridge:
Matt, can you merge this when we're ready to release.

PR #603, opened 2015-06-02
[WIP] Implement periodic resync.

fasaxc:
* Move etcd polling into a worker greenlet.
* Start another greenlet to periodically trigger a resync.

### DIFF
--- a/calico/felix/config.py
+++ b/calico/felix/config.py
@@ -168,6 +168,9 @@ class Config(object):
 
         self.add_parameter("StartupCleanupDelay", "Delay before cleanup starts",
                            30, value_is_int=True)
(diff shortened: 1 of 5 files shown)


fasaxc:
@plwhite FYI, work in progress on periodic resync.  Annoyingly, it doesn't diff cleanly (because I reordered the methods a bit) but I just moved the existing polling function out to a separate greenlet and used Events to trigger it to do things.

Seems to work in my kick-the-types testing here.

fasaxc:
Rebased onto current master.

fasaxc:
I'm going to split this out into a refactor PR and then a periodic resync PR.

PR #604, opened 2015-06-02
Previous fix broke IPv6 - unbreak it. While we're at it, beef up test…

plwhite:
…s substantially.


plwhite:
OK, this has been superseded by PR #605.

PR #607, opened 2015-06-02
Profile deletion

plwhite:
Self-explanatory. Such a minor bug that I think it is not worth release noting or raising an issue.

### DIFF
--- a/calico/felix/profilerules.py
+++ b/calico/felix/profilerules.py
@@ -71,7 +71,7 @@ class RulesManager(ReferenceManager):
 
     @actor_message()
     def on_rules_update(self, profile_id, profile):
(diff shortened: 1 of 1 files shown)


plwhite:
Shaun, I think this is uncontroversial (just noticed in passing), but you know this code better so over to you to tell me I've misunderstood completely if you like!

plwhite:
Ouch. Still a good fix, but merge disaster - let me redo.

plwhite:
I realised I botched the merge and had to redo - this is now the right version (and yes, it takes some effort to botch a one line merge, but I managed).

fasaxc:
LGTM

PR #610, opened 2015-06-02
Hopefully reduce occupancy.

fasaxc:
@plwhite this implements the optimization that we discussed.  Would be good if you or Alex could give it a spin in the simulated scale test to see if it actually helps.

* Replace endpoint dict with optimized struct in IpsetManager.
* Only store endpoints that contribute.

### DIFF
--- a/calico/felix/ipsets.py
+++ b/calico/felix/ipsets.py
@@ -50,6 +50,7 @@ class IpsetManager(ReferenceManager):
         # State.
         # Tag IDs indexed by profile IDs
         self.tags_by_prof_id = {}
+        # EndpointData "structs" indexed by EndpointId.
         self.endpoint_data_by_ep_id = {}
 
         # Main index.  Since an IP address can be assigned to multiple
@@ -389,7 +390,7 @@ class IpsetManager(ReferenceManager):
         """
         Notes in the index that an endpoint uses the given profiles.
 
-        :param set[str] prof_ids: List of profile IDs
+        :param set[str] prof_ids: set of profile IDs that the endpoint is in.
         :param EndpointId endpoint_id: ID of the endpoint
         """
         for prof_id in prof_ids:
@@ -400,7 +401,8 @@ class IpsetManager(ReferenceManager):
         Notes in the index that an endpoint no longer uses any of the
         given profiles.
 
-        :param set[str] prof_ids: List of profile IDs
+        :param set[str] prof_ids: set of profile IDs to remove the endpoint
+            from.
         :param EndpointId endpoint_id: ID of the endpoint
         """
         for prof_id in prof_ids:
--- a/calico/felix/ipsets.py
+++ b/calico/felix/ipsets.py
@@ -19,10 +19,10 @@ felix.ipsets
 
 IP sets management functions.
 """
-from collections import defaultdict
 
-import logging
+from collections import defaultdict
 from itertools import chain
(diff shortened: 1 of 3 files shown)


calico/felix/ipsets.py
```diff
@@ -50,8 +50,7 @@ def __init__(self, ip_type):
         # State.
         # Tag IDs indexed by profile IDs
         self.tags_by_prof_id = {}
-        # Endpoint dicts indexed by EndpointId objects
-        self.endpoints_by_ep_id = {}
+        self.endpoint_data_by_ep_id = {}
```
plwhite: You lost a comment here in the change!

calico/felix/ipsets.py
```diff
@@ -361,7 +391,7 @@ def _add_profile_index(self, prof_ids, endpoint_id):
         """
         Notes in the index that an endpoint uses the given profiles.
 
-        :param list[str] prof_ids: List of profile IDs
+        :param set[str] prof_ids: List of profile IDs
```
plwhite: Comment says both set and List. One of these is wrong.

calico/felix/ipsets.py
```diff
@@ -372,7 +402,7 @@ def _remove_profile_index(self, prof_ids, endpoint_id):
         Notes in the index that an endpoint no longer uses any of the
         given profiles.
 
-        :param list[str] prof_ids: List of profile IDs
+        :param set[str] prof_ids: List of profile IDs
```
plwhite: Again, comment inconsistent.

plwhite:
LGTM with a couple of minor comments about comments.

plwhite:
In my initial testing however, this fix made absolutely no difference whatsoever, to the point where I am wondering if the data is kept referenced somewhere else so that we cannot remove it from memory? Just can't explain why it doesn't seem to work.

To add to the sheer entertainment value I put a return at the top of "on_endpoint_update" in ipsets which also had no effect at all, which backs the idea above though I really cannot find any references to objects after being passed in.

alexwlchan:
@plwhite I have literally just emailed you to say that it made no difference in my testing as well. I was feeling a bit dubious, but maybe something else is going on.

fasaxc:
@alexwlchan @plwhite Yay for testing!

plwhite:
If we hadn't tested we'd be finished by now. Just sayin'.

fasaxc:
Pushed my markups.  I think we agreed that it was a good change, even if it didn't affect occupancy.

plwhite:
It turns out that the failure to reduce occupancy is a test artefact. When under heavy enough test load, Felix was just getting behind and so failing to process messages in a timely manner. The impact of this fix is that when it does process the messages it throws away unnecessary data.

Key headline is that in my scale test, creating 10s of thousands of small endpoints, this cut the memory use per endpoint from about 2.5KiB to about 1KiB (resident set - slightly smaller impact on virtual size). That means this is a good fix and does have benefit.

It also means that we should consider further improvements, namely throwing away the unneeded data earlier (we put data on queues that will later be thrown away - better to throw it away before putting on queues). @fasaxc - you should be aware of this.

Merge blocked pending release, then we'll merge this one.

matthewdupre:
Please add an entry to CHANGES.md for this fix.

fasaxc:
@matthewdupre done.

PR #611, opened 2015-06-02
Intern IDs that are likely to be repeated.

fasaxc:
@plwhite this one is a bit more speculative.  It interns IDs that I expect to be duplicated a lot (profile IDs and tag IDs get used at least 4-8 times each so that might help quite a bit).  I'd appreciate if you or Alex could pass it through the scale test to see if it helps.

* Intern white-listed fields while parsing JSON dicts and lists.
* Intern EndpointID fields.

### DIFF
--- a/calico/datamodel_v1.py
+++ b/calico/datamodel_v1.py
@@ -112,7 +112,36 @@ def get_profile_id_for_profile_dir(key):
     return final_node if prefix == PROFILE_DIR else None
 
 
-class EndpointId(namedtuple("EndpointId", ["host", "orchestrator",
-                                           "workload", "endpoint"])):
+class EndpointId(object):
+    __slots__ = ["host", "orchestrator", "workload", "endpoint"]
+
+    def __init__(self, host, orchestrator, workload, endpoint):
+        # Host and orchestrator can be sensibly interned, they appear
+        # regularly.  They shouldn't contain any utf-8 characters but we get
+        # back a unicode string so encode it.
+        self.host = intern(host.encode("utf8"))
+        self.orchestrator = intern(orchestrator.encode("utf8"))
+        # No sensible interning to be done for workload/endpoint.
+        self.workload = workload.encode("utf8")
+        self.endpoint = endpoint.encode("utf8")
+
     def __str__(self):
-        return self.__class__.__name__ + ("<%s/%s/%s/%s>" % self)
+        return self.__class__.__name__ + ("<%s>" % self.endpoint)
+
+    def __eq__(self, other):
+        if other is self:
+            return True
+        if not isinstance(other, EndpointId):
+            return False
+        if (other.endpoint == self.endpoint and
+                other.workload == self.workload and
+                other.host == self.host and
+                other.orchestrator == self.orchestrator):
+            return True
+        return False
+
+    def __ne__(self, other):
+        return not self.__eq__(other)
+
+    def __hash__(self):
+        return hash(self.endpoint) + hash(self.workload)
--- a/calico/datamodel_v1.py
+++ b/calico/datamodel_v1.py
@@ -116,14 +116,13 @@ class EndpointId(object):
     __slots__ = ["host", "orchestrator", "workload", "endpoint"]
 
     def __init__(self, host, orchestrator, workload, endpoint):
-        # Host and orchestrator can be sensibly interned, they appear
-        # regularly.  They shouldn't contain any utf-8 characters but we get
-        # back a unicode string so encode it.
+        # We intern these strings since they can occur in many IDs.  The
+        # host and orchestrator are trivially repeated for all endpoints
+        # on a host.  The others get repeated over time.
         self.host = intern(host.encode("utf8"))
         self.orchestrator = intern(orchestrator.encode("utf8"))
-        # No sensible interning to be done for workload/endpoint.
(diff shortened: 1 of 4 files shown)


fasaxc:
I ran a couple of basic FV tests against this and #610 merged together and they seemed to work.

calico/datamodel_v1.py
```diff
@@ -109,7 +109,36 @@ def get_profile_id_for_profile_dir(key):
...
+    def __init__(self, host, orchestrator, workload, endpoint):
+        # Host and orchestrator can be sensibly interned, they appear
+        # regularly.  They shouldn't contain any utf-8 characters but we get
+        # back a unicode string so encode it.
+        self.host = intern(host.encode("utf8"))
+        self.orchestrator = intern(orchestrator.encode("utf8"))
+        # No sensible interning to be done for workload/endpoint.
+        self.workload = workload.encode("utf8")
+        self.endpoint = endpoint.encode("utf8")
+
```
plwhite: Why no interning of endpoint ID / workload ID? Seems to me that we might have multiple endpoints in the same workload (say) - and it feels like it can't hurt. Is there a cost to interning that I'm missing?

calico/datamodel_v1.py
```diff
@@ -109,7 +109,36 @@ def get_profile_id_for_profile_dir(key):
...
+    def __eq__(self, other):
+        if other is self:
+            return True
+        if not isinstance(other, EndpointId):
+            return False
+        if (other.endpoint == self.endpoint and
+                other.workload == self.workload and
+                other.host == self.host and
+                other.orchestrator == self.orchestrator):
+            return True
```
plwhite: You could replace `other.host == self.host` with `other.host is self.host` because of the interning; would that be quicker, or is `==` smart enough to do much the same?

calico/felix/fetcd.py
```diff
@@ -221,10 +222,12 @@ def load_initial_dump(self):
         for child in initial_dump.children:
             profile_id, rules = parse_if_rules(child)
             if profile_id:
+                profile_id = intern(profile_id.encode("utf8"))
```
plwhite: Nit picky really, but it would be cleaner if the interning was done in `parse_if_blah` for all `blah`.

calico/felix/fetcd.py
```diff
@@ -418,7 +423,7 @@ def on_orch_delete(self, response, hostname, orchestrator):
         _log.info("Orchestrator dir %s/%s deleted, removing contained hosts",
                   hostname, orchestrator)
         for endpoint_id in list(self.endpoint_ids_per_host[hostname]):
-            if endpoint_id.orchestrator == orchestrator:
+            if endpoint_id.orchestrator == orchestrator.encode("utf8"):
```
plwhite: Sorry - I seem to be picking lots of nits today. More efficient to encode orchestrator once at the top.

calico/felix/fetcd.py
```diff
@@ -456,8 +461,45 @@ def _build_config_dict(cfg_node):
...
+
+    # Rules dicts.
+    "protocol",
+    "src_tag",
+    "dst_tag",
+    "action",
+])
 def intern_dict(d):
-    return dict((intern(str(k)), v) for k, v in d.iteritems())
+    out = {}
```
plwhite: This is a good utility function - better to put it in either `futils.py` or even `common.py`.

plwhite:
I've added quite a few comments, all of them quite minor persnickerty things. Worth us proceeding to test this though, since none of them will have a major impact, and I think generally it looks most excellent.

alexwlchan:
I'm running the scale tests against this PR. Hopefully they'll finish before I leave the office; if not, I'll get them first thing tomorrow.

calico/datamodel_v1.py
fasaxc: `intern()` basically puts the value into a WeakKeyDictionary so it adds some overhead (not sure exactly how much).  Probably 2-3 pointers amortised?

Might be worth interning them to deal with temporal replication though.

calico/datamodel_v1.py
fasaxc: I think it does make that optimisation.  One of the main uses for interning is to make dict keys go faster, which only works if strings make that optimisation.

calico/felix/fetcd.py
fasaxc: Yeh, I was thinking this was getting messy. Wanted to test it before cleaning up any further.

fasaxc:
Only mid-way through code review markups so not ready to merge (might be worth a spin through the scale test though).

fasaxc:
(Still some code review items to address)

fasaxc:
Finished my markups.

fasaxc:
@plwhite did you pass this back for me to fix up the conflict or something else?

plwhite:
Sorry, just to fix up the conflict. I thought I commented, but obviously failed to!

fasaxc:
Conflicts fixed up. Back to you @plwhite

plwhite:
Sorry, but this still conflicts - presumably with something that I merged earlier today! Can you please fix and merge (no need for me to push the button). I'd do it myself but I need to get out to visit a customer now and we'd like that in today.

plwhite:
Conflict was just in CHANGES.md, so I just resolved after all and am now merging.

PR #612, opened 2015-06-02
Limit our shell-out concurrency.

fasaxc:
@plwhite another (small) speculative change that might help with the scale testing.  It puts a cap on the number of concurrent `check_call()` calls.  I doubt it's optimal to try to issue 1000 shell-outs at once at start of day so it'd be interesting to run this and tweak the limit to see if it has an impact.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,8 @@
 
 ## 0.27-dev
 
+- Limit number of concurrent shell-outs in felix to prevent file descriptor
+  exhaustion.
 - Stop restarting Felix on Ubuntu if it fails more than 5 times in 10 seconds.
 
 ## 0.26
--- a/calico/felix/futils.py
+++ b/calico/felix/futils.py
@@ -27,11 +27,13 @@ import hashlib
 import logging
 import os
 from types import StringTypes
+import gevent.lock
 from gevent import subprocess
 import tempfile
 import time
 
 from collections import namedtuple
+
 CommandOutput = namedtuple('CommandOutput', ['stdout', 'stderr'])
 
 # Logger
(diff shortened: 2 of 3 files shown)


matthewdupre:
@fasaxc needs a CHANGES.md entry please.

fasaxc:
@plwhite I think this is ready.  I think there's a benefit in reducing our test envelope but it'd be good to push it in the scale test with a low FD limit to make sure that it solves the problem I set out to fix.

fasaxc:
No particular reason this needs to be in 0.24.

Lukasa:
This has still not been scale tested and also won't merge cleanly. I think the odds of it landing today aren't high either, so I'm knocking it back to 0.26. @eepyaich We really want to schedule some testing of this this week if at all possible.

eepyaich:
Leaving with @plwhite to decide when to schedule this testing.

plwhite:
I'm not convinced this is required for 0.26 - it might make it, but I don't want to commit to that. Let's put it in 1.0 rather than inventing a deadline.

fasaxc:
Rebased onto master.

fasaxc:
Ran a test on @matthewdupre's scale rig with 200+ local taps and the standard FD limit.  Reproed felix falling over.  Watching lsof, I see lots of pipes, which are almost certainly due to communicating with all our child processes.

fasaxc:
Verified that this prevents the cyclic-restart problem that we saw with low ulimit.  Allowing more than about 20 concurrent shell-outs seems to make no difference to performance so lets cap it at 32.

fasaxc:
@matthewdupre this is now verified and tuned to a sensible limit.  Please can you review/merge?

matthewdupre:
Looks good - merging.

PR #619, opened 2015-06-03
Sprinkle stats counters through felix.

fasaxc:
Thought adding some stats to felix might help with general high-level awareness of what's going on in there.

SIGUSR1 triggers a stats dump; which looks like this:
```
2015-06-03 15:53:31,214 [INFO][4854/19] calico.felix.futils 196: === DIAGNOSTICS ===
2015-06-03 15:53:31,214 [INFO][4854/19] calico.felix.futils 198: --- Actor framework counters ---
2015-06-03 15:53:31,215 [INFO][4854/19] calico.felix.futils 188: Batches processed: 28
2015-06-03 15:53:31,215 [INFO][4854/19] calico.felix.futils 188: Blocking calls completed: 13
2015-06-03 15:53:31,215 [INFO][4854/19] calico.felix.futils 188: Blocking calls started: 13
2015-06-03 15:53:31,215 [INFO][4854/19] calico.felix.futils 188: Messages completed: 30
2015-06-03 15:53:31,215 [INFO][4854/19] calico.felix.futils 188: Messages created: 32
2015-06-03 15:53:31,215 [INFO][4854/19] calico.felix.futils 188: Messages executed OK: 30
2015-06-03 15:53:31,215 [INFO][4854/19] calico.felix.futils 198: --- Actor framework ---
2015-06-03 15:53:31,215 [INFO][4854/19] calico.felix.futils 493: Current ref index: 32
2015-06-03 15:53:31,215 [INFO][4854/19] calico.felix.futils 495: Number of tracked messages outstanding: 2
2015-06-03 15:53:31,215 [INFO][4854/19] calico.felix.futils 198: --- IPv4 filter iptables updater ---
2015-06-03 15:53:31,215 [INFO][4854/19] calico.felix.futils 188: Batches finished: 6
2015-06-03 15:53:31,215 [INFO][4854/19] calico.felix.futils 188: Chain deletes: 1
2015-06-03 15:53:31,216 [INFO][4854/19] calico.felix.futils 188: Chain rewrites: 2
2015-06-03 15:53:31,216 [INFO][4854/19] calico.felix.futils 188: Cleanups performed: 1
2015-06-03 15:53:31,216 [INFO][4854/19] calico.felix.futils 188: Refreshed chain list: 3
2015-06-03 15:53:31,216 [INFO][4854/19] calico.felix.futils 188: Rule inserts: 2
2015-06-03 15:53:31,216 [INFO][4854/19] calico.felix.futils 188: iptables success: 4
...
```

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,7 @@
 
 ## 0.24-dev
 
+- Add Felix statistics logging on USR1 signal.
 - Add support for routing over IP-in-IP interfaces in order to make it 
   easier to evaluate Calico without reconfiguring underlying network.
 - Reduce felix occupancy by replacing endpoint dictionaries by "struct"
--- a/calico/felix/actor.py
+++ b/calico/felix/actor.py
@@ -110,7 +110,8 @@ import weakref
 
 from gevent.event import AsyncResult
 from gevent.queue import Queue
-
+from calico.felix import futils
+from calico.felix.futils import StatCounter
 
 _log = logging.getLogger(__name__)
 
@@ -119,6 +120,9 @@ ResultOrExc = collections.namedtuple("ResultOrExc", ("result", "exception"))
 
 # Local storage to allow diagnostics.
 actor_storage = gevent.local.local()
+
+# Global diagnostic counters.
+_stats = StatCounter("Actor framework counters")
 
 
 class Actor(object):
@@ -226,8 +230,10 @@ class Actor(object):
                 except BaseException as e:
                     _log.exception("Exception processing %s", msg)
                     results.append(ResultOrExc(None, e))
+                    _stats.increment("Messages executed with exception")
                 else:
                     results.append(ResultOrExc(result, None))
+                    _stats.increment("Messages executed OK")
                 finally:
                     self._current_msg = None
             try:
@@ -242,11 +248,13 @@ class Actor(object):
                 _log.warn("Splitting batch to retry.")
                 self.__split_batch(batch, batches)
                 num_splits += 1  # For diags.
+                _stats.increment("Split batches")
                 continue
             except BaseException as e:
                 # Most-likely a bug.  Report failure to all callers.
                 _log.exception("_finish_msg_batch failed.")
                 results = [(None, e)] * len(results)
+                _stats.increment("_finish_msg_batch() exception")
 
             # Batch complete and finalized, set all the results.
             assert len(batch) == len(results)
@@ -256,6 +264,9 @@ class Actor(object):
                         future.set_exception(exc)
                     else:
                         future.set(result)
+                    _stats.increment("Messages completed")
+
+            _stats.increment("Batches processed")
         if num_splits > 0:
             _log.warn("Split batches complete. Number of splits: %s",
                       num_splits)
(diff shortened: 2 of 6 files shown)


fasaxc:
I haven't written UTs for the stats stuff.  Thought I'd let you have a look at it first (and maybe try it under a load test to see if it highlights anything interesting).

fasaxc:
@alexwlchan might be a good idea to include this patch in your scale testing (merged together with whatever other PRs you're testing).  Once you've got felix chewing up CPU, you can send it a SIGUSR1 and it'll dump the stats to its log.

plwhite:
Generally, looks good, and happy to merge once we have a few tests. There are a couple of things that I think would help it though.

* It would be nice to be able to simply grep for all stats, so that I can easily pull out the stats from the log file - perhaps by starting all stat lines with a well known string such as "STAT". What do you think?

* If I have a stat "Number of widgets" and I mistype one of the increments as "Number of wodgets" then that will turn into two stats. Similarly, stats with 0 values do not show up. I wonder if we should explicitly specify the names of all stats? I'm in two minds about this, since the benefit is not gigantic, and it makes something lightweight a bit more tedious, but I thought I'd mention the possibility.

* It would be very nice to have periodic dumping of stats. Could we add this at some point easily? It seems to me it would be pretty trivial to do (at the cost of yet another greenlet), though the interval would have to be configurable and perhaps turned off by default to avoid log spam, so might be more work than we want right now.

* Finally, I thought about what stats would be useful and came up with a few more.

    * I'd quite like to see a count of every message type in the system split by actor name and message name (which means that you would increment `"Message sent from %s to %s of type %s" % (sending actor, receiving actor, method)`.

    * I also would like to know how many of X are being managed, so counts of how many local endpoints there are, how many programmed tags, etc.

    * Finally, it would be good to have a stat counter that isn't really stats in quite the same way, that measures things like memory, heap usage, etc. - system stuff.

    I suspect the first of these is trivial enough to be worthwhile, but the others are perhaps going beyond the scope of markups into "more work I just made up"!

fasaxc:
@plwhite 

* +1 for being able to grep
* I don't think there are any duplicate stats right now, how about we introduce constants at the point that we start logging the same stat from multiple places?
* Periodic dumping sounds like a good idea and should be straightforward.  I'll either add it to this if it's trivial or follow up with another PR.
* Agree that those extra stats sound good.  How about I do the easy ones and then I/we follow up with additional PRs rather than blocking this from going in?

plwhite:
Thanks. I'm fine with all that - and agree that worth just getting this in, and we can add more later.

plwhite:
@fasaxc This is with you to add markups, so putting ownership back to you.

fasaxc:
@plwhite Adding the message counts took me longer than expected so I've stopped there.  There's a good interface to hook into though, you can register a function to dump some extra diags using `futils.register_diags` 

Switched to using logger calico.stats to make it greppable.

plwhite:
I feel we're going to do lots more iteration on diags, but this is good stuff - lets get it checked in ow.

plwhite:
You forgot to update CHANGES.md, so I did it for you.

Issue #618, opened 2015-06-03
Remove conntrack entries for deleted endpoint?


fasaxc:
As per #616, customer was doing heavy churn, filling up the conntrack table.  One particular issue was that, each time they deleted a VM, they had to wait for the conntrack entries for that VM to time out.  the default timeout is very long so the table filled up.

Can we do something better than recommending a lower timeout.  For example, having felix remove the entries when an endpoint disappears?

fasaxc:
Had a quick look at the conntrack tools.  It looks like we could have felix issue a pair of calls for each removed IP address:
```
conntrack --delete --orig-src <ip>
conntrack --delete --orig-src <ip>
```
Those delete all the conntrack entries that are going to/from the relevant IP.

fasaxc:
@Lukasa, IIRC you tried something like that but it didn't fly for some reason; can you remember what that was?

Lukasa:
I have no recollection of having tried anything like this, so I'm not sure how to help. You sure it was me?

ed-harrison:
@fasaxc is addressing under #623

fasaxc:
With me to have a think if there's any race conditions with a new endpoint being added on the same host with reused IP.

Issue #617, opened 2015-06-03
Protect against full conntrack table


fasaxc:
As per #616 customer was doing heavy churn and managed to fill up the conntrack table.  #616 covers improving the install process to cover setting the conntrack limit appropriately.  This issue covers investigating whether there are other areas we can tune or rules we can install to avoid getting DoSed by a full table.

Presently, when the table fills up, you can't even ssh into the host to investigate and it's possible for a VM to deliberately try to fill up the table by creating lots of connections.

We may just want to document best practice or there may be impacts for our iptables programming to, for example, prioritise traffic to the host over traffic to the guests.

Lukasa:
We've done some documentation and #623, so I'm calling this done for now.

PR #623, opened 2015-06-04
When an IP (or whole endpoint) is removed, remove conntrack entries

fasaxc:
@plwhite this should improve #618 and #617 as well as closing a potential security hole caused by address reuse.

Please can you review, then I'll add UT if you agree it's the right approach?

TODO list:
- [x] Make markups
- [x] add the conntrack tool as a dependency of our debs
- [x] and RPMs
- [x] UT

### DIFF
--- a/calico/felix/devices.py
+++ b/calico/felix/devices.py
@@ -31,7 +31,7 @@ from calico import common
 from calico.felix import futils
 
 # Logger
-import re
+from calico.felix.futils import FailedSystemCall
 
 _log = logging.getLogger(__name__)
 
@@ -230,6 +230,30 @@ def interface_up(if_name):
         return False
 
     return bool(int(flags, 16) & 1)
+
+
+def remove_conntrack_flows(ip_addresses):
+    """
+    Removes any conntrack entries that use any of the given IP
+    addresses in their source/destination.
+    """
+    for ip in ip_addresses:
+        _log.debug("Removing conntrack rules for %s", ip)
+        for direction in ["--orig-src", "--orig-dst",
+                          "--reply-src", "--reply-dst"]:
+            try:
+                futils.check_call(["conntrack", "--delete", direction, ip])
+            except FailedSystemCall as e:
+                if e.retcode == 1 and "0 flow entries" in e.stderr:
+                    # Expected if there are no flows.
+                    _log.debug("No conntrack entries found for %s/%s.",
+                               ip, direction)
+                    continue
+                # Suppress the exception, conntrack entries will timeout and
+                # it's hard to think of an example where killing and
+                # restarting felix would help.
+                _log.exception("Failed to remove conntrack flows for %s. "
+                               "Ignoring.", ip)
 
 
 # These constants map to constants in the Linux kernel. This is a bit poor, but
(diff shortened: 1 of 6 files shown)


calico/felix/devices.py
```diff
@@ -226,6 +226,30 @@ def interface_up(if_name):
...
+    Removes any conntrack entries that use any of the given IP
+    addresses in their source/destination.
+    """
+    for ip in ip_addresses:
+        _log.debug("Removing conntrack rules for %s", ip)
+        for direction in ["--orig-src", "--orig-dst",
+                          "--reply-src", "--reply-dst"]:
+            try:
+                futils.check_call(["conntrack", "--delete", direction, ip])
+            except FailedSystemCall as e:
```
plwhite: My reading of the conntrack man page is that you need to explicitly specify IPv4 or IPv6 (and that it defaults to IPv4, which is why this will work for IPv4).

calico/felix/endpoint.py
```diff
@@ -226,6 +233,15 @@ def on_endpoint_update(self, endpoint):
...
 
+        # When an IP address is removed, squash any old conntrack entries for
+        # that IP address.  This prevents the table from becoming full under
+        # high VM churn and it also prevents a new endpoint that happens to
+        # get a reused IP from spoofing old streams.
+        old_endpoint = self.endpoint or {}
+        old_ip_nets = set(old_endpoint.get(self.nets_key, []))
+        removed_nets = old_ip_nets - new_ip_nets
+        remove_conntrack_flows([futils.net_to_ip(n) for n in removed_nets])
+
```
plwhite: You are removing the conntrack entries while the interface still has them in place. I think it would be better to move this call into devices.set_routes, so you delete the conntrack entry at the same point as you delete the IP address itself.

plwhite:
Couple of markups and back to you. I agree with the general idea of using the conntrack tool, but would like to see it moved into where we set the IP addresses up (devices.set_routes)

You'll also need to update our dependencies / docs (as you say), and add a line to CHANGE.md.

matthewdupre:
@fasaxc can we please get this marked-up, UTed and merged so we can close the issue down?  Thanks.

fasaxc:
@matthewdupre I'll try to get to this today/tomorrow but I don't think it's critical to get it into this release if it slips out.

calico/felix/endpoint.py
fasaxc: Sounds good, I'd thought that set_routes didn't get hit in that case but I was wrong.

fasaxc:
Made my markups, just UTs to fix and add.

fasaxc:
@plwhite @matthewdupre Made my markups and added UT.  Back to you for rereview...

fasaxc:
I haven't tested the deb/RPM changes, I figure the easiest (only?) way to test those is to let it build?

matthewdupre:
Deb / RPM changes look good.  Agreed there's no need for other testing there.

With you Peter to merge.

calico/felix/endpoint.py
```diff
@@ -249,6 +256,14 @@ def on_endpoint_update(self, endpoint):
         if endpoint != self.endpoint:
             self._dirty = True
 
+        # When an IP address is removed, squash any old conntrack entries for
```
plwhite: This is dead code - it should have been removed in the refactor but wasn't.

plwhite:
@matthewdupre I made a trivial markup to Shaun's fix - removing a few lines that he added, then failed to remove when he refactored. Can you please review the very final commit (should be utterly uncontroversial), after which I'll merge. Thanks

Issue #599, opened 2015-06-01
Diags script doesn't gather etcd diagnostics


matthewdupre:
A dump of the database would be useful (both to check its contents, and to check connectivity).

matthewdupre:
Fixed in #627.

PR #627, opened 2015-06-05
Extend diags script: add etcd and rabbit diags

matthewdupre:
Fixes issue #599


matthewdupre:
Thanks - good spot.  I've added a commit switching all the incorrect redirects.

neiljerram:
Looks good, thanks.

Issue #388, opened 2015-04-16
Bird logs spammed with "<WARN> Netlink: File exists"


Symmetric:
The bird logs are spammed with this sort of thing:

```
2015-04-16 20:32:47 <WARN> Netlink: File exists
2015-04-16 20:32:47 <WARN> Netlink: File exists
2015-04-16 20:32:49 <WARN> Netlink: File exists
2015-04-16 20:32:49 <WARN> Netlink: File exists
2015-04-16 20:32:51 <WARN> Netlink: File exists
2015-04-16 20:32:51 <WARN> ...
```

This makes it harder to spot actual issues, and nearly impossible to get a good feel for the general flow of BGP messages. 

(Of course, you can do `grep -v 'Netlink\|\.\.\.' bird.log`, but that's not very user-friendly)

fasaxc:
This looks like a bug in our BIRD config: https://gitlab.labs.nic.cz/labs/bird/wikis/FAQ#netlink-file-exists

We should set BIRD to prefer kernel routes.  @Symmetric do you have a system exhibiting the issue that you can try the fix out on?

Symmetric:
@fasaxc yep, looks like it's still occurring on my dev machine.

ed-harrison:
@matthewdupre Can you have a look at this - I think some of the comments in https://gitlab.labs.nic.cz/labs/bird/wikis/FAQ#netlink-file-exists may relate to the conversation we've been having about route preference anyway.

matthewdupre:
This is being caused by BIRD trying to reinstall the host's interface subnet routes (e.g. 172.18.203.0/24 for most of ours).

matthewdupre:
Fixed.

PR #630, opened 2015-06-05
BIRD config: remove 'device routes' to stop NETLINK: File Exists log spam

matthewdupre:
Issue #388 fix.  Also requires a calico-chef update (and possibly updates to Juju and calico-docker too).

BIRD tries to reinstall the route to the subnet the host is on (e.g. 172.18.203.0/24 for most of our test kit).  This fails, because it already exists, leading to a 'NETLINK: File Exists' log every 2 seconds.

The 'device routes' option in the BIRD config stops BIRD from ignoring these routes.  Honestly, I'm not exactly sure why BIRD still accepts the endpoint routes without 'device routes', but I've tried it and it seems to work, without the log spam.

### DIFF
--- a/etc/bird/calico-bird.conf.template
+++ b/etc/bird/calico-bird.conf.template
@@ -26,7 +26,6 @@ protocol kernel {
   learn;          # Learn all alien routes from the kernel
   persist;        # Don't remove routes on bird shutdown
   scan time 2;    # Scan kernel routing table every 2 seconds
-  device routes;
   import all;
   export all;     # Default is export none
 }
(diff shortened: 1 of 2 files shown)


matthewdupre:
Here's the associated calico-chef change: https://github.com/Metaswitch/calico-chef/pull/29

neiljerram:
I'm not sure about this.  I have a hunch this is related to routing in L3 fabrics, so I'd like to have some reassurance - either expertise or testing - that this change isn't going to break L3 fabric routing, before merging this.

matthewdupre:
OK, I'll try and get some

matthewdupre:
OK, I tried it on the L3 test rig and it works fine too.

neiljerram:
Great then, thanks.

PR #632, opened 2015-06-05
Remove a couple of spammy logs.

fasaxc:
The HTTP-related logs get very spammy when etcd is being migrated or rebuilt and those errors are pretty descriptive anyway so the stack is less useful.

The info-level log for endpoint updates is spammy when we're at high scale across the whole cluster.

### DIFF
--- a/calico/felix/ipsets.py
+++ b/calico/felix/ipsets.py
@@ -317,8 +317,6 @@ class IpsetManager(ReferenceManager):
             for ip in old_ips:
                 self._remove_mapping(tag, profile_id, endpoint_id, ip)
 
-        _log.info("Endpoint update complete")
-
     def _add_mapping(self, tag_id, profile_id, endpoint_id, ip_address):
         """
         Adds the given tag->IP->profile->endpoint mapping to the index.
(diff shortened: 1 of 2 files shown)


fasaxc:
@Lukasa  This is the other half of https://github.com/Metaswitch/python-etcd/pull/7/files.  That fix makes the exception propagate but I found we were still spamming the log with not-so-useful stack traces for the IncompleteRead error.  I've trimmed that to only log the exception type/message instead.  I think that's fine because those exceptions are all pretty descriptive.

CCing @ed-harrison

calico/felix/ipsets.py
```diff
@@ -317,8 +317,6 @@ def on_endpoint_update(self, endpoint_id, endpoint):
             for ip in old_ips:
                 self._remove_mapping(tag, profile_id, endpoint_id, ip)
 
-        _log.info("Endpoint update complete")
```
Lukasa: If this log provided any value it probably would have been better to drop the severity rather than remove it, but there's no good reason for that to block this merge.

PR #634, opened 2015-06-08
Handle 0 in a TCP port range by using standard port match.

### LINKED ISSUE #631: Rule with 0 in port range is rejected by kernel, prevents profile update
The multiport iptables extension doesn't like 0 as a tcp port but OpenStack allows 0 to be configured.  We should firewall against this in Felix:

* If 0 really is a valid port for the given profile, we should figure out how to match on it.
* It it is not a valid port, then we should probably just program range 0-n as 1-n instead.

The impact of trying to program a bad rule like this is that the iptables update fails.  We catch that error for profile and endpoint chain updates in order to limit the impact to the single profile or endpoint.

fasaxc:
Extract 0 out when splitting ports into chunks.

Fixes #631.  0 is a reserved but, I suppose, valid port number.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,7 @@
   objects.
 - Allow different hosts to have different interface prefixes for combined
   OpenStack and Docker systems.
+- Add missing support for 0 as a TCP port.
 
 ## 0.23
 
--- a/calico/common.py
+++ b/calico/common.py
@@ -494,7 +494,7 @@ def validate_rule_port(port):
     :returns: None or an error string if invalid
     """
     if isinstance(port, int):
-        if port < 1 or port > 65535:
+        if port < 0 or port > 65535:
             return "integer out of range"
         return None
 
@@ -513,7 +513,7 @@ def validate_rule_port(port):
     except ValueError:
         return "range invalid"
 
-    if start >= end or start < 1 or end > 65535:
+    if start >= end or start < 0 or end > 65535:
         return "range invalid"
 
     return None
--- a/calico/felix/frules.py
+++ b/calico/felix/frules.py
@@ -215,7 +215,11 @@ def _split_port_lists(ports):
     :return list[list[str]]: list of chunks.  If the input is empty, then
                              returns a list containing a single empty list.
     """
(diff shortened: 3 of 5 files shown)


plwhite:
This begs a question for me. What does it mean when OpenStack configures a port of 0? According to google, a port of 0 means "select a dynamic port" - which makes no sense in a firewall rule. I'm a bit concerned that the OpenStack behaviour might be "use port 0 to mean all ports" while we implement something different.

I realise this is mostly just FUD, but can you reassure me that we have implemented the same behaviour as for another OpenStack networking implementation?

Lukasa:
Port 0 is a port that can be set on a packet. It's not strictly valid, and I suspect most implementations of Berkeley sockets won't let you set it if you use the high-level API, but TCP doesn't explode if you set a port of 0.

fasaxc:
@Lukasa is there a reference implementation of OpenStack's security policy that I can compare with?  I remember seeing an agent in the codebase that did iptables rules at one point but can't seem to find it now.

Lukasa:
[Sure is](https://github.com/openstack/neutron/blob/745e5c68678a8c568d68a97336ebcbd932c648d7/neutron/agent/linux/iptables_firewall.py).

fasaxc:
@Lukasa @plwhite They use multiport, so either it blows up or they're filtering it out elsewhere.

plwhite:
I agree with the firewall changes, but I'm confused. I tried explicitly configuring 0 in a multiport rule, and it just worked. That seems to match what OpenStack are expecting. What am I missing here?

This is ubuntu 14.04, incidentally.

fasaxc:
@plwhite  Interesting, 0 on its own is accepted but 0 in a range is rejected.

fasaxc:
D'oh, the straightforward match has the same limitation, I must have tried it with a single value rather than a range too!

fasaxc:
@plwhite scratch that, I had the syntax wrong.  Both seem to allow ranges and single values.  Damn!

fasaxc:
Closing this, looks like I just need to change the validation and it'll all work!

PR #636, opened 2015-06-09
Separate FORWARD and INPUT chains in Felix

### LINKED ISSUE #196: Security hole in dhcp iptables rule
Currently, this rule is always assigned to a VM, regardless of security policy:
-A felix-from-dcf2bb9c-19 -p udp -m udp --sport 68 --dport 67 -j RETURN

The problem with this rule is that it does not specify any source and/or destination, so certain traffic could be allowed to/from the public internet (assuming that the IP assigned to the VM is public) on those ports, even if the security policy is setup to disallow that.  While they are typically used only for DHCP, this does not prevent someone from setting up a different service using these ports to get around security policies.

I believe a better rule would be something like:
-A felix-from-dcf2bb9c-19 -p udp -m udp -s 0.0.0.0 --sport 68 -d 255.255.255.255 --dport 67 -j RETURN

This rule should prevent off-subnet use.  This rule doesn't catch all DHCP traffic, but the rest (that has the actual IP as the source address) would get allowed by the other rules.

Lukasa:
This resolves #196.

The effect of this patch is to stop adding felix-FROM-ENDPOINT to the INPUT chain of the filter table in iptables, and instead to add another top-level chain that contains only rules that affect inbound packets. This creates a split between the endpoint security rules (which apply only to forwarded packets) and the rules that affect traffic intended to reach the host.

Right now this is set up to be extremely secure, but we may want to relax it slightly. ICMP packets from guests leap to mind as a possible exemption we'll want to grant.

I have to drop a few extra commits that will move some other rules from where they are to this new INPUT chain. In particular, the metadata rule needs to move in.

I haven't had a chance to live test this yet, and the UT coverage here isn't great, so I suggest not merging it until someone gets a chance to live-test it. Nevertheless, review would be welcome, particularly from @plwhite and @fasaxc.

### DIFF
--- a/calico/felix/dispatch.py
+++ b/calico/felix/dispatch.py
@@ -165,16 +165,16 @@ class DispatchChains(Actor):
         # Special case: allow the metadata IP through from all interfaces.
         if self.config.METADATA_IP is not None and self.ip_version == 4:
             # Need to allow outgoing Metadata requests.
-            root_from_upds.append("--append %s "
-                                  "--protocol tcp "
-                                  "--in-interface %s+ "
-                                  "--destination %s "
-                                  "--dport %s "
-                                  "--jump RETURN" %
-                                  (CHAIN_FROM_ENDPOINT,
-                                   self.config.IFACE_PREFIX,
-                                   self.config.METADATA_IP,
-                                   self.config.METADATA_PORT))
+            root_inbound_upds.append("--append %s "
+                                     "--protocol tcp "
+                                     "--in-interface %s+ "
+                                     "--destination %s "
+                                     "--dport %s "
+                                     "--jump RETURN" %
+                                     (CHAIN_INBOUND,
+                                      self.config.IFACE_PREFIX,
+                                      self.config.METADATA_IP,
+                                      self.config.METADATA_PORT))
 
         # Special case: allow DHCP inbound from all interfaces.
         if self.ip_version == 4:
--- a/calico/felix/dispatch.py
+++ b/calico/felix/dispatch.py
@@ -24,7 +24,7 @@ import logging
 from calico.felix.actor import Actor, actor_message, wait_and_check
 from calico.felix.frules import (
     CHAIN_TO_ENDPOINT, CHAIN_FROM_ENDPOINT, CHAIN_FROM_LEAF, CHAIN_TO_LEAF,
-    CHAIN_INBOUND, chain_names, interface_to_suffix
+    chain_names, interface_to_suffix
 )
 
 _log = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


Lukasa:
It'll almost certainly be easiest to review this change as separate commits working towards the final goal, rather than looking at the diff as a whole.

fasaxc:
@Lukasa sounds like a good idea, I'll try to give it a look-over today.

Lukasa:
Ok, I've done a quick bit of live test and this appears to function fine, without explosions and death (by which I mean DHCP continues to work).

Lukasa:
Temporarily assigning to @plwhite to review as he's got the right time zone, though @plwhite should feel free to reassign if it's inconvenient.

calico/felix/dispatch.py
```diff
@@ -164,16 +165,36 @@ def _calculate_update(self, ifaces):
...
-                                  "--protocol tcp "
-                                  "--in-interface %s+ "
-                                  "--destination %s "
-                                  "--dport %s "
-                                  "--jump RETURN" %
-                                  (CHAIN_FROM_ENDPOINT,
-                                   self.config.IFACE_PREFIX,
-                                   self.config.METADATA_IP,
-                                   self.config.METADATA_PORT))
+            root_inbound_upds.append("--append %s "
```
fasaxc: Sorry, didn't get to this yesterday; is there any reason this chain needs to be managed by the dispatch chain actor, seems like it's a static chain so we could even write the contents directly into felix-INPUT?

The reason might be that you can see things that we'll want to put in here later that depend on the list of tap devices.

fasaxc:
One comment from me in the code.  Only other thought is, will this break ping and ssh from the host to the VM, which I think we've agreed is a feature.

calico/felix/dispatch.py
Lukasa: The honest answer is that I took the path that made this look isomorphic to the rest of the work being done in Felix. It seemed like it didn't really cost anything, and had the advantage of providing a useful hook for extension.

If you want to take a different approach, that suits me.

Lukasa:
> Only other thought is, will this break ping and ssh from the host to the VM, which I think we've agreed is a feature.

No, conntrack ought to save those. It may not save ping, but I already asked whether we should add an ICMP exemption here.

calico/felix/dispatch.py
plwhite: I think we should just put all of the input rules directly into felix-INPUT - that to me seems like the simplest approach.

plwhite:
I'm a bit confused, but I think the right design is as follows.

Prior to this fix, we have `felix-INPUT` that references all the tap interfaces.

After the fix we should have `felix-INPUT` that contains : 

* conntrack rules (so ssh and ping to guests can work)
* DHCP rules
* Metadata rules

Somewhat to my surprise, that is all that should be in `felix-INPUT`; we should just not have any of the reference to `felix-FROM-ENDPOINT`.

I am about to review on this basis.

plwhite:
OK, so I think it will be simpler to keep going and finish this with just the minimal rules above - which means deleting some code that does unnecessary stuff. But that is a simpler, safer and more correct solution so I think we should hold off.

Back to you @Lukasa (but @fasaxc, if you have any comments let me know!).

fasaxc:
@Lukasa I think @plwhite meant to assign to you.

fasaxc:
@Lukasa +1 for ICMP exception from the host, I think.  That's useful.

Lukasa:
@fasaxc Can you clarify: do you mean an exception allowing ICMP traffic host -> guest? Because I was talking about an exception allowing ICMP guest -> host.

Lukasa:
Ok, rebased on top of the ip-in-ip work.

Lukasa:
Hmm, seems like the pull request build job is busted. @matthewdupre, you might want to take a look at that...

fasaxc:
@Lukasa I was meaning host to guest.  It's really useful to be able to ssh or ping the guest from the host.

matthewdupre:
@Lukasa I don't understand how pull-request-build is broken?  It's passed, the console output looks fine, and there are packages in the right place.  What am I missing?

Lukasa:
Yah, sorry, that was an old message. We fixed it. =)

Lukasa:
@fasaxc @plwhite Argh, this languished with me. I think this is good to go, so can I get another round of review please?

calico/felix/frules.py
```diff
@@ -148,9 +159,10 @@ def install_global_rules(config, v4_filter_updater, v6_filter_updater,
             },
             {
                 CHAIN_FORWARD: set([CHAIN_FROM_ENDPOINT, CHAIN_TO_ENDPOINT]),
-                CHAIN_INPUT: set([CHAIN_FROM_ENDPOINT]),
-            },
-            async=False)
+                CHAIN_INPUT: set(),
+            }
+        )
+
```
plwhite: You removed "async=False" here. Why was that?

In practice I think it is benign (you put some stuff on the iptables updater queue, then put something else on and wait for it to be processed - so it doesn't matter because when you wait for the second, the first must have been completed). However, less confusing to leave it as it was.

plwhite:
One comment to deal with, otherwise looking good.

I also note that you do not include conntrack rules in this chain. I think that is in fact correct, despite my earlier remarks which were completely mad.

calico/felix/frules.py
Lukasa: The answer is 'because I'm an idiot'. =D

Lukasa:
The missing conntrack was an accident. I think we want it though, so I'll add conntrack back in.

Lukasa:
Changes made.

PR #637, opened 2015-06-09
Use neutron agent-delete in install instructions

### LINKED ISSUE #517: Add "neutron agent-delete" to docs
Our docs say to shut down the various agents that calico doesn't require.  However, that doesn't clean up their presence in the neutron database.  The entries are created on demand by the agents when they check in for the first time.

The "neutron agent-delete" command removes those entries.  Doing so prevents all the now-stopped agents from appearing in the list and being reported as down.

Lukasa:
Resolves #517.

### DIFF
--- a/docs/source/redhat-opens-install.rst
+++ b/docs/source/redhat-opens-install.rst
@@ -314,15 +314,6 @@ On each control node, perform the following steps:
 
        service neutron-server restart
 
-7. Find all the currently running OpenStack Neutron agents and stop them.
-   First, run the following command to find them all::
-
-       neutron agent-list
-
-   Then, for each agent, delete them with the following command, replacing
-   ``<agent-id>`` with the ID of the agent::
-
-       neutron agent-delete <agent-id>
 
 Compute Node Install
 --------------------
@@ -400,6 +391,16 @@ On each compute node, perform the following steps:
 
            chkconfig openvswitch off
            chkconfig neutron-openvswitch-agent off
+
+   Then, on your control node, run the following command to find the agents
+   that you just stopped::
+
+       neutron agent-list
+
+   For each agent, delete them with the following command on your control node,
+   replacing ``<agent-id>`` with the ID of the agent::
+
+       neutron agent-delete <agent-id>
 
 4. Run ``yum update``. This will bring in Calico-specific updates to the
    OpenStack packages and to ``dnsmasq``.
--- a/docs/source/redhat-opens-install.rst
+++ b/docs/source/redhat-opens-install.rst
@@ -314,6 +314,16 @@ On each control node, perform the following steps:
 
        service neutron-server restart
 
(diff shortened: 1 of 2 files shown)


ed-harrison:
I'm afraid I don't feel qualified to comment on whether this is correct.  In particular, #517 suggests that we should modify the instructions which stop other agents - however, these instructions are still present after this PR (the steps are in the "Compute Node" part of the install).

Lukasa:
Ok, can someone else let me know whether we need to make Ed's suggested change? I choose Neil!

docs/source/redhat-opens-install.rst
```diff
@@ -314,6 +314,16 @@ On each control node, perform the following steps:
 
        service neutron-server restart
 
+7. Find all the currently running OpenStack Neutron agents and stop them.
+   First, run the following command to find them all::
+
+       neutron agent-list
+
+   Then, for each agent, delete them with the following command, replacing
+   ``<agent-id>`` with the ID of one of the agents::
```
alexwlchan: Should this be “the ID of the agent”? That seems like it would fit better with “for each agent”.

ed-harrison: +1

docs/source/ubuntu-opens-install.rst
```diff
@@ -236,6 +236,16 @@ perform the following steps:
 
         sudo service neutron-server restart
 
+6. Find all the currently running OpenStack Neutron agents and stop them.
+   First, run the following command to find them all::
+
+       neutron agent-list
+
+   Then, for each agent, delete them with the following command, replacing
+   ``<agent-id>`` with the ID of one of the agents::
```
alexwlchan: This sentence needs the same change as above

docs/source/redhat-opens-install.rst
```diff
@@ -314,6 +314,16 @@ On each control node, perform the following steps:
...
 
+7. Find all the currently running OpenStack Neutron agents and stop them.
+   First, run the following command to find them all::
+
+       neutron agent-list
+
+   Then, for each agent, delete them with the following command, replacing
+   ``<agent-id>`` with the ID of the agent::
+
+       neutron agent-delete <agent-id>
```
fasaxc: I think that the agents need to be stopped using systemdctl or upstart's equivalent before you remove them from the DB.  Otherwise the DB will get repopulated as soon as the agent reconnects.

Lukasa: Fair enough, I'll move the note.

Lukasa:
Ok, I've updated the pull request with those markups. =)

matthewdupre:
Package build is actually passing - it's build 681, not 675.  Bit worrying that it's looking up the wrong build though...

alexwlchan:
@matthewdupre The calico-jenkins build links have been out-of-sync with pull requests on GitHub for a while now…

PR #479, opened 2015-04-29
Address some RHEL 7 install instruction issues:

fasaxc:
- improve some wording.
- remove dependency on wget, which isn't installed by default
- add missing yum install lines for python-etcd deps
- use uuidgen to generate a unique etcd cluster token
- fix etcd mount point and use "mount -a" rather than a manual
  mount command.

PR #638, opened 2015-06-09
Use uuidgen in Ubuntu openstack install instructions

### LINKED ISSUE #478: When using etcd as a transient store need to give it a new cluster ID each time
Our install guide puts etcd in a ramdisk but we set the initial-cluster-token only at install time.  This is wrong because, each time etcd gets rebuilt in the ramdisk, it will use the same token but it will start with a  fresh etcd index.  That will break felix's watches if it all happens between felix polls.  That's a small window to hit but we should close it.

I've investigated how to do this on RHEL 7.x and it looks like the /etc/sysconfig/etcd does not allow dynamic config so we need to use a wrapper script.

I tried putting this simple wrapper script in usr/local/bin/start-etcd:
```
#!/bin/sh
export ETCD_INITIAL_CLUSTER_TOKEN=`uuidgen`
exec /usr/local/bin/etcd
```
chmodding it +x and then changing the systemd unit file (/usr/lib/systemd/system/etcd.service) to use it:
```
[Unit]
Description=Etcd
After=syslog.target network.target

[Service]
User=root
ExecStart=/usr/local/bin/start-etcd
EnvironmentFile=-/etc/sysconfig/etcd
KillMode=process
Restart=always

[Install]
WantedBy=multi-user.target
```
And I saw it generate anew cluster ID each time.

Lukasa:
This should finally resolve #478. It's basically a straight port of the fix from #479, adapted for upstart.

### DIFF
--- a/docs/source/ubuntu-opens-install.rst
+++ b/docs/source/ubuntu-opens-install.rst
@@ -133,14 +133,11 @@ please get in touch with us and we'll be happy to help you through the process.
 
     tmpfs /var/lib/etcd tmpfs nodev,nosuid,noexec,nodiratime,size=512M 0 0
 
-
 6. Edit ``/etc/init/etcd.conf``:
 
    - Find the line which begins ``exec /usr/bin/etcd`` and edit it,
-     substituting for ``<controller_fqdn>``, ``<controller_ip>``, and
-     ``<cluster_token>`` appropriately. Each time you create a new etcd cluster
-     you should use a new cluster_token, as this is checked by Calico
-     components to check whether etcd has moved.
+     substituting for ``<controller_fqdn>`` and ``<controller_ip>``
+     appropriately.
 
      ::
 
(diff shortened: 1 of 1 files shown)


fasaxc:
@Lukasa turns out Upstart lets you use shell commands in upstart jobs.  This works for me:
```
description "etcd"
author "etcd maintainers"

start on stopped rc RUNLEVEL=[2345]
stop on runlevel [!2345]

respawn

setuid etcd

env ETCD_DATA_DIR=/var/lib/etcd
export ETCD_DATA_DIR

exec /usr/bin/etcd --name calicobobcatci01gulfstreamdatconcouk                                                      \
                   --advertise-client-urls http://172.24.193.128:2379,http://172.24.193.128:4001      \
                   --listen-client-urls http://0.0.0.0:2379,http://0.0.0.0:4001                                          \
                   --listen-peer-urls http://0.0.0.0:2380                                                                \
                   --initial-advertise-peer-urls http://172.24.193.128:2380                                \
                   --initial-cluster-token $(uuidgen)                                                      \
                   --initial-cluster calicobobcatci01gulfstreamdatconcouk=http://172.24.193.128:2380      \
                   --initial-cluster-state new
```

Lukasa:
> turns out Upstart lets you use shell commands in upstart jobs

I can see why upstart got retired, then. :wink:

That said, this way is simpler, so let's use it.

Issue #635, opened 2015-06-09
Updated security group rules but no etcd update


fasaxc:
I updated the security group rules in the default group through horizon but Calico didn't seem to notice (I used `curl -v 'http://localhost:4001/v2/keys/calico?wait=true&stream=true&recursive=true'` to do a watch but saw no updates).  Updating rules in another group did work and I saw updates as expected.

The group is in use by a VM: http://calicobobcat-ci01/horizon/admin/instances/da8993cd-b433-468e-95a4-0b1ecb823d34/detail

@ed-harrison @neiljerram calicobobcat-ci01 still seems to be in the bad state.  Can someone with neutron debugging experience have a look at this tomorrow?  Just jump on the bobcat rig and debug there, it won't be in use during UK hours.  It's similar to the issue that a customer hit and one that we've hit in the SG churn tests on the bobcat rig.

Output to neutron log when I delete a security rule from the default group:
```
2015-06-08 17:13:15.001 29163 INFO neutron.wsgi [req-207a0fcd-322b-4ad5-afbb-605cd5e23d0d None] (29163) accepted ('172.24.193.128', 50492)

2015-06-08 17:13:15.002 29163 DEBUG keystoneclient.middleware.auth_token [-] Authenticating user token __call__ /usr/lib/python2.7/dist-packages/keystoneclient/middleware/auth_token.py:569
2015-06-08 17:13:15.003 29163 DEBUG keystoneclient.middleware.auth_token [-] Removing headers from request environment: X-Identity-Status,X-Domain-Id,X-Domain-Name,X-Project-Id,X-Project-Name,X-Project-Domain-Id,X-Project-Domain-Name,X-User-Id,X-User-Name,X-User-Domain-Id,X-User-Domain-Name,X-Roles,X-Service-Catalog,X-User,X-Tenant-Id,X-Tenant-Name,X-Tenant,X-Role _remove_auth_headers /usr/lib/python2.7/dist-packages/keystoneclient/middleware/auth_token.py:628
2015-06-08 17:13:15.003 29163 DEBUG keystoneclient.middleware.auth_token [-] Returning cached token _cache_get /usr/lib/python2.7/dist-packages/keystoneclient/middleware/auth_token.py:1034
[...]

PR #641, opened 2015-06-10
Setup the CalicoNotifierProxy early

Lukasa:
Ok, this is a proposed fix for #635. I'm going to try to test this today, but it'll not necessarily be all that easy to test, so @fasaxc may want to run this test as well.

### DIFF
--- a/calico/openstack/mech_calico.py
+++ b/calico/openstack/mech_calico.py
@@ -36,6 +36,9 @@ from neutron.plugins.ml2.drivers import mech_agent
 from neutron import context as ctx
 from neutron import manager
 
+# Monkeypatch import
+import neutron.plugins.ml2.rpc as rpc
+
 try:  # Icehouse, Juno
     from neutron.openstack.common import log
 except ImportError:  # Kilo
@@ -65,6 +68,15 @@ STARTUP_DELAY_SECS = 30
 SecurityProfile = namedtuple(
     'SecurityProfile', ['id', 'inbound_rules', 'outbound_rules']
 )
+
+
+# This terrible global variable points to the running instance of the
+# Calico Mechanism Driver. This variable relies on the basic assertion that
+# any Neutron process, forked or not, should only ever have *one* Calico
+# Mechanism Driver in it. It's used by our monkeypatch of the
+# security_groups_rule_updated method below to locate the mechanism driver.
+# TODO: Let's not do this any more. Please?
+mech_driver = None
 
 
 def requires_state(f):
@@ -138,6 +150,10 @@ class CalicoMechanismDriver(mech_agent.SimpleAgentMechanismDriverBase):
         self._periodic_resync_greenlet = None
         self._epoch = 0
 
+        # Tell the monkeypatch where we are.
+        global mech_driver
+        mech_driver = self
+
         # Make sure we initialise even if we don't see any API calls.
         eventlet.spawn_after(STARTUP_DELAY_SECS, self._init_state)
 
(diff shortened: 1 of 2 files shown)


fasaxc:
@Lukasa I thought that we needed to make the DB connection _after_ we're forked to avoid sharing the socket?

Lukasa:
We do, but this is a terribly named variable: it doesn't contain any DB connections. If you note we only use it with methods that explicitly pass a `context` object, and *that* object is where the DB connections are. So I think we'll be safe here, but I do want us to beat on it for a while to be sure.

(If I'm right, I should enhance this patch to call the variable what it actually is: `plugin`.)

fasaxc:
Gave this a spin on the bobcat rig.  Sadly, it deadlocks immediately.  I think the call to `NeutronManager.get_plugin()` tries to grab the manager lock to create the plugin but the manager lock is already held to do the initialisation (and apparently isn't reentrant).

Our subsequent timer pop to initialise the driver then deadlocks.

fasaxc:
Would it be crazy to monkey-patch that class rather than interposing a proxy?

matthewdupre:
What's the status of this pull request?

fasaxc:
@matthewdupre I tried it out but it didn't work.  Back to @Lukasa for a respin.  I think we need to fix this for the next release, it harms our reliability when we're being used in a low-churn test set-up.  That's the first thing anyone is going to do.

It's also really easy to hit when you're doing manual testing, it just happens that our FV tests tend to create enough VMs to kickstart the plugin before they mess around with SGs.

Lukasa:
No progress since @fasaxc's updates, it's not currently my highest priority.

> Would it be crazy to monkey-patch that class rather than interposing a proxy?

I don't think we can do that. The class needs a reference to a single instance of the mechanism driver, which needs to be the one running in the forked process in question. I don't *think* we can safely assume that the monkeypatched class will survive forking well.

It's *possible* we can come up with a solution where we say there is always only ever one CalicoMechanismDriver in a process, and then have a simple lookup table, but I don't *know* that that's going to be accurate, either now or in the future.

Lukasa:
Yup, this definitely doesn't work. =(

Lukasa:
There is an extra option here: we can actually consume the RPC messages that should be being generated for security group changes.

Lukasa:
Ok, so here's basically the problem as I understand it.

First, our current method of snaffling up security group updates involves attaching a proxy object to the ML2Plugin object. It seems to be extremely unreliable to do this before forking: attempting to do an `eventlet.spawn` call to it doesn't work, presumably because the fork happens while that eventlet is outstanding and the fork causes that eventlet to get blown up (thanks for the suggestion there @neiljerram).

The simplest way to do this *for now* is to do some truly terribly monkeypatching with global state, I think. In some quick tests on my end this appeared to work ok, which is good enough for now, but we should definitely give it a battering.

Longer term, we should think about whether the right thing to do is to actually get onto the end of the RPC queue that Neutron normally uses to handle security group updates. That's the architecturally correct approach, rather than taking the *extremely brittle* method of monkeypatching an object.

Lukasa:
Ok, this patch includes the monkeypatch solution.

I hate it, but a bit of live test suggests it might work. :disappointed_relieved:

@fasaxc, you'll probably want to test with this. Separately, we should aim to find a way to do this properly.

Lukasa:
Note that I've checked that the code is at least in the same place in Icehouse, Juno, and Kilo, so this 'fix' has a reasonable shot at working across all versions.

fasaxc:
Tested this on Ubuntu 14.04 icehouse and it seems to work.  I can restart neutron and then immediately make SG changes, which are picked up.

calico/openstack/mech_calico.py
```diff
@@ -138,6 +150,10 @@ def __init__(self):
         self._periodic_resync_greenlet = None
         self._epoch = 0
 
+        # Tell the monkeypatch where we are.
+        global mech_driver
```
fasaxc: Worth asserting that it's None so we catch it if our assumption gets broken?

Lukasa: Yeah, why not. Fixed in 8387191fa7f77c5b4991d64efdcd1d23237ccb3c.

Lukasa:
retest this please

Lukasa:
Merging this because @ed-harrison has waved it through. Retroactive code review from @fasaxc would still be valuable.

fasaxc:
LFTM*

* Looks Finessed To Me

PR #648, opened 2015-06-10
Add support for IP-in-IP interfaces.

fasaxc:
Allows Calico to be evaluated without reconfiguring the underlying fabric to peer with Calico.

This is split into two commits, the first does some refactoring to make it easier to re-use the ipsets code.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,8 @@
 
 ## 0.24-dev
 
+- Add support for routing over IP-in-IP interfaces in order to make it 
+  easier to evaluate Calico without reconfiguring underlying network.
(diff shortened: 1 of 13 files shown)


plwhite:
A reasonable amount of detailed feedback in line. For some reason it seems to have pulled my feedback out from the context here, but you should see context if you look at the two separate commits (I reviewed separately).

In summary, nothing particularly exciting, and should be OK to merge as soon as you have applied the (minor) feedback. Since we are in a hurry and the review comments are relatively minor for the most part, I would be OK with you merging yourself once you've applied feedback.

fasaxc:
Thanks @plwhite, I've addressed most of the feedback and added UT.  I'm going to drive the bird_ip issue separately.

PR #650, opened 2015-06-11
Add support for arbitrary IP protocols

fasaxc:
@plwhite I had to create a new PR with my code review markups.  This time it's based on master so good to merge once you're happy with it.

### DIFF
--- a/calico/common.py
+++ b/calico/common.py
@@ -46,6 +46,9 @@ FORMAT_STRING = '%(asctime)s [%(levelname)s][%(process)s/%(tid)d] %(name)s %(lin
 # string used by the logger.
 SYSLOG_FORMAT_STRING = '{excname}[%(process)s]: %(module)s@%(lineno)d %(message)s'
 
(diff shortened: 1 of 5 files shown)


calico/common.py
```diff
@@ -46,6 +46,21 @@
...
+KERNEL_PROTOCOLS.update(intern(str(p)) for p in xrange(1, 256))
+
+# Protocols that support a port match in iptables.  We allow the name and
+# protocol number.
+KERNEL_PORT_PROTOCOLS = set(["tcp", 6,
+                             "udp", 17,
+                             "udplite", 136,
+                             "sctp", 132,
+                             "dccp", 33])
+
```
plwhite: I think you need "6" in the list as well as 6; you are allowing an integer protocol to be passed as a string above, but not here.

plwhite:
Happy with this, but one trivial markup. I don't need to see this again - go ahead and make that markup and then you can merge yourself.

PR #651, opened 2015-06-12
Add IP masquerade support for private IPAM pools.

fasaxc:
@plwhite / @ed-harrison  Please can you coordinate to get this reviewed on Friday?  It's a dependency for the high-priority work that the calico-docker team is doing so we'd like to get it into the 0.24 release.

The change is to have felix monitor calico-docker's IPAM pools, watching for a new "masquerade" flag on the pool and keeping an ipset up-to-date with the set of pools that have masquerading enabled.  It then programs an iptables rule to masquerade traffic that is coming from a masqueraded pool, going to any non-calico IP (i.e. one that's not in any pool).

This is pretty tactical and it's an unsupported feature that should have no impact on OpenStack.  one concern I have is that we're watching ipam/v4/pool/..., which seems a bit internal but I don't think we'll get this all together if we start changing that right now.

It could do with a few more comments and I'll write some UTs tomorrow if you're broadly happy with it.

### DIFF
--- a/calico/common.py
+++ b/calico/common.py
@@ -82,6 +82,11 @@ KNOWN_RULE_KEYS = set([
 # valid characters is the same for endpoints, profiles, and tags.
 VALID_ID_RE = re.compile('^[a-zA-Z0-9_\.\-]+$')
 
+VALID_LINUX_IFACE_NAME_RE = re.compile(r'^[a-zA-Z0-9_]{1,15}$')
+
+# Not that thorough: we don't care if it's a valid CIDR, only that it doesn't
+# have anything malicious in it.
+VALID_IPAM_POOL_ID_RE = re.compile(r'^[0-9\.:a-fA-F\-]{1,43}$')
 
 tid_storage = gevent.local.local()
 tid_counter = itertools.count()
@@ -569,3 +574,41 @@ def validate_tags(profile_id, tags):
 
     if issues:
         raise ValidationFailed(" ".join(issues))
+
+def validate_ipam_pool(pool_id, pool, ip_version):
+    # Remove any keys that we're not expecting.  Stops unvalidated data from
+    # slipping through.  We ignore other keys since this structure is used
+    # by calicoctl for its own purposes too.
+    keys_to_remove = set()
+    for key in pool:
+        if key not in ("cidr", "ipip", "masquerade"):
+            keys_to_remove.add(key)
+    for key in keys_to_remove:
+        pool.pop(key)
+
+    issues = []
+    if "cidr" not in pool:
+        issues.append("'cidr' field is missing")
+    else:
+        cidr = pool["cidr"]
+        if cidr is None or not validate_cidr(cidr, ip_version):
+            issues.append("Invalid CIDR: %r" % cidr)
+        else:
+            pool["cidr"] = canonicalise_cidr(cidr, ip_version)
+
+    if "ipip" in pool:
+        ipip_name = pool["ipip"]
+        if not isinstance("ipip", StringTypes):
+            issues.append("'ipip' field should be a string, not %r" %
+                          ipip_name)
+        elif not VALID_LINUX_IFACE_NAME_RE.match(ipip_name):
+            issues.append("Invalid interface name for 'ipip': %r" % ipip_name)
+
+    if not isinstance(pool.get("masquerade", False), bool):
+        issues.append("Invalid 'masquerade' field: %r" % pool["masquerade"])
+
+    if not VALID_IPAM_POOL_ID_RE.match(pool_id):
+        issues.append("Invalid pool ID: %r" % pool)
+
+    if issues:
+        raise ValidationFailed(','.join(issues))
--- a/calico/common.py
+++ b/calico/common.py
@@ -87,7 +87,7 @@ VALID_LINUX_IFACE_NAME_RE = re.compile(r'^[a-zA-Z0-9_]{1,15}$')
 # Not that thorough: we don't care if it's a valid CIDR, only that it doesn't
 # have anything malicious in it.
 VALID_IPAM_POOL_ID_RE = re.compile(r'^[0-9\.:a-fA-F\-]{1,43}$')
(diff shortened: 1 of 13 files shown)


plwhite:
I may not have time to do this today (I am about to go to visit a customer and won't get back until late afternoon). Passing ownership over to you @ed-harrison to find somebody else.

calico/common.py
```diff
@@ -547,3 +552,41 @@ def validate_tags(profile_id, tags):
...
     if issues:
         raise ValidationFailed(" ".join(issues))
+
+def validate_ipam_pool(pool_id, pool, ip_version):
+    # Remove any keys that we're not expecting.  Stops unvalidated data from
+    # slipping through.  We ignore other keys since this structure is used
+    # by calicoctl for its own purposes too.
+    keys_to_remove = set()
+    for key in pool:
+        if key not in ("cidr", "ipip", "masquerade"):
```
Lukasa: Nothing wrong with this, but we should move this list to a variable with a helpful name. It's a bit magic where it is now.

calico/felix/splitter.py
```diff
@@ -55,23 +56,28 @@ def apply_snapshot(self, rules_by_prof_id, tags_by_prof_id,
...
 
         # Step 2: fire in update events into the endpoint manager, which will
         # recursively trigger activation of profiles and tags.
-        _log.info("Applying snapshot. STAGE 2: endpoints->endpoint mgr.")
+        _log.info("Applying snapshot. Queueing endpoints->endpoint mgr.")
         for ep_mgr in self.endpoint_mgrs:
             ep_mgr.apply_snapshot(endpoints_by_id, async=True)
 
+        # Step 3: send update to NAT manager.
+        self.ipv4_nat_manager.apply_snapshot(ipv4_pools_by_id, async=True)
```
matthewdupre: Should have a corresponding log, like the other steps.

calico/felix/splitter.py
```diff
@@ -33,18 +33,19 @@ class UpdateSplitter(Actor):
     and IPv6-specific actors.
     """
     def __init__(self, config, ipsets_mgrs, rules_managers, endpoint_managers,
-                 iptables_updaters):
+                 iptables_updaters, ipv4_nat_manager):
```
matthewdupre: This is a very unclear variable name - the caller passes 'v4_masq_manager' to this, but also has a 'v4_nat_updater'.  My first reading was that this would be the latter (which wouldn't work), not the former.  I suggest using [ip]v4_masq_manager or similar instead.

calico/felix/masq.py
```diff
@@ -0,0 +1,86 @@
...
+        # Calico-owned pool.  (We assume that NAT is not required for
+        # Calico-owned IPs.)
+        self._iptables_mgr.ensure_rule_inserted(
+            "POSTROUTING "
+            "--match set --match-set %s src "
+            "--match set ! --match-set %s dst "
+            "--jump MASQUERADE" % (MASQ_POOLS_SET_NAME,
+                                   ALL_POOLS_SET_NAME),
+            async=True
+        )
```
matthewdupre: This rule will have a data plane performance impact in the mainline case when we aren't using this feature.  Is this acceptable?  We at least need to retest it (or change the spec for this PR so the rule isn't created).

calico/felix/masq.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+ALL_POOLS_SET_NAME = FELIX_PFX + "all-ipam-pools"
+MASQ_POOLS_SET_NAME = FELIX_PFX + "masq-ipam-pools"
+
+
+class MasqueradeManager(Actor):
+    def __init__(self, ip_type, iptables_mgr):
+        super(MasqueradeManager, self).__init__(qualifier=str(ip_type))
+        assert ip_type in (IPV4, IPV6)
```
matthewdupre: We only actually support IPv4 at present.  I don't think this will work with IPv6 anyway, so little point pretending like it might (the rule we install only uses one pair of ipsets, when I think we'd need v4 and v6 versions?).  I'm also not entirely sure if IPv6 NAT is enough of a thing that it even makes sense to do this.

calico/felix/masq.py
fasaxc: Sounds good, I'll make it so the rule is only present if needed.

calico/common.py
```diff
@@ -547,3 +552,41 @@ def validate_tags(profile_id, tags):
...
+    if "cidr" not in pool:
+        issues.append("'cidr' field is missing")
+    else:
+        cidr = pool["cidr"]
+        if cidr is None or not validate_cidr(cidr, ip_version):
+            issues.append("Invalid CIDR: %r" % cidr)
+        else:
+            pool["cidr"] = canonicalise_cidr(cidr, ip_version)
+
+    if "ipip" in pool:
```
matthewdupre: This doesn't really belong in this commit surely?

matthewdupre:
OK, made a few comments, otherwise looks fine.  Back to you for markups and UTs.

calico/felix/splitter.py
fasaxc: Yep, missed this when I was renaming from an earlier revision.

fasaxc:
@matthewdupre markups pushed, still working on UT.  I've given it a spin on my test system and watched the rules come and go as I add and remove pools that need masq.

calico/common.py
fasaxc: We think we'll use this soon (to decide whether to turn on the IPIP function for a particular pool or not) but it's not yet used.

fasaxc:
@matthewdupre UTs pushed.  Assuming you've gone home, will discuss here whether to merge locally or wait for a re-review on Monday.

fasaxc:
@robbrockbank Please can you sanity-check this (especially my markups in the most recent commits)?

robbrockbank:
LGTM

fasaxc:
Ta.  @matthewdupre we merged this after Rob gave it a sanity check so it would get an FV run before Monday.  You and @plwhite are a lot closer to this code, though, so feel free to give it another look over.

PR #653, opened 2015-06-15
Allow failure to install diag dump handler

Lukasa:
The diag-dump function broke unit tests on Windows. More importantly, there's no real reason to worry if we fail to install the diag dump handler, so just allow that to pass without anything more than a warning level log.


fasaxc:
LGTM

PR #655, opened 2015-06-15
Update RHEL install instructions.

matthewdupre:
I did a RHEL install, but there were a few things in the instructions that were unecessary or slightly wrong.

There's also some inconsistency about use of systemctl vs service (and the equivalent chkconfig vs systemctl enable).  I haven't fixed this yet.

### DIFF
--- a/docs/source/redhat-opens-install.rst
+++ b/docs/source/redhat-opens-install.rst
@@ -397,13 +397,7 @@ On each compute node, perform the following steps:
 4. Run ``yum update``. This will bring in Calico-specific updates to the
    OpenStack packages and to ``dnsmasq``.
 
-5. Install build dependencies:
-
-   ::
-
-       yum groupinstall 'Development Tools'
-
-6. Install and configure the DHCP agent on the compute host:
+5. Install and configure the DHCP agent on the compute host:
 
    ::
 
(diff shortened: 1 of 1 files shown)


docs/source/redhat-opens-install.rst
```diff
@@ -111,6 +111,9 @@ please get in touch with us and we'll be happy to help you through the process.
 
    - Download, unpack, and install the binary::
 
+.. warning:: We've seen certificate errors downloading etcd - you may need to
+             add --insecure to the `curl` command to ignore this.
+
```
ed-harrison: This ends up with everything in the rest of step 1 in the warning box - its not immediately obvious to me how to fix this given the rest of the indenting, but maybe you should just not bother with the warning and add --insecure to the example command.

docs/source/redhat-opens-install.rst
```diff
@@ -394,13 +397,7 @@ On each compute node, perform the following steps:
...
    OpenStack packages and to ``dnsmasq``.
 
-5. Install build dependencies:
-
-   ::
-
-       yum groupinstall 'Development Tools'
-
-6. Install and configure the DHCP agent on the compute host:
+5. Install and configure the DHCP agent on the compute host:
```
ed-harrison: Why is this not required (and why did we think it used to be)?

ed-harrison:
One particular comment for @matthewdupre to resolve (plus one question).

docs/source/redhat-opens-install.rst
matthewdupre: It's not so much that it's not required - it's that the etcd part of the instructions has already installed it.

docs/source/redhat-opens-install.rst
alexwlchan: @ed-harrison I’ve made a commit which tries to fix the indentation, by moving the warning to after the curl command. I think it’s worth having the warning, because just dropping the `--insecure` flag without explanation is bound to raise some eyebrows.

This is what the rendered output looks like on my machine; could you double-check the change please?

screen shot 2015-06-18 at 20 09 43

ed-harrison:
LGTM, but I think the --insecure warning needs to be in the etcd proxy section too.  I've added that - over to @matthewdupre to confirm and merge.

matthewdupre:
Thanks all - this looks good.  Merging.

PR #658, opened 2015-06-18
Changes.md update for pr 636

Lukasa:
Missing from #636.


plwhite:
Is that the right wording? I'm inclined to feel that the right wording is why we did it rather than how we did it. In other words "Improve security between endpoint and host and simplify INPUT chain logic".

Lukasa:
Why not?

PR #661, opened 2015-06-18
Reinstate from-endpoint policy in INPUT chain

### LINKED ISSUE #659: New felix-INPUT chain prevents exposing service from host
Piston Cloud expose more than just the metadata service from the host to VM.  Our tightening up of the rules under #636 has broken that because the packets from VMs hit our "DENY from tap+" rule before getting to Piston's own rules.

In general, I think we need some supported way to do this.  Ideally, a way that encompasses the special-case of the metadata service in a generic way.  We also need a pragmatic solution for @nbartos for hte short term.

### LINKED ISSUE #660: Make optional for felix to drop unmatched packets in INPUT chain
Currently, felix drops any packets from a VM that don't match our whitelist in the INPUT chain, this is too aggressive if the operator has deployed services on the host that should be accessible to VMs. It's also asymmetric, because it has no impact on a VM's ability to contact another host (which will see the packet come in over eth0 instead, for example).

Suggest we make the action on this rule configurable in the config file so that a knowledgable operator can set it to RETURN instead of DROP if they have their own rules in the INPUT chain after felix's.

fasaxc:
Reinstate the jump to the felix-FROM-ENDPOINT chain for traffic from endpoints going to the host.  This allows a host's outbound policy to limit its outgoing traffic before it reaches the host.

Make the action for packets that are accepted by the felix-FROM-ENDPOINT chain configurable so that a knowledgable operator can implement their own policy after felix's

Fixes #659 Fixes #660

### DIFF
--- a/calico/felix/config.py
+++ b/calico/felix/config.py
@@ -173,6 +173,9 @@ class Config(object):
         self.add_parameter("MetadataPort", "Metadata Port",
                            8775, value_is_int=True)
         self.add_parameter("InterfacePrefix", "Interface name prefix", None)
+        self.add_parameter("DefaultEndpointToHostAction",
+                           "Action to take for packets that arrive from"
+                           "an endpoint to the host.", "DROP")
         self.add_parameter("LogFilePath",
                            "Path to log file", "/var/log/calico/felix.log")
         self.add_parameter("LogSeverityFile",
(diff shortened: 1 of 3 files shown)


fasaxc:
@Lukasa I'd like to get this into the next release, if possible, please can you review?

Lukasa:
The diff looks reasonable to me. We're having some FV failures at the moment that haven't been diagnosed yet so I won't merge until those are cleaned up.

Lukasa:
I'm going to have to merge this by hand because I have other fixes that have to go in from #636.

Lukasa:
Ok, merged by hand. =) Please do confirm that b7120af959ffbdddafebf1062775911cab2cec95 looks right @fasaxc.

Separately, I'm going to add a new GH issue for all this, because I think this code is now a dense tangle of sadness and it suggests we need a better approach.

fasaxc:
@Lukasa that commit looks good to me.

PR #665, opened 2015-06-19
Allow ICMPv6 traffic needed for DHCP to the host

Lukasa:
When merging #636 I believe we broke DHCPv6 because we circumvented some ipv6 rules that were previously in the endpoint-specific chains. This PR returns support for IPv6 by adding exemptions into the INPUT chain.


Lukasa:
@fasaxc @plwhite Note that I'm duplicating the rules that are currently in the relevant filter chains for VMs. We may want to remove them from those chains, I'm honestly not sure, but right now I just want to get the FVs passing again.

neiljerram:
LGTM.

PR #667, opened 2015-06-19
Allow DNS through the INPUT table as well

Lukasa:
Another follow-on from #636 (much like #665). This time, we need to allow DNS through as well.


Lukasa:
@neiljerram Can I arrange for you to take a swing at this for review, please?

PR #672, opened 2015-06-19
Fix that profile chains weren't being stubbed out.

fasaxc:
Spotted while live testing calico-docker with a bad profile that failed validation.

* When a profile chain was deleted or failed validation, we were aliasing that to {}, which is the empty chain, instead of deleting the chain.
* Spotted that we were still catching CalledProcessError instead of FailedSystemCall in a few places, fixed those up.
* Fixed up some comments and docstrings.

Impact is that, if you have multiple profiles attached to an endpoint and you're using an early chain for deny rules followed by more broad accept rules and the deny rules failed validation then the deny rules would be ignored rather than being replaced with a safe stub chain.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,8 @@
 
 ## 0.25-dev
 
+- #672: Fix bug where profile chain was left empty instead of being 
+  stubbed out.
 - Improve security between endpoint and host and simplify INPUT chain logic.
 
 ## 0.24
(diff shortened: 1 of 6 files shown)


Lukasa:
I did a quick "does this look vaguely sensible" review and have no feedback. I still think @plwhite needs to review.

plwhite:
@Cory LGTM except for some minor markups. I made them and committed - can you confirm that you are happy and then merge?

Markups are : 

- a couple of comment clarifications
- one of them included into "this parameter must not be None" so I added an assert and a test

Lukasa:
Ok, I rebased this onto master and then merged manually in 446f229eee144c31b5887538c992f1ecdc0cce7f. Thanks chaps!

PR #677, opened 2015-06-22
Warn when using non-Calico network type

### LINKED ISSUE #132: Calico mechanism driver should log more helpfully when user misconfigures network.
I installed Calico on a system that already had a vxlan network configured - the mechanism driver couldn't create it, so the instance failed to find a host, but the issue was difficult to diagnose.

Lukasa:
Resolves #132.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## 0.26-dev
+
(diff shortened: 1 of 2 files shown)


matthewdupre:
Please add an identifier for the network.

PR #678, opened 2015-06-22
Documentation to increase conntrack table size

### LINKED ISSUE #616: Default conntrack table sizes are too small
The conntrack table, which we make heavy use of, has a configurable size, which is fairly low (65k) for a heavily used VM host.

A customer doing heavy churn testing (2 VMs/s with 800 TCP connections per VM) found they had to increase the table size to 1m with a 1 hour hang timeout to deal with connections to VMs that have been deleted.

At a minimum, our docs should explain how to tune the conntrack table size.

Lukasa:
Resolves #616.

This does not change any form of timeout. For the simple case that's probably good enough, but @fasaxc could provide good advice here.

### DIFF
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -33,6 +33,20 @@ task-based approach, when installing Calico with OpenStack on Ubuntu or
 Red Hat, please see :doc:`ubuntu-opens-install` or
 :doc:`redhat-opens-install`.
 
+System configuration
+--------------------
+
+A common problem on Linux systems is running out of space in the conntrack
+table, which can cause poor iptables performance. This can happen if you run a
+lot of workloads on a given host, or if your workloads create a lot of TCP
+connections or bidirectional UDP streams.
+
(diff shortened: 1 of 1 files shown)


docs/source/configuration.rst
```diff
@@ -33,6 +33,16 @@ task-based approach, when installing Calico with OpenStack on Ubuntu or
 Red Hat, please see :doc:`ubuntu-opens-install` or
 :doc:`redhat-opens-install`.
 
+System configuration
+--------------------
+
+A common problem on Linux systems is running out of space in the conntrack
+table, which can cause poor iptables performance. For this reason, we recommend
```
matthewdupre: I think this needs a bit more explanation how you can hit it (e.g. 'if you're running workloads that open many TCP/UDP connections').

Lukasa:
Ok, updated with elaboration. Back to you @matthewdupre.

Lukasa:
There is *no way* this broke the package build.

Issue #675, opened 2015-06-20
Only configure conntrack rules for tap+ interfaces


Lukasa:
Raised by @nbartos on the mailing list:

> A lot of those connection tracking rules seem redundant.  Instead of all the current instances of "--ctstate INVALID -j DROP" and "-m conntrack --ctstate RELATED,ESTABLISHED -j RETURN", can't we just do this at the top level?

>     -A felix-FORWARD -i tap+ -m conntrack --ctstate INVALID -j DROP
>     -A felix-FORWARD -i tap+ -m conntrack --ctstate RELATED,ESTABLISHED -j ACCEPT
>     -A felix-FORWARD -i tap+ -j felix-FROM-ENDPOINT
>     -A felix-FORWARD -o tap+ -m conntrack --ctstate INVALID -j DROP
>     -A felix-FORWARD -o tap+ -m conntrack --ctstate RELATED,ESTABLISHED -j ACCEPT
>     -A felix-FORWARD -o tap+ -j felix-TO-ENDPOINT
>     
>     -A felix-INPUT -i tap+ -m conntrack --ctstate INVALID -j DROP
>     -A felix-INPUT -i tap+ -m conntrack --ctstate RELATED,ESTABLISHED -j ACCEPT

This is a good simplification, and we should make it.

Lukasa:
Resolved in #681.

PR #681, opened 2015-06-23
Streamline conntrack rules, move to top-level chains

fasaxc:
@Lukasa this is still a work in progress but I wanted to ask if you know why we need a blanket ICMPv6 ACCEPT in the FORWARD chain.  I can see why we'd want it i the INPUT chain, since we need to play nice as a router, but is it needed in the FORWARD chain?

Also, I wanted to make sure that it was correct to match on tap+ for the metadata rule.  That was missing before.

### DIFF
--- a/calico/felix/endpoint.py
+++ b/calico/felix/endpoint.py
@@ -438,27 +438,8 @@ def _get_endpoint_rules(endpoint_id, suffix, ip_version, local_ips, mac,
     to_chain_name, from_chain_name = chain_names(suffix)
 
     # First build the chain that manages packets to the interface.
-    # Common chain prefixes.  Allow IPv6 ICMP and conntrack rules.
-    to_chain = ["--flush %s" % to_chain_name]
-    if ip_version == 6:
-        #  In ipv6 only, there are 6 rules that need to be created first.
-        #  RETURN ipv6-icmp anywhere anywhere ipv6-icmptype 130
-        #  RETURN ipv6-icmp anywhere anywhere ipv6-icmptype 131
-        #  RETURN ipv6-icmp anywhere anywhere ipv6-icmptype 132
-        #  RETURN ipv6-icmp anywhere anywhere ipv6-icmp router-advertisement
-        #  RETURN ipv6-icmp anywhere anywhere ipv6-icmp neighbour-solicitation
-        #  RETURN ipv6-icmp anywhere anywhere ipv6-icmp neighbour-advertisement
-        #
-        #  These rules are ICMP types 130, 131, 132, 134, 135 and 136.
-        for icmp_type in ["130", "131", "132", "134", "135", "136"]:
-            to_chain.append("--append %s --jump RETURN "
-                            "--protocol ipv6-icmp "
-                            "--icmpv6-type %s" % (to_chain_name, icmp_type))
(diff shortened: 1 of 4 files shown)


Lukasa:
As I noted in #665, [I have no idea](https://github.com/Metaswitch/calico/pull/665#issuecomment-113452011) if we want them or not. I was really working on expediency there because I wanted to get that stuff fixed up before the end of the week. ;) I'm prepared to believe we can safely remove the ICMPv6 stuff from the FORWARD chain.

Yeah, the metadata side of things is fine too.

This is actually a fix for #675, right?

And, finally, is this actually a WIP? Because it kinda looks done to me...

fasaxc:
Yep, it covers #675, but I thought I'd also check the other rules too and clean up what I could.

Pretty-much code complete, but I didn't get a chance to kick the tires on it and I had that design question about ICMPv6.  Also, would benefit from some UT.

Lukasa:
Cool. Well, I can sit on this for a day or so to see if you find time for UT here.

calico/felix/endpoint.py
```diff
@@ -438,27 +438,8 @@ def _get_endpoint_rules(endpoint_id, suffix, ip_version, local_ips, mac,
...
-        #  RETURN ipv6-icmp anywhere anywhere ipv6-icmptype 130
-        #  RETURN ipv6-icmp anywhere anywhere ipv6-icmptype 131
-        #  RETURN ipv6-icmp anywhere anywhere ipv6-icmptype 132
-        #  RETURN ipv6-icmp anywhere anywhere ipv6-icmp router-advertisement
-        #  RETURN ipv6-icmp anywhere anywhere ipv6-icmp neighbour-solicitation
-        #  RETURN ipv6-icmp anywhere anywhere ipv6-icmp neighbour-advertisement
-        #
-        #  These rules are ICMP types 130, 131, 132, 134, 135 and 136.
-        for icmp_type in ["130", "131", "132", "134", "135", "136"]:
-            to_chain.append("--append %s --jump RETURN "
```
fasaxc: I don't think these are needed any more.  only the router should be sending these in a way that bypasses the incoming policy of the VM, and the router is the host so its packets only hit the OUTPUT chain anyway.

fasaxc:
@Lukasa back to you for review.

calico/felix/frules.py
```diff
@@ -378,13 +366,34 @@ def _rule_to_iptables_fragment(chain_name, rule, ip_version, tag_to_ipset,
...
                        dhcp_src_port, dhcp_dst_port, ipv6=False,
-                       default_action="DROP"):
+                       default_action="DROP", hosts_set_name=None):
     """
-    Returns a list of rules that should be applied to the INPUT chain.
+    Returns a list of rules that should be applied to the felix-INPUT chain.
+    :returns Tuple: list of rules and set of deps.
     """
+    # Optimisation: return immediately if the traffic is not from one of the
+    # interfaces we're managing.
```
Lukasa: This comment got detached from its relevant code, which is now down on line 387.

Lukasa:
One problem with a comment going a-wandering. However, because this pull request adds tests (:joy:) I'll do you a favour and fix that up myself. =D

PR #682, opened 2015-06-23
Fix log rotation problems

### LINKED ISSUE #249: Log files are all confused
Our default logging configuration for Felix (at least on Ubuntu) causes logs to end up all over the place.

Felix chooses to log to `/var/log/calico`, and wants to rotate its own log files. However, at least one other process wants to do that too, which means that `/var/log/calico` looks like this:

```
ubuntu@calico04:~$ ls -lah /var/log/calico
total 16M
drwxr-xr-x  2 root root   4.0K Mar  9 06:47 .
drwxrwxr-x 14 root syslog 4.0K Mar  9 06:47 ..
-rw-r--r--  1 root root      0 Mar  9 06:47 felix.log
-rw-r--r--  1 root root    14M Mar  9 08:45 felix.log.1
-rw-r--r--  1 root root      0 Feb 27 06:55 felix.log.2015-02-26
-rw-r--r--  1 root root      0 Feb 28 06:28 felix.log.2015-02-27
-rw-r--r--  1 root root      0 Mar  1 06:53 felix.log.2015-02-28
-rw-r--r--  1 root root      0 Mar  2 06:36 felix.log.2015-03-01
-rw-r--r--  1 root root      0 Mar  3 06:26 felix.log.2015-03-02
-rw-r--r--  1 root root      0 Mar  4 06:27 felix.log.2015-03-03
-rw-r--r--  1 root root      0 Mar  5 06:34 felix.log.2015-03-04
-rw-r--r--  1 root root      0 Mar  6 06:30 felix.log.2015-03-05
-rw-r--r--  1 root root      0 Mar  7 06:43 felix.log.2015-03-06
-rw-r--r--  1 root root      0 Mar  8 06:39 felix.log.2015-03-07
-rw-r--r--  1 root root   892K Mar  8 11:40 felix.log.2.gz
-rw-r--r--  1 root root   895K Mar  7 11:40 felix.log.3.gz
-rw-r--r--  1 root root   898K Mar  6 11:40 felix.log.4.gz
```

Note the large numbers of empty files being rotated by some other process, while felix logs to `felix.log.1` and correctly rotates those logs out.

Additionally, Felix logs to stdout by default, which means another set of logs are available in `/var/log/upstart`:

```
ubuntu@calico04:~$ ls -lah /var/log/upstart
-rw-r-----  1 root root   1.3M Mar  9 08:49 calico-felix.log
-rw-r-----  1 root root   890K Mar  9 06:47 calico-felix.log.1.gz
-rw-r-----  1 root root   888K Mar  8 06:39 calico-felix.log.2.gz
-rw-r-----  1 root root   902K Mar  7 06:43 calico-felix.log.3.gz
-rw-r-----  1 root root   894K Mar  6 06:30 calico-felix.log.4.gz
-rw-r-----  1 root root   905K Mar  5 06:34 calico-felix.log.5.gz
-rw-r-----  1 root root   899K Mar  4 06:26 calico-felix.log.6.gz
-rw-r-----  1 root root   899K Mar  3 06:26 calico-felix.log.7.gz
```

This is really excessive, and doesn't gain us anything: we're just doing work we don't need to do.

My personal preference would be to disable Felix's logging to file, allowing the init process to control how Felix logs, but I'm open to other suggestions.

Lukasa:
This resolves #249.

Generally we have an inconsistent attitude to logging. Felix rotates its own logs. Additionally, we install logrotate files for Ubuntu, but not for Red Hat. The upshot of this is that:

- When installed from source, Felix rotates its own logs but does not compress them.
- When installed from RPM, Felix rotates its own logs but does not compress them.
- When installed from .deb, Felix attempts to rotate its logs, so does logrotate, and it all goes to crap. Also, the logs rotated by logrotate *do* get compressed, but the ones rotated by Felix simply don't.

This patch resolves this problem in two ways.

Firstly, it stops Felix rotating its logs. Log rotation is simply not Felix's problem, so it should stop doing it.

Secondly, it adds logrotate handling to the RPM files so that Felix's logs are consistently rotated. logrotate is better at log rotation than Felix (it compresses logs, for example), so we should be using it.

Longer term I'd like to talk about whether Felix should log to file *at all*, but for now this is a good start.

Interested parties: @matthewdupre for review, @plwhite because he cares a lot about logging, @fasaxc because he cares a lot about Felix and also RPM.

### DIFF
--- a/rpm/calico-felix.logrotate
+++ b/rpm/calico-felix.logrotate
@@ -3,5 +3,5 @@
     missingok
     compress
     delaycompress
-    notifempty
+    minsize 1M
 }
--- a/rpm/calico-felix.logrotate
+++ b/rpm/calico-felix.logrotate
@@ -1,0 +1,7 @@
+/var/log/calico/felix.log {
+    daily
+    missingok
+    compress
+    delaycompress
+    notifempty
+}
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,8 @@
 - Provide warning log when attempting to use Neutron networks that are not of
   type 'local' or 'flat' with Calico.
 - Handle invalid JSON in IPAM key in etcd.
+- Move all log rotation into logrotate and out of Felix, to prevent conflicts.
+- Change log rotation strategy for logrotate to not rotate small log files.
 
 ## 0.25
 
--- a/calico/common.py
+++ b/calico/common.py
@@ -263,7 +263,7 @@ def complete_logging(logfile=None,
                 root_logger.removeHandler(handler)
             else:
                 handler.setLevel(stream_level)
-        elif isinstance(handler, logging.handlers.TimedRotatingFileHandler):
+        elif isinstance(handler, logging.handlers.WatchedFileHandler):
             file_handler = handler
             if file_level is None:
                 root_logger.removeHandler(handler)
@@ -275,9 +275,7 @@ def complete_logging(logfile=None,
         if not file_handler:
             mkdir_p(os.path.dirname(logfile))
             formatter = logging.Formatter(FORMAT_STRING)
-            file_handler = logging.handlers.TimedRotatingFileHandler(
-                logfile, when="D", backupCount=10
-            )
+            file_handler = logging.handlers.WatchedFileHandler(logfile)
             file_handler.addFilter(GreenletFilter())
             file_handler.setLevel(file_level)
             file_handler.setFormatter(formatter)
--- a/debian/calico-felix.logrotate
+++ b/debian/calico-felix.logrotate
@@ -3,5 +3,5 @@
     missingok
     compress
     delaycompress
-    notifempty
+    minsize 1M
 }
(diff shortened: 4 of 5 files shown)


rpm/calico-felix.logrotate
```diff
@@ -0,0 +1,7 @@
+/var/log/calico/felix.log {
+    daily
+    missingok
+    compress
+    delaycompress
+    notifempty
```
matthewdupre: I think it would be better to use `minsize`

fasaxc:
Sounds good, wasn't aware of WatchedFileHandler.  Glad it's in 2.6+.

plwhite:
LGTM. That was a very small fix for a very annoying problem.
:shipit:

rpm/calico-felix.logrotate
Lukasa: Can do, but we need to do it in Ubuntu as well. =)

Lukasa:
Updated, back to you for review @matthewdupre.

Issue #325, opened 2015-04-07
Implement periodic resync


fasaxc:
etcd-based felix already supports resync but it doesn't trigger it on a timer, which I think we'd like to do.  This could be as simple as aborting stopping the polling loop so that we drop out and go round the resync loop again every n seconds.

matthewdupre:
Can I suggest considering adding some jitter to the timer?  The old zeroMQ Felixes could put very high load on the controller if they all resynced at once (after the whole cluster came up at roughly the same time).

fasaxc:
Indeed, I've got an issue to track avoiding the thundering herd too.

fasaxc:
I'm pondering how to do the timer for this, using time.time() may bite us later because it's non-monotonic.  I usually use monotime but it's not packaged.

Lukasa:
We can package our own dependencies if we need to, it's not that tricky.

neiljerram:
What about sleep() (within a dedicated thread) or add_timeout().  Surely gevent provides something like those?

plwhite:
How much do we really care about the monotonic issue? It seems to me like it's very marginal; not much risk of a problem, and not much that could go wrong if time.time() slips backwards by a second or something.

Lukasa:
@plwhite There is no guarantee that the slip will only be a second. We cannot trust a non-monotonic clock not to jump around wildly. Given that we don't care about what the actual wall clock time is, we definitely want a monotonic clock.

neiljerram:
Why would anyone implement a periodic X in any other way than

- either a loop that calls sleep() and then does X

- or using a library call of the style of glib's g_timeout_add (https://developer.gnome.org/glib/stable/glib-The-Main-Event-Loop.html#g-timeout-add).

In both cases, dealing with possibly non-monotonic time is Someone Else's Problem.  So why are we discussing it here?

fasaxc:
@neiljerram we already have a loop that's doing the polling so the natural way that occurred to me to structure that code is to calculate the time at which we want to stop polling and resync and then check if we're past that time each time around the loop.

You're right, we could use something like gevent.spawn_later(), which is callback function that we have simple access to, and somehow kick the etcd polling loop.  I'll probably go with something like that if monotime isn't available.

neiljerram:
Think we need to go back to first principles on this issue - i.e. why are we doing anything?  I don't think "which I think we'd like to do" is good enough.

fasaxc:
[...]

PR #687, opened 2015-06-24
Implement felix periodic resync.

fasaxc:
* [x] Start another greenlet to periodically trigger a resync.
* [x] Make resync force an update to the dataplane in case anything has been broken by another process
* [x] UT

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,7 @@
 
 - Limit number of concurrent shell-outs in felix to prevent file descriptor
   exhaustion.
+- Have felix periodically resync from etcd and force-refresh the dataplane.
 - Stop restarting Felix on Ubuntu if it fails more than 5 times in 10 seconds.
 
 ## 0.26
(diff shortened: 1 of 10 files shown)


fasaxc:
@matthewdupre care to give this a look-over before I write the UT?  We now:

* Add config for the resync interval.
* Start a worker greenlet to trigger the resyncs.
* Pass a force_reprogram flag when doing the resync to ensure the dataplane gets reprogrammed.  (Normally, felix assumes the dataplane won't change under its feet.)

I haven't done anything to interrupt the current outstanding poll, so we'll actually resync after the current poll completes.

matthewdupre:
This looks OK - can you please update the doc as well.

I'm interested in how much testing you've done of this - the reason I'm particularly curious is because most of our tests don't run for long enough to execute this, so we'll have much less live regression testing than usual.

fasaxc:
@matthewdupre I've been running it with a much lower period for my testing, so I see it execute regularly.  I'll touch base with @lwr20 to see if there's a good way to work this into the tests.

matthewdupre:
Just one docs typo, UTs look OK although I haven't reviewed them in detail.

fasaxc:
Fixed.

matthewdupre:
Thanks

fasaxc:
Fixes #325

PR #690, opened 2015-06-24
Tolerate missing keys during atomic delete

### LINKED ISSUE #689: Timing window can cause mechanism driver to explode when deleting absent endpoints.
Observed on our FV rig. We managed to hit the perfect interleaving combination of etcd calls, such that the resync thread attempted to delete an endpoint that had been destroyed midway through the resync process. This took out the process with an EtcdKeyNotFound error.

The fix is simple: just tolerate EtcdKeyNotFound when deleting stuff. Who cares if the thing you went to delete doesn't exist?

Lukasa:
Resolves #689.

### DIFF
--- a/calico/openstack/t_etcd.py
+++ b/calico/openstack/t_etcd.py
@@ -265,8 +265,10 @@ class CalicoTransportEtcd(object):
     @_handling_etcd_exceptions
     def atomic_delete_endpoint(self, endpoint):
         """
-        Atomically delete a given endpoint. This method allows exceptions from
-        etcd to bubble up.
+        Atomically delete a given endpoint.
+
+        This method tolerates attempting to delete keys that are already
+        missing, otherwise allows exceptions from etcd to bubble up.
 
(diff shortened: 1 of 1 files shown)


calico/openstack/t_etcd.py
```diff
@@ -276,9 +276,17 @@ def atomic_delete_endpoint(self, endpoint):
...
         )
-        self.client.delete(
-            endpoint.key, prevIndex=endpoint.modified_index, timeout=5
-        )
+
+        try:
+            self.client.delete(
+                endpoint.key, prevIndex=endpoint.modified_index, timeout=5
+            )
+        except etcd.EtcdKeyNotFound:
```
matthewdupre: 'This method allows exceptions from etcd to bubble up' contradicts this (slightly).  Please adjust the comment.

matthewdupre:
One little comment, otherwise looks good.

Lukasa:
Docstring updated

PR #700, opened 2015-06-29
Restart Felix on failure on RHEL systems

### LINKED ISSUE #699: Red Hat systems may restart killed Felix processes
In our systemd service file we set `Restart=always`. According to the systemd docs, [this might not be a good idea](http://www.freedesktop.org/software/systemd/man/systemd.service.html#Restart=):

> If set to `always`, the service will be restarted regardless of whether it exited cleanly or not, got terminated abnormally by a signal, or hit a timeout.

> Setting this to `on-failure` is the recommended choice for long-running services, in order to increase reliability by attempting automatic recovery from errors. 

I suspect that we actually want `on-failure`, not `always`.

Lukasa:
Resolves #699.

### DIFF
--- a/rpm/calico-felix.service
+++ b/rpm/calico-felix.service
@@ -8,7 +8,7 @@ User=root
 ExecStartPre=/usr/bin/mkdir -p /var/run/calico
 ExecStart=/usr/bin/calico-felix --config-file=/etc/calico/felix.cfg
 KillMode=process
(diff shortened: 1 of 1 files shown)


matthewdupre:
(Note that Felix did restart on failure anyway - this pull request simply stops it from restarting on successful exit.)

Issue #40, opened 2014-11-11
DHCP not working due to bad udp checksum - TX offloading problem


localpref:
My CirrOS based virtual machines  `Linux cirros 3.2.0-68-virtual #102-Ubuntu SMP Tue Aug 12 22:14:39 UTC 2014 x86_64 GNU/Linux` ignore the DHCP reply sent by dnsmasq. When tcpdumping on the VM's tap interface, I can see the DHCP request AND the DHCP reply, but the reply is not accepted.

It appears to be down to a problem with checksum generation, which usually only happens on the physical interface (offloading), but with Calico all DHCP happens inside the same compute node and only goes from the dnsmasq ns to the relevant tap ports...

A workaround that helped me make my instances accept the DHCP reply is adding the following iptables/mangle entry on my compute node:

`iptables -A POSTROUTING -t mangle -p udp --dport bootpc -d 100.64.0.0/10 -j CHECKSUM --checksum-fill`

the one odd thing that I see in the attached tcpdump files is that the IP source address is that of the compute node (10.13.0.140), while the DHCP server address contained is that of the virtual gateway 100.64.0.1, and maybe that's where the wrong checksum comes from?

however, I think the problem needs to be debugged further - maybe there is a way to signal that there is no offloading support, and maybe dnsmasq needs to be patched to properly calculate checksums on request?
Or is there something more sinister going on here?

It appears as though I am hitting something related to
http://article.gmane.org/gmane.linux.kernel/1003853

more info for a busybox patch from 2012:
http://lists.busybox.net/pipermail/busybox/2012-September/078348.html

more info:
https://github.com/mseknibilel/OpenStack-Folsom-Install-guide/issues/14
https://bugzilla.novell.com/show_bug.cgi?id=668194

tcpdump with broken reply:
12:48:43.085347 IP (tos 0xc0, ttl 64, id 61166, offset 0, flags [none], proto UDP (17), length 336)
    10.13.0.140.67 > 100.64.0.4.68: [bad udp cksum 0x702a -> 0x342f!] BOOTP/DHCP, Reply, length 308, xid 0xc8b8dc3d, Flags [none] (0x0000)
	  Your-IP 100.64.0.4
	  Server-IP 100.64.0.1
	  Client-Ethernet-Address fa:16:3e:57:03:66
	  Vendor-rfc1048 Extensions
	    Magic Cookie 0x63825363
	    DHCP-Message Option 53, length 1: Offer
[...]

PR #710, opened 2015-07-01
Remove DHCP checksum fix from Calico compute node install process (Neutron does it instead)

matthewdupre:
First part of the fix to #40.  Must not be merged until the Neutron change is backported to Icehouse and Juno.


matthewdupre:
OK - Neutron fixes merged and building.  Ready for this to go in.

Issue #571, opened 2015-05-20
etcd profile out of sync, doesn't seem to get fixed


fasaxc:
Ran an FV run on the bobcat rig 

* using my smc-multi-pids branch, which supports multiple neutron workers, 
* bobcat rig was patched with calico master commit d649e65
* running 8 API workers on the control node

The FV run failed: http://calicobobcat-jenkins.gulfstream:8080/job/fv-tests/791/.  The test_security_group failures seem to be genuine ping failures.  

I manually ran the I ran test_security_group from the command line and paused the test when pings started to fail.

It looks like horizon's picture of the group2 SG differs from that in etcd.  In horizon, it has an allow rule from the default group; in etcd that is missing.  I'm not sure how that got out of sync but the periodic resync doesn't seem to fix it.

fasaxc:
Re-ran that same test and it looked like it was working before I left the office.  Still, pretty worrying to get two failures in a row.

fasaxc:
From code reading, it looks like that test should re-add the rule at the start of the test so I'm not sure how it could fail twice in a row if that re-add was giving us a kick to re-write the profile.

Lukasa:
Periodic resync won't fix this at the moment, it's a known problem.

What's on the smc-multi-pids branch?

Lukasa:
Also, when testing this, make sure to watch the relevant key in etcd. I wonder if we're having a write ordering problem.

matthewdupre:
@Lukasa - what do you think the known problem is?

Lukasa:
We don't update existing keys during re sync, we either write missing ones or delete extra ones.

matthewdupre:
OK, understood - am I right in inferring that you don't know what the original cause of this is (other than potentially writes overtaking each other)?

Lukasa:
Correct.

fasaxc:
My branch is in calico-test. It fixes some tests that were asserting that there was only one neutron process.

fasaxc:
I've been running with 8 workers

ed-harrison:
@Lukasa This is currently with you.  Are you actively thinking about how we could modify our resync strategy to catch cases where the keys are present, but incorrect?

Lukasa:
Depends what you mean by actively. I have a number of ways to fix this, but I don't know what the urgency on them is. We should discuss on Monday.

ed-harrison:
Indeed...

matthewdupre:
[...]

PR #711, opened 2015-07-02
Fix up incorrect data in etcd during Neutron resync

Lukasa:
This change is a pretty stupid solution to #571. Right now this is about half the work: it adds the function, but doesn't test it. I also want to do a refactor because a couple of these methods are massive now.

### DIFF
--- a/calico/openstack/mech_calico.py
+++ b/calico/openstack/mech_calico.py
@@ -732,7 +732,15 @@ class CalicoMechanismDriver(mech_agent.SimpleAgentMechanismDriverBase):
             if etcd_data != neutron_data:
                 # Write to etcd.
                 LOG.warning("Resolving error in port %s", endpoint.id)
-                self.transport.endpoint_created(port)
+                try:
+                    self.transport.write_port_to_etcd(
+                        port, prev_index=endpoint.modified_index
+                    )
+                except ValueError:
+                    # If someone wrote to etcd they probably have more recent
+                    # data than us, let it go.
+                    LOG.info("Atomic CAS failed, no action.")
+                    continue
 
     def resync_profiles(self, context):
         """
@@ -829,7 +837,19 @@ class CalicoMechanismDriver(mech_agent.SimpleAgentMechanismDriverBase):
             if (etcd_rules != neutron_rules) or (etcd_tags != neutron_tags):
                 # Write to etcd.
                 LOG.warning("Resolving error in profile %s", profile.id)
-                self.transport.write_profile_to_etcd(neutron_profile)
+
+                try:
+                    self.transport.write_profile_to_etcd(
+                        neutron_profile,
+                        prev_rules_index=neutron_profile.rules_modified_index,
+                        prev_tags_index=neutron_profile.tags_modified_index,
+                    )
+                except ValueError:
+                    # If someone wrote to etcd they probably have more recent
+                    # data than us, let it go.
+                    LOG.info("Atomic CAS failed, no action.")
+                    continue
+
 
     def add_port_interface_name(self, port):
         port['interface_name'] = 'tap' + port['id'][:11]
--- a/calico/openstack/mech_calico.py
+++ b/calico/openstack/mech_calico.py
@@ -48,7 +48,9 @@ except ImportError:  # Kilo
 
 # Calico imports.
 import etcd
-from calico.openstack.t_etcd import CalicoTransportEtcd, port_etcd_data
+from calico.openstack.t_etcd import (
+    CalicoTransportEtcd, port_etcd_data, profile_rules, profile_tags
+)
 
 LOG = log.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


Lukasa:
Ok, I believe this is now good to go. The refactor code is in #712, which should be considered only once this is merged.

I'm assigning this to @fasaxc for review, but it'd probably be good to do some live test of this before we merge. Let's not merge it until next week.

calico/openstack/mech_calico.py
```diff
@@ -675,6 +683,52 @@ def resync_endpoints(self, context):
...
+                LOG.info("Failed to update deleted endpoint %s", endpoint.id)
+                continue
+
+            with context.session.begin(subtransactions=True):
+                port = self.db.get_port(context, endpoint.id)
+
+            # Get the data for both.
+            try:
+                etcd_data = json.loads(endpoint.data)
+            except Exception:
```
fasaxc: Would `ValueError` and `TypeError` be enough here?  I _think_ that's all loads can raise.

calico/openstack/mech_calico.py
```diff
@@ -675,6 +683,52 @@ def resync_endpoints(self, context):
...
+                # If the JSON data is bad, just ignore it. We can't blow up
+                # here because that will trigger a new resync on another node
+                # that will blow *it* up as well. Just tolerate it.
+                LOG.exception("Bad JSON data in key %s", endpoint.key)
+                continue
```
fasaxc: `continue` doesn't seem right.  If the JSON data is bad shouldn't we be fixing it?

calico/openstack/mech_calico.py
```diff
@@ -675,6 +683,52 @@ def resync_endpoints(self, context):
...
+
+            self.add_port_gateways(port, context)
```
fasaxc: Ignore if already covered in your refactor... These few lines feel duplicative.  Seems like there should be a neutron_port_to_etcd_endpoint function

Lukasa: Already covered in the refactor. =)

calico/openstack/mech_calico.py
```diff
@@ -730,6 +786,61 @@ def resync_profiles(self, context):
                 # means the profile was created or updated elsewhere.
                 continue
 
+        # Finally, reconcile the security profiles. This involves looping over
+        # them, grabbing their data, and then comparing that to what Neutron
+        # has.
+        changed_profiles = (p for p in profiles if p.id in reconcile_groups)
```
fasaxc: How about `profiles_to_reconcile`?  `changed_profiles` seems a bit off

calico/openstack/mech_calico.py
```diff
@@ -730,6 +786,61 @@ def resync_profiles(self, context):
...
+            # Convert the etcd data into in-memory data structures.
+            try:
+                etcd_rules = json.loads(etcd_profile.rules_data)
+                etcd_tags = json.loads(etcd_profile.tags_data)
+            except Exception:
+                # If the JSON data is bad, just ignore it. We can't blow up
+                # here because that will trigger a new resync on another node
+                # that will blow *it* up as well. Just tolerate it.
+                LOG.exception("Bad JSON data in key %s", etcd_profile.key)
+                continue
```
fasaxc: Ditto, shouldn't we fix?

fasaxc:
Few comments but didn't spot anything major.  Do you think we're stretching those named tuples a bit?  They're getting a bit unwieldy now they (sometimes) carry the data too.

calico/openstack/mech_calico.py
Lukasa: Let's assume the answer is yes. =D

calico/openstack/mech_calico.py
Lukasa: Yeah, we should. We do it in the refactor, weirdly, so I'm inclined to leave it in place for now and add it in the refactor code.

Lukasa:
Ok @fasaxc, markups made. =)

calico/openstack/mech_calico.py
fasaxc: OK

calico/openstack/t_etcd.py
```diff
@@ -235,6 +256,27 @@ def provide_felix_config(self):
...
+        """
+        Get data for an endpoint out of etcd. This should be used on endpoints
+        returned from functions like ``get_endpoints``.
+
+        :param endpoint: An ``Endpoint`` class.
+        :return: A ``Endpoint`` class with ``data`` not None.
+        """
+        LOG.debug("Getting endpoint %s", endpoint.id)
+
+        result = self.client.read(endpoint.key, timeout=5)
```
fasaxc: We're using these timeouts in a few places, worth wrapping up so we don't forget them and only have one place to change  at 3am when the timeout is too short.

fasaxc:
Markups look good, one thought that just occurred to me but I don't need to see it again.

fasaxc:
Good to merge after the 0.27 release, I think.

calico/openstack/t_etcd.py
Lukasa: Yeah, good idea.

Lukasa:
And done. Will merge once we release.

PR #712, opened 2015-07-03
Refactor the resync code.

Lukasa:
This change follows #711, and includes a straight refactor of the resync code. The goal here is to improve readability by preventing the person reading the code from having to look at a 100+ line function and wonder what stuff is and isn't in scope.

This should not be merged until #711 is, and should in all respects be considered a separate change that is dependent on #711, not a superset of it.


fasaxc:
@Lukasa please could you put up a PR that has #711 as the base so I can review as a delta?

Lukasa:
Sure. =) See #713.

PR #714, opened 2015-07-07
Get all fixed IPs for a port

Lukasa:
It turns out #538 was not a one-off. In addition to security group information potentially being out of date on a port during a port update, the IP address information for that port could be out of date as well. This change fixes that problem, by querying the `IPAllocation` table.

I've done some quick live test on Icehouse and nothing appears to explode, but we need to test this more widely.

### DIFF
--- a/calico/openstack/mech_calico.py
+++ b/calico/openstack/mech_calico.py
@@ -679,7 +679,7 @@ class CalicoMechanismDriver(mech_agent.SimpleAgentMechanismDriverBase):
                 # Fill out other information we need on the port and write to
                 # etcd.
                 port['fixed_ips'] = self.get_fixed_ips_for_port(
-                    context._plugin_context, port
+                    context, port
                 )
                 port['security_groups'] = self.get_security_groups_for_port(
                     context, port
--- a/calico/openstack/mech_calico.py
+++ b/calico/openstack/mech_calico.py
@@ -31,6 +31,7 @@ from functools import wraps
 
 # OpenStack imports.
 from neutron.common import constants
+from neutron.db import models_v2
 from neutron.plugins.ml2 import driver_api as api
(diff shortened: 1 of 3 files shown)


Lukasa:
Alright, with live test I believe this works. Throwing it over the fence to @neiljerram for code review.

calico/openstack/mech_calico.py
```diff
@@ -668,11 +678,14 @@ def resync_endpoints(self, context):
...
-                self.add_port_interface_name(port)
+                port['fixed_ips'] = self.get_fixed_ips_for_port(
+                    context, port
+                )
                 port['security_groups'] = self.get_security_groups_for_port(
                     context, port
                 )
+                self.add_port_gateways(port, context)
+                self.add_port_interface_name(port)
                 self.transport.endpoint_created(port)
```
neiljerram: We've made the same change in 4 very similar blocks of code here.  Surely time for a subroutine!?

neiljerram:
Just the one possible markup, otherwise LGTM

calico/openstack/mech_calico.py
Lukasa: There actually is a subroutine, it's just in a different pull request. =D This merge will break that one, but that's fine, I'll fix it up by hand.

neiljerram: So what's the sequence of events?  Should I merge this PR first?

Lukasa: Yeah, this one should go in first, but we might want to hold off until the FVs are back in a good state.

neiljerram: OK.  If I don't notice myself, feel free to prod me when that happens.

Lukasa: Ok Neil, FVs are green.

PR #722, opened 2015-07-13
Pin mock and coverage versions.

### LINKED ISSUE #720: Mock has finally broken Python 2.6
We need to adjust our testing dependencies because mock no longer works on Python 2.6.

Lukasa:
This fixes our UTs.

Resolves #720.


Lukasa:
@matthewdupre Can you find a time in the release process that it'll be safe to merge this?

PR #727, opened 2015-07-15
Python-etcd does not like keyword arguments

Lukasa:
This fixes a current FV failure.


fasaxc:
LGTM.  Any adjustment we can make to the UTs to catch future versions of this bug?  I think we mock out the client so presumably we can set the signature of the write method.

Lukasa:
@fasaxc I've added a UT catch as well.

Issue #645, opened 2015-06-10
Empty SG chain after power cycle 


DaveLangridge:
The test creates some VMs, does some pings, then power cycles one of the compute nodes. Then it does some more ping tests. The test fails more times than it passes (2:1 at a guess). A similar test which reboots the host passes most/all of the time.

When it fails, the test script can’t ssh into a VM on the host that’s been cycled. Looking at the iptables, the structure seems to be present, but most of the SG stuff is missing:

Chain felix-from-cd250fed-c7 (1 references)
pkts bytes target     prot opt in     out     source               destination
    0     0 DROP       all  --  *      *       0.0.0.0/0            0.0.0.0/0            ctstate INVALID
    0     0 RETURN     all  --  *      *       0.0.0.0/0            0.0.0.0/0            ctstate RELATED,ESTABLISHED
    2   630 RETURN     udp  --  *      *       0.0.0.0/0            0.0.0.0/0            udp spt:68 dpt:67
   11   924 MARK       all  --  *      *       0.0.0.0/0            0.0.0.0/0            MARK and 0x0
   11   924 felix-p-_a0438770e0e8efc-o  all  --  *      *       10.28.1.171          0.0.0.0/0            MAC FA:16:3E:20:5A:19
    0     0 RETURN     all  --  *      *       0.0.0.0/0            0.0.0.0/0            mark match ! 0x1/0x1 /* No mark means profile accepted packet */
    0     0 DROP       all  --  *      *       0.0.0.0/0            0.0.0.0/0            /* Default DROP if no match (endpoint cd250fed-c73e-41b9-8dd2-e7f01512cbe7): */

Chain felix-p-_a0438770e0e8efc-i (1 references)
pkts bytes target     prot opt in     out     source               destination
   59  4584 DROP       all  --  *      *       0.0.0.0/0            0.0.0.0/0            /* WARNING Missing chain DROP: */

Chain felix-p-_a0438770e0e8efc-o (1 references)
pkts bytes target     prot opt in     out     source               destination
   11   924 DROP       all  --  *      *       0.0.0.0/0            0.0.0.0/0            /* WARNING Missing chain DROP: */

Chain felix-to-cd250fed-c7 (1 references)
pkts bytes target     prot opt in     out     source               destination
    0     0 DROP       all  --  *      *       0.0.0.0/0            0.0.0.0/0            ctstate INVALID
[...]

PR #729, opened 2015-07-15
For robustness, periodically refresh iptables chains

fasaxc:
Make us robust to other processes messing with iptables as per #645.

* [x] Implement
* [x] Fix up UTs
* [x] UT new function

### DIFF
--- a/calico/felix/actor.py
+++ b/calico/felix/actor.py
@@ -360,7 +360,8 @@ class Actor(object):
             self._op_count = 0
 
     def __str__(self):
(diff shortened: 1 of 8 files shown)


fasaxc:
@plwhite can you take this code review?

fasaxc:
Added UT, including filling in some existing gaps in fiptables coverage.

eepyaich:
Looks good to me - merging it now, so that we can get some good FV runs before the next release.

PR #732, opened 2015-07-16
Add status reporting to Felix

miroslav21:
Add status reporting to Felix.


alexwlchan:
ok to test

calico/felix/config.py
```diff
@@ -343,6 +354,38 @@ def _validate_cfg(self, final=True):
...
+        if self.REPORTING_TTL_SECS <= self.REPORTING_INTERVAL_SECS\
+                and self.REPORTING_TTL_SECS != 0:
+            raise ConfigException("Reporting TTL ({} sec) less or equal reporting "
+                                  "interval ({} sec). "
+                                  .format(self.REPORTING_INTERVAL_SECS,
+                                          self.REPORTING_TTL_SECS),
+                                  self.parameters["ReportingIntervalSecs"])
+
+        # If status report ttl is not set or set to 0,
+        # it's vlaue is set to 2.5 times interval
```
alexwlchan: "vlaue" ~> "value". And I think the apostrophe in "it's" should be removed.

calico/felix/test/test_fetcd_api.py
```diff
@@ -0,0 +1,157 @@
...
+    Used to check whether status has expected format
+    i.e. whether timestamp is in ISO 8601 Zulu format
+    """
+    def __eq__(self, other):
+        import re
+        timestamp_regex = re.compile('.*"status_time": "\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}Z?".*')
+        is_timestamp =  timestamp_regex.match(str(other)) is not None
+        return is_timestamp
+    def __repr__(self):
+        return "<TestIfStatus()>"
```
alexwlchan: Two tips for \__repr__():

1. This should return a string which, if evaluated, returns an equal object to this one. So you shouldn't have the angle brackets; this should just return `TestIfStatus()`.
2. Rather than hard-coding the name of the class, you can access it through a special parameter: `self.__class__.__name__`. Using this makes the class more robust to subclassing, because the subclass can just inherit repr(). Unlikely to matter here, but it's a good habit to get into.

So I'd suggest writing the repr() (and the one below) like this:

    def __repr__(self):
        return '%s()' % self.__class__.__name__

calico/felix/test/test_fetcd_api.py
```diff
@@ -0,0 +1,157 @@
...
+        self.m_config.REPORTING_INTERVAL_SECS = 0
+        self.m_config.REPORTING_TTL_SECS = 0
+
+    @patch('calico.felix.fetcd.EtcdAPI.write_to_etcd')
+    @patch('calico.felix.fetcd._EtcdWatcher')
+    @patch('calico.felix.fetcd.etcd')
+    def finish_setup(self, m_etcd, m_EtcdWatcher, m_write_to_etcd, **kwargs):
+        # Set configuration attributes and start etcd_api
+        for key, value in kwargs.iteritems():
+            setattr(self.m_config, key, value)
```
alexwlchan: This isn't wrong, but I think it could be neater. Rather than updating the key/value pairs one-by-one, you could just use the update() method:

> <code><strong>update</strong>([<em>other</em>])</code>

> Update the dictionary with the key/value pairs from other, overwriting existing keys. Return None. ([Python docs](https://docs.python.org/2/library/stdtypes.html#dict.update))

So instead you'd write:

```python
self.m_config.update(kwargs)
```

calico/felix/fetcd.py
```diff
@@ -442,6 +626,26 @@ def on_workload_delete(self, response, hostname, orchestrator,
...
+        port = 4001
+    if etcd_communicator.client and copy_cluster_id:
+        old_cluster_id = etcd_communicator.client.expected_cluster_id
+        _log.info("(Re)connecting to etcd. Old etcd cluster ID was %s.",
+                  old_cluster_id)
+    else:
+        _log.info("(Re)connecting to etcd. No previous cluster ID.")
+        old_cluster_id = None
+    etcd_communicator.client = etcd.Client(host=host, port=port,
+                              expected_cluster_id=old_cluster_id)
```
alexwlchan: I'm nitpicking, but please could you tidy up the indentation of this line? Ideally the second line of arguments should line up with the parens on the first line.

matthewdupre:
I've added you to the other pull request job's whitelist.  You shouldn't see any more admin verification requests.

alexwlchan:
You've changed the permissions on a bunch of these files &ndash; you should fix those up before this gets merged.

calico/datamodel_v1.py
```diff
@@ -33,13 +33,19 @@
 # All Calico data is stored under this path.
 ROOT_DIR = "/calico"
 
+# Current version
+VERSION = "/v1"
+
 # OpenStack data is stored under this path.
 OPENSTACK_DIR = ROOT_DIR + "/openstack"
 OPENSTACK_VERSION_DIR = OPENSTACK_DIR + "/v1"
```
emmagordon: Use VERSION here rather than "/v1"

calico/datamodel_v1.py
```diff
@@ -82,11 +88,22 @@ def dir_for_per_host_config(hostname):
...
+
+def key_for_uptime(hostname):
+    return dir_for_status(hostname) + "/uptime"
+
+
 def key_for_endpoint(host, orchestrator, workload_id, endpoint_id):
     return (HOST_DIR + "/%s/workload/%s/%s/endpoint/%s" %
             (host, orchestrator, workload_id, endpoint_id))
 
-
```
emmagordon: Did you mean to delete this blank line? Not consistent with spacing in the rest of this file.

calico/felix/config.py
```diff
@@ -374,6 +380,38 @@ def _validate_cfg(self, final=True):
                 self.parameters["DefaultEndpointToHostAction"]
             )
 
+        # Reporting interval and TTL need to be integers.
+        try:
+            self.REPORTING_TTL_SECS = int(self.REPORTING_TTL_SECS)
+        except ValueError:
+            raise ConfigException("Reporting interval was not integer.",
```
emmagordon: error message should refer to ttl, not interval

calico/felix/config.py
```diff
@@ -374,6 +380,38 @@ def _validate_cfg(self, final=True):
...
+                                  self.parameters["ReportingIntervalSecs"])
+        try:
+            self.REPORTING_INTERVAL_SECS = int(self.REPORTING_INTERVAL_SECS)
+        except ValueError:
+            raise ConfigException("Reporting TTL was not integer.",
```
emmagordon: error message should refer to interval, not ttl

calico/felix/config.py
```diff
@@ -374,6 +380,38 @@ def _validate_cfg(self, final=True):
...
+                                  self.parameters["ReportingTTLSecs"])
+
+        # For non-positive time values of reporting interval we set both
+        # interval and ttl set to 0 - i.e. status reporting is disabled.
```
emmagordon: typo - extra 'set' in here, should just be '.. ttl to 0 ...'

calico/felix/config.py
```diff
@@ -374,6 +380,38 @@ def _validate_cfg(self, final=True):
...
+        if self.REPORTING_INTERVAL_SECS <= 0:
+            self.REPORTING_TTL_SECS = 0
+            self.REPORTING_INTERVAL_SECS = 0
+
+        # Status report time to live must be more than status reporting interval.
+        if self.REPORTING_TTL_SECS <= self.REPORTING_INTERVAL_SECS\
+                and self.REPORTING_TTL_SECS != 0:
+            raise ConfigException("Reporting TTL ({} sec) less or equal reporting "
+                                  "interval ({} sec). "
```
emmagordon: error message isn't grammatically correct

calico/felix/config.py
```diff
@@ -374,6 +380,38 @@ def _validate_cfg(self, final=True):
...
+
+        # For non-positive time values of reporting interval we set both
+        # interval and ttl set to 0 - i.e. status reporting is disabled.
+        if self.REPORTING_INTERVAL_SECS <= 0:
+            self.REPORTING_TTL_SECS = 0
+            self.REPORTING_INTERVAL_SECS = 0
+
+        # Status report time to live must be more than status reporting interval.
+        if self.REPORTING_TTL_SECS <= self.REPORTING_INTERVAL_SECS\
+                and self.REPORTING_TTL_SECS != 0:
```
emmagordon: the reasoning behind the 'and ttl != 0' part wasn't immediately clear until I carried on reading below - if you move the setting ttl to default value if not set (or 0) to above this test then you can remove the 'and' here

calico/felix/felix.py
```diff
@@ -49,7 +49,6 @@
 
 _log = logging.getLogger(__name__)
 
-
```
emmagordon: The only changes in this file seem to de adding/deleting blank lines - intentional?

calico/felix/fetcd.py
```diff
@@ -89,15 +92,19 @@ class EtcdAPI(Actor):
     Since the python-etcd API is blocking, we defer API watches to
     a worker greenlet and communicate with it via Events.
 
-    As and when we add status reporting, to avoid needing to interrupt
-    in-progress polls, I expect we'll want a second  worker greenlet that
-    manages an "upstream" connection to etcd.
+    To avoid needing to interupt in-progress polls, another working
+    greenlet is introduced to manage status updating and writing to
+    etcd.
```
emmagordon: '... another working greenlet is introduced ...' - remember this docstring needs to continue to make sense in future, so better to say something like '... a different worker greenlet is used ...'

calico/felix/fetcd.py
```diff
@@ -108,11 +115,24 @@ def __init__(self, config, hosts_ipset):
...
+
     @logging_exceptions
     def _periodically_resync(self):
         """
         Greenlet: if enabled, periodically triggers a resync from etcd.
+
+        :return: Does not return, unless periodic resync disabled.
         """
+
+
```
emmagordon: Why the blank lines here?

calico/felix/fetcd.py
```diff
@@ -108,11 +115,24 @@ def __init__(self, config, hosts_ipset):
...
 
+        # Create a client to write to etcd.
+        self.client = None
+        _reconnect(self)
+
+        # Start up a reporting greenlet.
+        self._status_reporting_greenlet = gevent.spawn(self._status_reporting)
+        self._status_reporting_greenlet.link_exception(self._on_worker_died)
+
+
```
emmagordon: 1 blank line only between class method definitions .

calico/felix/fetcd.py
```diff
@@ -129,6 +149,67 @@ def _periodically_resync(self):
...
+        while True:
+            try:
+                self.update_felix_status(ttl)
+                _log.info("Status updated.")
+                gevent.sleep(interval)
+            except EtcdException as e:
+                _log.info("Status updating failed. %r. Reconnecting... ", e)
+                _reconnect(self)
+                gevent.sleep(RETRY_DELAY)
+
```
emmagordon: 1 blank line only between class method definitions .

calico/felix/fetcd.py
```diff
@@ -129,6 +149,67 @@ def _periodically_resync(self):
...
+
```
emmagordon: 1 blank line only between class method definitions .

calico/felix/fetcd.py
```diff
@@ -129,6 +149,67 @@ def _periodically_resync(self):
...
+        ttl = self._config.REPORTING_TTL_SECS
+        interval = self._config.REPORTING_INTERVAL_SECS
+        _log.info("Config loaded, reporting interval: %s, TTL: %s", interval, ttl)
+        if interval == 0:
+            _log.info("Interval is 0, status reporting disabled.")
+            return
+        while True:
+            try:
+                self.update_felix_status(ttl)
+                _log.info("Status updated.")
```
emmagordon: This will log every interval seconds - depending on the value of that interval, this could easily swamp the logs and make it hard to pick out other useful stuff. Should we only log the case where there is an error? If there's a good reason to log the success, I'd at least make it debug level rather than info.

calico/felix/fetcd.py
```diff
@@ -129,6 +149,67 @@ def _periodically_resync(self):
...
+        """
+        Greenlet: periodically writes to etcd
+
+        :return: Does not return, unless reporting disabled.
+        """
+        _log.info("Started status reporting thread.")
+        ttl = self._config.REPORTING_TTL_SECS
+        interval = self._config.REPORTING_INTERVAL_SECS
+        _log.info("Config loaded, reporting interval: %s, TTL: %s", interval, ttl)
+        if interval == 0:
```
emmagordon: A blank line before and after this if block would break up the code into logical sections and make it more readable.

calico/felix/fetcd.py
```diff
@@ -129,6 +149,67 @@ def _periodically_resync(self):
...
+                gevent.sleep(interval)
+            except EtcdException as e:
+                _log.info("Status updating failed. %r. Reconnecting... ", e)
+                _reconnect(self)
+                gevent.sleep(RETRY_DELAY)
+
+
+    def update_felix_status(self, ttl):
+        """
+        Writes status info to etcd
```
emmagordon: Should explain that we're writing two things to etcd - uptime and a json doc containing the status info (which is currently just the current time).

calico/felix/fetcd.py
```diff
@@ -140,6 +221,7 @@ def load_config(self):
         return self._watcher.configured
 
     @actor_message()
+
```
emmagordon: Presumably an accidental blank line?

calico/felix/fetcd.py
```diff
@@ -257,7 +339,9 @@ def _run(self):
         self.load_config.wait()
         while True:
             _log.info("Reconnecting and loading snapshot from etcd...")
-            self._reconnect(copy_cluster_id=False)
+
+            _reconnect(self, copy_cluster_id=False)
+
```
emmagordon: You seem to be quite a fan of blank lines! :) Not convinced they add anything to readability here.

calico/felix/fetcd.py
```diff
@@ -352,7 +419,9 @@ def _load_config(self):
                            "retry.", e)
                 gevent.sleep(RETRY_DELAY)
             else:
-                self.config.report_etcd_config(host_dict, global_dict)
+
+                self._config.report_etcd_config(host_dict, global_dict)
+
```
emmagordon: Again, these blank lines are unnecessary for readability here.

calico/felix/fetcd.py
```diff
@@ -457,7 +526,6 @@ def _wait_for_etcd_event(self):
                                             timeout=Timeout(connect=10,
                                                             read=90),
                                             check_cluster_uuid=True)
-                _log.debug("etcd response: %r", response)
```
emmagordon: Is there a reason to remove this debug logging?

calico/felix/fetcd.py
```diff
@@ -659,6 +728,26 @@ def on_workload_delete(self, response, hostname, orchestrator,
             del self.endpoint_ids_per_host[hostname]
 
 
+def _reconnect(etcd_communicator, copy_cluster_id=True):
+    """ Parameter etcd_communicator can be either etcd_watcher or etcd_api"""
```
emmagordon: function doctring should explain what the function does and then detail the params

calico/felix/test/data/felix_section.cfg
```diff
@@ -9,3 +9,6 @@ MetadataPort = 246
 
 [log]
 InterfacePrefix = whatever
+
+[log]
+ReportingIntervalSecs = 5
```
emmagordon: You've not got a test for setting the ttl in the config file. So at the moment this tests that ttl defaults correctly from the interval read from file. If testing that defaulting is not duplicated in another test, then make sure you still cover both after adding ttl to this test config file.

calico/felix/test/test_config.py
```diff
@@ -208,25 +210,31 @@ def test_file_sections(self):
...
 
     def test_env_var_override(self):
         """
-        Test environment variables override config options,
+        Test environment variables override config options.
         """
         with mock.patch.dict("os.environ", {"FELIX_ETCDADDR": "9.9.9.9:1234",
-                                            "FELIX_METADATAPORT": "999"}):
+                                            "FELIX_METADATAPORT": "999",
+                                            "FELIX_REPORTINGINTERVALSECS": "10"}):
```
emmagordon: You should also test setting ttl in the environment variables.

calico/felix/test/test_config.py
```diff
@@ -208,25 +210,31 @@ def test_file_sections(self):
...
             with mock.patch('calico.common.complete_logging'):
                 config = Config("calico/felix/test/data/felix_section.cfg")
 
         host_dict = { "InterfacePrefix": "blah",
                       "StartupCleanupDelay": "42",
                       "MetadataAddr": "4.3.2.1",
-                      "MetadataPort": "123" }
+                      "MetadataPort": "123",
+                      "ReportingIntervalSecs": "10",
```
emmagordon: The point of this test is to check that the environment value takes precedence - but you've set it to the same value in both places...

calico/felix/test/test_config.py
```diff
@@ -208,25 +210,31 @@ def test_file_sections(self):
...
+                      "ReportingTTLSecs": "20"}
 
         global_dict = { "InterfacePrefix": "blah",
                         "StartupCleanupDelay": "99",
                         "MetadataAddr": "5.4.3.2",
-                        "MetadataPort": "123" }
+                        "MetadataPort": "123",
+                        "ReportingIntervalSecs": "10"}
```
emmagordon: Again, this needs to be different to the environment interval value for the test to be useful.

emmagordon: You've not included ttl here.

calico/felix/test/test_config.py
```diff
@@ -237,6 +245,8 @@ def test_env_var_override(self):
         self.assertEqual(config.METADATA_PORT, 999)
         self.assertEqual(config.METADATA_IP, "1.2.3.4")
         self.assertEqual(config.STARTUP_CLEANUP_DELAY, 42)
+        self.assertEqual(config.REPORTING_INTERVAL_SECS, 10)
+        self.assertEqual(config.REPORTING_TTL_SECS, 20)
```
emmagordon: These asserts are to check that the environment values were used, but you've not currently got ttl set in the environment.

calico/felix/test/test_config.py
```diff
@@ -273,4 +283,115 @@ def test_ip_in_ip_enabled_bad(self):
...
+    def test_reporting_ttl_not_int(self):
+        """
+        Test exception is rased if status report ttl has invalid (non-integer) value.
+        """
+        with mock.patch('calico.common.complete_logging'):
+            config = Config("calico/felix/test/data/felix_missing.cfg")
+        cfg_dict = { "InterfacePrefix": "blah",
+                     "ReportingTTLSecs": "NaN"}
+        with self.assertRaisesRegexp(ConfigException,
+                                     "Field was not integer.*"):
```
emmagordon: Your exception raising code in config.py claims to raise an exception with the message 'Reporting TTL was not integer'. If your test is hitting a 'Field was not integer' exception then it seems there is already more generic error handling code and you do not need the integer checking for ttl/interval in config.py. Alternatively, your test is not working.

calico/felix/test/test_config.py
```diff
@@ -273,4 +283,115 @@ def test_ip_in_ip_enabled_bad(self):
...
+    def test_reporting_interval_not_int(self):
+        """
+        Test exception is rased ifstatus reporting interval has invalid (non-integer) value.
+        """
+        with mock.patch('calico.common.complete_logging'):
+            config = Config("calico/felix/test/data/felix_missing.cfg")
+        cfg_dict = { "InterfacePrefix": "blah",
+                     "ReportingIntervalSecs": "NaN"}
+        with self.assertRaisesRegexp(ConfigException,
+                                     "Field was not integer.*"):
```
emmagordon: Ditto above comment about your more specific error checking code in config.py either not being necessary or this test isn't working correctly.

calico/felix/test/test_config.py
```diff
@@ -273,4 +283,115 @@ def test_ip_in_ip_enabled_bad(self):
...
+        Test that exception is raised when status reporting ttl is less than interval.
+        """
+        with mock.patch('calico.common.complete_logging'):
+            config = Config("calico/felix/test/data/felix_missing.cfg")
+        cfg_dict = { "InterfacePrefix": "blah",
+                     "ReportingIntervalSecs": 47,
+                     "ReportingTTLSecs": 4 }
+
+        with self.assertRaisesRegexp(ConfigException,
+                                     "Reporting TTL.*?less or equal reporting interval.*"):
```
emmagordon: Note this will need updating once you change this grammatically incorrect error message as per above comment in config.py

calico/felix/test/test_fetcd.py
```diff
@@ -45,6 +45,7 @@ class TestEtcdAPI(BaseTestCase):
     @patch("gevent.spawn", autospec=True)
     def test_create(self, m_spawn, m_etcd_watcher):
         m_config = Mock(spec=Config)
+        m_config.ETCD_ADDR = 'localhost:4001'
```
emmagordon: Move 'localhost:4001' into a constant somewhere rather than repeatedly hard coding it. Makes life easier if etcd_addr ever changes in future!

calico/felix/test/test_fetcd_api.py
```diff
@@ -0,0 +1,163 @@
...
+from mock import Mock, call, patch
+from calico.datamodel_v1 import key_for_status, key_for_uptime
+from calico.felix.fetcd import EtcdAPI
+from calico.felix.test.base import BaseTestCase
+
+
+_log = logging.getLogger(__name__)
+
+# To make testing continuous status reporting more convenient,
+# we speed up sleeping.
```
emmagordon: This code is not inside a function/class, so importing this module will cause sleeping to speed up, perhaps unexpectedly - this could really catch someone out. 

Would be better to move this into a function, and then the single test that needs it can call that.

Also, do we need to set it back again afterwards?

calico/felix/test/test_fetcd_api.py
```diff
@@ -0,0 +1,163 @@
...
+            self.etcd_api.write_to_etcd('key', 'value', async=True)
+            self.run_actor_loop()
+            self.assertTrue(self.etcd_api.client.write.called)
+
+    def test_update_felix_status_disabled(self):
+        """
+        Test reporting is disabled for reporting interval 0
+        """
+        self.finish_setup()
+        gevent.sleep(1)
```
emmagordon: I'm not convinced by the logic of this test - if the interval was set to more than the time you're sleeping for you'd also expect write_to_etcd not to have been called during this period. Would be better to check that the status reporting greenlet returned (which only happens if reporting is disabled).

calico/felix/test/test_fetcd_api.py
```diff
@@ -0,0 +1,163 @@
...
+        status_key = key_for_status(hostname)
+        uptime_key = key_for_uptime(hostname)
+        ttl = self.etcd_api._config.REPORTING_TTL_SECS
+
+        gevent.sleep(50)
+
+        status_call = call(status_key, TestIfStatus(), async=True)
+        uptime_call = call(uptime_key, TestIfUptime(), ttl=ttl, async=True)
+        self.etcd_api.write_to_etcd.assert_has_calls(16 * [status_call,
+                                                           uptime_call])
```
emmagordon: We could also test that uptime has incremented by <status reporting interval> seconds between each call to write status to etcd. Similarly for timestamps.

calico/felix/test/test_fetcd_api.py
```diff
@@ -0,0 +1,163 @@
...
+
+        gevent.sleep(1)
+
+        self.assertTrue(_reconnect.called)
+
+
+class TestIfStatus(object):
+    """
+    Used to check whether status has expected format
+    i.e. whether timestamp is in ISO 8601 Zulu format
```
emmagordon: This should be extended to test that status is the expected json doc, and that it contains a 'status_time' whose value is in the expected time stamp format.

calico/felix/test/test_fetcd_api.py
miroslav21: As m_config is Mock object here, calling update method does not work as it should. Using setattr is the only way I could think of to circumvent this.

calico/openstack/mech_calico.py
```diff
@@ -196,14 +205,155 @@ def _init_state(self):
 
         self._my_pid = current_pid
 
-        # Start our resynchronization process.  Just in case we ever get two
-        # threads running, use an epoch counter to tell the old thread to die.
+        # Start our resynchronization process and status updating.  Just in
```
joe-marshall: You've added an extra 'same' into the next line - not sure why.

calico/felix/felix.py
```diff
@@ -48,7 +48,7 @@
 from calico.felix.fetcd import EtcdAPI
 
 _log = logging.getLogger(__name__)
-
+_log.setLevel(logging.DEBUG)
```
alexwlchan: I'm confused by the rearrangement of newlines in this file:

* There should be two newlines about the definition of `_main_greenlet`, as there were before.
* It's not clear why you've added the newlines below

calico/felix/felix.py
```diff
@@ -204,6 +207,7 @@ def main():
 
     try:
         config = Config(options.config_file)
+        _log.debug("felix cfg loaded")
```
alexwlchan: Perhaps it would be nice for this log to say where the config was loaded from; include options.config_file in the message?

Probably of marginal utility, but it might be useful and I see no reason not to.

calico/openstack/mech_calico.py
```diff
@@ -150,11 +154,16 @@ def __init__(self):
         # Initialize fields for the database object and transport.  We will
         # initialize these properly when we first need them.
         self.db = None
```
joe-marshall: Why isn't db_context a private attribute? It's not accessed anywhere outside the class.

calico/openstack/mech_calico.py
```diff
@@ -150,11 +154,16 @@ def __init__(self):
...
+        self.db_context = None
         self.transport = None
         self._my_pid = None
         self._periodic_resync_greenlet = None
         self._epoch = 0
 
+        # Prepare client for accessing status updates in etcd.
+        self._client = None
```
joe-marshall: Same question as above - should this be private?

calico/openstack/mech_calico.py
```diff
@@ -196,14 +205,163 @@ def _init_state(self):
...
+        """
+        This method acts as a status updates handler logic for the
+        Calico mechanism driver. Watches for felix updates in etcd
+        and passes info to Neutron database.
+        """
+        LOG.info("Handle status updating thread started")
+        self.db_context = ctx.get_admin_context()
+
+        # We only read initial felix instances in etcd if we are master.
+        if self.transport.is_master:
```
joe-marshall: Don't you need to check the epoch before registering Felixes? It's a pretty small window, but the epoch could have changed by this point, couldn't it?

calico/openstack/mech_calico.py
```diff
@@ -196,14 +205,163 @@ def _init_state(self):
...
+            self._client = etcd.Client(host=cfg.CONF.calico.etcd_host,
+                                       port=cfg.CONF.calico.etcd_port)
+            self._register_initial_felixes()
+
+        while self._epoch == expected_epoch:
```
joe-marshall: Why create the client here, when you may never use it (i.e. if you are not master)? For that matter, why recreate the client at all, when you already created it above? Either re-use your client, or create it in _poll_and_handle_felix_updates.

calico/openstack/mech_calico.py
```diff
@@ -196,14 +205,163 @@ def _init_state(self):
...
+            self._client = etcd.Client(host=cfg.CONF.calico.etcd_host,
+                                       port=cfg.CONF.calico.etcd_port)
+
+            # Only handle updates if we are the master node.
+            if self.transport.is_master:
+                LOG.info("I am master: polling felix updates in etcd")
+                self._poll_and_handle_felix_updates(expected_epoch)
+            else:
+                # Short sleep interval before we check if we've become
+                # the master.
```
joe-marshall: Should you log in this branch?

calico/openstack/mech_calico.py
```diff
@@ -196,14 +205,163 @@ def _init_state(self):
...
+                # Since we're polling on a subtree, we can't just
+                # increment the index, we have to look at the
+                # modifiedIndex.
+                self.next_etcd_index = max(self.next_etcd_index,
+                                           int(response.modifiedIndex)) + 1
+        except (ReadTimeoutError, SocketTimeout) as e:
+            # This is expected when we're doing a poll and nothing
+            # happened. socket timeout doesn't seem to be caught by
+            # urllib3 1.7.1. Simply reconnect.
+            LOG.debug("Read from etcd timed out (%r), retrying.", e)
```
joe-marshall: Note that this client will never be used - because a new one gets created in the while loop in _status_updating_thread.

calico/openstack/mech_calico.py
```diff
@@ -196,14 +205,163 @@ def _init_state(self):
...
+                                       port=cfg.CONF.calico.etcd_port)
+        except (etcd.EtcdClusterIdChanged, etcd.EtcdEventIndexCleared) as e:
+            # Complete resync is needed.
+            LOG.warning("Out of sync with etcd (%r).  Reconnecting "
+                     "for full sync.", e)
+            try:
+                # Connect the client
+                self._client = etcd.Client(host=cfg.CONF.calico.etcd_host,
+                                           port=cfg.CONF.calico.etcd_port)
+                context = ctx.get_admin_context()
```
joe-marshall: The next few lines of code should be refactored into a method, which can be called both from here, and from the periodic resync thread. Speaking of which...

What happens if the resync thread gets scheduled while this thread is resyncing? The answer might be that everything is fine, but if the two threads can trip each other up, then you will need to have some kind of flag saying 'resync in progress', to stop the two threads interfering with each other.

calico/openstack/mech_calico.py
```diff
@@ -196,14 +205,163 @@ def _init_state(self):
...
+                # Resync profiles.
+                self.resync_profiles(context)
+                # Now, set the config flags.
+                self.transport.provide_felix_config()
+            except Exception:
+                LOG.exception("Error in full resync.")
+        except Exception as e:
+            ## What other exceptions do we expect?
+            ## Is full resync needed on unexpected exception?
+            LOG.info("Handling status update failed. ({})".format(e))
```
joe-marshall: As above, this client will never be used.

calico/openstack/mech_calico.py
```diff
@@ -196,14 +205,163 @@ def _init_state(self):
...
+
+    def _handle_status_update(self, response):
+        """
+        Handle updates of felix instances written to etcd. Pass status
+        updates to neutron database.
+
+        :param response: update written to etcd by one of felix instances
+        """
+        # Only get info from uptime keys - since those have ttl, we know
+        # if key was updated or created.
```
joe-marshall: I'm not keen on the magic number 7 here - please define a constant, or use a more explicit regex to show your assumptions about the key structure.

calico/openstack/mech_calico.py
```diff
@@ -1038,3 +1196,18 @@ def profiles_match(etcd_profile, neutron_profile):
...
+
+def _hostname_from_status_key(key):
+    '''
+    Help function to get hostname from a status key.
+
+    :param: key for felix status
+            expected key format: <STATUS_DIR>/<hostname>/<rest of status key>
+    '''
+    key = list(key)
+    in_host_dir = key[len(STATUS_DIR + '/'):]
```
joe-marshall: This will work, but it's nasty. You should use a regex to parse the key.

joe-marshall:
Reviewed the mech_calico changes. Mostly good, but I have a couple of concerns about the interaction between the status_update and periodic_resync threads, and about the inefficient way that client objects get created but not used.

calico/openstack/test/test_plugin_etcd.py
```diff
@@ -95,14 +109,23 @@ def assertEtcdDeletes(self, expected):
             self.assertEqual(self.recent_deletes, expected)
         self.recent_deletes = set()
 
-    def etcd_read(self, key, recursive=False, timeout=None):
+    def etcd_read(self, key, wait=False, waitIndex=None, recursive=False, timeout=None):##add wait&waitIndex
```
alexwlchan: Two notes:

1. Python convention for variable names is lowercase_with_underscores, not camelCase. (Exceptions and classes use PascalCase.) The variables throughout this function should be renamed accordingly.
2. There's an extra comment on the end of this line, but it doesn't seem to add anything. Please remove it.

calico/openstack/test/test_plugin_etcd.py
```diff
@@ -95,14 +109,23 @@ def assertEtcdDeletes(self, expected):
...
         """Read from the accumulated etcd database.
         """
         self.maybe_reset_etcd()
 
+        ## Slow down reading from etcd status subtree to allow threads to run more often
+        if wait and key == STATUS_DIR:##
```
alexwlchan: Comments should begin with a single octothorpe, and the empty comment should be removed.

calico/openstack/test/test_plugin_etcd.py
```diff
@@ -129,7 +152,15 @@ def etcd_read(self, key, recursive=False, timeout=None):
             print "children: %s" % [child.key
                                     for child in read_result.children]
             if read_result.value is None and read_result.children == []:
-                raise lib.m_etcd.EtcdKeyNotFound()
+                raise lib.m_etcd.EtcdKeyNotFound(self.etcd_data)
+
+            read_result._children = []##actual direct children of the dir in etcd response. needed for status_dir, where children are dirs and gotta be iterated
```
alexwlchan: As above, comment should be prefixed with a single hash. Since this is a long comment, it should probably be on its own line and wrapped to 72 characters.

calico/openstack/test/test_plugin_etcd.py
```diff
@@ -129,7 +152,15 @@ def etcd_read(self, key, recursive=False, timeout=None):
...
+            list_of_statuses = [{"key": K} for K in self.etcd_data.keys()]
+            read_result._children.append({"nodes": list_of_statuses})
+            #takto vyzeraju _children
+            # {'nodes': [list of status]}
+            # kde status:  {'key': '/../uptime'}
```
alexwlchan: I don't speak Slovak, so this comment isn't useful to me. Please could you rewrite it in English?

calico/openstack/test/test_plugin_etcd.py
```diff
@@ -185,6 +216,7 @@ def test_start_no_ports(self):
                 ["SGID-default"]
         })
 
+
```
alexwlchan: Style nitpick: methods should only be separated by a single newline, not two.

calico/felix/test/test_fetcd_api.py
```diff
@@ -0,0 +1,177 @@
...
+class TestIfUptime(object):
+    """
+    Used to check whether uptime has correct format (i.e. whether it is
+    non-negative integer)
+    """
+    def __eq__(self, other):
+        is_int = type(other) == int
+        is_non_negative = other >= 0
+        return is_int and is_non_negative
+    def __repr__(self):
```
alexwlchan: Style nitpick: should be a single newline between methods; same as the method above. I would also suggest adding some parens to make the booleans easier to read; i.e.,

```python
is_int = (type(other) == int)
is_non_negative = (other >= 0)
```

calico/felix/test/test_fetcd_api.py
```diff
@@ -0,0 +1,177 @@
...
+        return '%s()' % self.__class__.__name__
+
+def is_json(object):
+  try:
+    json.loads(object)
+  except ValueError, e:
+    return False
+  return True
```
alexwlchan: A few nitpicks:

* This function should have a docstring
* Indentation should be four spaces, not 2
* There's no need to capture the exception as `e` in `except ValueError, e:`; you can just do `except ValueError`.

  (And the preferred form for this construction is `except ValueError as e`. That form works with Python 2.6 to 3.5, but the comma syntax has been dropped in Python 3.)

calico/felix/test/test_fetcd_api.py
```diff
@@ -0,0 +1,177 @@
...
+class TestIfStatus(object):
+    """
+    Used to check whether status has JSON format and contains expected
+    attributes, whis are:
+    - timestamp in ISO 8601 Zulu format
+    """
+    def __eq__(self, other):
+        import re
+        timestamp_regex = re.compile('.*"status_time": "\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}Z?".*')
+        has_timestamp =  timestamp_regex.match(str(other)) is not None
```
alexwlchan: Nitpick: lose the extra space.

Separately: I feel like I've commented on this block before (but perhaps I just thought it and never wrote it down). Using a regex to check that a block is JSON formatted and a timestamp is prone to errors. It would be better to:

* Try to parse as JSON (if it throws a parsing error => incorrectly formatted)
* Try to get the `status_time` key (if it doesn't exist => incorrectly formatted)
* Use the `strptime()` method from the datetime module to try to parse the time (if it doesn't work => incorrectly formatted)

A regex is fragile for this task. Some examples:

```python
'<"status_time": "2012-07-04T09:13:26Z>'  # not valid JSON, but will pass this regex
'{"status_time":"2012-07-04T09:13:26Z}'   # valid JSON, but will fail this regex because it's minified
'{"status_time":"9999-99-99T00:00:00Z}'   # valid JSON, will pass the regex, but an impossible timestamp
```

I'm also not clear why this isn't a function, like `is_json` below.

calico/datamodel_v1.py
```diff
@@ -32,13 +32,19 @@
 # All Calico data is stored under this path.
 ROOT_DIR = "/calico"
 
+# Current version
+VERSION = "/v1"
```
fasaxc: I don't think I agree with this change.  It makes it look like the version of the two subtrees should be incremented together, which is not necessarily the case.  We may want to rev the felix version without revving the openstack version, for example.

calico/felix/config.py
```diff
@@ -147,7 +147,7 @@ def __init__(self, config_path):
         """
         Create a config. This reads data from the following sources.
         - Environment variables
-        - Configuration file
+        - Configuration file - (/etc/calico/felix.cfg)
```
fasaxc: Nitpick: for consistency with the lines below, probably don't want a dash and a paren `- (`.

calico/felix/config.py
```diff
@@ -382,6 +389,28 @@ def _validate_cfg(self, final=True):
...
+        # interval and ttl to 0 - i.e. status reporting is disabled.
+        if self.REPORTING_INTERVAL_SECS <= 0:
+            self.REPORTING_TTL_SECS = 0
+            self.REPORTING_INTERVAL_SECS = 0
+
+        # Status report time to live must be more than status reporting interval,
+        # unless TTL is default.
+        if self.REPORTING_TTL_SECS <= self.REPORTING_INTERVAL_SECS\
+                and self.REPORTING_TTL_SECS != 0:
+            raise ConfigException("Reporting TTL ({} sec) is less than or equal "
```
fasaxc: I think I'd default the TTL in this case rather than raising an exception.  Imagine fat-fingering the config in a 1000-host cluster and watching your felixes die en masse.

calico/felix/felix.py
```diff
@@ -192,6 +192,7 @@ def dump_top_level_actors(log):
 
 
 def main():
+    _log.debug("Felix main started.")
```
fasaxc: The log statement comes before we've initialised the log (which is on the line below) so it'll likely just cause an error to stderr saying "no loggers found"

calico/felix/fetcd.py
```diff
@@ -87,20 +90,24 @@
...
+    etcd.
     """
 
     def __init__(self, config, hosts_ipset):
         super(EtcdAPI, self).__init__()
         self._config = config
 
+        # Timestamp storing when the EtcdAPI started. This info is needed
+        # in order to report uptime to etcd.
+        self._start_time = time.time()
```
fasaxc: `time.time()` shouldn't be used for this; if the clock is changed then we'll misreport our uptime.  We should use a monotonic clock instead.  I made a start on adding such a clock here: https://github.com/projectcalico/calico/compare/monotonic-time

calico/felix/fetcd.py
```diff
@@ -111,10 +118,20 @@ def __init__(self, config, hosts_ipset):
         self._resync_greenlet = gevent.spawn(self._periodically_resync)
         self._resync_greenlet.link_exception(self._on_worker_died)
 
+        # Create a client to write to etcd.
+        self.client = None
+        _reconnect(self)
```
fasaxc: Good thinking in trying to keep the code DRY but passing self to a function is a big "code smell", especially if the function is accessing `self._xxx` (private) fields.

In this case, you could probably use the mix-in pattern or a wrapper/proxy object that delegates to the etcd client and holds the config.

calico/felix/fetcd.py
```diff
@@ -132,6 +149,67 @@ def _periodically_resync(self):
             gevent.sleep(sleep_time)
             self.force_resync(reason="periodic resync", async=True)
 
+    @logging_exceptions
+    def _status_reporting(self):
```
fasaxc: Nitpick: `_status_reporting` isn't a verb phrase.  `_periodically_report_status` is wordy but matches the method above

calico/felix/fetcd.py
```diff
@@ -132,6 +149,67 @@ def _periodically_resync(self):
...
+        _log.debug("Reporting interval: %s, TTL: %s", interval, ttl)
+
+        if interval == 0:
+            _log.info("Interval is 0, status reporting disabled.")
+            return
+
+        while True:
+            try:
+                self.update_felix_status(ttl)
+                gevent.sleep(interval)
```
fasaxc: You should jitter this interval.  At scale it's important to prevent multiple felixes from getting synchronised.

calico/felix/fetcd.py
```diff
@@ -132,6 +149,67 @@ def _periodically_resync(self):
...
+        Writes two keys to etcd:
+        uptime in secs
+        felix status in JSON - containing current time in ISO 8601 Zulu format
+
+        :param: ttl int: time to live in sec - lifetime of the status report
+        """
+        status_key = key_for_status(self._config.HOSTNAME)
+        status = {}
+
+        time_now = datetime.datetime.fromtimestamp(time.time())
```
fasaxc: Simpler to do `datetime.datetime.now()`.

calico/felix/fetcd.py
```diff
@@ -132,6 +149,67 @@ def _periodically_resync(self):
...
+
+        if interval == 0:
+            _log.info("Interval is 0, status reporting disabled.")
+            return
+
+        while True:
+            try:
+                self.update_felix_status(ttl)
+                gevent.sleep(interval)
+            except EtcdException as e:
```
fasaxc: This except block cannot be hit because you pass `async=True` to the `write_to_etcd` method; which will make it return an `AsyncResult` instead of raising the exception.  that means that any exceptions will take down the process instead.

Also, there are other exceptions that need to be dealt with.  The etcd client may raise a read or connection timeout, for example.

calico/felix/fetcd.py
fasaxc: Sorry, `datetime.datetime.utcnow()`

calico/felix/fetcd.py
```diff
@@ -132,6 +149,67 @@ def _periodically_resync(self):
...
+        time_now = datetime.datetime.fromtimestamp(time.time())
+        time_formatted = time_now.replace(microsecond=0).isoformat()+'Z'
+        status['status_time'] = time_formatted
+
+        status_value = json.dumps(status)
+
+        uptime_key = key_for_uptime(self._config.HOSTNAME)
+        uptime_value = int(time.time() - self._start_time)
+
+        self.write_to_etcd(status_key, status_value, async=True)
```
fasaxc: I'm not sure we're getting any benefit from passing this back to the actor as a message. The etcd client is already threadsafe so we should just be abel to use it as `self.client.xxx`

calico/felix/fetcd.py
```diff
@@ -191,7 +269,7 @@ class _EtcdWatcher(gevent.Greenlet):
 
     def __init__(self, config, hosts_ipset):
         super(_EtcdWatcher, self).__init__()
-        self.config = config
+        self._config = config
```
fasaxc: You're renaming this from being a public field to being a private field, so your external function can access it; doesn't seem right to me

calico/felix/test/test_fetcd.py
```diff
@@ -325,7 +331,7 @@ def test_dispatch_delete_resync(self):
 
     def test_per_profile_del(self):
         """
-        Test profile deletion triggers dleetion for tags and rules.
+        Test profile deletion triggers deletion for tags and rules.
```
fasaxc: Ta

calico/felix/test/test_fetcd_api.py
```diff
@@ -0,0 +1,185 @@
+# -*- coding: utf-8 -*-
```
fasaxc: We usually keep all the tests for foo.py in test_foo.py, any reason not to combine this with the above?

calico/felix/test/test_fetcd_api.py
```diff
@@ -0,0 +1,185 @@
...
+        # Set configuration attributes and start etcd_api
+        for key, value in kwargs.iteritems():
+            setattr(self.m_config, key, value)
+
+        self.m_hosts_ipset = Mock()
+        self.etcd_api = EtcdAPI(self.m_config, self.m_hosts_ipset)
+
+        self.etcd_api.write_to_etcd = Mock()
+
+    def run_actor_loop(self):
```
fasaxc: You should use `self.step_actor(self.etcd_api)`.  That does almost exactly what you're doing here but it also patches the actor's greenlet to avoid trigerring assertions on certain codepaths.

calico/felix/test/test_fetcd_api.py
```diff
@@ -0,0 +1,185 @@
...
+
+    def test_update_felix_status_disabled(self):
+        """
+        Test reporting is disabled for reporting interval 0
+        """
+        self.finish_setup()
+        # We call gevent.sleep so that _status_reporting thread is executed -
+        # since no interval and ttl is given, _status_reporting thread returns
+        # instead of writing to etcd.
+        gevent.sleep(1)
```
fasaxc: We shouldn't use sleep for this.  Since it returns, you could just call it instead.  Even better, use gevent.timeout to call it and make sure you don't block forever if it doesn't return.

calico/felix/test/test_fetcd_api.py
```diff
@@ -0,0 +1,185 @@
...
+        self.finish_setup(REPORTING_INTERVAL_SECS=15,
+                          REPORTING_TTL_SECS=37)
+
+        hostname = self.etcd_api._config.HOSTNAME
+        status_key = key_for_status(hostname)
+        uptime_key = key_for_uptime(hostname)
+        ttl = self.etcd_api._config.REPORTING_TTL_SECS
+
+        # We call gevent.sleep so that _status_reporting thread is executed
+        gevent.sleep(1)
```
fasaxc: Again, I'm not keen on using sleep for this, it's very brittle and it slows down the test runs.  Instead, you can split out the "step" function of the loop and execute it manually (this is exactly why `Actor._step` exists).  Then, separately, you can test that the loop calls the step function.

calico/felix/test/test_fetcd_api.py
```diff
@@ -0,0 +1,185 @@
...
+        hostname = self.etcd_api._config.HOSTNAME
+        status_key = key_for_status(hostname)
+        uptime_key = key_for_uptime(hostname)
+        ttl = self.etcd_api._config.REPORTING_TTL_SECS
+
+        status_call = call(status_key, TestIfStatus(), async=True)
+        uptime_call = call(uptime_key, TestIfUptime(), ttl=ttl, async=True)
+
+        # We call gevent.sleep so that _status_reporting thread is executed
+        gevent.sleep(5)
```
fasaxc: This test will take at least 10s if I'm reading it right.  That's *far* too long.  Right now all the tests run in 12s and that really reduces development cycles.  You need to find a way to avoid needing to sleep.

calico/felix/test/test_fetcd_api.py
```diff
@@ -0,0 +1,185 @@
...
+
+        hostname = self.etcd_api._config.HOSTNAME
+        status_key = key_for_status(hostname)
+        uptime_key = key_for_uptime(hostname)
+        ttl = self.etcd_api._config.REPORTING_TTL_SECS
+
+        # We call gevent.sleep so that _status_reporting thread is executed
+        gevent.sleep(1)
+
+        status_call = call(status_key, TestIfStatus(), async=True)
```
fasaxc: Can't decide if I like TestIfStatus() or not, I'd usually do this to unpack the arguments, which is fewer lines of code, but they are more ugly:
```
fn_name, args, kwargs = self.etcd_api.write_to_etcd.mock_calls[0]
status_key, if_status = args
self.assertEqual(status_key, ...)
```

calico/felix/test/test_fetcd_api.py
```diff
@@ -0,0 +1,185 @@
...
+        self.finish_setup(REPORTING_INTERVAL_SECS=4,
+                          REPORTING_TTL_SECS=12)
+        self.etcd_api.write_to_etcd = Mock(side_effect=EtcdException)
+
+        gevent.sleep(1)
+
+        self.assertTrue(_reconnect.called)
+
+
+class TestIfStatus(object):
```
fasaxc: Thinking on these more, I think a good approach would be to define a generic class that converts a "validation" function into an object with an `__eq__`.  That'd be really useful as a general solution to this problem.  Right now, there's a lot of boilerplate to define one of these classes.
```
def is_valid_json(value):
    json.loads(value)  # Exception if not valid
    return True
c = call("foo", Check(is_valid_json), bar="baz")
```

calico/felix/test/test_fetcd_api.py
```diff
@@ -0,0 +1,185 @@
...
+        is_int = (type(other) == int)
+        is_non_negative = (other >= 0)
+        return is_int and is_non_negative
+
+    def __repr__(self):
+        return '%s()' % self.__class__.__name__
+
+def is_json(object):
+    """
+    Help function, returns wheteher given object is json.
```
fasaxc: Typo: wheteher

calico/openstack/mech_calico.py
```diff
@@ -150,10 +154,16 @@ def __init__(self):
...
         self.db = None
+        self._db_context = None
         self.transport = None
         self._my_pid = None
         self._periodic_resync_greenlet = None
         self._epoch = 0
+        self.in_resync = False
+
+        # Prepare client for accessing status updates in etcd.
+        self._client = None
```
fasaxc: I think the split between the mech_calico and transport modules is supposed to be that transport owns the etcd client(s).  why have we created another one in mech_calico?

calico/openstack/mech_calico.py
```diff
@@ -196,14 +206,148 @@ def _init_state(self):
...
+        """
+        This method acts as a status updates handler logic for the
+        Calico mechanism driver. Watches for felix updates in etcd
+        and passes info to Neutron database.
+        """
+        LOG.info("Handle status updating thread started")
+        self._db_context = ctx.get_admin_context()
+
+        # We only read initial felix instances in etcd if we are master.
+        if self.transport.is_master:
```
fasaxc: It's incorrect to do this before starting the loop.  There may be no master when the thread is started; it may take 60s for a master to be elected.

calico/openstack/mech_calico.py
fasaxc: I think that gets spammy, so debug log if anything.

calico/openstack/mech_calico.py
```diff
@@ -196,14 +206,148 @@ def _init_state(self):
...
+            # Only handle updates if we are the master node.
+            if self.transport.is_master:
+                LOG.info("I am master: polling felix updates in etcd")
+                self._poll_and_handle_felix_updates(expected_epoch)
+            else:
+                # Short sleep interval before we check if we've become
+                # the master.
+                eventlet.sleep(MASTER_CHECK_INTERVAL_SECS)
+        else:
+            LOG.warning("Unexpected fork. "
```
fasaxc: `fork` has s technical meaning so best to avoid it.  I'd probably say: "Unexpected: epoch changed; update thread exiting"

calico/openstack/mech_calico.py
```diff
@@ -196,14 +206,148 @@ def _init_state(self):
...
+                                           int(response.modifiedIndex)) + 1
+        except (ReadTimeoutError, SocketTimeout) as e:
+            # This is expected when we're doing a poll and nothing
+            # happened. socket timeout doesn't seem to be caught by
+            # urllib3 1.7.1. Simply reconnect.
+            LOG.debug("Read from etcd timed out (%r), retrying.", e)
+        except (etcd.EtcdClusterIdChanged, etcd.EtcdEventIndexCleared) as e:
+            # Complete resync is needed.
+            LOG.warning("Out of sync with etcd (%r).  Reconnecting "
+                     "for full sync.", e)
```
fasaxc: Alignment

calico/openstack/mech_calico.py
```diff
@@ -196,14 +206,148 @@ def _init_state(self):
...
+        :param response: update written to etcd by one of felix instances
+        """
+        # Only get info from uptime keys - since those have ttl, we know
+        # if key was updated or created.
+        key = response.key
+        if not is_uptime_key(key):
+            return
+        # Ignore deleted and expired keys - their deletion is noticed by
+        # other timeout algorithm before reaching the horizon UI.
+        if response.action == "delete" or response.action == "expire":
```
fasaxc: You missed "compareAndDelete".  You should use `etcdutils.ACTION_MAPPING`

calico/openstack/mech_calico.py
```diff
@@ -196,14 +206,148 @@ def _init_state(self):
...
+            return
+
+
+        hostname = _hostname_from_status_key(key)
+        agent_state = {'agent_type': AGENT_TYPE_FELIX,
+                       'binary': '',
```
fasaxc: What is binary supposed to be, looks like we're skipping it?

calico/openstack/mech_calico.py
```diff
@@ -196,14 +206,148 @@ def _init_state(self):
...
+                       'host': hostname,
+                       'topic': constants.L2_AGENT_TOPIC}
```
fasaxc: Hmm, seems fishy that we use the L2_AGENT_TOPIC, when Calico is an L3 solution.  What is this used for.  

CC @neiljerram

calico/openstack/mech_calico.py
fasaxc: If it's meant to be the version, it should be pretty straightforward to get felix to fill that into the data it puts in etcd.

calico/openstack/mech_calico.py
```diff
@@ -196,14 +206,148 @@ def _init_state(self):
...
+        response = self._client.read(STATUS_DIR, recursive=True)
+
+        # Read through hosts
+        for host in response._children:
+            # Read through status updates of the node.
+            for status in host['nodes']:
+                key = status['key']
+                # If host has uptime key, we pass the host to the neutron
+                # database with start_flag set True.
+                if key[-7:] == '/uptime':
```
fasaxc: Not keen on the magic number -7 here and this will break if we add any other keys that happen to end in uptime.  Better to parse this properly (probably with a regex in the datamodel module).

calico/openstack/mech_calico.py
```diff
@@ -196,14 +206,148 @@ def _init_state(self):
...
+                    hostname = _hostname_from_status_key(key)
+                    agent_state = {'agent_type': AGENT_TYPE_FELIX,
```
fasaxc: this code is duplicated, you should wrap it up in a function

calico/openstack/mech_calico.py
```diff
@@ -591,6 +727,23 @@ def periodic_resync_thread(self, expected_epoch):
         else:
             LOG.warning("Periodic resync thread exiting.")
 
+    def complete_resync(self, context):
+        """Go through complete resync.
+        """
+        if not self.in_resync:
```
fasaxc: I don't think this is correct.  There's no guarantee that because this resync is started, it's safe to ignore another resync that comes in.  You could make it more correct by using a proper lock so the two threads would queue up to enter this method.

However, I'm very uneasy about having two threads with overlapping responsibilities.  It doesn't seem right that your new thread calls resync_enpoints via this method, for example.

Is there anything wrong with leaving the two threads as "ships in the night", responsible for different subtrees?

calico/openstack/mech_calico.py
```diff
@@ -1038,3 +1191,24 @@ def profiles_match(etcd_profile, neutron_profile):
         (etcd_rules == neutron_group_rules) and
         (etcd_tags == neutron_group_tags)
     )
+
+def _hostname_from_status_key(key):
```
fasaxc: These utility functions belong in datamodel.py

calico/openstack/mech_calico.py
```diff
@@ -1038,3 +1191,24 @@ def profiles_match(etcd_profile, neutron_profile):
...
+    '''
+    Help function to get hostname from a status key.
```
fasaxc: Typo: "Help"

calico/datamodel_v1.py
```diff
@@ -81,6 +87,18 @@ def dir_for_per_host_config(hostname):
...
+def dir_for_status(hostname):
+    return STATUS_DIR + "/%s" % hostname
+
+
+def key_for_status(hostname):
+    return dir_for_status(hostname) + "/last_reported_status"
+
+
+def key_for_uptime(hostname):
+    return dir_for_status(hostname) + "/uptime"
```
fasaxc: Since we'll have to live with this API for a long time, I'm wondering if it should have some mention of felix in it since "host/uptime" looks like the uptime of the host, rather than felix.   Maybe just change this to "felix_uptime"?

Also, the string "(felix_)uptime" appears all over the place, it should be in a constant in this file.

calico/openstack/mech_calico.py
neiljerram: Yes, it is.  Resolving that will be part of the forthcoming big conflab about how Calico-style connectivity should really be represented in Neutron.

alexwlchan:
@fasaxc Any reason not to close this in favour of #813?

fasaxc:
Replaced by #814

PR #738, opened 2015-07-21
Tolerate ports being deleted mid-resync

Lukasa:
This fixes a problem we saw in our automated testing, where the resync fired mid-port-delete. This led the earlier etcd read to succeed, but the port lookup to fail.

### DIFF
--- a/calico/openstack/mech_calico.py
+++ b/calico/openstack/mech_calico.py
@@ -32,6 +32,7 @@ from functools import wraps
 
 # OpenStack imports.
 from neutron.common import constants
+from neutron.common.exceptions import PortNotFound
 from neutron.db import models_v2
 from neutron.plugins.ml2 import driver_api as api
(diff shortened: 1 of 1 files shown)


fasaxc:
Change looks good but I had a skim over the code and it looked like there were a few places where we call `self.db.get_xxx` without catching an exception.  The ones in the post_commit hooks are probably OK because the request would fail, which is pretty reasonable.

What about where we get the SG rules, for example.  Is it a problem if the SG was deleted under us of will we get an empty list?

Lukasa:
>What about where we get the SG rules, for example. Is it a problem if the SG was deleted under us of will we get an empty list?

Yeah, I checked that: we'll get an empty list. This strictly means that we'll have the security group in etcd for one resync cycle longer, but I'd rather do that than run the risk of missing that a group no longer has any rules in it.

PR #739, opened 2015-07-21
Don't log about bad JSON at such a high level.

Lukasa:
There's no reason for this log to be error level, it's scary for no reason.


matthewdupre:
I'll accept that it's a warning (because we'll rewrite anyway, so who cares what's in etcd); but if this is something that can actually be hit in a non-tampered with system then I think the log needs to be made clearer.

Merging for now to get the tests passing, but if we can hit this without someone having written nonsense to etcd, I think we need to improve the log a bit.

matthewdupre:
(Also the pull request builder seems to be lying about the build failing.)

PR #740, opened 2015-07-22
Some frequently asked questions.

Lukasa:


docs/source/faq.rst
```diff
@@ -89,6 +92,94 @@ state. All of our components can be shutdown and restarted without risk,
 because they resynchronize state as necessary. This makes modelling
 their behaviour extremely simple, reducing the complexity of bugs.
 
+"Does Calico use microsegmentation?"
```
matthewdupre: I don't understand this (and it sounds like you're not clear either).  We think we should drop it for now, and ask Chris if (/what) he thinks we should put back in.

Lukasa: Dealt with.

PR #746, opened 2015-07-29
Update links to reflect the organisation/repo change

alexwlchan:
All these links redirect correctly, but this saves people a redirect and means we'll be consistent with new links going forward.

@Lukasa Assigning to you because you've been taking the lead on the repository migration.


Lukasa:
LGTM. :cake:

PR #748, opened 2015-07-29
Add (IPS trace-like) message tracking to Felix

PatrickKidger:
Not looking to merge before 1.0.

Adds additional message logging options, which can be read by an external program to generate an IPS trace. This is toggled with an additional config option MessageTracking (defaults to False, i.e. off), and the location of the log set with an additional config option MessageTrackingLogFilePath (defaults to /var/log/calico/calico-message-tracker.log)

Also updates some actor names for increased specificity.

@joe-marshall @fasaxc


alexwlchan:
ok to test

calico/common.py
```diff
@@ -281,6 +286,20 @@ def complete_logging(logfile=None,
...
+            if not _message_log.handlers:
+                mkdir_p(os.path.dirname(messagelogfile))
+                _message_log.setLevel(logging.DEBUG)
+                _fh = logging.handlers.WatchedFileHandler(messagelogfile)
+                _fh.setLevel(logging.DEBUG)
+                _fh.setFormatter(logging.Formatter('%(message)s'))
+                _message_log.addHandler(_fh)
+                _message_log.disabled = False
+        else:
+            _message_log.disabled = True
```
alexwlchan: It looks as if `_message_log.disabled` won't be set if we pass the first `if` statement, but not the second. Could that be a problem?

calico/felix/actor.py
```diff
@@ -226,12 +237,30 @@ def _step(self):
...
                 actor_storage.msg_name = msg.name
                 try:
+
+                    # Keeps track of how many times we retry a message.
+                    try:
+                        # Increment the counter if it already exists
+                        msg_retries[msg.uuid] += 1
+                    except KeyError:
+                        # Else set it up
+                        msg_retries[msg.uuid] = 1
```
alexwlchan: Rather than doing this yourself, you might want to look at the [`collections` module](https://docs.python.org/2/library/collections.html). The `defaultdict` class provides essentially the same functionality as what you've written here, but it's a bit cleaner.

calico/common.py
```diff
@@ -271,6 +275,7 @@ def complete_logging(logfile=None,
                 handler.setLevel(file_level)
 
     # If we've been given a log file, log to file as well.
+
```
alexwlchan: There are several blank lines like this, seemingly added at random &ndash; presumably accidental?

calico/felix/actor.py
```diff
@@ -232,11 +250,31 @@ def _step(self):
...
+                                      "'function': '%s', "
+                                      "'tries': %s, "
+                                      "'exception': '%s'"
+                                      "}" % (msg.uuid,
+                                             int(time.time() * 1000),
+                                             msg.caller,
+                                             msg.recipient,
+                                             msg.name,
+                                             msg_tries[msg.uuid],
+                                             message_exception))
```
alexwlchan: This is fine, but perhaps it could be rewritten to use json.dumps() instead? My thinking:

* Using the json module is probably a bit more robust when it comes to quoting strings (for example, if there's an apostrophe in the message_exception variable)
* The keys and values would be directly together, rather than split via %s-formatting tokens. Reduces the chance that pairs are inadvertently swapped.
* It would be a little easier to add new fields here.

You can't quite do everything with the json module (`"key1": {...}` isn't valid JSON), but you can construct the big dictionary this way.

calico/felix/actor.py
```diff
@@ -385,7 +426,8 @@ class Message(object):
     """
     Message passed to an actor.
     """
-    def __init__(self, msg_id,  method, results, caller_path, recipient,
+
+    def __init__(self, msg_id, method, results, caller_path, recipient,
```
alexwlchan: You've added more blank lines here (although good catch on the extra space). Any reason for that?

alexwlchan:
We'd like to get this merged at some point, but Patrick has finished his internship.

I'm going to own this patch and get it reviewed/merged, but I don't have a firm timetable for doing so.

alexwlchan:
The only non-trivial merge conflict is in calico/felix/fiptables.py:

```python
<<<<<<< HEAD
    def __init__(self, table, ip_version=4):
        super(IptablesUpdater, self).__init__(qualifier="%s v%d" % (table, ip_version))
=======
    def __init__(self, table, config, ip_version=4):
        super(IptablesUpdater, self).__init__(qualifier="v%d-%s" %
                                                        (ip_version, table))
>>>>>>> master
```

I can't see any difference between the two, so I'm going to resolve in favour of master and assume everything will be okay.

alexwlchan:
@fasaxc Assigning this to you for review, but this is a lower priority than Miroslav's changes (since he's still around in the office).

matthewdupre:
ok to test

calico/common.py
```diff
@@ -35,6 +35,8 @@
 from types import StringTypes
 from netaddr.strategy import eui48
 
+MESSAGE_LOG_NAME = "message_tracking_log"
```
fasaxc: This should be under "calico.".  You could even put it under the actor namespace at "calico.felix.actor.messages", for example.

calico/felix/actor.py
```diff
@@ -232,11 +249,28 @@ def _step(self):
                 except BaseException as e:
                     _log.exception("Exception processing %s", msg)
                     results.append(ResultOrExc(None, e))
+                    message_exception = e.__class__.__name__ +\
+                                                      ': ' + str(e)
```
fasaxc: I think `repr(e)` usually does what you want here

calico/felix/actor.py
```diff
@@ -232,11 +249,28 @@ def _step(self):
...
                     _stats.increment("Messages executed with exception")
                 else:
+                    message_exception = None
                     results.append(ResultOrExc(result, None))
                     _stats.increment("Messages executed OK")
                 finally:
+                    batch_uuids.add(msg.uuid)
+                    # Log the message info
+                    # Strings are enclosed in quotes. Numbers aren't.
+                    _message_log.info('"{uuid}": {data}'.format(
```
fasaxc: This will do all the work to format this message even if logging isn't enabled, including translating the object to JSON.  That'll add a big overhead.  Since you're dumping the JSON, which requires a function call, you could guard this call with `if _message_log.isEnabledFor(logging.INFO):` so that it doesn't do any work if logging is turned off.

calico/felix/actor.py
```diff
@@ -232,11 +249,28 @@ def _step(self):
...
+                        uuid=msg.uuid,
+                        data=json.dumps({
+                            "received-time": int(time.time() * 1000),
```
fasaxc: "received-time" isn't a good name for a JSON field because "-" isn't a valid identifier character in most languages.  You should use an underscore.

fasaxc: Why store milliseconds instead of just leaving the timestamp as seconds?  Presumably you'll be parsing this with Python, so you'll just have to convert it back again.

calico/felix/actor.py
```diff
@@ -218,6 +227,10 @@ def _step(self):
             batch = self._start_msg_batch(batch)
             assert batch is not None, "_start_msg_batch() should return batch."
             results = []  # Will end up same length as batch.
+
+            # Will store the logs for the messages from this batch.
```
fasaxc: This variable doesn't store logs, it stores the UUIDs from the messages.  I'm not sure why it's needed, don't we already have a list of the messages in the batch?

calico/felix/actor.py
```diff
@@ -232,11 +249,28 @@ def _step(self):
...
+                    message_exception = e.__class__.__name__ +\
+                                                      ': ' + str(e)
                     _stats.increment("Messages executed with exception")
                 else:
+                    message_exception = None
                     results.append(ResultOrExc(result, None))
                     _stats.increment("Messages executed OK")
                 finally:
+                    batch_uuids.add(msg.uuid)
+                    # Log the message info
```
fasaxc: These comments aren't so useful.  Why are you logging the message info and what's special about this log?  Also, there's no need to describe JSON as "Strings are enclosed in quotes. Numbers aren't.", you can call it JSON.

calico/felix/actor.py
```diff
@@ -259,6 +293,13 @@ def _step(self):
                 # Most-likely a bug.  Report failure to all callers.
                 _log.exception("_finish_msg_batch failed.")
                 results = [(None, e)] * len(results)
+                for uuid in batch_uuids:
```
fasaxc: Better to loop over `batch` instead, I think, since that will give you the messages in order (whereas a set is not consistent in its order).

fasaxc:
General thoughts:

* Unfortunately, the logging as structured will be too expensive because it does lots of work to stringify the log messages as JSON before handing it off to Logger.info().
* There's a lot of duplication between the log calls.  We should have a utility method for logging that something happened to a message.  That method should return immediately if the function is disabled and it should handle formatting a generic JSON object.  This could even be a method on the Message object itself.
* We're using a logger but we're duplicating its functions (such as logging time in our own format).  

i.e.
```
msg.log_event("Sent")
msg.log_event("Received")
msg.log_event("Attempted")
msg.log_event("Succeeded")
msg.log_event("Failed", exc=e)
msg.log_event("Batch failed", exc=e)
```

fasaxc:
I think this has enough problems that we should close this PR and revisit if we decide to add this function later.

My number one concern is that this is all on the critical path of message passing so it needs to be very defensive and low overhead, especially when turned off.

Issue #725, opened 2015-07-14
Felix doesn't immediately apply config changes to etcd


spikecurtis:
Modifying configuration settings in the `/calico/vX/host/<host>/config/` directory doesn't trigger Felix to react to the change.  You have to manually bounce the Felix process.

Felix should immediately apply config changes made to etcd.

fasaxc:
@spikecurtis It'd be nice to have but the steer I was getting a while back was that it wasn't MVP.  We could fairly easily do something blunt such as getting felix to restart if it detects a config change.  Risk is that all your felixes end up in a restart loop if something goes wrong.

Any particular use-case you had in mind?  Tuning logging on the fly is what I'd really like.

spikecurtis:
My view is that this is a manageability issue.

Administrators are going to want to fiddle with all the little Felix options and it's going to drive people nuts that you have to restart to pick up config changes.  It's an extremely common complaint to support for other products I've worked on.

I think everyone has learned to live with the idea that processes don't pick up changes to config files without a kick, but the expectation is that if you change a config database, the change happens immediately.  It surprised me that Felix didn't pick up the change.

fasaxc:
Some design thoughts on this:

* Logging and the like are easy to make dynamic.
* Some things are not straightforward:
  * the interface prefix is assumed to be static, we program it once and then never touch it again.
  * various timers that are configurable, such as the periodic resync timer, can't easily be changed because we abort the resync loop if resync is disabled.
  * etcd address awkward to change
  * felix hostname awkward to change, we discard a lot of data that we read if it doesn't match our hostname

Considering all that, I err towards restarting felix when any of the hard-to-change pieces change.  That could be as simple as committing suicide and being restarted or we could try to re-exec the process to avoid bothering the init daemon.

spikecurtis:
From a requirements standpoint, I'd say

 - it MUST happen automatically with no extra kick from the user
 - it MUST be clear from the logs what is happening (e.g. if restarting because of a config change, rather than for some other reason) 
 - it SHOULD minimize disruption as much as possible

The actual implementations of different kinds of changes are going to have trade offs in terms of code complexity, amount of disruption, etc.

spikecurtis:
@fasaxc there's a specific use case for this that we're now running up against: IP-in-IP support.

We'd like to allow users to turn this on with a `calicoctl` command.  But that doesn't work because it currently requires restarting every Felix in the cluster manually.

fasaxc:
[...]

PR #749, opened 2015-07-31
Check config on resync and restart if changed

fasaxc:
The simple brute-force fix for #725:

* [x] Resync if tour watch detects a config change.
* [x] Each time we resync with etcd, check if the config has changed.
* [x] If config has changed, die (and allow init daemon to restart us).
* [x] UTs
* [x] CHANGELOG

### DIFF
--- a/calico/felix/fetcd.py
+++ b/calico/felix/fetcd.py
@@ -194,6 +194,11 @@ class _EtcdWatcher(gevent.Greenlet):
         self.config = config
         self.hosts_ipset = hosts_ipset
 
+        # Keep track of the config loaded from etcd so we can spot if it
+        # changes.
+        self.last_global_config = None
+        self.last_host_config = None
+
         # Events triggered by the EtcdAPI Actor to tell us to load the config
         # and start polling.  These are one-way flags.
         self.load_config = Event()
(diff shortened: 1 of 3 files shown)


calico/felix/fetcd.py
```diff
@@ -352,7 +371,22 @@ def _load_config(self):
...
             else:
-                self.config.report_etcd_config(host_dict, global_dict)
+                if self.configured.is_set():
+                    # We've already been configured.  We don't yet support
+                    # dynamic config update so instead we check if the config
+                    # has changed and die if it has.
+                    _log.info("Checking configuration for changes...")
+                    if (host_dict != self.last_host_config or
+                            global_dict != self.last_global_config):
+                        _log.critical("Felix configuration has changed, "
```
matthewdupre: I agree that this should be high priority, but not that high - I'd probably say error or warning.

calico/felix/fetcd.py
```diff
@@ -659,6 +693,16 @@ def on_workload_delete(self, response, hostname, orchestrator,
             del self.endpoint_ids_per_host[hostname]
 
 
+def die_and_restart():
+    # Sleep so that we can't die more than 5 times in 10s even if someone is
+    # churning the config.  This prevents our upstart/systemd jobs from giving
+    # up on us.
+    gevent.sleep(2)
```
matthewdupre: Urgh, horrible.  Probably the best (simple) option though.

fasaxc:
@matthewdupre UTs (including covering pre-existing _load_config method) and markups made.

PR #776, opened 2015-08-21
Update bundle to use Project Calico versions of charms, and Kubernetes version of etcd charm

matthewdupre:


matthewdupre:
Cory says it's perfect.

PR #778, opened 2015-08-24
Replace fork() with posix_spawnp()

fasaxc:
Improves performance when felix has a medium/large VM size.

### DIFF
--- a/calico/felix/futils.py
+++ b/calico/felix/futils.py
@@ -27,10 +27,14 @@ import hashlib
 import logging
 import os
 from types import StringTypes
+import types
+import gc
 import gevent.lock
 from gevent import subprocess
+from gevent.subprocess import Popen
 import tempfile
 import time
+from posix_spawn import posix_spawnp, FileActions
 
 try:
     import resource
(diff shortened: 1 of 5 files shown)


calico/felix/futils.py
```diff
@@ -95,6 +99,122 @@ def call_silent(args):
...
+            if gc_was_enabled:
+                gc.enable()
+            raise
+        finally:
+            for fd in fds_to_close_in_parent:
+                os.close(fd)
+
+        # Parent
+        self._watcher = self._loop.child(self.pid)
+        self._watcher.start(self._on_child, self._watcher)
```
Lukasa: I assume you copied this implementation from gevent, but I'd like to point out that it appears to be possible for the code to explode here and never re-enable the garbage collector. Which is sad.

fasaxc:
@Lukasa I'd like to get this into master soon so that we can let it bake with the FVs for the rest of the week.

I haven't tested the RHEL install instructions yet.

Lukasa:
LGTM: I'm happy to merge for now and see how the tests feel about it.

PR #781, opened 2015-08-25
Rev Calico version to sort out pre-release build dependencies

fasaxc:
This should make the calico-jenkins'hosted package strictly preferred to the other.


matthewdupre:
OK, I'm happy to do this on a basis of always bumping to an rc after each release.  In general we should update the setup.py and rpm versions as well, but this will do for now.

PR #788, opened 2015-09-01
Rework packet marking logic to use the MARK to indicate a match

fasaxc:
* Enables the kernel's RPF checks explicitly for our interfaces.
* Simplifies the iptables profile dispatch rules and uses MARK==1 to indicate acceptance, which avoids the class of bug that was present before.
* Adds UT for _get_endpoint_rules().

### DIFF
--- a/calico/felix/devices.py
+++ b/calico/felix/devices.py
@@ -33,6 +33,23 @@ from calico.felix.futils import FailedSystemCall
 _log = logging.getLogger(__name__)
 
 
+def check_kernel_config():
+    """
+    Checks the kernel configuration for problems that would break our
+    security guarantees, for example.
+
+    :raises BadKernelConfig if a problem is detected.
+    """
+    ps_name = "/proc/sys/net/ipv4/conf/all/rp_filter"
+    rp_filter = int(_read_proc_sys(ps_name).strip())
+    if rp_filter > 1:
+        _log.critical("Kernel's RPF check is set to 'loose'.  This would "
+                      "allow endpoints to spoof their IP address.  Calico "
+                      "requires net.ipv4.conf.all.rp_filter to be set to "
+                      "0 or 1.")
+        raise BadKernelConfig("net.ipv4.conf.all.rp_filter set to 'loose'")
+
+
 def interface_exists(interface):
     """
     Checks if an interface exists.
@@ -107,18 +124,10 @@ def configure_interface_ipv4(if_name):
     """
     # Enable the kernel's RPF check, which ensures that a VM cannot spoof
     # its IP address.
-    with open('/proc/sys/net/ipv4/conf/%s/rp_filter' % if_name, 'wb') as f:
-        f.write('1')
-
-    with open('/proc/sys/net/ipv4/conf/%s/route_localnet' % if_name,
-              'wb') as f:
-        f.write('1')
-
-    with open("/proc/sys/net/ipv4/conf/%s/proxy_arp" % if_name, 'wb') as f:
-        f.write('1')
-
-    with open("/proc/sys/net/ipv4/neigh/%s/proxy_delay" % if_name, 'wb') as f:
-        f.write('0')
+    _write_proc_sys('/proc/sys/net/ipv4/conf/%s/rp_filter' % if_name, 1)
+    _write_proc_sys('/proc/sys/net/ipv4/conf/%s/route_localnet' % if_name, 1)
+    _write_proc_sys("/proc/sys/net/ipv4/conf/%s/proxy_arp" % if_name, 1)
+    _write_proc_sys("/proc/sys/net/ipv4/neigh/%s/proxy_delay" % if_name, 0)
 
 
 def configure_interface_ipv6(if_name, proxy_target):
@@ -126,7 +135,6 @@ def configure_interface_ipv6(if_name, proxy_target):
     Configure an interface to support IPv6 traffic from an endpoint.
       - Enable proxy NDP on the interface.
       - Program the given proxy target (gateway the endpoint will use).
-      - Enable the kernel's RPF check.
 
     :param if_name: The name of the interface to configure.
     :param proxy_target: IPv6 address which is proxied on this interface for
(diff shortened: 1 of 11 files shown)


neiljerram:
Can you add a reference to the issue?

calico/felix/endpoint.py
```diff
@@ -437,42 +437,43 @@ def _get_endpoint_rules(endpoint_id, suffix, ip_version, local_ips, mac,
...
 
     # Jump to each profile in turn.  The profile will do one of the
     # following:
     # * DROP the packet; in which case we won't see it again.
-    # * RETURN without MARKing the packet; indicates that the packet was
-    #   accepted.
-    # * RETURN with a mark on the packet, indicates reaching the end
-    #   of the chain.
+    # * RETURN the packet with MARK==1, indicating it accepted the packet.
+    # * RETURN the packet with MARK==0, indicating it did not match the packet.
```
neiljerram: I can't easily review this change, because - it appears - we don't have an overall comment that presents the iptables design as a whole.

For example, the change right begs the question "why not just ACCEPT, instead of RETURN with MARK==1".

Could you add, somewhere, an overall comment that presents the whole design?

fasaxc:
@neiljerram I've added a large explanatory comment to frules.py.  Hope that helps.

fasaxc:
@neiljerram @nbartos has convinced me that we should revise this patch and switch to mandating rp_filter=1; it's simply far more robust than the iptables solution can be made.

fasaxc:
@neiljerram I've made the changes to use the Kernel's RPF checks for IPv4 (using the sysctl) and IPv6 (using the iptables rpfilter match in the raw/PREROUTING chain).

I also added a sanity check to verify that the global rp_filter setting is sane.  Better to die loudly than continue insecurely.

calico/felix/devices.py
```diff
@@ -33,6 +33,23 @@
...
+
+    :raises BadKernelConfig if a problem is detected.
+    """
+    ps_name = "/proc/sys/net/ipv4/conf/all/rp_filter"
+    rp_filter = int(_read_proc_sys(ps_name).strip())
+    if rp_filter > 1:
+        _log.critical("Kernel's RPF check is set to 'loose'.  This would "
+                      "allow endpoints to spoof their IP address.  Calico "
+                      "requires net.ipv4.conf.all.rp_filter to be set to "
+                      "0 or 1.")
```
neiljerram: Huh?  Why is 0 an acceptable value?  I thought we just concluded that we always want RPF check to be on, i.e. rp_filter = 1.

calico/felix/devices.py
```diff
@@ -135,6 +150,16 @@ def configure_interface_ipv6(if_name, proxy_target):
                            "proxy", str(proxy_target), "dev", if_name])
 
 
+def _read_proc_sys(name):
+    with open(name, "rb") as f:
+        return f.read()
```
neiljerram: I think it would be nice to do .strip() here, so that callers don't have to.  Are there any downsides to that?

calico/felix/endpoint.py
```diff
@@ -431,75 +426,63 @@ def __str__(self):
...
-        profile_in_chain = profile_to_chain_name("inbound", profile_id)
-        to_deps.add(profile_in_chain)
-        to_chain.append("--append %s --jump MARK --set-mark 0" %
-                        to_chain_name)
-        to_chain.append("--append %s --jump %s" %
-                        (to_chain_name, profile_in_chain))
-        to_chain.append('--append %s --match mark ! --mark 1/1 '
-                        '--match comment --comment "No mark means profile '
-                        'accepted packet" --jump RETURN' % to_chain_name)
+        profile_in_chain = profile_to_chain_name(direction, profile_id)
```
neiljerram: Maybe rename profile_in_chain to profile_chain, given that it's not now necessarily inbound?

calico/felix/frules.py
```diff
@@ -16,7 +16,93 @@
...
+by dispatch.py.
+
+The dispatch chains direct packets to per-endpoint ("felix-to/from")
+chains, which are responsible for policing IP addresses.  Those chains are
+managed by endpoint.py.  Since the actual policy rules can be shared by
+multiple endpoints, we put each set of policy rules in its own chain and
+the per-endpoint chains send packets to the relevant policy
+(felix-p-xxx-i/o) chains in turn.  Policy profile chains are managed by
+profilerules.py.
+
```
neiljerram: This is all great, but I still wasn't really clearly seeing the big picture, and in particular it wasn't obvious to me why a profile chain has 3 possible return values, rather than just 2 (deny or accept)?

I've worked it all out now, and the main problem was that I didn't have enough of the etcd data model in my head, and needed to revise that.  But, perhaps we could add even more hear to make that immediately clearer to other readers as well.

Here are my notes from working this out:

    Endpoint outbound security
     Profile 1
     Profile 2
     ...
     DROP if not matched by any profiles
    
    
    
    Endpoint inbound security
     Profile 1
     Profile 2
     ...
     DROP if not matched by any profiles
    
    
    Profile:
     Tags
     Rules
      Allow/deny stuff to/from IP address/prefix/tag
    
    
    Matched, Deny -> DROP
    Matched, Allow -> RETURN with mark 1
    No match -> RETURN with no mark (or mark 0)
    
    
    "Default behaviour if no rules are matched or if there are no rules in
    the list or it is missing is to deny."

I think the key points are just that

- an individual profile doesn't have an overall 'deny' or 'accept' default behaviour; it only expresses an opinion on the particular IP addresses, prefixes or tags that it matches

- if a given packet isn't matched by any profiles, there is a global overall default of 'deny'.

I wonder if it would be worth adding an expression of these points, and/or some of my notes, here?

calico/felix/endpoint.py
fasaxc: Yep, missed that one

neiljerram:
Nice change.  I've left some detailed line notes, but none of them fundamentally questions the change.

In terms of commit structure, though, I wonder about organizing this a bit better.  I think there are the following separable changes.

- Use rp_filter to guard against IP address spoofing, instead of our similar but less robust iptables rules.

- Deduplicate the code for building from- and to-endpoint chains.

- Invert the use of the mark.

- Add big explanation of our iptables design.

- Add (excellent) new tests.

What do you think?

calico/felix/devices.py
fasaxc: The kernel uses the maximum value between the "all" and "<interface>" values but the ordering of the values is

* 0 == no checking
* 1 == strict
* 2 == loose

We want strict checking so we need the all value to be set to 0 or 1, then we can set 1 on the interface and get the right result.  If it's set to 2 then that overrides our per-interface setting.

fasaxc: I'll add a comment.

calico/felix/devices.py
fasaxc: No, sounds good.

fasaxc:
I'm not sure it can be unpicked without writing intermediate code intentionally to be thrown away.  I'll see if I can break it up into 2-3 chunks.

fasaxc:
Replaces by #794.

Issue #786, opened 2015-08-27
ipset code could be smarter


plwhite:
Some rough testing shows that : 

- flushing and creating either a 1k or a 10k entry ipset takes around 20-30ms CPU time (double that real time)
- adding up to 100 new entries to a 10k entry ipset takes around 1-2ms (about 3ms real time)

That implies that the code in ipset updater that flushes and recreates the ipset to get it to the correct state should only apply deltas, and that the change is both worth doing and very simple.

PR #791, opened 2015-09-02
Incrementally update ipsets rather than wipe and reset.

plwhite:
Fixes issue #786.

### DIFF
--- a/calico/felix/ipsets.py
+++ b/calico/felix/ipsets.py
@@ -500,7 +500,7 @@ class IpsetActor(Actor):
         self._ipset = ipset
         # Members - which entries should be in the ipset.
         self.members = set()
-        # Members which really are in the ipset.
+        # Members which really are in the ipset. None means "don't know yet"
         self.programmed_members = None
         self._force_reprogram = False
         self.stopped = False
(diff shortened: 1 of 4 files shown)


calico/felix/ipsets.py
```diff
@@ -544,12 +544,16 @@ def _finish_msg_batch(self, batch, results):
...
-        _log.info("Rewriting %s with %d members.", self, len(self.members))
         _log.debug("Setting ipset %s to %s", self, self.members)
-        # Defer to our helper.
-        self._ipset.replace_members(self.members)
-        # We have got the set into the correct state.
-        self.programmed_members = self.members.copy()
+        # Defer to our helper to actually make the changes.
+        if self._force_reprogram or self.programmed_members is None:
+            self._ipset.replace_members(self.members)
+            self.programmed_members = self.members.copy()
```
fasaxc: Since we're on an occupancy push, why don't we rework the code a little to avoid these copies?  We could adjust the contract of replace_members to require that the input is a copy (it already is, in fact), then directly store it as self.members rather than doing a clear/update; then directly copy it to self.programmed_members.

I wish Python had an immutable set type, which would make it much harder to shoot ourselves in the foot when passing sets around between actors.

fasaxc:
Code looks like it should work but I'm wondering if we should push the (programmed_)members fields down into Ipset and have IpsetActor be a simpler lightweight wrapper.  We won't get the benefit of this new function in the MasqueradeManager, for example.

calico/felix/ipsets.py
alexwlchan: > I wish Python had an immutable set type

What about frozenset? Looks like it implements a very similar API to set(), except it's immutable and hashable. No idea how using frozenset affects occupancy.

(Standard library, apparently available at least as far back as 2.6)

plwhite:
I agree with all of the above comments about how we could make the code a bit better, but I don't think I'll have time to make them this week (since I'm away a chunk of it). Can we merge this now, and raise a separate issue for tidying this up?

fasaxc:
@neiljerram's doing a release (which will go out tomorrow).  I think we'd better check this in after that.

neiljerram:
Yes, I'd prefer that.  Thanks @fasaxc !

fasaxc:
@plwhite Please can you update the changelog and then merge at your convenience?

PR #792, opened 2015-09-03
Improve netlink-based generation of interface update notifications

neiljerram:


calico/felix/devices.py
```diff
@@ -315,46 +322,68 @@ def watch_interfaces(self):
...
+                    ifname in if_last_flags):
+                    # An interface that we've previously signalled is
+                    # being deleted, so remove our record of it.
+                    del if_last_flags[ifname]
+
+                if (ifname and
+                    msg_type == RTM_NEWLINK and
+                    operstate == IF_OPER_UP and
+                    (ifname not in if_last_flags or
+                     if_last_flags[ifname] != flags)):
```
fasaxc: * Why do we store the flags when we only seem to care about the operstate?
* Assuming that storing the flags is the right thing to do,  shouldn't we include the operstate in the if_last_flags cache.  Otherwise, we won't notify if the flags remain constant while the operstate goes from down to up.

fasaxc: Hmm, I suppose we don't store it until we see it oper up, so maybe the latter is OK.  Can you confirm that going oper down isn't enough to remove the routes and require reprogramming?

fasaxc:
Not sure I understand the logic behind storing the flags but testing the operstate, please can you explain?

neiljerram:
Idea of storing the flags is that I think it makes sense to re-emit an interface_update if the flags change, even if oper state doesn't change.  It's conceivable (just about) that there could be an interface state change that causes routes or other settings to be lost - so why not get Felix to check and reprogram in that case, so long as this isn't happening so frequently as to impact performance.   In practice we don't see this in the traces I've looked at - i.e. once oper state is UP, the flags do not change - so I would say that this will not be a performance hit.

neiljerram:
Following our discussion earlier, I added debug logging to look at the RTA data for all other attributes provided, and none of that is changing - once oper state is UP - from one notification to the next, except for the IFLA_MAP attribute.  I haven't been able to find clear doc of what IFLA_MAP means, but it looks like it's to do with traffic control, so unlikely to be anything relevant to our processing here.

Therefore I've just added the delta that we mooted earlier, to check for oper state going DOWN (or anything other than UP) and to delete the if_last_flags entry in that case, same as we do for DELLINK.

With that delta, I think we can call this done...

neiljerram:
Needed rebasing...

neiljerram:
@plwhite @fasaxc - I believe this is ready for merging now.  I've used my Devstack/GCE setup to verify that it passes the FV's test_mainline_connectivity.py.  Should we do any more OpenStack testing before merging, and any container system testing before or after merging?  If so, can you recommend how to do that?

neiljerram:
@plwhite @fasaxc - I believe this is ready for merging now.  I've used my Devstack/GCE setup to verify that it passes the FV's test_mainline_connectivity.py.  Should we do any more OpenStack testing before merging, and any container system testing before or after merging?  If so, can you recommend how to do that?

calico/felix/devices.py
```diff
@@ -299,6 +299,8 @@ def remove_conntrack_flows(ip_addresses, ip_version):
 RTM_DELLINK = 17
 
 IFLA_IFNAME = 3
+IFLA_OPERSTATE = 16
+IF_OPER_UP = 6
```
fasaxc: Why is up == 6?  Just wanted to check that it's not a bit field that we should be checking against a mask.

fasaxc:
One question, otherwise the code looks good.  Are there any UTs to beef up here, now that the logic is more complex?

calico/felix/devices.py
neiljerram: From /usr/include/linux/if.h:

    /* RFC 2863 operational status */
    enum {
    	IF_OPER_UNKNOWN,
    	IF_OPER_NOTPRESENT,
    	IF_OPER_DOWN,
    	IF_OPER_LOWERLAYERDOWN,
    	IF_OPER_TESTING,
    	IF_OPER_DORMANT,
    	IF_OPER_UP,
    };

neiljerram:
No, no existing UTs for InterfaceWatcher.  So, can we merge this now?

fasaxc:
@neiljerram I've merged this but it'd be good to get some UTs for this area if you have time.

PR #795, opened 2015-09-04
Add monotonic_time() function

alexwlchan:
This is the monotonic-time branch rebased against master and with tweaked tox settings.

It's required for #732. I'm trying to get a clean UT pass, so that we might be able to merge it.

UT failures from the original monotonic-time branch [are in a Gist](https://gist.github.com/alexwlchan/c99921d9b8af21d416a0). Seems to be passing now, but I want to get Jenkins to double check before I take this further.

### DIFF
--- a/calico/common.py
+++ b/calico/common.py
@@ -188,6 +188,7 @@ class GreenletFilter(logging.Filter):
         record.tid = greenlet_id()
         return True
 
+
 def default_logging():
     """
     Sets up the Calico default logging, with default severities.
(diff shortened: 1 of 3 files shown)


fasaxc:
@alexwlchan The UT was the issue.  I believe that we mock out time.sleep() somewhere and do not restore it, which was making the test fail.  Not sure if that's already fixed but the UT was failing 50% of the time when run with the rest of the UTs.

alexwlchan:
@fasaxc Jenkins appears to be happy, and I've run this branch at least a dozen times without seeing a UT failure. Probably ready to merge, and then it's available for the status reporting work.

alexwlchan:
Closing as a dupe of #813

PR #802, opened 2015-09-08
Update Redhat install to reference python-posix-spawn 0.2.post7

fasaxc:


fasaxc:
Tested that the install works on one of our RH test machines.

matthewdupre:
Need to move this from the etcd section to the compute section (was wrong before).

Otherwise LGTM.

matthewdupre:
Change looks OK

PR #810, opened 2015-09-11
Juju: Give Cinder, Keystone and Nova cloud controller their own nodes

matthewdupre:
I was seeing haproxy problems when deploying with the old bundle - haproxy was only set up for keystone, not nova (which was on the same node).  As a result attempts to connect to nova failed because haproxy wasn't listening on the right ports.

I found a similar problem with Cinder.

It might be possible to move Keystone back onto the rabbitmq-server node, or perhaps colocate some of these services with something less picky (say etcd or BIRD, which shouldn't rely on HAProxy), but for now this bundle works for me.


neiljerram:
LGTM

PR #818, opened 2015-09-18
Update Juju bundles

matthewdupre:
Add a kilo bundle (WIP), and fix up the Icehouse bundle to use the latest charms.

Still need to test Kilo, add Juno and rework the doc a bit.


matthewdupre:
Tested Kilo successfully.

All that's left is to add a Juno bundle, and update the doc to mention that the various bundles.

matthewdupre:
OK, I've updated the bundles; so let's get this merged.

fasaxc:
Looks fine, in as much as I can review it.

PR #819, opened 2015-09-18
Refactor LocalEndpoint to be a batched actor.

fasaxc:
Should improve performance under load/flapping data and make it easier to add status reporting.

### DIFF
--- a/calico/felix/endpoint.py
+++ b/calico/felix/endpoint.py
@@ -357,6 +357,7 @@ class LocalEndpoint(RefCountedActor):
                     (self.endpoint[self.nets_key] !=
                      pending_endpoint[self.nets_key])):
                 # IP addresses have changed, need to update the routing table.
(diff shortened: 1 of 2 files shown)


fasaxc:
@plwhite please can you review this PR for me?  It refactors the LocalEndpoint actor to use batches.  The motivation is to keep better track of whether the routes and iptables are in sync.  That will help with reporting felix per-port status because the end of _finish_msg_batch will always know what the state is.

It also coalesces duplicate/too frequent updates, as usual for a batched actor.

plwhite:
I find these refactoring PRs really hard to review, but I've convinced myself that it looks right. So long as it passes the FVs (the FV has not completed yet), I'm fine with it. Go ahead and merge if the FVs complete clean.

plwhite:
OK, I noticed it had completed, so did the merge.

PR #821, opened 2015-09-22
Rev python-etcd to 0.4.1+calico.1.

fasaxc:
Main change in the API is that exceptions are now wrapped, which allows us
to combine a lot of except clauses.


matthewdupre:
LGTM

Issue #811, opened 2015-09-14
Reduce number of unnecessary copies in ipset code


plwhite:
We on several occasions do unnecessary copies. For example, in _sync_to_ipset, programmed_members is discarded and replaced with a copy of members. This leads to churning of the heap, which has to have a CPU and possible occupancy cost.

We should see if we can cut a few of those copies.

plwhite:
I might not do this, but I have a couple of hours and an idea for how to do it simply, so will own it until I either do it or decide that it's better to leave to somebody else.

PR #826, opened 2015-09-24
Reduce amount of unneeded copying in ipset code.

plwhite:
Fixes issue #811.

### DIFF
--- a/calico/felix/ipsets.py
+++ b/calico/felix/ipsets.py
@@ -94,8 +94,8 @@ class IpsetManager(ReferenceManager):
         """
         assert self._is_starting_or_live(tag_id)
         active_ipset = self.objects_by_id[tag_id]
-        members = self.ip_owners_by_tag.get(tag_id, {}).keys()
-        active_ipset.replace_members(set(members),
+        members = frozenset(self.ip_owners_by_tag.get(tag_id, {}).iterkeys())
+        active_ipset.replace_members(members,
                                      force_reprogram=self._force_reprogram,
                                      async=True)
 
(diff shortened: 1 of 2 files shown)


plwhite:
No changelog updates, because there is already a changelog update saying "improve ipset performance".

calico/felix/ipsets.py
```diff
@@ -528,32 +530,35 @@ def replace_members(self, members, force_reprogram=False):
...
         # Defer to our helper to actually make the changes.
-        if self._force_reprogram or self.programmed_members is None:
+        if self._force_reprogram:
+            _log.debug("Replacing content of ipset %s with %s", self,
+                       self.members)
             self._ipset.replace_members(self.members)
-            self.programmed_members = self.members.copy()
-        elif self.members != self.programmed_members:
+        elif self.programmed_members != self.members:
+            assert(self.programmed_members is not None)
```
fasaxc: Nitpick, I think assert shouldn't have parens because they can be misleading.  In particular, `assert(False, "message")` looks right but it's actually equivalent to `x = (False, "message); assert x`, which is wrong.

fasaxc:
One minor nitpick.  Otherwise LGTM.  Please merge when ready.

calico/felix/test/test_ipsets.py
```diff
@@ -491,46 +491,66 @@ def setUp(self):
         self.ipset = Mock(spec=Ipset)
         self.actor = IpsetActor(self.ipset)
 
+
```
alexwlchan: Super nitpicky: there should only be one newline between method definitions.

plwhite:
I have picked both those nits, and shall now merge.

PR #828, opened 2015-09-25
Install yum-plugin-priorities on CentOS

matthewdupre:


docs/source/redhat-opens-install.rst
```diff
@@ -59,6 +59,10 @@ Instructions for installing OpenStack on RHEL can be found here --
 Configure YUM repositories
 ~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+If you're using CentOS, check the yum priorities plugin is installed:
+
+    yum install yum-plugin-priorities
```
alexwlchan: This doesn't actually render as a monospaced block; see http://calico-jenkins/pr-docs/centos-install/redhat-opens-install.html. You need to double up the colon on the previous line.

alexwlchan:
Does it matter that this is the only mention of CentOS in our docs? Should we add something to say this is RHEL or CentOS, or can we let the mention of Red Hat stand in for both of them?

matthewdupre:
I think it's OK - anyone using CentOS should be aware of what it is.

I've avoided using just 'EL7', because we don't support Scientific Linux (probably).

alexwlchan:
I can’t imagine the package build/FV tests were broken by this docs change, so I’m going to go ahead and merge.

PR #838, opened 2015-10-06
Optimization: use logging.disable() to set logging level.

fasaxc:
This reduces the overhead of each logging call.

### DIFF
--- a/calico/common.py
+++ b/calico/common.py
@@ -282,6 +282,15 @@ def complete_logging(logfile=None,
             file_handler.setFormatter(formatter)
             root_logger.addHandler(file_handler)
 
+    # Optimization: disable all logging below the minimum level that we care
+    # about.  The global "disable" setting is the first thing that gets checked
+    # in the logging framework so it's the fastest way to disable logging.
+    levels = [file_level, syslog_level, stream_level]
+    # Map None to something greater than the highest logging level.
+    levels = [l if l is not None else logging.CRITICAL + 1 for l in levels]
+    min_log_level = min(levels)
+    logging.disable(min_log_level - 1)
+
     _log.info("Logging initialized")
 
 


fasaxc:
retest this please

calico/common.py
```diff
@@ -282,6 +282,12 @@ def complete_logging(logfile=None,
             file_handler.setFormatter(formatter)
             root_logger.addHandler(file_handler)
 
+    # Optimization: disable all logging below the minimum level that we care
+    # about.  The global "disable" setting is the first thing that gets checked
+    # in the logging framework so it's the fastest way to disable logging.
+    min_log_level = min(file_level, syslog_level, stream_level)
+    logging.disable(min_log_level - 1)
+
```
plwhite: What happens if one of the levels is "None"?

It turns out that it cannot be (the calling code never sets any level to None). Worth removing the test against file_level being None a few lines above then. Arguably we could assert, but I don't think it is worth it.

plwhite:
Trivial markup, then ready for merge. Go ahead and apply markup and merge.

calico/common.py
fasaxc: Actually, I think that' s a good spot!  "none" is a valid value to use in the config
```
LOGLEVELS = {"none":      None,
             "debug":     logging.DEBUG,
             "info":      logging.INFO,
             "warn":      logging.WARNING,
             "warning":   logging.WARNING,
             "err":       logging.ERROR,
             "error":     logging.ERROR,
             "crit":      logging.CRITICAL,
             "critical":  logging.CRITICAL}
```

plwhite:
It's a bit odd that you can have log levels of None, and the code does not explain how that could happen. If you add a comment to the method explaining that it is valid for the value to be None, then fine to merge.

plwhite:
It's a bit odd that you can have log levels of None, and the code does not explain how that could happen. If you add a comment to the method explaining that it is valid for the value to be None, then fine to merge the updated version.

PR #840, opened 2015-10-08
Quick improvement to UT coverage

alexwlchan:
**Extend the UT coverage of calico.felix.devices**

This adds the (previously uncovered) case to check that interface_up returns False if it hits an IOError trying to read the `/sys/class/net/<if_name>/flags` file, and tidies up the name/docstring for the two related tests.


fasaxc:
LGTM

PR #845, opened 2015-10-15
Truncate long stdin/stout/stderr output from failing commands.

### LINKED ISSUE #769: FailedSystemCall logs entire input even if it's huge
When #768 was hit, I found that the log output included the whole input to the ipset restore command, which was several MB.  It wrapped a couple of logs just to hold that output.  I think we have to truncate that output.  Ideally, we'd get the first and last few lines with ellipses in-between.

fasaxc:
Fixes #769.

### DIFF
--- a/calico/felix/futils.py
+++ b/calico/felix/futils.py
@@ -67,6 +67,8 @@ SHORTENED_PREFIX = "_"
 MAX_CONCURRENT_CALLS = 32
 _call_semaphore = gevent.lock.Semaphore(MAX_CONCURRENT_CALLS)
 
+DEFAULT_TRUNC_LENGTH = 1000
+
 
 class FailedSystemCall(Exception):
     def __init__(self, message, args, retcode, stdout, stderr, input=None):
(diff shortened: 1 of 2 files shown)


calico/felix/test/test_futils.py
```diff
@@ -125,6 +125,22 @@ def test_uniquely_shorten(self):
                                           "%r but got %r" %
                                           (inp, length, exp, output))
 
+    def test_safe_truncate(self):
+        self.assert_safe_truncate("foobarbazb", 10, "foobarbazb")
+        # Yes, this get's longer, which is silly.  However, there's no point
```
alexwlchan: /s/get's/gets

calico/felix/futils.py
```diff
@@ -67,6 +67,7 @@
 MAX_CONCURRENT_CALLS = 32
 _call_semaphore = gevent.lock.Semaphore(MAX_CONCURRENT_CALLS)
 
+DEFAULT_TRUNC_LENGTH = 1000
```
alexwlchan: Nitpick: two newlines before a class.

calico/felix/futils.py
```diff
@@ -79,12 +80,29 @@ def __init__(self, message, args, retcode, stdout, stderr, input=None):
         self.input = input
 
     def __str__(self):
+        input = self.input
```
alexwlchan: It can't cause problems here (I think), but I always get a little twitchy whenever I see variables with the same name as the Python builtins.

I'm also confused &ndash; where are we actually using this variable? What's the difference between this and `safe_truncate()`?

calico/felix/test/test_futils.py
fasaxc: Put that in just for you @alexwlchan

calico/felix/futils.py
fasaxc: Oh, that was the prototype for safe_truncate.  Just forgot to snip it out.

fasaxc:
@alexwlchan I think you may as well take the review proper!

alexwlchan:
Those two markups made, looks fine to me. I assume the FV failure is an FV test problem, not something buggy in this change.

Happy for you to merge as soon as you have a clean test run.

fasaxc:
retest this please

alexwlchan:
retest this please

alexwlchan:
We're temporarily code frozen, but I'm happy to merge this once we cut the release.

PR #850, opened 2015-10-21
Report interface oper down for interfaces that we've never seen up.

### LINKED ISSUE #851: Felix fails to report interface going down if it's already up when Felix starts
* Restart Felix with an existing interface that is up.
* Felix reads the interface state when it restarts, reports interface as up, so far so good.
* Flap interface down, Felix doesn't notice.

Problem is that devices.py only reports interfaces going down if it saw them go up earlier.

fasaxc:
Closes a race where the LocalEndpoint reads that the interface is up at start of day but then it subsequently goes down.

Fixes #851


matthewdupre:
LGTM

PR #852, opened 2015-10-21
[Work in progress] Implement concurrent resync

fasaxc:
Rather than stopping the world to resync with etcd, process the resync in parallel with watching etcd.  Improves scalability under high churn.

* [x] Introduce driver threads
* [x] Merge updates into consistent stream
* [x] Pass logging config to driver
* [x] Driver initialises logging
* [x] Pass status messages from driver to Felix
* [x] Handle config key updates in Felix
* [x] Handle Ready key update in driver
* [x] Special-case top-level directory deletions?  If a delete nukes the ready flag, it's better to resync than to send in 1000s of deletes and spot the ready flag in the middle.
* [x] Rework snapshot processing to use in-sync message instead
  * [x] do cleanup after in-sync message
  * [x] ipset manager defers update to dirty tags until we're in-sync
  * [x] dispatch chains defers updates until in-sync
  * [x] profile rules manager defers programming (at least deletions) until we're in sync
  * [x] hosts ipset
  * [x] status reporting cleanup
* [x] Explicit logfile name in config
* [x] Handle trie corner cases: encode invalid characters?
* [x] Handle etcd error cases in driver:
  * [x] Snapshot error response
  * [x] Event error response
* [x] Tidy up 
  * [x] Driver code: comments, logging
  * [x] ijson loop
  * [x] fetcd code
  * [x] Check syslog process name
* [x] Make simple etcd response handling common (reuse python-etcd objects?)
* [x] logrotate config for new driver log file
* [x] Fix existing UTs
* [x] UTs for driver
* [x] UTs for felix
* [x] Package ijson for Ubuntu
* [x] Package datrie for Ubuntu
* [x] Package deps for Redhat?
* [x] Remove private version number
* [x] Docs
  * ~~[ ] Update install docs.~~ (Covered by #880.)
  * [x] Update release docs
  * [x] Update config docs

Extras?:
Let's do these as follow-ups:
* ~~Optimize incremental parsing perf?~~
* ~~Use faster json library for parsing events?~~

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,5 +1,6 @@
 [run]
 include =
+    calico/etcddriver/*
     calico/felix/*
     calico/openstack/*
     calico/*.py
--- a/.coveragerc
+++ b/.coveragerc
@@ -4,7 +4,9 @@ include =
     calico/openstack/*
     calico/*.py
 omit =
+    calico/test/*
     calico/felix/test/*
     calico/openstack/test/*
+    calico/etcddriver/test/*
 branch = True
 concurrency = eventlet
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,8 @@
 # Changelog
 
+- Felix now parses the etcd snapshot in parallel with the event stream;
+  this dramatically increases scale when under load.
+
 ## 1.2.0
 
 - Add liveness reporting to Felix.  Felix now reports its liveness into
--- a/calico/common.py
+++ b/calico/common.py
@@ -161,13 +161,13 @@ def mkdir_p(path):
     except TypeError:
         try:
             os.makedirs(path)
-        except OSError as exc: # Python >2.5
+        except OSError as exc:  # Python >2.5
             if exc.errno == errno.EEXIST and os.path.isdir(path):
                 pass
             else: raise
 
 
-def default_logging(gevent_in_use=True):
+def default_logging(gevent_in_use=True, syslog_executable_name=None):
     """
     Sets up the Calico default logging, with default severities.
 
@@ -188,7 +188,7 @@ def default_logging(gevent_in_use=True):
     root_logger = logging.getLogger()
     root_logger.setLevel(logging.DEBUG)
 
-    executable_name = os.path.basename(sys.argv[0])
+    executable_name = syslog_executable_name or os.path.basename(sys.argv[0])
     syslog_format = SYSLOG_FORMAT_STRING.format(excname=executable_name)
     syslog_formatter = logging.Formatter(syslog_format)
     if os.path.exists("/dev/log"):
@@ -388,6 +388,7 @@ def validate_endpoint(config, combined_id, endpoint):
     if issues:
         raise ValidationFailed(" ".join(issues))
 
+
 def validate_rules(profile_id, rules):
     """
     Ensures that the supplied rules are valid. Once this routine has returned
@@ -573,6 +574,7 @@ def validate_tags(profile_id, tags):
 
     if issues:
         raise ValidationFailed(" ".join(issues))
+
 
 def validate_ipam_pool(pool_id, pool, ip_version):
     """
--- a/calico/datamodel_v1.py
+++ b/calico/datamodel_v1.py
@@ -190,6 +190,12 @@ class EndpointId(object):
     def __str__(self):
         return self.__class__.__name__ + ("<%s>" % self.endpoint)
 
+    def __repr__(self):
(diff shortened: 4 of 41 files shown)


fasaxc:
skip ci

fasaxc:
retest this please

fasaxc:
@plwhite I think this is worth a first look while I'm on vacation.  Some areas that I know still need work:

* There's duplicated etcd request/response code in the driver.  Might make sense to use python-etcd for the simpler requests.
* I have to fake up a Node object to pass to the PathDispatcher, which feels a bit ugly.  Maybe that's liveable-with though.
* Feels like the socket read/write code could go into utility functions.

calico/felix/dispatch.py
```diff
@@ -47,6 +47,7 @@ def __init__(self, config, ip_version, iptables_updater):
         self.ifaces = set()
         self.programmed_leaf_chains = set()
         self._dirty = False
+        self._datamodel_in_sync = False
```
plwhite: I found the name of this variable confusing. It really means "is there a snapshot in progress", but I read it as "we have no changes to make". Worth renaming it, I think.

plwhite:
Mostly this just looks right, though had a poke round rather than a full review. Some comments.

* Can we rely in messages arriving in order with SOCK_STREAM? Probably this is OK since there is only one thing on the other end, so it must be fine, but should we use something with guaranteed ordering to make the point?

* If the driver process dies, how does Felix notice? I assume it must, but I couldn't see the code.

I need to do more, but net of this is that I'm comfortable with what I've seen.

fasaxc:
* SOCK_STREAM is guaranteed in-order.  It's basically equivalent to TCP.
* If the driver dies, Felix should get an exception on its `recv()`.

Lukasa:
@plwhite For the record, SOCK_STREAM is definitionally in-order, regardless of the underlying protocol (AF_UNIX, AF_APPLETALK, whatever), which means your desire to "use something with guaranteed ordering to make the point" is actually already done by this code! This is one of the few good decisions of the socket API: you can ask for "whatever is equivalent to TCP" on your chosen protocol and get it. Specifically, from the man page:

> SOCK_STREAM provides sequenced, reliable, two-way, connection-based byte streams.

Note also:

> SOCK_DGRAM supports datagrams (connectionless, unreliable messages of fixed maximum length).

Though I have no idea how AF_UNIX and SOCK_DGRAM combine to produce *unreliability*: feels like a fun thing to investigate some time: can you actually lose a datagram on an AF_UNIX socket?

@matthewdupre might be interested by the possibility of combining AF_UNIX and SOCK_SEQPACKET:

> SOCK_SEQPACKET provides a sequenced, reliable, two-way connection-based data transmission path for datagrams of fixed maximum length; a consumer is required to read an entire packet with each input system call.

In other words: SCTP you can actually use! Clearly @matthewdupre needs to get this into the Felix code sometime.

calico/felix/dispatch.py
fasaxc: I think your first reading was right: it means "first snapshot complete".

fasaxc:
@Lukasa I started with `SOCK_SEQPACKET` but it puts a maximum size limit on the size of a message.  Although that limit is large, I'm always wary of limits that could bite at some random time in the future so I switched to `SOCK_STREAM` instead.  (msgpack had good stream support so there wasn't any pain in our code to use a stream anyway.)

fasaxc:
retest this please

plwhite:
I concluded that it probably was OK, but I'm happy with being told that it definitely is! So if that's not the subtle, hard-to-find and yet critical bug that's hidden in the code, where is it?

fasaxc:
retest this please

fasaxc:
retest this please

fasaxc:
retest this please

matthewdupre:
retest this please

matthewdupre:
retest this please

matthewdupre:
I suggest you add package the RPM python debs up to the list too.  While we haven't done it for the others, we can at least prevent the problem from growing.

matthewdupre:
Another required pre-merge item: remove the 1.3.0~~smc version.

fasaxc:
retest this please

fasaxc:
@matthewdupre the debs were cleanly built with py2dsc so I say we defer the build until there's something for Jenkins to build.

plwhite:
This PR is merged, but we should still discuss the queue size issue. That is a small enough detail not to block anything.

PR #853, opened 2015-10-22
Add a delay to Actor scheduling when under load.

fasaxc:
This limits the maximum frequency at which we'll schedule each actor.  This:

* Encourages batching when we're under load (while processing messages immediately when we're not).
* Reduces starvation.

In addition:

* work around gevent issue where sleep() does not necessarily yield to another greenlet by calling sleep(0.000001) instead
* reduce number of loop iterations before yielding to reduce starvation.

### DIFF
--- a/calico/felix/actor.py
+++ b/calico/felix/actor.py
@@ -99,6 +99,7 @@ inconsistent state.
 """
 import collections
 import functools
+from calico.monotonic import monotonic_time
 import gevent
 import gevent.local
 import logging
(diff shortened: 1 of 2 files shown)


plwhite:
This looks right technically, but leaves me a bit nervous. It effectively adds 0.01s to each message transfer in an unstressed system, adding probably 0.1s to the time to network an endpoint. On the other hand, it encourages batching and saves processing in a heavily stressed system, which is a good thing.

I want to think a bit more about this before merging it.

fasaxc:
If the Actor hasn't been scheduled for more than 0.01s, we'll schedule it immediately without a delay.  We only start delaying once we get multiple messages within 0.01s.

fasaxc:
@plwhite worth a try on a scale run?

plwhite:
I am going to just merge this now so we can get a run or two with it in.

PR #858, opened 2015-10-26
Fix typo in calico.felix.config

alexwlchan:
Assigning to @plw because he wrote the line originally, and so is most likely to know whether this is a genuine mistake, or whether I've misunderstood something.

h/t Matt Meredith for spotting.

### DIFF
--- a/calico/felix/config.py
+++ b/calico/felix/config.py
@@ -437,11 +437,11 @@ class Config(object):
             # config to read.
             return
 
-        for name, parameter in self.parameters.iteritems():
+        for parameter in self.parameters.itervalues():
             if parameter.value is None:
                 # No value, not even a default
                 raise ConfigException("Missing undefaulted value",
-                                      self.parameters["InterfacePrefix"])
+                                      parameter)
 
     def _warn_unused_cfg(self, cfg_dict, source):
         # Warn about any unexpected items - i.e. ones we have not used.


calico/felix/config.py
```diff
@@ -441,7 +441,7 @@ def _validate_cfg(self, final=True):
             if parameter.value is None:
                 # No value, not even a default
                 raise ConfigException("Missing undefaulted value",
-                                      self.parameters["InterfacePrefix"])
+                                      name)
```
fasaxc: Doesnt' the opbject in self.parameters stringify to a nice description of the param?  I.e. maybe the line was correct?

alexwlchan: It looks to me (and admittedly I only gave this a cursory glance) that `self.parameters["InterfacePrefix"]` will be the same, no matter which parameter has been undefaulted.

Looking a bit closer, I think `parameter` is what I should have passed in here. I think I got confused because `name` isn't being used in the loop.

matthewdupre:
retest this please

calico/felix/config.py
plwhite: The second parameter to a config exception should be a parameter object, and so `parameter` is the right thing to pass. Always logging "InterfacePrefix" is clearly daft.

plwhite:
@alexwlchan I think this comes back to you to make that markup, retest and get it merged.

alexwlchan:
Markups made, tests passed, so merging.

PR #872, opened 2015-11-07
Add support for etcd with SSL

TrimBiggs:
Add etcd scheme, key, certificate, and CA certificate config params
Update EtcdClientOwner to pass in new config params when creating client
Modify subclasses of EtcdClientOwner based on changes


calico/etcdutils.py
```diff
@@ -135,7 +149,8 @@ class EtcdWatcher(EtcdClientOwner):
     def __init__(self,
                  etcd_authority,
                  key_to_poll):
-        super(EtcdWatcher, self).__init__(etcd_authority)
+        super(EtcdWatcher, self).__init__(etcd_authority, "http", None, None,
+                                          None)
```
TrimBiggs: I'm not thrilled about this workaround, TBH... Was thinking about having a "EtcdHttpClientOwner" that used the basic __init__ for EtcdClientOwner or allowing the __init__ of EtcdClientOwner to have the keys and certs as keyword args that default to None.

caseydavenport:

calico/etcdutils.py
TrimBiggs: Note, this is the Etcd class used by OpenStack

calico/etcddriver/driver.py
```diff
@@ -170,7 +169,12 @@ def _handle_init(self, msg):
         # OK to dump the msg, it's a one-off.
         _log.info("Got init message from Felix %s", msg)
         self._etcd_base_url = msg[MSG_KEY_ETCD_URL].rstrip("/")
-        self._etcd_url_parts = urlparse(self._etcd_base_url)
+        self._etcd_scheme = msg[MSG_KEY_ETCD_URL].split(":")[0]
```
robbrockbank: I don't think you need _scheme - it is available as _etcd_url_parts.scheme

robbrockbank: hmm i see you pass in the scheme to urlparse ... so perhaps you can't seems odd though.

TrimBiggs: I think you're right actually. I think scheme might only be passed in if it's not part of the URL string itself (such as "www.google.com" scheme="http" vs "http://www.google.com")

calico/etcdutils.py
```diff
@@ -96,9 +97,14 @@ class EtcdClientOwner(object):
     reconnecting, optionally copying the cluster ID.
     """
 
-    def __init__(self, etcd_authority):
+    def __init__(self, etcd_authority, etcd_scheme, etcd_key, etcd_cert,
```
robbrockbank: Surely better to add these new parameters as keyword args, then it isn't necessary to update all of the derived classes  unless you need to pass that info in.

TrimBiggs: Yup, I commented this below.

calico/etcdutils.py
```diff
@@ -135,7 +149,8 @@ class EtcdWatcher(EtcdClientOwner):
     def __init__(self,
```
robbrockbank: I assume we should add the keyword args here too and pass them through.

calico/felix/fetcd.py
```diff
@@ -677,7 +686,11 @@ class EtcdStatusReporter(EtcdClientOwner, Actor):
     """
 
     def __init__(self, config):
-        super(EtcdStatusReporter, self).__init__(config.ETCD_ADDR)
+        super(EtcdStatusReporter, self).__init__(config.ETCD_ADDR,
+                                                 config.ETCD_SCHEME,
```
robbrockbank: Don't forget to update these to kw args if you change them.,

alexwlchan:
ok to test

calico/felix/config.py
```diff
@@ -165,6 +165,14 @@ def __init__(self, config_path):
...
         self.add_parameter("FelixHostname", "Felix compute host hostname",
                            socket.gethostname(), sources=[ENV, FILE])
+        self.add_parameter("EtcdScheme", "Protocol type for http or https",
+                           "http", sources=[ENV, FILE])
+        self.add_parameter("EtcdKeyFile", "Path to etcd key file",
+                           None, sources=[ENV, FILE])
+        self.add_parameter("EtcdCertFile", "Path to etcd certificate file",
+                           None, sources=[ENV, FILE])
+        self.add_parameter("EtcdCaFile", "Path to etcd CA certificate file",
+                           None, sources=[ENV, FILE])
```
alexwlchan: All these parameters will need documenting in configuration.rst

calico/felix/config.py
```diff
@@ -373,6 +385,36 @@ def _validate_cfg(self, final=True):
...
+
+                # If Certificate Authority cert provided, check it's readable
+                if self.ETCD_CA_FILE and \
+                        not os.access(self.ETCD_CA_FILE, os.R_OK):
+                    raise ConfigException("Cannot read CA certificate. Value "
+                                          "must be readable file path.",
+                                          self.parameters["EtcdCaFile"])
+        elif self.ETCD_SCHEME != "http":
+            raise ConfigException("Invalid protocol scheme. Value must be one "
+                                  "of: \"\", \"http\", \"https\".",
```
alexwlchan: This exception message doesn’t seem right. If self.ETCD_SCHEME is set to an empty string, won’t we hit this config exception?

TrimBiggs: If it has reached this point and the environment var was set to `""`, it would interpret this value as not being set and give it the default value of "http".

We should never reach this point with the value being set to `""`, unless I am misinterpreting how `add_parameter` works with defaults for environment variables in config.py.

calico/etcddriver/test/test_driver.py
```diff
@@ -608,6 +608,9 @@ def send_init_msg(self):
             {
                 MSG_KEY_ETCD_URL: "http://localhost:4001",
                 MSG_KEY_HOSTNAME: "thehostname",
+                MSG_KEY_KEY_FILE: "",
```
fasaxc: Shouldn't we use `None` for not-present?

calico/etcdutils.py
```diff
@@ -119,9 +129,17 @@ def reconnect(self, copy_cluster_id=True):
         else:
             _log.info("(Re)connecting to etcd. No previous cluster ID.")
             old_cluster_id = None
+
+        key_pair = None
+        if self.etcd_ca and self.etcd_key:
```
fasaxc: Feels a bit odd to test etcd_ca and then use etcd_cert.  Is that correct?

calico/felix/config.py
```diff
@@ -373,6 +385,45 @@ def _validate_cfg(self, final=True):
...
+                                          "must be a readable path.",
+                                          self.parameters["EtcdKeyFile"])
+                if not (os.path.isfile(self.ETCD_CERT_FILE) and
+                        os.access(self.ETCD_CERT_FILE, os.R_OK)):
+                    raise ConfigException("Cannot read cert file. Cert file "
+                                          "must be a readable path.",
+                                          self.parameters["EtcdCertFile"])
+
+                # If Certificate Authority cert provided, check it's readable
+                if self.ETCD_CA_FILE and \
```
fasaxc: Nitpick: prefer parens to `\`

calico/felix/test/test_config.py
```diff
@@ -76,6 +76,10 @@ def test_default_config(self):
 
             # Test defaulting.
             self.assertEqual(config.ETCD_ADDR, "localhost:4001")
+            self.assertEqual(config.ETCD_SCHEME, "http")
+            self.assertEqual(config.ETCD_KEY_FILE, "")
```
fasaxc: I think elsewhere we allow the user to use the string "none" in the config file and we translate that to Python's `None` to pass around.

We should be consistent with that, I think.

calico/etcddriver/driver.py
```diff
@@ -576,6 +578,13 @@ def _stop_watcher(self):
...
 
     def get_etcd_connection(self):
+        if self._etcd_url_parts.scheme == "https":
+            return HTTPSConnectionPool(self._etcd_url_parts.hostname,
+                                       self._etcd_url_parts.port or 2379,
+                                       key_file=self._etcd_key_file,
+                                       cert_file=self._etcd_cert_file,
+                                       ca_certs=self._etcd_ca_file,
+                                       maxsize=1)
         return HTTPConnectionPool(self._etcd_url_parts.hostname,
```
fasaxc: Nitpick: I'd probably use an else: here for symmetry but that's totally aesthetic.

calico/felix/test/test_config.py
```diff
@@ -101,6 +105,49 @@ def test_invalid_action(self):
...
+        }
+
+        for filename in data:
+            log.debug("Test filename : %s", filename)
+            with self.assertRaisesRegexp(ConfigException,
+                                         data[filename]):
+                config = Config("calico/felix/test/data/%s" % filename)
+
+    def test_unreadable_etcd_key(self):
+        with mock.patch("os.path.isfile", autospec=True) as m_isfile, \
```
fasaxc: Doesn't this style of with block break on Python 2.6?

alexwlchan: 1. Yes
2. This is the only UT failure at the moment; everything else is passing

calico/felix/test/test_fetcd.py
```diff
@@ -74,6 +74,10 @@ def setUp(self):
         super(TestEtcdAPI, self).setUp()
         self.m_config = Mock(spec=Config)
         self.m_config.ETCD_ADDR = ETCD_ADDRESS
+        self.m_config.ETCD_SCHEME = "http"
+        self.m_config.ETCD_KEY_FILE = "none"
```
fasaxc: Shouldn't this be `None`?

fasaxc:
Just a minor UT quibble but otherwise good to go in.  I don't need to see it again.

eepyaich:
@neiljerram Can you create an issue for networking-calico to add support for SSL based etcd to our OpenStack driver?

neiljerram:
@eepyaich The link for doing this is at https://bugs.launchpad.net/networking-calico/+filebug.

Did you ask me because you think only I can do it (false), or because you really want me specifically to do it?

TrimBiggs:
I've created PR #915 to rebase this branch off of master, rather than the smc-concurrent-resync branch.

Closing.

calico/felix/test/test_fetcd.py
TrimBiggs: Fixed in new PR

PR #876, opened 2015-11-09
[WIP] Run the Calico UTs inside a container

alexwlchan:
Turns out I'd already written this as a learning exercise. Still has some kinks to work out. Example invocation:

    $ make calicout.created  # builds the container
    $ make ut                # runs the test

Still to do:

* [x] Fix the bug that prevents you using tox on the host and in a container (I was forgetting to pass the `-r` flag to run-unit-tests.sh)
* [x] Use run-unit-tests.sh instead of plain tox
* [x] Work out where to put this in the repo (any suggestions?)
* [x] Update README
* [x] Add a Makefile


alexwlchan:
One other thing that springs to mind &ndash; are we happy that UTs passing on one Linux flavour is sufficient, or do we want a RHEL variant as well?

alexwlchan:
@fasaxc Turns out it was fairly easy to tidy this up. Could you review please?

My only concern is that it's a *big* image. (630MB) Unfortunately, I don't see an easy way to slim down the layers: https://imagelayers.io/?images=alexwlchan%2Fcalico- Will that be a problem?

alexwlchan:
Actually, it occurs to me that I should rebase with your changes for tox coverage before review.

alexwlchan:
I’ve rebased to include Shaun’s changes to tox/coverage. I’d test it, but this container doesn’t work on OS X because distutils tries to create a hard link when it installs the package, and VirtualBox/Docker really don’t like that. That means these UTs will only work on Linux, boo.

It is possible to do some distutils jiggery-pokery to make this work, but I want to think carefully before I actually do it.

utils/calico-uts.Dockerfile
```diff
@@ -0,0 +1,34 @@
+FROM ubuntu:14.04
+
+MAINTAINER Alex Chan <alexchan@projectcalico.org>
+
+RUN apt-get update
+RUN echo "APT::GET::Assume-Yes \"true\";" >> /etc/apt/apt.conf
+
+# tox installs python-posix-spawn and python-etcd from Git
+RUN apt-get install git
```
fasaxc: All these separate RUN stanza will each lay down a filesystem layer.  It's usually better to put them all together to minimize the size of the image.

fasaxc: You could move them to a setup script and run that as one step.  That'd also make it easier to create another container based on CentOS, for example.

alexwlchan: Would it be okay to pull these commands into a shell script, copy that shell script into the image, and just run that script? I'm not keen on stuffing them into one line -- it makes it harder to see what's actually needed.

fasaxc:
I think this is a good start but I'm wondering if we need to join it up with a few other areas before we merge it:

* @tomdee has #849  to add a Dockerfile for Felix.  This adds another one with different setup for the UTs , which feels like a lot of duplication :-(
* How does this relate to the run-unti-tests.sh script that is used by Jenkins?  Should we make containers the official way to run the UTs and integrate this all together?
* PyCharm 5.0 has container support (that's all I know) but I'm not sure if it works with this approach.  Feels like we should close that loop.
* Once we have containers, do we still want tox?  Would it be better to run separate containers with the different levels of Python and do real installs instead of hokey virtualenv installs?

Apart from that, I think you need to address cleanup of unused container images and containers.  They quickly stack up and fill up the disk IME.

To get the size of the container down, you could use one of these ("-slim") images: https://hub.docker.com/_/python/.

fasaxc:
Also: the UTs run as root inside the container, which means any files you map in or copy out will have bad permissions (my .tox dir ended up owned by root after the first run)

alexwlchan:
> tomdee has #849 to add a Dockerfile for Felix. This adds another one with different setup for the UTs , which feels like a lot of duplication

Yeah, this was mostly written a while back and in isolation. We should discuss if/how we want to tie this stuff together.

> How does this relate to the run-unti-tests.sh script that is used by Jenkins? Should we make containers the official way to run the UTs and integrate this all together?

It calls that script, but that's about it. I would prefer making this the one and only way to run UTs, but it needs a bit of work first. Having multiple approaches feels like asking for trouble.

> Once we have containers, do we still want tox? Would it be better to run separate containers with the different levels of Python and do real installs instead of hokey virtualenv installs?

tox has quite nice result output (IMO). Could we replicate that with containers?

> Apart from that, I think you need to address cleanup of unused container images and containers. They quickly stack up and fill up the disk IME.

Yep. The initial version weighs in at >700 MB, which seems excessive. I also haven't thought about how to get coverage information *out*.

I don't have any more time to work on this today. I'm going to push another commit which replaces most of the RUN stanzas with a shell script and tries to cut down the image size.

alexwlchan:
The current image is still ~750MB. That's just too big. :-/

alexwlchan:
via @fasaxc, this seems like it would be worth looking into: https://github.com/openstack/dox

alexwlchan:
I’m closing this PR because it’s languishing, I’m unlikely to work on it any further, and at this point, anybody wanting to do this should probably be starting from scratch.

PR #877, opened 2015-11-09
Rev coverage in tox; ensure code runs with correct concurrency type.

fasaxc:
This brings the UT config into line with my upcoming changes on the smc-concurrent-resync branch.  It should resolve the failing PR UTs.

### DIFF
--- a/run-unit-test.sh
+++ b/run-unit-test.sh
@@ -7,6 +7,17 @@
 # script as './run-unit-test.sh -r'.
 set -e
 
+if [ -n "$VIRTUAL_ENV" ]; then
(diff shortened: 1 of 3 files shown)


alexwlchan:
Looks okay to me, but I'd like to see a clean run of UTs on Jenkins before it gets merged.

matthewdupre:
Does gevent b6 have a fix for the issue we had with the previous b's?

fasaxc:
@matthewdupre Yes, I think so.  I think it's what we've been using for a while because we only excluded explicit versions.  IIRC, it failed at import time before but b6 seems fine.

PR #909, opened 2015-11-20
Add logs for felix status reporting events.

fasaxc:


fasaxc:
@matthewdupre this would be useful for diagnosing @DaveLangridge's scale-testing issue

DaveLangridge:
Does this log when the status of a felix instance changes? (Doesn't look like it.) That would be useful. Should be an unusual event.

matthewdupre:
Marking this do not merge until I can check up on the interaction with the mech driver move.

fasaxc:
@DaveLangridge we already log if the key times out.  This adds a log when a felix checks in too as well as top and tail logs for processing a resync.

fasaxc:
Closed in favour of https://review.openstack.org/249904

PR #921, opened 2015-11-25
Use setup.py to install calico-diags, instead of deb/rpm incantations

neiljerram:
As well as simplicity, this has the advantage that calico-diags will
be correctly installed (including in a virtualenv, if applicable) if
the calico code is installed using 'pip install'.


neiljerram:
FYI this has been through a successful DevStack/Calico test now.

neiljerram:
test this please

PR #922, opened 2015-11-25
Make test call to hit HTTPSConnectionPool for coverage

TrimBiggs:


fasaxc:
ok to test

fasaxc:
Ignoring FV failure: this was a UT-only change.

PR #925, opened 2015-11-30
Remove a few non-deferred stringifications

alexwlchan:


fasaxc:
LGTM

fasaxc:
I thought I had a good regex for tracking these down but it didn't find these ones!

PR #927, opened 2015-12-01
Update install and upgrade instructions for networking-calico packagi…

neiljerram:
…ng change

### DIFF
--- a/docs/source/opens-upgrade.rst
+++ b/docs/source/opens-upgrade.rst
@@ -74,6 +74,16 @@ Now, restart etcd::
 On each machine running the Calico compute software (the component called
 Felix), run the following upgrade steps.
 
+Uninstall pip-installed networking-calico
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+If present, uninstall any pip-installed networking-calico package::
+
+    pip uninstall networking-calico
+
+(networking-calico function is now installed as a Debian or RPM package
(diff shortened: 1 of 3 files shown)


docs/source/opens-upgrade.rst
```diff
@@ -74,6 +74,16 @@ Now, restart etcd::
...
 
+Uninstall pip-installed networking-calico
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+If present, uninstall any pip-installed networking-calico package::
+
+    pip uninstall networking-calico
+
+(networking-calico function is now installed as a Debian or RPM package
+instead.)
```
alexwlchan: Perhaps I’ve missed something, but where’s the step where we install it as a Deb or RPM?

neiljerram: networking-calico is now a Deb/RPM dependency of both calico-compute and calico-control.  So it is pulled in automatically either by a new install of calico-compute/control, or by an upgrade (actually dist-upgrade in the Deb case) of an existing system that has calico-compute/control.

(Also, FYI, the calico-compute and calico-control Deb/RPM packages are now generated by the networking-calico repo.  Whereas previously there were generated from the core Calico repo.)

alexwlchan: Right, makes sense. I didn't realise they were pulled in as a dependency.

:+1:

PR #943, opened 2015-12-14
Fix race in driver FV tests.

fasaxc:
Wait for the driver to be blocked on the watcher queue before
sending the resync request.  This makes the test deterministic.


fasaxc:
Also added a few logs while I was diagnosing the issue.

matthewdupre:
Looks fine.  Will merge shortly.

PR #961, opened 2016-01-18
Add DHCP agent information to data model

neiljerram:


neiljerram:
Hi @fasaxc - The associated networking-calico change feels close enough now that it's worth looking at getting this data model change into the core Calico repository.  Please could you review it?

calico/datamodel_v1.py
```diff
@@ -85,6 +86,10 @@
 ENDPOINT_STATUS_DOWN = "down"
 ENDPOINT_STATUS_ERROR = "error"
 
+# Information intended for use by the DHCP agent.
+DHCP_DIR = ROOT_DIR + "/dhcp" + DHCP_VERSION
+SUBNET_DIR= DHCP_DIR + "/subnet"
```
fasaxc: Does it make sense to go straight to subnet rather than having an intermediate network directory?

fasaxc: I'm thinking that we may, later, want to have some config attached to the whole OpenStack-style network and this prevents that

neiljerram: Well I think so, from a DHCP perspective, for the same reason as commented at lines 121-128 of https://review.openstack.org/#/c/241310/17/networking_calico/agent/dhcp_agent.py.

neiljerram: But would such config go under the 'dhcp' subtree, with 'dhcp' being explicitly specified as information for the DHCP agent?

fasaxc: Not right now, but I'm wondering if we could get caught out later.  The only thing that I can see that we're concretely missing is the admin status for the network object.  OpenStack seems to leave everything else up to the subnet object.

PR #969, opened 2016-01-28
Only release ipsets after updating iptables.

### LINKED ISSUE #967: ipsets can be leaked when a tag is no longer referenced by any profile
When a `ProfileRules` object no longer requires a `TagIpset`, it decrefs it.  However, it issues the decref _before_ it updates the iptables rules for the profile.  If the `TagIpset`tries to delete the ipset before the iptables update completes then the ipset is still referenced and the deletion fails.  We ignore the deletion failure because we assume that the ipset was already gone.

Restarting felix cleans up the left-over ipsets.

The fix is to defer the decref of the `TagIpset` until after we've successfully updated iptables.

The overall impact is that rule churn can result in leaked ipsets which are only cleaned up by restarting Felix.

fasaxc:
Fixes #967

### DIFF
--- a/calico/felix/profilerules.py
+++ b/calico/felix/profilerules.py
@@ -127,6 +127,7 @@ class ProfileRules(RefCountedActor):
         self._pending_profile = None
         # Currently-programmed profile dictionary.
         self._profile = None
+        self._required_tags = set()
 
         # State flags.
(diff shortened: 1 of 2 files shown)


matthewdupre:
LGTM

PR #984, opened 2016-02-26
Update docs for 1.3.0.

fasaxc:
* Prefer individual package updates over dist-upgrade.
* Manually install calico-dhcp-agent in this release.

### DIFF
--- a/docs/source/opens-upgrade.rst
+++ b/docs/source/opens-upgrade.rst
@@ -87,27 +87,17 @@ instead.)
 Ubuntu 14.04
 ^^^^^^^^^^^^
 
-First, upgrade packaged components::
+First, use ``apt-get`` to install the updated packages.  On each compute host
+upgrade the Calico packages, as follows::
 
     apt-get update
-    apt-get dist-upgrade
+    apt-get install calico-compute calico-felix calico-common python-etcd \
+                    networking-calico
 
-We recommend upgrading the whole distribution as shown here.  In case you
-prefer to upgrade particular packages only, those needed for a Calico compute
-node are the following.
-
-::
-
-    calico-common
-    calico-compute
-    calico-dhcp-agent
-    calico-felix
-    dnsmasq-base
-    networking-calico
-    neutron-dhcp-agent
-    nova-api-metadata
-    nova-compute
-    python-etcd
+.. warning:: Running ``apt-get upgrade`` is not sufficient to upgrade Calico
+             due to new dependent packages added in version 1.3.  If you want
+             to upgrade Calico as part of a system-wide update, you must use
+             ``apt-get dist-upgrade``.
 
 Then, restart Felix to ensure that it picks up any changes::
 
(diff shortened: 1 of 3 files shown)


docs/source/opens-upgrade.rst
```diff
@@ -87,27 +87,16 @@ instead.)
 Ubuntu 14.04
 ^^^^^^^^^^^^
 
-First, upgrade packaged components::
+First, use ``apt-get`` to install the updated packages.  On each compute host
+you can upgrde only the Calico packages, as follows::
```
neiljerram: typo 'upgrde'

neiljerram: also what does 'can' mean here?

docs/source/opens-upgrade.rst
```diff
@@ -87,27 +87,16 @@ instead.)
...
 
     apt-get update
-    apt-get dist-upgrade
+    apt-get install calico-compute calico-felix calico-common python-etcd
```
neiljerram: Surely this is insufficient, as it misses lots of packages from the list below (being deleted).

docs/source/opens-upgrade.rst
```diff
@@ -119,12 +108,23 @@ process when the binary has been updated since it started running::
...
+the Neutron-provided one.  The Calico DHCP agent is backed by etcd, allowing
+it to scale to higher numbers of hosts::
 
     service neutron-dhcp-agent stop
-    service calico-dhcp-agent restart
+    echo manual > /etc/init/neutron-dhcp-agent.override
+    apt-get install calico-dhcp-agent
 
-Or if you are using an earlier OpenStack release::
+Check that the only the Calico DHCP agent is now running::
```
neiljerram: typo: 'the'

docs/source/opens-upgrade.rst
```diff
@@ -182,22 +196,16 @@ neutron-server), run the following upgrade steps.
...
+First, use ``apt-get`` to install the updated packages.  On each control host
+you can upgrade only the Calico packages, as follows::
 
     apt-get update
-    apt-get dist-upgrade
-
-We recommend upgrading the whole distribution as shown here.  In case you
-prefer to upgrade particular packages only, those needed for a Calico control
-node are the following.
+    apt-get install calico-control calico-common python-etcd
```
neiljerram: Again, surely wrong, as omits networking-calico?

docs/source/ubuntu-opens-install.rst
```diff
@@ -351,13 +352,15 @@ perform the following steps:
...
+   For OpenStack Liberty and later, install the Calico DHCP agent (which
+   uses etcd, allowing it to scale to higher numbers of hosts) and disable
+   the Neutron-provided one:
 
    ::
 
-       sudo service neutron-dhcp-agent stop
+    service neutron-dhcp-agent stop
+    echo manual > /etc/init/neutron-dhcp-agent.override
+    apt-get install calico-dhcp-agent
```
neiljerram: Indentation looks wrong here, have you checked RST compilation?

docs/source/opens-upgrade.rst
fasaxc: The other packages should get pulled in by the version dependencies in the "main" packages.  Any package that you explicitly `apt-get install` gets set to "manually installed" in the APT database, which we don't want for all our dependencies.

I've included felix and common here because I wasn't sure that calico-compute had an explicit version dependency in it.  I included python-etcd because we've made 0.4.3 available but our dependency is still 0.4.1 so it wouldn't get upgraded by default.

docs/source/ubuntu-opens-install.rst
neiljerram: Also if you're removing sudo, we should do that consistently throughout.

docs/source/opens-upgrade.rst
fasaxc: Doesn't calico-control depend on (the right version of) networking-calico?

PR #992, opened 2016-03-15
Add diff-cover output to run-unit-test.sh output.

fasaxc:
Shows which lines of coverage you've missed in your patch.


neiljerram:
Nice.  Happy to merge this once testing has gone through.

neiljerram:
FV failure can't possibly be related to this change, so merging anyway.

PR #995, opened 2016-03-18
Add stack-trace to port status error message.

fasaxc:
Exception was expected so previously we didn't log the trace.
Now we're seeing an unexpected exception.


neiljerram:
LGTM.

fasaxc:
retest this please

PR #998, opened 2016-03-23
Explicitly track which IPs should be removed from conntrack

### LINKED ISSUE #986: First conntrack call can fail with EINVAL
Seen in an overnight FV run:
```
2016-03-03 03:28:11,775 [ERROR][4973/19] calico.felix.devices 353: Failed to remove conntrack flows for 10.28.0.62. Ignoring.
  File "/usr/lib/python2.7/dist-packages/calico/felix/devices.py", line 342, in remove_conntrack_flows
    direction, ip])
  File "/usr/lib/python2.7/dist-packages/calico/felix/futils.py", line 340, in check_call
    args, retcode, stdout, stderr, input=input_str)
FailedSystemCall: Failed system call (retcode : 1, args : ('conntrack', '--family', 'ipv4', '--delete', '--orig-src', '10.28.0.62'))
  stdout  : 
  stderr  : conntrack v1.4.1 (conntrack-tools): Operation failed: invalid parameters

  input  : None

2016-03-03 03:28:11,792 [INFO][4973/21] calico.felix.endpoint 562: Interface tap5bac9c15-61 deconfigured
```
Hunch is that the first call to conntrack triggers loading of the `nf_conntrack_netlink` kernel module but there's a race causing the first call to fail.  Subsequent calls to conntrack then succeed.

We should ensure the kernel module is loaded at start of day.

### LINKED ISSUE #987: Suspiciously few "Interface deconfigured" logs
While investigating #986, it seemed suspicious that we only ran our first conntrack command a long way into the test.  We should be running conntrack commands every time an interface is removed to clean up the relevant conntrack state.

fasaxc:
Previously we relied on loading the IPs from the routing table, but that's not sound when an interface is being torn down because its routes may have already gone by the time we read them.  (Fixes #987.)

In addition, pre-load the `nf_conntrack_netlink` kernel module to work around #986 (Fixes #986).

### DIFF
--- a/calico/felix/devices.py
+++ b/calico/felix/devices.py
@@ -293,7 +293,6 @@ def set_routes(ip_type, ips, interface, mac=None, reset_arp=False):
     removed_ips = (current_ips - ips)
     for ip in removed_ips:
         del_route(ip_type, ip, interface)
-    remove_conntrack_flows(removed_ips, 4 if ip_type == futils.IPV4 else 6)
     for ip in (ips - current_ips):
         add_route(ip_type, ip, interface, mac)
     if reset_arp:
--- a/calico/felix/endpoint.py
+++ b/calico/felix/endpoint.py
@@ -388,6 +388,8 @@ class LocalEndpoint(RefCountedActor):
         self._pending_endpoint = None
         self._endpoint_update_pending = False
         self._mac_changed = False
+        # IPs that no longer belong to this endpoint and need cleaning up.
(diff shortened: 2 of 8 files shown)


matthewdupre:
Looks good.  Nicely done.

I'd be tempted to assert that _removed_ips was empty at the start of each batch, mostly for symmetry.

fasaxc:
Ta.  I can't see a good place to add that assert so I'm going to leave it out.

PR #1001, opened 2016-03-30
Fix RPM pyparsing dependency.

fasaxc:
Package name was incorrect, should just be "pyparsing", not "python-pyparsing".


neiljerram:
Thanks, LGTM.  This change isn't going to be covered by our testing, so not waiting for that.

PR #1003, opened 2016-03-31
Add support for negated match criteria.

fasaxc:
For example, in addition to `src_tag`, we now allow `!src_tag`, which matches if the source IP is not a member of the tag.

Note:

- Even negated matches for, for example, ports, require a non-negated protocol match.  This is because a port match is only valid (negated or not) if the protocol has ports.
- `!icmp_type` and `!icmp_code` are treated together as one match, meaning `not(icmp_type==<type> && icmp_code==<code>)` rather than the more obvious `not(icmp_type==<type>) && not(icmp_code==<code>)`.  This is a kernel limitation; the kernel can only match on ICMP type and code together.

### DIFF
--- a/calico/common.py
+++ b/calico/common.py
@@ -69,9 +69,7 @@ KERNEL_PORT_PROTOCOLS = set([
     "dccp", 33, "33",
 ])
 
-# Valid keys for a rule JSON dict.
-KNOWN_RULE_KEYS = set([
-    "action",
+NEGATABLE_MATCH_KEYS = [
     "protocol",
     "src_net",
     "src_tag",
@@ -83,8 +81,17 @@ KNOWN_RULE_KEYS = set([
     "dst_ports",
     "icmp_type",
     "icmp_code",
-    "ip_version",
-])
+]
+
+# Valid keys for a rule JSON dict.
+KNOWN_RULE_KEYS = set(
+    [
+        "action",
+        "ip_version",
+    ] +
+    NEGATABLE_MATCH_KEYS +
+    ["!%s" % k for k in NEGATABLE_MATCH_KEYS]
+)
 
 # Valid actions to see in a rule.
 KNOWN_ACTIONS = set(["allow", "deny", "next-tier"])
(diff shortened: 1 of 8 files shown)


calico/common.py
```diff
@@ -601,105 +608,132 @@ def _validate_rules(rules_dict, issues):
...
+    if icmp_code is not None:
+        if not isinstance(icmp_code, int):
+            issues.append("ICMP code is not an integer in rule %s." %
+                          rule)
+        elif not 0 <= icmp_code <= 255:
+            issues.append("ICMP code is out of range.")
+        if icmp_type is None:
+            # ICMP code without ICMP type not supported by iptables;
+            # firewall against that.
+            issues.append("ICMP code specified without ICMP type.")
```
alexwlchan: Do you want `in rule %s` in these issue descriptions?

(Not actually reviewing this, I just glanced at the code and happened to spot this.)

fasaxc: Hmm, it's not ideal to have all those format strings here.  It'd be better to stringify the rule once when we raise the exception.  Since I'm only moving this code around, I'd rather leave as-is for now.

fasaxc:
@matthewdupre Please can you take this PR?

calico/common.py
```diff
@@ -601,105 +608,132 @@ def _validate_rules(rules_dict, issues):
...
+
+    # Check the (possibly negated) profile.
+    protocol_key = neg_pfx + 'protocol'
+    maybe_neg_proto = rule.get(protocol_key)
+    if maybe_neg_proto is not None and maybe_neg_proto not in KERNEL_PROTOCOLS:
+        issues.append("Invalid %s %s in rule %s" %
+                      (protocol_key, maybe_neg_proto, rule))
+    elif maybe_neg_proto is not None:
+        maybe_neg_proto = intern(str(maybe_neg_proto))
+        rule[protocol_key] = str(maybe_neg_proto)
```
matthewdupre: Do we really want the rule object to contain '!protocol': '!tcp'?  It seems superfluous and confusing to have two !s?  (I also found this test tricky to read.)

calico/common.py
```diff
@@ -601,105 +608,132 @@ def _validate_rules(rules_dict, issues):
...
+
+    ip_version = rule.get('ip_version')
+    if ip_version == 4 and protocol == "icmpv6":
+        issues.append("Using icmpv6 with IPv4 in rule %s." % rule)
+    if ip_version == 6 and protocol == "icmp":
+        issues.append("Using icmp with IPv6 in rule %s." % rule)
+
+    for tag_type in (neg_pfx + 'src_tag', neg_pfx + 'dst_tag'):
+        tag = rule.get(tag_type)
```
matthewdupre: Are we really allowing both 'src_tag' and '!src_tag' in the same rule?  If not we should police it (which probably breaks the model of calling this function once per negationness).

calico/common.py
```diff
@@ -601,105 +608,132 @@ def _validate_rules(rules_dict, issues):
...
+
+    icmp_type = rule.get(neg_pfx + 'icmp_type')
+    if icmp_type is not None:
+        if not isinstance(icmp_type, int):
+            issues.append("ICMP type is not an integer in rule %s." %
+                          rule)
+        elif not 0 <= icmp_type <= 255:
+            issues.append("ICMP type is out of range in rule %s." %
+                          rule)
+    icmp_code = rule.get(neg_pfx + "icmp_code")
```
matthewdupre: This doesn't validate that the negationness of the type and code match.

calico/common.py
matthewdupre: Looking elsewhere, it looks like you do intend to allow this.  How are endpoints in both tags treated?  This definitely needs to be documented.

calico/common.py
matthewdupre: Looks like I've confused what was in KERNEL_PROTOCOLS.

calico/felix/plugins/fiptgenerator.py
```diff
@@ -783,70 +784,96 @@ def _rule_to_iptables_fragments_inner(self, chain_name, rule, ip_version,
...
+                        (proto, ports_key, rule)
+                    if neg_pfx == '':
+                        # Positive match; caller has already chunked the
+                        # ports list up into blocks of suitable length for
+                        # _ports_to_multiport.  We only see one chunk.
+                        ports_str = self._ports_to_multiport(ports)
+                        append("--match multiport", "--%s-ports" % direction,
+                               ports_str)
+                    else:
+                        # This is a negative match.  While an individual
```
matthewdupre: We don't really need to support this.  We could also support positive >15 matches if we wanted, by adding more rules in.  It probably won't scale that far anyway - I imagine there'll be a limit to how many different matches we can use anyway.

matthewdupre:
I have some concerns about allowing the same criteria in a rule in both positive and negative form - I think this could be confusing.  I expect the behaviour as it stands will be 'negative wins' when something hits both.

Do we expect people to make much use of 'is in tag A and isn't in tag B' type rules?

matthewdupre:
Other than a worry about perhaps giving people a bit too much rope the implementation is pretty simple and looks fine.

calico/common.py
fasaxc: Match criteria in the same rule are anded together so `{"!src_tag": "foo", "src_tag": "bar"}` matches IPs that belong to endpoints in tag `bar`, unless they're in tag `foo`.

`{"!src_tag": "foo", "src_tag": "foo"}` wouldn't match anything.

calico/common.py
fasaxc: Not sure I follow.  Either `neg_pfx` is `""` or it's `"!"` so we'll validate both together?

calico/felix/plugins/fiptgenerator.py
fasaxc: We already do support >15 positive matches, that's handled in the calling function, which splits up the rules.  Rationale for supporting that is that it avoids needing to feed errors back up through, say OpenStack, which we didn't know how to do at the time (maybe we still don't?).  Also 15, is a pretty low limit, easy to see someone wanting to throw 16 or more in there.

According to the iptables docs, you can have one rule with a million matches in it if you want to.

fasaxc:
> I expect the behaviour as it stands will be 'negative wins' when something hits both.

Match criteria in the same rule are always "anded" together so having both a positive and negative match criteria in the rule has pretty reasonable behaviour: both criteria must be satisfied.

> Do we expect people to make much use of 'is in tag A and isn't in tag B' type rules?

Main reason I did it that way was that I didn't need to police anything.  Second reason was it felt a little richer (and simple/orthogonal given the anding behaviour above).  I can't think of a _great_ example use-case but how about something where you've got a "team" and a "user", you could do something like `{"src_tag": "ocelot", "!src_tag": "shaun"}` to match traffic from other ocelots.

fasaxc:
Pushed a new commit with beefed up docs and a self-review fix (spotted that I failed to update `profilerules.py` to extract the new dependencies).

calico/common.py
matthewdupre: It took me a minute to get the correct context of 'match' looking back too!  I mean that it doesn't reject 'icmp_type':'3', '!icmp_code':'5' - where only one is negated.

We could declare etcd trusted and punt the problem to calicoctl (etc), but since we are doing validation here, we should try and do all of it.

matthewdupre:
OK - I'm happy, except for the lack of validation of mismatched codes and types (code && !type etc).

docs/source/etcd-data-model.rst
```diff
@@ -339,23 +339,49 @@ Each rule sub-object has the following JSON-encoded structure:
...
       "dst_net": "<CIDR>",
       "dst_ports": [1234, "2048:4000"],
-      "icmp_type": <int>,
-      "action": "deny|allow",
+      "icmp_type": <int>, "icmp_code": <int>,  # Treated together, see below.
+
+      # Negated matches:
+      "!protocol": ...,
+      "!src_tag": ...,
+      "!src_selector": ...,
```
matthewdupre: What does this mean?  As I understand it, src_selectors should support negation internally, so this is perhaps superfluous?

matthewdupre:
I spoke to Casey, and one question came out of that w.r.t. the meaning of !src_selectors.

calico/common.py
fasaxc: I think it gets validated a few lines below:
```
        if icmp_type is None:
            # ICMP code without ICMP type not supported by iptables;
            # firewall against that.
            issues.append("ICMP code specified without ICMP type.")
```

docs/source/etcd-data-model.rst
fasaxc: A selector with negation inside it will render the known (endpoint) IPs that don't match the selector.  "!src_selector" renders the positive ipset and then negates the match so it'll match IPs that are not assigned to endpoints too.

fasaxc:
@matthewdupre Think I've addressed all your comments now, back to you for the OK.

fasaxc:
Rebased onto master.

docs/source/etcd-data-model.rst
matthewdupre: @caseydavenport and I chatted about this.  We have mixed feelings on this:
- we both think that having !src_selector(...) be different to src_selector(!...) is unintuitive.
- we're not agreed on whether this behaviour is logically correct or not.
- we're not sure what we need it for anyway (I'm not entirely clear what the specific use case for this PR is at all tbh).

Therefore, can we please remove support for !src_selector?

fasaxc:
I think we lost this comment from @matthewdupre  in a rebase or something:

> 
> In docs/source/etcd-data-model.rst:
> 
> >        "src_net": "<CIDR>",
> >        "src_ports": [1234, "2048:4000"],
> >        "dst_tag": "<tag_name>",
> >        "dst_net": "<CIDR>",
> >        "dst_ports": [1234, "2048:4000"],
> > -      "icmp_type": <int>,
> > -      "action": "deny|allow",
> > +      "icmp_type": <int>, "icmp_code": <int>,  # Treated together, see below.
> > +
> > +      # Negated matches:
> > +      "!protocol": ...,
> > +      "!src_tag": ...,
> > +      "!src_selector": ...,
> @caseydavenport and I chatted about this. We have mixed feelings on this:
> 
> we both think that having !src_selector(...) be different to src_selector(!...) is unintuitive.
> we're not agreed on whether this behaviour is logically correct or not.
> we're not sure what we need it for anyway (I'm not entirely clear what the specific use case for this PR is at all tbh).
> Therefore, can we please remove support for !src_selector?

I think there are good use-cases for both.  It really depends whether you're trying to negate the whole match or whether you're negating the selection.  Negation "outside" the selector is useful in rules that drop traffic if it doesn't come from the expected source.  For example, "drop if not from my security group".  Another approach would be to do "accept if from my security group" but that does an immediate accept so it doesn't allow further rules to refine the match.

Negation "inside" the selector makes the selectors more capable so you can express things like "has(bar) && !has(foo)".  It just falls out that once you add negation, it applies to other operations too.

matthewdupre:
I think our concern was about unintuitiveness of !src_selector(X) not being the same as src_selector(!X).

Logically speaking I actually think the current way is right: if you treat selectors as k8s API objects that only apply to pods, src_selector(!X) really means is_pod() && src_selector(!X) which negates to !is_pod() || !src_selector(!X | it's a pod) which is the same as !is_pod() || src_selector(X) == !src_selector(X) as above..  I don't think it's reasonable to expect users to grasp that subtlety though.

In a sense, unless we're planning on adding selector support to calicoctl it doesn't really matter; we just need to be careful about the wording of how selectors work in the edge cases where things are unlabelled.

lxpollitt:
We _are_ planning to add selector based policies to calicoctl.

caseydavenport:
I'm concerned about the vagueness of the API w.r.t selectors (which has been brought to light by this PR).

From an API standpoint, a Calico policy selector is (should be?) essentially a function that takes an Endpoint and returns `True` or `False`.  It isn't defined on an IP address.  The fact that `!src_selector: has(key)` will allow traffic from all IP addresses that don't belong to any Calico Endpoints is leaking our current implementation back up to the API.  If (somehow) we used something other than IP addresses to filter traffic from endpoints, would we still define the behavior this way?  Same thing goes for `!tag: sometag`.

Other thoughts:
- I think `src_selector` would be more appropriately called `src_endpoint_selector` or similar to keep the API explicit and more easily extensible (potentially hard to do this if we wait any longer).
- What are the use cases for `!src_selector` and `!tag`? Are we trying to solve a problem that may or may not exist?  I'd argue that the API should state that `!src_selector: has(k)` should equal `src_selector: !has(k)`, which makes the two syntaxes redundant.  I'm in favor of only supporting one.

@lxpollitt @fasaxc @matthewdupre - We should have a call to discuss this and sort out concerns.

matthewdupre:
Comment looks fine.

PR #1007, opened 2016-04-05
Doc: update to cover Calico plugin for Fuel 7.0

neiljerram:


docs/source/fuel-integration.rst
```diff
@@ -12,14 +12,31 @@
...
 
-Integration with Mirantis Fuel 6.1
-==================================
+==============================
+Integration with Mirantis Fuel
+==============================
 
-This document describes our integration of Calico with the
-Mirantis Fuel system for rapidly deploying an OpenStack cluster.
+Calico plugins are available for Fuel 6.1 and 7.0.
```
fasaxc: If we know 8.0 doesn't work, should we drop in a warning?

eepyaich:
LGTM - but as @fasaxc says, maybe a note to say Fuel 8 is not currently supported (but is expected to come in summer 2016 - get in touch if this is important to you) would be a good idea.

neiljerram:
@eepyaich @fasaxc Does that added note look OK?

PR #1015, opened 2016-05-05
Fix that next-tier action also accepted packets.

### LINKED ISSUE #1014: next-tier action of final tier accepting packet?
Casey put a DROP rule in a profile and used a next-tier rule in his policy, which should have deferred to the profile but the packet seems to have been accepted.

fasaxc:
Previously, we were setting the accept mark at the start of the profile chain and then unsetting it at the end of the chain if no rules matched.  That was correct while the only returning rule type was "allow", but it meant that when we added "next-tier" it also implicitly accepted the packet.

Fixes #1014

### DIFF
--- a/calico/felix/plugins/fiptgenerator.py
+++ b/calico/felix/plugins/fiptgenerator.py
@@ -425,12 +425,10 @@ class FelixIptablesGenerator(FelixPlugin):
 
         # Generates an inbound and an outbound chain for each profile.
         # Within each chain, the logic is as follows:
-        # * Start by marking all packets.
-        # * If we hit an allow rule, we'll return with the Accept mark still
-        #   set, to indicate that we matched.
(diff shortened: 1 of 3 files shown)


matthewdupre:
I think the change is good, although it did take me a few minutes to figure out how it actually worked.

Do we maintain a description of how the rules are supposed to look anywhere?  If not we probably should.

fasaxc:
@matthewdupre I think that code is pretty well commented; might help to start in `_build_to_or_from_chain()`, which documents what it's expecting from the profile chains.

matthewdupre:
I think it's OK for now, and I'll merge once we've got FV approval: what I meant was there's nowhere you can read 1-2 pages and understand what the rules are.  The micro commenting is fine, but I had to think for a bit about things like how multiple tiers fit into iptables, and build the 'output' picture in my head.  It's more about information density than presence.

PR #1021, opened 2016-05-20
Fix that we were using the chain name suffix as the interface suffix.

fasaxc:
We were using a utility function to calculate the suffix part of an interface name but, apart from extracting the suffix, the utility function also scrambled it if the name was too long.

That behaviour was correct for a chain name, but not for calculating the suffix of the interface name.

### DIFF
--- a/calico/felix/test/test_dispatch.py
+++ b/calico/felix/test/test_dispatch.py
@@ -106,23 +106,28 @@ class TestDispatchChains(BaseTestCase):
         d.programmed_leaf_chains.add("felix-FROM-EP-PFX-a")
         d.programmed_leaf_chains.add("felix-FROM-EP-PFX-z")
         ifaces = ['tapa1', 'tapa2', 'tapa3',
-                  'tapb1', 'tapb2',
+                  'tapb1', 'tapb20123456789012345',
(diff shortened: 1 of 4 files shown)


fasaxc:
@neiljerram Please can you take this small review?

calico/felix/test/test_dispatch.py
```diff
@@ -106,23 +106,28 @@ def test_tree_building(self):
         d.programmed_leaf_chains.add("felix-FROM-EP-PFX-a")
         d.programmed_leaf_chains.add("felix-FROM-EP-PFX-z")
         ifaces = ['tapa1', 'tapa2', 'tapa3',
-                  'tapb1', 'tapb2',
+                  'tapb1', 'tapb20123456789012345',
                   'tapc']
         to_delete, deps, updates, new_leaf_chains = d._calculate_update(ifaces)
         self.assertEqual(to_delete, set(["felix-FROM-EP-PFX-z"]))
+        print "Deps", pformat(deps)
```
neiljerram: Do you mean to leave this Print in?

neiljerram:
Looks great to me, other than the Print.

PR #1030, opened 2016-06-07
Update to CLA process

tomdee:


fasaxc:
@tom, I'm going to merge this as-is but how about tweaking the wording to make it clear that the first step is to submit the PR and to give a link to the CLA so they can review it before they start any work?  I know I'd want to know what I'd have to sign up to before I started drafting a patch.

PR #1040, opened 2016-06-23
Add commit hook that checks copyright and license.

fasaxc:
@neiljerram since you're so fond of copyright statements... ;-)


neiljerram:
LGTM, except note that both files will need to be 755.  (Not sure if GitHub shows me that.)

PR #1048, opened 2016-06-24
Add tool to remove all felix iptables/ipsets changes.

fasaxc:


fasaxc:
@neiljerram Please can you review this new tool to remove felix iptables changes/ipsets?

calico/felix/cleanup.py
```diff
@@ -0,0 +1,100 @@
...
+
+def clean_up_iptables(iptables_cmd, iptables_save_cmd):
+    # We can't delete chains until they're unreferenced.  First, remove felix
+    # jump rules from kernel chains.
+    ipt_lines = check_output([iptables_save_cmd]).splitlines()
+    table = None
+    for line in ipt_lines:
+        if line.startswith("*"):
+            # Start of a new table, save off the name.
+            table = line[1:]
```
neiljerram: Think should just have line[1] here, and assert that there aren't more words on the line.  If we don't check, and there ever are extra words, line[1:] won't substitute properly where it's used in further iptables commands below.

PR #1050, opened 2016-06-27
Fix example policy in bare metal docs to be valid json

elfchief:


matthewdupre:
ok to test

matthewdupre:
Oops - good catch.

Thanks for contributing :)

PR #1051, opened 2016-06-28
Use a different conntrack command to trigger module load.

fasaxc:
Old -S command caused error on SL6.5 with conntrack 1.4.3.


matthewdupre:
LGTM

PR #1054, opened 2016-06-29
Missing conntrack requires conntrack, not iptables

elfchief:
Error message previously read "Failed to execute conntrack; Calico
requires iptables to be installed.", which is obviously incorrect.
Fixed to correctly ask for conntrack.


fasaxc:
ok to test

PR #1067, opened 2016-07-19
Add log action and ability to log any rule.

fasaxc:
If the "log_prefix" field is present in a rule, we emit an iptables LOG action in addition to the action for the rule itself.

### DIFF
--- a/calico/felix/plugins/fiptgenerator.py
+++ b/calico/felix/plugins/fiptgenerator.py
@@ -516,8 +516,7 @@ class FelixIptablesGenerator(FelixPlugin):
                     self._profile_to_chain_name("outbound", profile_id)])
 
     def profile_updates(self, profile_id, profile, ip_version, tag_to_ipset,
-                        selector_to_ipset, on_allow="ACCEPT", on_deny="DROP",
-                        comment_tag=None):
+                        selector_to_ipset, comment_tag=None):
         """
         Generate a set of iptables updates that will program all of the chains
         needed for a given profile.
@@ -550,9 +549,7 @@ class FelixIptablesGenerator(FelixPlugin):
                         r,
                         ip_version,
                         tag_to_ipset,
-                        selector_to_ipset,
-                        on_allow=on_allow,
-                        on_deny=on_deny))
+                        selector_to_ipset))
             updates[chain_name] = fragments
 
         return updates, deps
@@ -782,9 +779,7 @@ class FelixIptablesGenerator(FelixPlugin):
                                                  inbound_or_outbound[:1])
 
     def _rule_to_iptables_fragments(self, chain_name, rule, ip_version,
-                                    tag_to_ipset, selector_to_ipset,
-                                    on_allow="ACCEPT",
-                                    on_deny="DROP"):
+                                    tag_to_ipset, selector_to_ipset):
         """
         Convert a rule dict to a list of iptables fragments suitable to use
         with iptables-restore.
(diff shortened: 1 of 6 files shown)


docs/source/etcd-data-model.rst
```diff
@@ -547,6 +551,14 @@ requiring the protocol to be specified):
                all ICMP traffic apart from traffic that matches **both** type
                and code.
 
+``log_prefix``
```
matthewdupre: This doc should also be updated for the new `log` action.

calico/felix/plugins/fiptgenerator.py
```diff
@@ -551,6 +551,39 @@ def profile_updates(self, profile_id, profile, ip_version, tag_to_ipset,
...
+        """
+        drop_rules = self.drop_rules(ip_version, chain_name,
+                                     rule_spec=rule_spec, comment=comment,
+                                     ipt_action=ipt_action)
+        if log_pfx is not None:
+            log_rule = " ".join([ipt_action,
+                                 chain_name,
+                                 rule_spec,
+                                 '--jump LOG --log-prefix "%s: " '
+                                 '--log-level 4' % log_pfx])
```
matthewdupre: This level should be pulled out into a constant.  I also think 4 is a bit too high (particularly since we use this for things that aren't drops).  Maybe 3?

We might need to make it configurable in the rule in future anyway.

matthewdupre:
I think it's probably best not to do it by default, but I can see that people might want the ability to add rate limits to their log rules (`-m limit`).

matthewdupre:
fiptgenerator is starting to get difficult to follow.  I spotted a couple of concretely unpleasant things:
- drop_rules is wrongly named: it only produces one rule, and other code depends on this (such as the equally poorly named logged_drop_rules).
- ipt_targets (~1000) are built in 3 different ways: inline, call out, or passed in.  Relatedly, _rule_to_iptables_fragments_inner() doesn't even use the on_allow parameter!

Any other refactoring ideas would be good.

calico/felix/plugins/fiptgenerator.py
```diff
@@ -1006,14 +1038,21 @@ def _rule_to_iptables_fragments_inner(self, chain_name, rule, ip_version,
...
             rules = [" ".join(["--append", chain_name, rule_spec_str,
                                "--jump", ipt_target])]
-            if extra_rule:
-                rules.append(extra_rule)
+            if extra_rules:
+                rules.extend(extra_rules)
         return rules
 
+    def _log_target(self, rule):
+        log_pfx = rule.get("log_prefix", "calico")
```
matthewdupre: I think we need a slightly more informative default here.

Perhaps 'calico-iptables-packet'?

matthewdupre:
Thanks Shaun - a few comments on this one above.  Actions:

- I think we should add the rate limiting.
- Make / reject the specific fixes.
- If you can think of anything good to do, I feel like there's room to refactor and make it clearer.

fasaxc:
> drop_rules is wrongly named: it only produces one rule, and other code depends on this (such as the equally poorly named logged_drop_rules).

Actually, it returns a list containing a single rule.  It's designed to be overridden in a plugin (and our neflow plugin does hook that function).  that's why I didn't add the logging function to drop_rules directly; that API is part of the plugin API so I can't just add  a new parameter.  `logged_drop_rules` does handle there being multiple rules in `drop_rules()` return value.  It inserts its rule at the start of the list using slice assignment: `foo[0:0]=prefix_list`.

> ipt_targets (~1000) are built in 3 different ways: inline, call out, or passed in. Relatedly, _rule_to_iptables_fragments_inner() doesn't even use the on_allow parameter!

Ugh, I don't think we ever pass in different on_deny values either.  I'll see if it'd be trivial to snip those out; I think we can remove a parameter from the API without breaking plugins.

> Any other refactoring ideas would be good. I found I needed to read the entirety of rule_to_iptables_fragment() and its subfunctions to have confidence in it.

I'm slammed today so I don't have time to do more cleanup.

> I think we should add the rate limiting.

Agreed, that's be really great to have, but I'd rather follow up with that in another PR.  We need to do a release this week and I don't want this to slip out due to feature-creep.

PR #1091, opened 2016-09-01
use projectcalico fork of python-etcd.  it supports etcd v3 server

mikev:


mikev:
Note - this file https://github.com/projectcalico/calico/blob/master/pyi/calico-felix.spec, which was created July 2016.  Doesn't allows us to use GitHub forks.

I've verified this by executing run-unit-test.sh, which fails.

We'll need to fix this first.  Why did we need calico-felix.spec?

matthewdupre:
This isn't needed for a calico-containers release.

PR #1105, opened 2016-09-19
Reduce default etcd watcher timeout to 10s and make configurable.

fasaxc:
This makes us more responsive to connection failures and reduces the
upper bound on how long we'll wait to process the resync_after_current_poll
flag.

At very low churn rates, this increases the number of etcd requests, but
only to a level that we already support in order to handle higher churn.

At high churn rate, this makes no difference because we'll never hit the
timeout.

### DIFF
--- a/calico/etcdutils.py
+++ b/calico/etcdutils.py
@@ -32,6 +32,9 @@ from calico.datamodel_v1 import READY_KEY
 
 _log = logging.getLogger(__name__)
 
+# Since this module does long-polling, we expect read timeouts from etcd but
+# urllib3 logs timeouts at warning level.  Disable that to avoid log spam.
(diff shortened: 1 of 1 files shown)


neiljerram:
retest this please

neiljerram:
I'm happy for this to be merged as soon as we've had a successful test run with it.  Test run 262 (http://localhost:8080/job/pull-request-build/262/) showed a 'UDP stream suffered packet loss' regression, but it's possible that that was just a random glitch.  So we need at least one retest.

neiljerram:
retest this please

lwr20:
retest this please

lwr20:
retest this please

Issue #1106, opened 2016-09-20
Improve code quality by using static code analysis tools


cutwater:
Currently calico project does not provide tools to check code for correctness and PEP8 coding style. The suggestion is to integrate static checks into development process, using **flake8** utility (it includes all pep8 checks and brings some additional ones). It will include several possible stages: manual check performed by developer during working with code, automated check using CI.

**Action items:**
- [ ] Add tox job (env) called "pep8" to run checks
- [ ] Fix major flake8 issues, except those ones, that require code rework
- [ ] Integrate "pep8" job with CI
- [ ] Fix remaining issues

fasaxc:
@cutwater We've now retired the python code on master so PEP-8 doesn't apply any more.  It'd be great to get a golang linter in place though.

fasaxc:
Added gometalinter under #1306

PR #1102, opened 2016-09-15
[WiP] Port "brains" of Felix to golang

fasaxc:
- Golang process becomes the main calico-felix process.
- Python code now becomes the "iptables dataplane driver".
- Golang process handles access to etcd via libcalico, paving the way for etdv3 support as well as other datastores.
- IP sets calculation moved to the golang process.
- New protobuf API between the Golang process and the Python process.
- New build process based on wrapping up pyinstaller bundles in the debs and RPMs.  Allows us to include the golang binary as well as vendoring all our dependencies, including Python itself.

TODOs:

- [ ] Calico/node
- [ ] Usage reporting
- [ ] Scale testing
- [ ] UT
- [x] FV
- [x] Validation

PR #1107, opened 2016-09-20
Add pep8 checks

cutwater:
Related issue: #1106 

Add tox job `pep8`

### DIFF
--- a/run-unit-test.sh
+++ b/run-unit-test.sh
@@ -27,7 +27,7 @@ fi
 
 coverage erase
 
-tox "$@"
+tox -e py27 "$@"
 
 # Make sure we run the following coverage html command with the recent
 # coverage.
--- a/tox.ini
+++ b/tox.ini
@@ -1,6 +1,6 @@
 [tox]
 # TODO: Reinstate pypy once gevent segfault is resolved.
-envlist = py27
+envlist = py27,pep8
 
 [testenv]
 deps =
@@ -38,3 +38,20 @@ deps =
     coverage>=4.0.2,<4.1
     unittest2
     gevent>=1.1b6
+
+[testenv:pep8]
+usedevelop = True
+deps = flake8
+commands =
+    flake8 {posargs:calico}
+
+[flake8]
+# TODO: Fix remaining issues and remove ignored errors list
+# E501: line too long
+# F821: Undefined name
+# F405: <object> may be undefined, or defined from star imports
+# F403: '*' import used; unable to detect undefined names
+ignore = E501,F403,F405,F821
+show-source = True
+show-pep8 = True
+count = True


calico/felix/test/test_fiptgenerator.py
```diff
@@ -365,63 +361,68 @@
     '--append felix-from-abcd --jump MARK --set-mark 0/0x1000000',
     # From chain polices the MAC address.
     '--append felix-from-abcd --match mac ! --mac-source aa:22:33:44:55:66 '
-               '--jump DROP -m comment --comment '
-               '"Incorrect source MAC"',
+    '--jump DROP -m comment --comment '
```
cutwater: pep8 produces warning if trying to indent second line. it can be resolved by using common indentation for wrapping long lines (visually separated by empty lines) or by using parenthesis. Not sure which style is preferable.

fasaxc:
@cutwater Thanks for this; it's something we've been thinking of doing for a long time.  That said, I'm in the middle of a big rework under #1102 and I'm worried that this would make a lot of conflicts (since I've moved a lot of files around).  Was it a manual process to make all these fixes or did you use a tool?  If it was a tool, I think it'd be best to hold off on this PR for now.

PS. Are you on our calico-users #slack?  We have a calico-dev channel on there that's a great place to discuss changes in more detail (or to discuss upcoming changes/get design feedback etc).  Here's an invite link: https://slack.projectcalico.org/

fasaxc:
ok to test

cutwater:
@fasaxc Trivial changes were done automatically, mostly using PyCharm. But some changes have to be done manually, since it's is better not to reformat all code, but fix issues locally and automated tools usually do more than expected. And there are problems like '*' imports that have to be checked manually.

marvin-tigera:
Test FAILed.

marvin-tigera:
Test FAILed.

cutwater:
@fasaxc I updated the patch and left only changes for tox that add new `pep8` checks. Due to conflicts possibility I'll split pep8 issues fixes into smaller patch that can be easily merged.

By some reason one of CI jobs failed and I don't have access to it's logs. Can you please check it.

lwr20:
CI output for this run here: https://gist.github.com/lwr20/8bdb63f6d10be86de783dddbd532166b

Looks rather like the new PEP8 checks failed?

marvin-tigera:
Test FAILed.

cutwater:
pep8 checks should fail at this time, but they should not be executed by CI. Uploaded fix.

marvin-tigera:
Test FAILed.

cutwater:
But it looks like that tox command is executed from jenkins scripts. So this job will fail until either jenkins job is updated or all issues are fixed.

lwr20:
The jenkins job runs the unit tests (and later on an FV test - the FV test is expensive, and unit tests are cheap, so it runs the unit tests to avoid bothering with an FV test for totally broken code).  The CircleCI job only runs the unit tests.  

In this PR, you've added PEP8 checking to the unit tests, right?  So I think it is correct that the Jenkins and Circle jobs fail until all the issues are fixed.  Or have I missed something?

As @fasaxc says, we probably won't be merging this for a little while - and certainly not until after the PEP8 fixes are in, so I think its OK that this PR fails for now.

marvin-tigera:

fasaxc:
@cutwater I'm going to close this for now because it's been broken by the felix port to golang.

As a head up: there's a good chance that we'll complete that port in the coming month or so, so there may not be any python code left after that.

PR #1121, opened 2016-10-03
Swap out docker_build_and_run.sh for a Makefile.

fasaxc:
Synchronizes the build with that used for the new Golang version of felix.


fasaxc:
Decided to make SemaphoreCI config condition on Makefile instead

PR #1125, opened 2016-10-05
Update urllib3 version for "containers" builds

caseydavenport:


marvin-tigera:
Build finished.

caseydavenport:
@fasaxc Cool, thanks for the context.

I'll see if adding to the calico/felix image is sufficient.

caseydavenport:
@fasaxc I'm not really sure what goes into our OpenStack builds or not.

It looks like to make this work I need to install the right version of urllib3 and the other deps before the pyinstaller build takes place, which means I can't just do it in the root Dockerfile since that takes the output of the pyinstaller build as an input.

Will the latest commit here be acceptable?

marvin-tigera:
Test PASSed.

fasaxc:
Ah, forgot we'd moved over to using the pyinstaller in calico/node already.  Yes, this should work fine to get the right code into calico/node, then the golang work will supercede once it's ready.

PR #1132, opened 2016-10-18
Pin pycparser version.

fasaxc:
Pinned in the tox.ini file because it's not currently a direct dependency of felix.  Installing it before installing felix itself is enough.


marvin-tigera:
Test PASSed.

marvin-tigera:
Test PASSed.

PR #1135, opened 2016-10-27
Do floating IP DNATs from POSTROUTING as well as from PREROUTING

neiljerram:
This is needed so that an endpoint can be pinged from its own host using
its floating IP.  One reason for wanting that is that that is what
OpenStack Tempest tests do.  More generally it adds utility without any
downside.

The PREROUTING chain is only traversed for packets that are received on
the compute host from elsewhere; so our use of PREROUTING to implement
floating IPs works for such packets, but not for packets that originate
from the compute host.  To cover the latter, we can traverse the
floating IP DNAT rules from the POSTROUTING chain as well.

Note that we also (already) do floating IP-related SNATs in the
POSTROUTING chain.  For those to operate correctly, the DNATs have to
happen first.

### DIFF
--- a/python/calico/felix/frules.py
+++ b/python/calico/felix/frules.py
@@ -300,9 +300,9 @@ def install_global_rules(config, filter_updater, nat_updater, ip_version,
                 (iface_match, CHAIN_PREROUTING),
                 async=False)
 
-    # Both IPV4 and IPV6 nat tables need felix-PREROUTING and
-    # felix-POSTROUTING, along with the dependent DNAT and SNAT tables
-    # required for NAT/floating IP support.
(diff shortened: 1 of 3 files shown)


neiljerram:
Yay, this in conjunction with https://review.openstack.org/#/c/390908/ allows Calico to pass a useful number of Tempest 'scenario' tests, as you can see at http://logs.openstack.org/08/390908/7/experimental/gate-tempest-dsvm-networking-calico-nv/476eda3/logs/testr_results.html.gz

neiljerram:
Merging this now myself, given that @fasaxc already approved the changes.

PR #1140, opened 2016-11-01
Update README for Golang felix.

fasaxc:


README.md
```diff
@@ -58,59 +52,72 @@ Before you do so, you should check out our contributing guidelines in the
...
+Felix uses Docker for builds.  We develop on Ubuntu 16.04 but other Linux
+distributions should work (we have not tries OS X).  To build felix, you will need to install:
+
+- Docker 1.12+
+- GNU make.
+- Plenty of disk space (since the builds need to use some heavyweight 
+  full-OS containers in order to build debs and RPMs).
+
+Then, run `make felix-docker-image`, for example, to build the `calico/felix`
+container or `make help` for other options.
```
fasaxc: `rkt` does support docker images but they're not its native format.  Might be nice to add a `make felix-rkt-image` that does spit out the tar-based rkt image so I think the terminology is correct.

README.md
```diff
@@ -58,59 +52,72 @@ Before you do so, you should check out our contributing guidelines in the
...
+By using docker, the build does not need to be run on the target platform.
+```
+make trusty-deb
+make xenial-deb
+make rpm
+```
+The packages (and source packages) are output to the dist directory.
+
+By default, debs are built with a "snapshot" version number that is
+greater than the current release version.
```
fasaxc: I don't think I've done that for RPMs yet.  It was needed to make PPA uploads work well so I was forced to do it for debs.

PR #1145, opened 2016-11-02
Clean up config parsing: remove unused config parameters.

fasaxc:
Remove the configuration parameters from the dataplane driver that are no longer used.

CC @caseydavenport This should fix the etcd endpoints issue you were seeing.

### DIFF
--- a/python/calico/felix/config.py
+++ b/python/calico/felix/config.py
@@ -25,10 +25,8 @@ On instantiation, this module automatically parses the configuration file and
 builds a singleton configuration object. That object may (once) be changed by
 etcd configuration being reported back to it.
 """
-import os
 from numbers import Number
 
-import ConfigParser
 import logging
 import socket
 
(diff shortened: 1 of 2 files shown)


python/calico/felix/config.py
```diff
@@ -234,17 +214,8 @@ def __init__(self):
                            False, value_is_bool=True)
         self.add_parameter("ClusterGUID", "Unique cluster identifier",
                            "baddecaf")
```
neiljerram: Think ClusterGUID can be removed too.

python/calico/felix/config.py
```diff
@@ -173,24 +169,8 @@ def __init__(self):
         self.parameters = {}
```
neiljerram: Please update block comment above to reflect the new general reality, and specifically not to mention ETCD_ADDRS.

PR #1146, opened 2016-11-02
Fix executable paths inside Dockerfile.

fasaxc:
The calico-felix binary wasn't able to find the calico-iptables-plugin binary because it wasn't on the path.


robbrockbank:
LGTM

fasaxc:
Not much point waiting for CI.  It doesn't test this, I don't think.

PR #1149, opened 2016-11-02
Rev Felix version to be a dev version of 2.0.0.

fasaxc:
Thought revving this now might flush out any build issues.


neiljerram:
Still LGTM!

PR #1153, opened 2016-11-03
Build review

neiljerram:


Makefile
```diff
@@ -84,20 +132,20 @@ MY_UID:=$(shell id -u)
 MY_GID:=$(shell id -g)
 
 # Build a docker image used for building our go code into a binary.
-.PHONY: golang-build-image
-golang-build-image:
+.PHONY: calico-build/golang
+calico-build/golang:
```
fasaxc: Like the streamlining of naming.  Couple of thoughts on the scheme though:
- our dockerhub prefix is `calico/` so we wouldn't be able to push these to dockerhub for caching purposes without retagging them (unless we grab calico-build too).
- should we put felix in there somewhere to distinguish from other projects and avoid name clashes?

neiljerram:
@fasaxc Hooray, this has now passed. I'm not sure about including the 'Version 2.0.0-beta' release commit in this PR, though, so please don't merge until I've thought more about that.

debian/build-debs
```diff
@@ -41,8 +41,8 @@ cd ${unpacked_src_dir}
 # *.so exclusion removed.
 excludes='-I*/*~ -I,,* -I.[#~]* -I.git -I.gitattributes -I.gitignore -I.gitmodules -I.gitreview'
 
-debuild -i="" $excludes -us -uc -b
-debuild -i="" $excludes -us -uc -S
+debuild --no-lintian -i="" $excludes -us -uc -b
```
neiljerram: (Well, at the moment it's really just generating a ton of errors in our build output, and our Debian packages are parsecs away from meeting Debian policy!)

utils/create-release-commit.sh
```diff
@@ -0,0 +1,69 @@
...
+
+# Read the new version back again from CHANGES.md - in case the user changed
+# it.
+nextrel=`grep '##' CHANGES.md | head -1 | awk '{print $2;}'`
+
+# Set the source package name.  (Some library functions need this.)
+package=felix
+
+# Update the Debian changelog.
+series=__STREAM__ update_debian_changelog
```
neiljerram: update_debian_changelog generates a changelog stanza with a version like

    2.0.0~beta-__STREAM__

And this is what gets committed into our Git history.

Then make deb handles converting __STREAM__ to either trusty or xenial.

Make sense?

utils/create-release-commit.sh
```diff
@@ -0,0 +1,69 @@
+#!/bin/bash -x
+
+# Automatically update the codebase for a new release, and (locally) commit
+# those changes.
+#
+# Usage: create-release-commit.sh [<NEW-VERSION>]
+#
+# If <NEW-VERSION> isn't specified on the command line, we prompt for it.
+
```
neiljerram: Sure, will do.

Makefile
```diff
@@ -284,26 +316,15 @@ $(BUNDLE_FILENAME): dist/calico-felix/calico-iptables-plugin dist/calico-felix/c
...
-	docker exec felix-pyi-build sh -c 'cd python && pip install .'
-	# As the current user, build the PyInstaller bundle.
-	docker exec -u $(MY_UID):$(MY_GID) \
-	            felix-pyi-build \
-	            /code/docker-build-images/pyi/run-pyinstaller.sh
-	# Finally, shut down the container.
-	docker rm -f felix-pyi-build
+	# Create and run build container.
+	$(DOCKER_RUN_RM_ROOT) -w /code/python calico-build/python sh -c \
+	    'pip install . && ../docker-build-images/pyi/run-pyinstaller.sh && rm -rf ../build `find . -name "*.pyc"` && chown -R $(MY_UID):$(MY_GID) ../dist'
```
neiljerram: Yes.  It's tricky but I found a way to do this.

Makefile
```diff
@@ -326,40 +347,27 @@ update-tools:
...
-	docker exec --user $(MY_UID):$(MY_GID) felix-ut python2.7 ./setup.py egg_info
-	# Do the install as root.
-	docker exec felix-ut sh -c 'pip install -e .'
-	# Run the UTs as non-root.
-	docker exec --user $(MY_UID):$(MY_GID) felix-ut sh -c 'COMPARE_BRANCH=$(UT_COMPARE_BRANCH) ./run-unit-test.sh'
-	# Tear down the container.
-	docker rm -f felix-ut
+	$(MAKE) calico-build/python
+	$(DOCKER_RUN_RM_ROOT) -w /code/python calico-build/python sh -c \
+	    "pip install . && COMPARE_BRANCH=$(UT_COMPARE_BRANCH) ./run-unit-test.sh && chown -R $(MY_UID):$(MY_GID) coverage.xml .coverage htmlcov"
```
neiljerram: Ditto.

PR #1163, opened 2016-11-04
Pin logrus to desired version

caseydavenport:


fasaxc:
Happy for this to go in as a stop-gap but we should rev to 0.11.0 and fix up our code after the beta.

fasaxc:
FV failure looks like a single flaky test to me.

PR #1167, opened 2016-11-07
Reset versioning and pins for ongoing development

neiljerram:


neiljerram:
@fasaxc I've merged this so as not to block any further Felix development - but would still appreciate if you could review that I got it right, when you're back in.

python/setup.py
```diff
@@ -45,7 +45,7 @@ def collect_requirements():
 
 setuptools.setup(
     name="felix",
-    version="2.0.0-beta3",
+    version="2.0.0.dev1",
```
fasaxc: Python, OTOH, does special-case some pre-release labels so `2.0.0.dev1 < 2.0.0-beta3`.  One option that will sort after the beta but before the next beta/full release is `2.0.0b4.dev1`: See: https://www.python.org/dev/peps/pep-0440/#developmental-releases

neiljerram:
Think this will shortly be overtaken by more proposed (by Tom) 'make release' change anyway.

PR #1169, opened 2016-11-10
[WIP] Untracked

neiljerram:


python/calico/felix/profilerules.py
```diff
@@ -246,9 +258,11 @@ def _delete_chains(self):
         """
         # Need to block here: have to wait for chains to be deleted
         # before we can decref our ipsets.
-        self._iptables_updater.delete_chains(
-            self.iptables_generator.profile_chain_names(self.id),
-            async=False)
+        chain_names = self.iptables_generator.profile_chain_names(self.id)
+        if self._untracked:
```
fasaxc: These conditionals feel like a code smell to me. Why not make the decision on raw/filter in the manager class and then only pass in the relevant iptables updater?

One way to make that work would be to adjust the profile ID to include the raw/filter bool. that'd have a couple of advantages:
- The create method would know which updater to pass in.
- The raw/filter nature would be part of the ID rather than part of the data so handling a change to that flag becomes a delete, then an add.

python/calico/felix/profilerules.py
```diff
@@ -149,6 +156,11 @@ def on_profile_update(self, profile, force_reprogram=False):
         assert not self._dead, "Shouldn't receive updates after we're dead."
         self._pending_profile = profile
         self._dirty |= force_reprogram
+        if profile:
+            if self._untracked is None:
+                self._untracked = profile['untracked']
+            else:
+                assert profile['untracked'] == self._untracked
```
fasaxc: Does this mean that we'll blow up if someone changes a policy from tracked to untracked (or is that handled elsewhere)?  That feels pretty mainline to me.

You could handle that in the manager by simulating a delete than a re-add.  We do that in the endpoint manager to support some awkward mutations.

go/felix/proto/felixbackend.proto
```diff
@@ -145,8 +145,9 @@ message PolicyID {
 }
 
 message Policy {
-  repeated Rule inbound_rules = 1;
-  repeated Rule outbound_rules = 2;
+  bool untracked = 1;
```
fasaxc: As a general rule, (for back-compat) you shouldn't change the field indexes of protobuf fields.  Even if this new field is written first for clarity, you should give it index 3.

python/calico/felix/endpoint.py
```diff
@@ -887,16 +893,26 @@ def _apply_endpoint_update(self):
...
             tiers = pending_endpoint.get("tiers", [])
             tier_dict = OrderedDict()
+            raw_tier_dict = OrderedDict()
             for tier in tiers or []:
                 pols = []
+                raw_pols = []
                 for pol in tier["policies"] or []:
                     pol_id = TieredPolicyId(tier["name"], pol)
-                    pols.append(pol_id)
+                    if is_untracked(pol_id):
```
fasaxc: I don't think using a global here is going to fly; the global might not have been updated when this code runs.  I think you'll have to plumb through the untracked-ness of the policies in the protobuf message.  I'd do that by adding an "untracked_tiers" field to the HostEndpoint object, then plumbing it through.

go/felix/proto/felixbackend.proto
fasaxc: Also, I don't see any (golang) code to set this flag being updated.

python/calico/felix/endpoint.py
```diff
@@ -887,16 +893,26 @@ def _apply_endpoint_update(self):
...
+                        raw_pols.append(pol_id)
+                    else:
+                        pols.append(pol_id)
                 tier_dict[tier["name"]] = pols
+                raw_tier_dict[tier["name"]] = raw_pols
```
fasaxc: This might not work as expected.  Felix's normal rule is that a packet skips a tier if none of the policies match the endpoint, and there's a default drop at the end of the tier.

I think this might create a tier with no policies in the raw table for each tier in the filter table.  I think that tier will then drop by default.

neiljerram:
@fasaxc Thanks for your review!  I'll work through the points one-by-one...

python/calico/felix/profilerules.py
neiljerram: Yes, but I did try hard not to hide this: https://github.com/neiljerram/felix/blob/untracked/python/calico/felix/profilerules.py#L135

Anyway, I was thinking of it as an initial simplification that we could later remove, and it sounds like it might disappear as I address some of your other points.

python/calico/felix/endpoint.py
neiljerram: That might be what is causing existing FVs to fail - let me look further...

neiljerram: Ah no, I believe I've avoided that problem, already, by adding a 'default_drop=True' parameter to host_endpoint_updates in fiptgenerator.py, and plumbing that through from LocalEndpoint:_update_chains (https://github.com/neiljerram/felix/blob/untracked/python/calico/felix/endpoint.py#L948).

Do you agree, or do you still see a hole here?

python/calico/felix/profilerules.py
neiljerram: I did also like and consider encoding the rawness in TieredPolicyID - assuming that that is also what you mean here - but it seemed to me that there were places that construct a TieredPolicyID that do not have access to the policy content, and hence don't know if it's untracked or not.  Let me check...

neiljerram: Yes, for example, at https://github.com/neiljerram/felix/blob/untracked/python/calico/felix/endpoint.py#L901...

Or have I misunderstood your idea here?

go/felix/proto/felixbackend.proto
fasaxc: Adding the untracked flag here ads it to the API between the go and python parts but I don't see any changes on the golang side of the API to actually fill in the flag with the correct value.  I think you need to set the flag around event_buffer.go:209:
```
		Policy: &proto.Policy{
			InboundRules:  parsedRulesToProtoRules(rules.InboundRules),
			OutboundRules: parsedRulesToProtoRules(rules.OutboundRules),
		},
```
You'll also need to add the flag to the datamodel in libcalico-go so that it's available here.  Felix uses the backend datamodel, so you'll need to add it to the Policy struct in `lib/backend/model/policy.go`:

```
type Policy struct {
	Order         *float64 `json:"order,omitempty" validate:"omitempty"`
	InboundRules  []Rule   `json:"inbound_rules,omitempty" validate:"omitempty,dive"`
	OutboundRules []Rule   `json:"outbound_rules,omitempty" validate:"omitempty,dive"`
	Selector      string   `json:"selector" validate:"selector"`
}
```

fasaxc: There's also another place you'll need to update: we convert Policy objects into ParsedRule objects (which contain rules modified to use ipsets instead of selectors/tags) so you'll need to pass the flag via those objects.

python/calico/felix/profilerules.py
neiljerram: Handling this in the rebased 1.4.x version of this PR.

python/calico/felix/profilerules.py
neiljerram: Handling this in the rebased 1.4.x version of this PR.

neiljerram:
Replaced by https://github.com/projectcalico/felix/pull/1170

PR #1177, opened 2016-11-22
Make release

neiljerram:


fasaxc:
A few thoughts:
- OK, I can live with a bit of shell.
- Please validate the incoming version is of the expected format and fail if not.  I've noticed that being a problem on almost every release.  I prefer using semver format for the canonical version number (i.e. the one in the git tag) since it's very well specified (and the Python code might be retired soon so it'd be odd to use PEP440 for our standard).
- I'm worried about forgetting to push the packaging branch.  Not sure what the right solution is.  We could run `git fetch|pull` and `git push`commands as part of the process but that makes it non-local.  WDYT?
- When doing a non-release build, please can you put in a fragment of the git hash into the package names?  git is non-linear so you can't assume that your 1.0.0-post.9 is the same as mine.
- I think you found a bug when giving me a demo so back to you to fix up.

neiljerram:
@fasaxc I think that's everything now - please could you take another look?

tomdee:
calico/Felix container image is a release artifact but it doesn't seem to be mentioned. It's how calico/node consumes Felix releases so it needs to be there.

neiljerram:
Thanks @tomdee .  I'm happy to extend 'make release' so that it builds and tags calico/felix.  What should be the instruction for publishing that image?  Or is there some other automated process (e.g. semaphore) that already covers that?

utils/lib.sh
```diff
@@ -1,104 +1,119 @@
...
+    git describe --tags --abbrev=0
+}
 
-    # Generate RPM version and release.
-    IFS=- read version qualifier <<< ${nextrel}
-    if test "${qualifier}"; then
-	rpmver=${version}
-	rpmrel=0.1.${qualifier}
+# Autogenerate PEP 440 version based on current Git state.
+function git_auto_version {
```
caseydavenport: Just glancing at this, so could be totally wrong.

The other make release targets use `git describe --tags --dirty` to get something very similar to this.  Any reason we can't just use that?

caseydavenport: Ok.  Just curious as to why we went for that approach in the other repos but it isn't good enough here.  @tomdee can probably answer.

neiljerram: Mostly I think just because we still have some Python code here (in Felix) for the moment - so still want a PEP 440 version.

There is another small reason, though: if you're working iteratively with uncommitted code, `git describe` on its own doesn't give you a monotonically increasing version.

PR #1178, opened 2016-11-22
Use new CalicoAPIConfig to specify connection info

robbrockbank:


fasaxc:
I think this was broken by the fact that I broke the Syncer on libcalico-go master.  Put up my own PR to try this patch with the Syncer fix: #1179

fasaxc:
Changes LGTM.  I'll merge #1179 if that passes CI.

PR #1182, opened 2016-11-24
Update README.md

lxpollitt:
Add contributor/user documentation header notice and tightened content in a few places to be more contributor focussed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,33 +4,19 @@
 [![IRC Channel](https://img.shields.io/badge/irc-%23calico-blue.svg)](https://kiwiirc.com/client/irc.freenode.net/#calico)
 # Project Calico
 
-Project Calico provides
-
-- A simple, pure layer 3 networking approach with no overlays for networking
-  "workloads" such as VMs and containers.
-- A distributed firewall implementing rich and flexible network policy,
-  imposed at ingress/egress to each workload.
-
-For more information see [the Project Calico website](http://www.projectcalico.org/learn/).
+<blockquote>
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -4,44 +4,27 @@
...
 
-Project Calico provides
-
-- A simple, pure layer 3 networking approach with no overlays for networking
-  "workloads" such as VMs and containers.
-- A distributed firewall implementing rich and flexible network policy,
-  imposed at ingress/egress to each workload.
-
-For more information see [the Project Calico website](http://www.projectcalico.org/learn/).
+<blockquote>
```
caseydavenport: A bit weird to use these tags in a markdown file, I think.

How about this:

```markdown
> Note that the documentation in this repo is targeted at Calico contributors.
> ## Documentation for Calico can be found at [docs.projectcalico.org](http://docs.projectcalico.org)
```

Formats like this:

> Note that the documentation in this repo is targeted at Calico contributors.
> ## Documentation for Calico can be found at [docs.projectcalico.org](docs.projectcalico.org)

lxpollitt: HTML fragments in markdown is an officially supported thing.  The HTML fragment proposed gives a better formatted notice for our purposes I think.

caseydavenport: Certainly supported and works.  Happy to merge if that's the way you'd prefer, but I much prefer the formatting of the way I suggested (both the markdown and the final rendered result).

Doesn't really matter though.

caseydavenport:
LGTM.

PR #1188, opened 2016-11-28
Update for current logrus API (fixes #1162)

### LINKED ISSUE #1162: Latest Logrus changes method
When updating felix to use pinned libcalico-go, we inadvertently updated logrus which changed functionality of a method felix uses when logging.

I lost the build output to prove it, but I think it can be reproed by `glide up` then building the binary. UT's will display an error.

neiljerram:


go/glide.lock
```diff
@@ -174,7 +174,6 @@ imports:
 - name: github.com/satori/go.uuid
   version: b061729afc07e77a8aa4fad0a2fd840958f1942a
 - name: github.com/Sirupsen/logrus
-  version: 4b6ea7319e214d98c938f12692336f7ca9348d6b
```
fasaxc: I think you should run `make update-vendor` to regenerate this file with the new commit hash.  I'm not sure it's valid as it stands.

go/glide.yaml
```diff
@@ -25,7 +25,6 @@ import:
   subpackages:
   - lib
 - package: github.com/Sirupsen/logrus
-  version: 4b6ea7319e214d98c938f12692336f7ca9348d6b
```
fasaxc: I think logrus has semver tags now, should we pin to the current version (or patch updates thereof)?

neiljerram:
retest this please

neiljerram:
retest this please

PR #1189, opened 2016-11-29
use calico's python-etcd fork, may fix #1092 for 1.4 branch.

### LINKED ISSUE #1092: Fix incompatibility with etcd3 server v2 api
Calico uses a Python etcd client.  It is currently incompatible with the etcd3 server.  The Etcd3 server correctly enforces best security practices and refuses connections from TLSv1. 

Root cause issue is documented here:
https://github.com/jplana/python-etcd/pull/196

and
https://github.com/projectcalico/calico-containers/issues/1107

vixns:
Allow felix 1.4.x to connect to etcd+tls, not fully tested yet.

### DIFF
--- a/felix_requirements.txt
+++ b/felix_requirements.txt
@@ -1,7 +1,6 @@
 gevent
 greenlet
 netaddr
-python-etcd>=0.4.1
 posix-spawn>=0.2.post6
 datrie>=0.7
 ijson>=2.2
--- a/pyi/Dockerfile
+++ b/pyi/Dockerfile
@@ -16,7 +16,7 @@ RUN ./build-prereqs.sh
 ENV LD_LIBRARY_PATH=/usr/local/lib
 RUN yum install -y libffi-devel libffi yajl
 ADD felix_requirements.txt felix_requirements.txt
-RUN pip install -U -r felix_requirements.txt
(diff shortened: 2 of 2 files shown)


marvin-tigera:

felix_requirements.txt
```diff
@@ -1,7 +1,6 @@
 gevent
 greenlet
 netaddr
-python-etcd>=0.4.1
```
fasaxc: I think this requirement should still be here.

pyi/Dockerfile
```diff
@@ -16,7 +16,7 @@ RUN ./build-prereqs.sh
 ENV LD_LIBRARY_PATH=/usr/local/lib
 RUN yum install -y libffi-devel libffi yajl
 ADD felix_requirements.txt felix_requirements.txt
-RUN pip install -U -r felix_requirements.txt
```
fasaxc: If you split this into two `pip` commands and run the `python-etcd` one first then you should get the github version of python-etcd.

Should we pin this to a particular commit for repeatability?

Why do you add -e here?  That installs python-etcd as a development egg, which shouldn't be needed.

marvin-tigera:

vixns:
I've tried two pip commands as suggested, but still some warnings.
Anyway, I've upgraded to 2.0, TLS support just works, thanks for the advice.

I no longer use the 1.4 branch, I'm closing this PR.

PR #1198, opened 2016-12-01
Add 5 minutes to initial usage reporting delay.

### LINKED ISSUE #1194: Increase initial usage reporting delay to 5 minutes
Will prevent spam in case of cyclic restart.

fasaxc:
Prevents spamming server under cyclic-restart scenarios.

Testing performed:

- Add UT for delay calculation.
- Manual test that sensible delay values printed to log.

Fixes #1194

### DIFF
--- a/go/felix/usagerep/usagerep.go
+++ b/go/felix/usagerep/usagerep.go
@@ -36,6 +36,7 @@ func PeriodicallyReportUsage(interval time.Duration, hostname, clusterGUID, clus
 	log.Info("Usage reporting thread started, waiting for size estimate")
 	stats := <-statsUpdateC
 	log.WithField("stats", stats).Info("Initial stats read")
+	// To avoid thundering herd, inject some startup jitter.
 	initialDelay := calculateInitialDelay(stats.NumHosts)
 	log.WithField("delay", initialDelay).Info("Waiting before first check-in")
 	time.Sleep(initialDelay)
@@ -51,10 +52,12 @@ func PeriodicallyReportUsage(interval time.Duration, hostname, clusterGUID, clus
 }
 
 func calculateInitialDelay(numHosts int) time.Duration {
-	// To avoid thundering herd, inject some startup jitter.
+	// Clamp numHosts so that we don't pass anything out-of-range to rand.Intn().
 	if numHosts <= 0 {
-		// Intn requires a positive argument so clamp numHosts to be >= 1.
 		numHosts = 1
+	}
+	if numHosts > 10000 {
+		numHosts = 10000
 	}
 	initialJitter := time.Duration(rand.Intn(numHosts*1000)) * time.Millisecond
 	// To avoid spamming the server if we're in a cyclic restart, delay the first report by
--- a/go/felix/usagerep/usagerep.go
+++ b/go/felix/usagerep/usagerep.go
@@ -36,13 +36,9 @@ func PeriodicallyReportUsage(interval time.Duration, hostname, clusterGUID, clus
 	log.Info("Usage reporting thread started, waiting for size estimate")
 	stats := <-statsUpdateC
 	log.WithField("stats", stats).Info("Initial stats read")
-	if stats.NumHosts > 25 {
-		// Avoid thundering herd by adding jitter to startup for a large
-		// cluster.
-		preJitter := time.Duration(rand.Intn(stats.NumHosts)) * time.Second
-		log.WithField("delay", preJitter).Info("Waiting before first check-in")
-		time.Sleep(preJitter)
-	}
+	initialDelay := calculateInitialDelay(stats.NumHosts)
+	log.WithField("delay", initialDelay).Info("Waiting before first check-in")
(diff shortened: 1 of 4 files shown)


go/felix/usagerep/usagerep_test.go
```diff
@@ -60,4 +61,27 @@ var _ = Describe("Usagerep", func() {
 		Expect(q.Get("guid")).To(Equal("baddecaf"))
 		Expect(q.Get("cluster_type")).To(Equal("unknown"))
 	})
+	It("should delay at least 5 minutes", func() {
+		Expect(calculateInitialDelay(0)).To(BeNumerically(">=", 5*time.Minute))
+		Expect(calculateInitialDelay(1)).To(BeNumerically(">=", 5*time.Minute))
+		Expect(calculateInitialDelay(1000)).To(BeNumerically(">=", 5*time.Minute))
+	})
+	It("should have a random component", func() {
+		Expect(calculateInitialDelay(1000000)).ToNot(Equal(calculateInitialDelay(1000000)))
```
neiljerram: This test - with 1 million hosts - seems unrealistic.  Is the number so large because you're trying to reduce the likelihood of a possible collision?  If so, I think a better test would be to call calculateInitialDelay(50) 10 times, and expect that at least one of the values is different.

go/felix/usagerep/usagerep_suite_test.go
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package usagerep_test
+package usagerep
```
neiljerram: Why this change?

go/felix/usagerep/usagerep.go
```diff
@@ -54,6 +50,19 @@ func PeriodicallyReportUsage(interval time.Duration, hostname, clusterGUID, clus
 	}
 }
 
+func calculateInitialDelay(numHosts int) time.Duration {
+	// To avoid thundering herd, inject some startup jitter.
+	if numHosts <= 0 {
+		// Intn requires a positive argument so clamp numHosts to be >= 1.
+		numHosts = 1
+	}
+	initialJitter := time.Duration(rand.Intn(numHosts*1000)) * time.Millisecond
```
neiljerram: Presumably the range of a Go integer is large enough that we don't need to worry about numHosts*1000 overflowing?

neiljerram: Also I'm wondering if a paren is misplaced here, and it should be:

    initialJitter := time.Duration(rand.Intn(numHosts*1000) * time.Millisecond)

(Because it appears that the other `time.` methods around here don't handle plain integers.)

go/felix/usagerep/usagerep_test.go
```diff
@@ -60,4 +61,27 @@ var _ = Describe("Usagerep", func() {
...
+	})
+	It("should have an average close to expected value", func() {
+		var total time.Duration
+		for i := int64(0); i < 1000000; i++ {
+			total += calculateInitialDelay(60)
+			if i > 100 {
+				average := time.Duration(int64(total) / (i + 1))
+				if average > (5*time.Minute+20*time.Second) &&
```
neiljerram: I think it's worth commenting here that the average jitter is 0.5s per host, and so 30s for 60 hosts - and hence the test for the average being between 20s and 40s.

neiljerram: This is a very nice test, though!

go/felix/usagerep/usagerep.go
fasaxc: Think it's OK, should be good enough for 2m hosts.

go/felix/usagerep/usagerep_test.go
fasaxc: Done

go/felix/usagerep/usagerep.go
fasaxc: It's correct: `time.Duration` is a type, not a method.  `time.Duration(...)` is a cast, which is required by go's (rather minimal) promotion rules.  `time.Millisecond` is a `time.Duration` constant.

go/felix/usagerep/usagerep_test.go
```diff
@@ -60,4 +61,42 @@ var _ = Describe("Usagerep", func() {
 		Expect(q.Get("guid")).To(Equal("baddecaf"))
 		Expect(q.Get("cluster_type")).To(Equal("unknown"))
 	})
+	It("should delay at least 5 minutes", func() {
+		Expect(calculateInitialDelay(0)).To(BeNumerically(">=", 5*time.Minute))
+		Expect(calculateInitialDelay(1)).To(BeNumerically(">=", 5*time.Minute))
+		Expect(calculateInitialDelay(1000)).To(BeNumerically(">=", 5*time.Minute))
+	})
+	It("should delay at most 10000 seconds", func() {
```
neiljerram: This should really say '5 minutes + 10000 seconds'.

PR #1199, opened 2016-12-01
Fix RPM changelog version '2.0.0' - should be '2.0.0-rc2'

neiljerram:
Caused by accidental overwriting of the ${version} variable.


neiljerram:
retest this please

PR #1203, opened 2016-12-01
Release process tidy-ups

### LINKED ISSUE #1193: Clean up release process dirt
Umbrella issue to track niggles found during release process:

- [x] Add reminder to rev libcalico-go
- [x] Add more artifact details
- [x] Recommend adding a description line to the top of the github release note (else the first bullet becomes the subject line of the email).
- [x] Move dockerhub/quay push from semaphore to release process
- [ ] ~~Sync our version number style with other projects.  "2.0.0" -> "v2.0.0" ?~~

fasaxc:
- More precise list of artifacts in RELEASING.md.
- Include instructions to push to Dockerhub/Quay.

Fixes #1193

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -10,8 +10,9 @@ through creating and distributing the following artifacts:
   Github release.
 - Docker container images: `calico/felix:$VERSION` and 
   `quay.io/calico/felix:$VERSION` containing the Felix binaries.  These
-  are ready to push to Dockerhub and Quay.  They also form the input
-  to the downstream `calico/node` build process.
+  are ready to push to Dockerhub and Quay.  They primarily form the input
+  to the downstream `calico/node` build process but they could also
+  be used to run Felix as a stand-alone container.
   
 As a second step, running `make deb rpm` after `make release`, produces
 debs and RPMs for the release which can be uploaded to our PPAs and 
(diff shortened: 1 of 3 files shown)


RELEASING.md
```diff
@@ -1,8 +1,22 @@
...
+through creating and distributing the following artifacts:
+
+- A git tag for the release, annotated with the release note; ready to 
+  push to Github.
+- File `dist/calico-felix-$VERSION.tgz`; the "PyInstaller bundle" 
+  containing the Felix binaries.  We attach that file to the 
+  Github release.
+- Docker container images: `calico/felix:$VERSION` and 
+  `quay.io/calico/felix:$VERSION` containing the Felix binaries.  These
+  are ready to push to Dockerhub and Quay.  They also form the input
```
neiljerram: I understood from what you said yesterday that their _only_ purpose was as input to the downstream process.  If that's right, I'd prefer that to be more clearly stated here.  Or if they have other purposes, what are those?

(For example, if they have no other purpose, why should we not consider simplifying our processes by using the bundle as input for calico/node instead?  My point is that we should do that right now - but that someone might consider that in future, and so would benefit from as full a description as possible of the artifacts' purposes.)

RELEASING.md
```diff
@@ -24,9 +38,19 @@ In a nutshell:
 
 So, to make a Felix release:
 
+- Consider whether you should update the libcalico-go pin in glide.yaml.
+  If you do so, you should run `make update-vendor` to update the 
+  `glide.lock` file.  Be wary of any additional libraries that get 
+  revved if they aren't being pulled in by the libcalico-go update. At 
+  this late stage, it's safer to only update commit IDs that you're
+  explictly expecting (i.e. undo any changes that `make update-vendor`
```
neiljerram: s/explictly/explicitly/

RELEASING.md
```diff
@@ -24,9 +38,19 @@ In a nutshell:
...
+  makes that you weren't expecting).
+
 - Run `make release VERSION=<new version>` and follow the instructions.  This
   creates the annotated release tag, builds the release artifacts, and tells
-  you what else you need to do to publish those.
+  you what else you need to do to publish those.  The release script
+  expects a version number of the form "2.0.0", with optional suffixes
+  such as "-beta.1-rc3".
```
neiljerram: If it's OK with you, I'd rather not recommend extra '.'s, so would say `-beta1-rc3` here.

PR #1204, opened 2016-12-02
Felix 2.0.0-rc2

neiljerram:


rpm/felix.spec
```diff
@@ -3,7 +3,7 @@
 Name:           felix
 Summary:        Project Calico virtual networking for cloud data centers
 Version:        2.0.0
-Release:        0.3.beta3%{?dist}
+Release:        0.1.rc2%{?dist}
```
fasaxc: This has gone backwards from `0.3.beta3` so I think the new package will be a downgrade.

neiljerram: Indeed.  In our own felix-2.0-testing RPM repo, I had to work around that by deleting the 0.3.beta3 RPMs.

How much do you think this matters?  I guess it depends if people outside configured the felix-2.0-testing RPM repo and will upgrade (as opposed to reinstall from scratch) when we say that 2.0.0 is GA.

I guess we could change the hardcoded '0.1.' to be '0.4.' instead, but that doesn't seem very elegant.

PR #1205, opened 2016-12-02
Add a make patch-script target.

fasaxc:
Outputs a script that patches the calico-felix executable on a test system.

### DIFF
--- a/utils/make-patch-script.sh
+++ b/utils/make-patch-script.sh
@@ -1,0 +1,21 @@
+#!/bin/bash
+
+set -e
+
+url=$1
+echo "URL: $url"
+sha_sum=$(cd bin; sha256sum calico-felix)
+
+echo
+echo "Run the following script on the target server to patch the"
+echo "/opt/calico-felix/calico-felix binary:"
+echo
+echo "curl -o calico-felix $url &&"
+echo "    if echo '$sha_sum' | sha256sum --status --check; then \\"
+echo "      sudo stop calico-felix || sudo systemctl stop calico-felix; "
+echo "      sudo cp calico-felix /opt/calico-felix/calico-felix && \\"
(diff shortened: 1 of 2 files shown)


utils/make-patch-script.sh
```diff
@@ -0,0 +1,21 @@
...
+url=$1
+echo "URL: $url"
+sha_sum=$(cd bin; sha256sum calico-felix)
+
+echo
+echo "Run the following script on the target server to patch the"
+echo "/opt/calico-felix/calico-felix binary:"
+echo
+echo "curl -o calico-felix $url &&"
+echo "    if echo '$sha_sum' | sha256sum --status --check; then \\"
```
neiljerram: Should the `sha256sum` invocation include `calico-felix` somewhere?  Otherwise how does it know what file to check?

utils/make-patch-script.sh
```diff
@@ -0,0 +1,21 @@
...
+echo "      sudo stop calico-felix || sudo systemctl stop calico-felix; "
+echo "      sudo cp calico-felix /opt/calico-felix/calico-felix && \\"
+echo "      sudo chmod +x /opt/calico-felix/calico-felix && \\"
+echo "      sudo start calico-felix || sudo systemctl start calico-felix;"
+echo "    else"
+echo "      echo 'Incorrect hash, file may have been tampered with!'"
+echo "    fi"
```
neiljerram: For next time, I recommend `cat <<EOF` instead of multiple `echo`s.  But this is fine.

utils/make-patch-script.sh
fasaxc: `$sha_sum` has the hash and file name in it.

neiljerram: Ah, OK then - thanks.

PR #1210, opened 2016-12-06
Make the iptables log-prefix configurable. Defaults to current contents.

wmuizelaar:


marvin-tigera:

python/calico/felix/config.py
```diff
@@ -479,6 +484,12 @@ def _validate_cfg(self, final=True):
                 self.parameters["DefaultEndpointToHostAction"]
             )
 
+        if len(self.LOG_PREFIX) < 1:
+            raise ConfigException(
```
fasaxc: I'd err of the side of not raising an exception here.  If the user _wants_ to set an empty log prefix, why not let them?

I'd also prevent defaulting over raising since, if you set this config in etcd, _every_ felix will validate it here and crash if the config is bad.

wmuizelaar: Good points. Basically, that'll mean just removing those 2 lines. Will fix.

wmuizelaar:
@fasaxc will work on the go version as well!

fasaxc:
ok to test

wmuizelaar:
Yes, added a test!

wmuizelaar:
looks like a test failed - but I can't see the details. Anything still needs to be changed?

fasaxc:
@wmuizelaar We use semaphore to run the UTs, which covers the one you added.  The `Tigera build debs and run fv test` job also runs additional automated system tests; which run on our Jenkins server.  Sorry the results aren't public, that's because exposing jenkins is a bit of a nightmare!

In this case, the test looks like an intermittent failure rather than a bug with your code.  I'm going to rerun the tests just to make sure...

fasaxc:
retest this please

wmuizelaar:
@fasaxc heh, I understand the jenkins-issue ;-) 

Please let me know if things need to be changed!

PR #1211, opened 2016-12-06
Make rule generation tolerate missing IP version for ICMP.

fasaxc:
Previously, ICMP rules without an IP version would cause an assert when being rendered on the "wrong" IP version.  Now they are ignored.

Also, reinstate the UTs for rule generation and fix up in light of golang port.

Fixes #1206

PR #1213, opened 2016-12-06
Skip rules that have CIDRs for different IP versions

### LINKED ISSUE #1212: Rules with IPv4 CIDRs and no IP version render unintuitively in IPv6
If you have a rule with `"src_net": "10.0.0.0/16"` but no IP version field then, when rendering the IPv6 version of the rule, we skip over the IPv4 CIDR resulting in a less-constrained rule.  If the CIDR is the only constraint, we can end up with a blanket allow/deny rule, which is unlikely to be what the user is expecting.

Workaround is to always specify the "ip_version" on the rule in a mixed IPv4/v6 environment.

fasaxc:
Previously, we skipped the CIDR only, but that leads to a very non-intuitive behaviour for allow rules, which can become vastly more broad than expected, as described in #1212 

New behaviour assumes that IPv4 packets should never match an IPv6 CIDR and vice-versa.  Hence, if the user says "packet should match 10.0.0.0/16" then they're implicitly saying "this rule shouldn't match any IPv6 packets".

This PR is based on #1211 so it should be reviewed after that one is merged.

Fixes #1212

### DIFF
--- a/python/calico/felix/plugins/fiptgenerator.py
+++ b/python/calico/felix/plugins/fiptgenerator.py
@@ -982,10 +982,19 @@ class FelixIptablesGenerator(FelixPlugin):
 
                 # Network (CIDR).
                 net_key = neg_pfx + dirn + "_net"
-                if net_key in rule and rule[net_key] is not None:
+                if rule.get(net_key):
                     ip_or_cidr = rule[net_key]
                     if (":" in ip_or_cidr) == (ip_version == 6):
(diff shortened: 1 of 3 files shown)


fasaxc:
Tentatively assigning to 2.0.0.  I feel like this is counter-intuitive enough to be called a bug and the impact is fairly severe.

lwr20:
Retested OK.

PR #1214, opened 2016-12-06
If an interface is down, make sure we remove its routes.

### LINKED ISSUE #1208: Neutron port updates for ports that are status down, are not reflected on the hypervisor by Felix.
Hi.

Using calico 1.4 with openstack Mitaka.

We can have multiple IPs on the same port, a VIP, in addition to the *main* instance IP.
If an instance is shutdown, or fails for some reason, we need to move the VIP to another instance.  We can update the Neutron port configs, remove the IP from the first port, and add it to the second port.  Everything works for the new instance.
However, the hypervisor for the first instance [ which may or may not be a different hypervisor used for the second instance ] still has a local route, for the VIP, on the original tap interface for the first instance.
This can cause problems, especially if an instance on the first hypervisor needs to communicate with the second instance [ which may be on a new hypervisor ], which fails, as the hypervisor essentially black holes the traffic by routing it to the tap interface of the first instance [ which is down ].

Essentially, we can end up with a situation like this on two hypervisors:

```
hypervisor1 # ip r s |grep tape3957b74-5a
10.1.2.188 dev tape3957b74-5a  scope link 
10.1.2.195 dev tape3957b74-5a  scope link 
```

```
hypervisor2 # ip r s |grep tap677220a2-ad
10.1.2.194 dev tap677220a2-ad  scope link 
10.1.2.195 dev tap677220a2-ad  scope link 
```

Hopefully this makes sense - please let me know if you need to clarify further :)

This might all be by design, but the above scenario is problematic for us.

It might be nice to have a config option available in felix to withdraw the routes for interfaces that are status down, or something like that.

If we start the first instance again, the route for the VIP is removed from the tap interface on the hypervisor, but we might not be able or want to do that for various reasons.
I presume deleting the instances would also remove the old route tap, but again that might not be desirable.

Thanks in advance for any help and thoughts on this.

Cheers,
Just

fasaxc:
We were assuming the kernel would do this for us, but in some cases, an interface can be oper down but still have routes.

Make sure that the routes get cleaned up in that case by calling `_deconfigure_interface()`.

Fixes #1208

### DIFF
--- a/python/calico/felix/endpoint.py
+++ b/python/calico/felix/endpoint.py
@@ -704,7 +704,7 @@ class LocalEndpoint(RefCountedActor):
         if not self._device_in_sync and self._iface_name:
             # Try to update the device configuration.  If successful, will set
             # the _device_in_sync flag.
-            if self._admin_up:
+            if self._admin_up and self._device_is_up:
                 # Endpoint is supposed to be live, try to configure it.
                 _log.debug("Device is out-of-sync, trying to configure it")
                 self._configure_interface()
@@ -1012,10 +1012,6 @@ class WorkloadEndpoint(LocalEndpoint):
         """
         Applies sysctls and routes to the interface.
         """
-        if not self._device_is_up:
-            _log.debug("Device is known to be down, skipping attempt to "
-                       "configure it.")
-            return
         try:
             if self.ip_type == IPV4:
                 devices.configure_interface_ipv4(self._iface_name)
--- a/python/calico/felix/test/test_endpoint.py
+++ b/python/calico/felix/test/test_endpoint.py
@@ -1012,14 +1012,16 @@ class TestWorkloadEndpoint(BaseTestCase):
                 mock.patch('calico.felix.devices.set_routes'),
                 mock.patch('calico.felix.devices.configure_interface_ipv4'),
                 mock.patch('calico.felix.devices.interface_up'),
-        ) as [m_set_routes, m_conf, m_iface_up]:
+                mock.patch('calico.felix.devices.interface_exists'),
+        ) as [m_set_routes, m_conf, m_iface_up, m_iface_exists]:
                 m_iface_up.return_value = False
                 local_ep.on_endpoint_update(data, async=True)
                 self.step_actor(local_ep)
                 self.assertEqual(local_ep._mac, data['mac'])
                 self.assertFalse(m_conf.called)
-                self.assertFalse(m_set_routes.called)
-                self.assertFalse(local_ep._device_in_sync)
+                self.assertTrue(local_ep._device_in_sync)
+                # Interface is down so we should explicitly remove the routes.
+                m_set_routes.assert_called_once_with(ip_type, set(), iface, None)
 
         # Now pretend to get an interface update - does all the same work.
         with mock.patch('calico.felix.devices.set_routes') as m_set_routes:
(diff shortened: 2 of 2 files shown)


fasaxc:
I've added UT but I'd like this to get more testing before merging in case it breaks another case.  Main concern is that we're now calling `set_routes()` when we know the interface is down.  I believe it's robust to the interface being down but we used to avoid that.

fasaxc:
Actually, looks like we were always calling set_routes on the delete path and we catch the relevant exceptions.  Maybe this could go in to 2.0.0.

caseydavenport:
Does this need to be in v2.0.0? It's not labeled as such.

fasaxc:
@caseydavenport We're just debating that, we think it's fairly low risk but it hasn't passed FV yet

neiljerram:
@fasaxc 1 FV failure here, caused by these error logs during TestManySecurityGroups:

=== LOG 2016-12-06 19:09:53 ERROR [pid 4920] ===
calico.felix.ipsets 467: Failed to check if ipset exists
File "site-packages/calico/felix/ipsets.py", line 461, in exists
File "site-packages/calico/felix/futils.py", line 355, in check_call
FailedSystemCall: Failed system call (retcode : 1, args : ('ipset', 'list', u'felix-4tt:o4JENdDneXDaiVNZ52-oI'))
stdout  :
stderr  :
input  : None


=== LOG 2016-12-06 19:09:53 ERROR [pid 4920] ===
calico.felix.ipsets 467: Failed to check if ipset exists
File "site-packages/calico/felix/ipsets.py", line 461, in exists
File "site-packages/calico/felix/futils.py", line 355, in check_call
FailedSystemCall: Failed system call (retcode : 1, args : ('ipset', 'list', u'felix-4tt:o4JENdDneXDaiVNZ52-oI'))
stdout  :
stderr  :
input  : None


=== LOG 2016-12-06 19:09:53 ERROR [pid 4920] ===
calico.felix.actor 395: _finish_msg_batch failed.
File "site-packages/calico/felix/actor.py", line 382, in _step
File "site-packages/calico/felix/ipsets.py", line 425, in _finish_msg_batch
File "site-packages/calico/felix/ipsets.py", line 345, in _finish_msg_batch
File "site-packages/calico/felix/ipsets.py", line 384, in _sync_to_ipset
File "site-packages/calico/felix/ipsets.py", line 511, in replace_members
File "site-packages/calico/felix/ipsets.py", line 461, in exists
File "site-packages/calico/felix/futils.py", line 355, in check_call
FailedSystemCall: Failed system call (retcode : 1, args : ('ipset', 'list', u'felix-4tt:o4JENdDneXDaiVNZ52-oI'))
stdout  :
stderr  :
input  : None


=== LOG 2016-12-06 19:09:53 ERROR [pid 4920] ===
calico.felix.actor 395: _finish_msg_batch failed.
File "site-packages/calico/felix/actor.py", line 382, in _step
File "site-packages/calico/felix/ipsets.py", line 425, in _finish_msg_batch
File "site-packages/calico/felix/ipsets.py", line 345, in _finish_msg_batch
File "site-packages/calico/felix/ipsets.py", line 384, in _sync_to_ipset
File "site-packages/calico/felix/ipsets.py", line 511, in replace_members
File "site-packages/calico/felix/ipsets.py", line 461, in exists
File "site-packages/calico/felix/futils.py", line 355, in check_call
FailedSystemCall: Failed system call (retcode : 1, args : ('ipset', 'list', u'felix-4tt:o4JENdDneXDaiVNZ52-oI'))
stdout  :
stderr  :
input  : None


=== LOG 2016-12-06 19:09:53 ERROR [pid 4909] ===
felix.go 476: Failed to read from front-end socket error=EOF

fasaxc:
retest this please

neiljerram:
Past Jenkins results show that we've hit this 'Failed to check if ipset exists' on a few past PR builds - including for a case (build 460 - #1149 - just a version number change) where the PR wasn't making any real change.

So I'm persuaded that it's just a glitch and nothing to do with this PR.

PR #1218, opened 2016-12-08
Fix nil pointer deref in policy sorter.

### LINKED ISSUE #1217: nil pointer dereference in policy_sorter.go
Found by @matthewdupre: (only) when debug level logging is enabled, we print out the policies and their orders:
```
	if log.GetLevel() >= log.DebugLevel {
		names := make([]string, len(tierInfo.OrderedPolicies))
		for ii, kv := range tierInfo.OrderedPolicies {
			names[ii] = fmt.Sprintf("%v(%v)",
				kv.Key.Name, *kv.Value.Order)
		}
		log.Infof("Before sorting policies: %v", names)
	}
```
Unfortunately, `kv.Value.Order` can (legitimately) be nil so this code can try to dereference a nil pointer.

fasaxc:
Fixes #1217.

### DIFF
--- a/go/felix/calc/policy_sorter_test.go
+++ b/go/felix/calc/policy_sorter_test.go
@@ -1,0 +1,40 @@
+// Copyright (c) 2016 Tigera, Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package calc_test
+
+import (
+	. "github.com/projectcalico/felix/go/felix/calc"
+
+	. "github.com/onsi/ginkgo/extensions/table"
+	. "github.com/onsi/gomega"
+	"github.com/projectcalico/libcalico-go/lib/backend/model"
+)
+
+var (
+	tenPointFive = 10.5
+)
+
+var _ = DescribeTable("PolKV should stringify correctly",
+	func(kv PolKV, expected string) {
+		Expect(kv.String()).To(Equal(expected))
+	},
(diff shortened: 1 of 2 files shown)


go/felix/calc/policy_sorter_test.go
```diff
@@ -0,0 +1,40 @@
...
+	. "github.com/onsi/ginkgo/extensions/table"
+	. "github.com/onsi/gomega"
+	"github.com/projectcalico/libcalico-go/lib/backend/model"
+)
+
+var (
+	tenPointFive = 10.5
+)
+
+var _ = FDescribeTable("PolKV should stringify correctly",
```
neiljerram: If I understand correctly, you need to remove the leading 'F' here (as it causes the 'Detected Programmatic Focus' status, which our CI thinks is a failure), and then retest.

fasaxc:
> I think Order is only a pointer so that we can use the nil value to indicate 'default', right? That seems a bit hacky; is there a better alternative?

This is the backend datamodel, which maps onto our JSON; using a pointer is the natural way for the parser library to represent the presence/absence of an optional field.  (In the json, we represent default by omitting the order from the JSON, which is pretty natural, I think.)

> The new `func (p PolKV) String()` string is OK for this context - but more generally would we not want a stringification function for policies that included more information than just the name and order?

If you just do `fmt.Printf("%v", aPolicy)` then you actually get quite a lot more detail (it'll print all the fields) so the defaults are fairly good for most tracing.  In this case, I wanted to limit the verbosity to just the fields I cared about.

neiljerram:
> > I think Order is only a pointer so that we can use the nil value to indicate 'default', right? That seems a bit hacky; is there a better alternative?

> This is the backend datamodel, which maps onto our JSON; using a pointer is the natural way for the parser library to represent the presence/absence of an optional field.

OK, so we're following the parser library here - fine.

> (In the json, we represent default by omitting the order from the JSON, which is pretty natural, I think.)

Definitely no issue there - my only query was with the code representation.

neiljerram:
> > The new `func (p PolKV) String() string` is OK for this context - but more generally would we not want a stringification function for policies that included more information than just the name and order?

> If you just do `fmt.Printf("%v", aPolicy)` then you actually get quite a lot more detail (it'll print all the fields) so the defaults are fairly good for most tracing. In this case, I wanted to limit the verbosity to just the fields I cared about.

OK - thanks.

PR #1220, opened 2016-12-08
Fixes for 1.4

neiljerram:


neiljerram:
@fasaxc WDYT?  I know that at least Logan would like the removal of the pycparser pin to be in our official 1.4 branch; and I'd like the floating IP fix in so that OpenStack CI can reference projectcalico/felix:1.4 instead of neiljerram/felix:fip-output-dnat-1.4-unpin-pycparser; and I believe jocado would like the interface down fix.

Another possible fix to add in here is #1189, but the raiser says they've now moved on to 2.0, so I'd rather not add that one in.

fasaxc:
SGTM.

PR #1222, opened 2016-12-08
Reduce length of usage reporting keys.

### LINKED ISSUE #1221: Reduce length of usage tracking fields
To minimise server storage impact.

fasaxc:
Quick fix to reduce server-size storage requirements.

Fixes #1221


go/felix/usagerep/usagerep.go
```diff
@@ -105,15 +105,14 @@ func calculateURL(hostname, clusterGUID, clusterType string, stats calc.StatsUpd
...
-		"git_revision":       {buildinfo.GitRevision},
-		"felix_type":         {"go"},
+		"hostname": {hostname},
+		"guid":     {clusterGUID},
+		"type":     {clusterType},
+		"size":     {fmt.Sprintf("%v", stats.NumHosts)},
+		"weps":     {fmt.Sprintf("%v", stats.NumWorkloadEndpoints)},
+		"heps":     {fmt.Sprintf("%v", stats.NumHostEndpoints)},
+		"version":  {buildinfo.GitVersion},
+		"rev":      {buildinfo.GitRevision},
```
neiljerram: OK, but I think it would be neater to trim a little further, 'hostname' -> 'host' and 'version' -> 'ver'.

caseydavenport:
What is the risk for putting this in v2.0 + what is the motivation?

go/felix/usagerep/usagerep.go
fasaxc: As discussed, those are expected by the server so we can't update them easily.

PR #1226, opened 2016-12-09
Fix slow Felix builds.

fasaxc:
- Add a cache directory for $GOPATH/pkg that is mounted as a volume into the build container.
- As part of building the build container image, rebuild the standard library. This ensures the standard library is built with CGO disabled so that it isn't considered stale by subsequent builds.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -281,11 +281,9 @@ bin/calico-felix: $(GO_FILES) \
 	    -v $${PWD}:/go/src/github.com/projectcalico/felix:rw \
 	    -v $${PWD}/.go-pkg-cache:/go/pkg/:rw \
 	    calico-build/golang \
-	    go build -i -o $@ -v $(LDFLAGS) "github.com/projectcalico/felix/go/felix"
-
-	@# Check that the executable is correctly statically linked.
-	@ldd bin/calico-felix | grep -q "not a dynamic executable" || \
-	     echo "Error: bin/calico-felix wasn't statically linked"
+	    sh -c 'go build -i -o $@ -v $(LDFLAGS) "github.com/projectcalico/felix/go/felix" && \
+               ( ldd bin/calico-felix | grep -q "not a dynamic executable" || \
+	             ( echo "Error: bin/calico-felix was not statically linked"; false ) )'
 
 # Build the pyinstaller bundle, which is an output artefact in its own right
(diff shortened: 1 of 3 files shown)


fasaxc:
retest this please

fasaxc:
retest this please

Makefile
```diff
@@ -270,12 +275,15 @@ bin/calico-felix: $(GO_FILES) \
 	# freshness checking for us.
 	$(MAKE) calico-build/golang
 	mkdir -p bin
+	mkdir -p .go-pkg-cache
```
neiljerram: Hm, not crazy about having another directory in the root, with naming that doesn't seem related to anything else.  Two thoughts:

1. Is this going to be reworked further anyway when Tom's build container comes along?

2. Why not map from $GOPATH/pkg outside the container, instead of from .go-pkg-cache?

Makefile
```diff
@@ -270,12 +275,15 @@ bin/calico-felix: $(GO_FILES) \
...
 	$(DOCKER_RUN_RM) \
 	    -v $${PWD}:/go/src/github.com/projectcalico/felix:rw \
+	    -v $${PWD}/.go-pkg-cache:/go/pkg/:rw \
 	    calico-build/golang \
-	    go build -o $@ $(LDFLAGS) "./go/felix/felix.go"
-	# Check that the executable is correctly statically linked.
-	ldd bin/calico-felix | grep -q "not a dynamic executable"
+	    sh -c 'go build -i -o $@ -v $(LDFLAGS) "github.com/projectcalico/felix/go/felix" && \
+               ( ldd bin/calico-felix | grep -q "not a dynamic executable" || \
+	             ( echo "Error: bin/calico-felix was not statically linked"; false ) )'
```
neiljerram: Do we really need this ldd test?  Is the static linking something that could easily regress and (otherwise) be missed?

fasaxc: Yes, I think so; we're not explicitly saying "do static linking", we're saying "disable this flag, which forces dynamic linking".  There is a flag for "do static linking", but that causes other problems so this seemed like a good pragmatic approach.

neiljerram: OK, thanks.

fasaxc:
retest this please

fasaxc:
Merging, since I think it's a big improvement even if we're going to revisit shortly.

PR #1244, opened 2016-12-19
Add support for policing forwarded traffic with host endpoint rules

fasaxc:
Includes the fix from #1242, which was in the same function.  That PR should be reviewed/merged first.

- [x] Double check that host endpoint chains use RETURN in compatible way
- [x] Add automated ST to cover

### DIFF
--- a/calico/felix/plugins/fiptgenerator.py
+++ b/calico/felix/plugins/fiptgenerator.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# Copyright (c) 2015-2016 Tigera, Inc. All rights reserved.
+# Copyright (c) 2015-2017 Tigera, Inc. All rights reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -488,7 +488,18 @@ class FelixIptablesGenerator(FelixPlugin):
                 (CHAIN_FORWARD, iface_match),
             ])
 
-        return forward_chain, set([CHAIN_FROM_ENDPOINT, CHAIN_TO_ENDPOINT])
+        # If we get to this point in the chain, the packet is not from or to
+        # a workload endpoint.  Hence, it must be being forwarded through
+        # this host, potentially via host endpoints that we want to police.
+        # Send the traffic to the host endpoint chains for processing.
+        # If there are no host endpoints configured, these chains are no-ops.
+        forward_chain.append("--append {chain} --jump {jump}".format(
+                chain=CHAIN_FORWARD, jump=CHAIN_FROM_IFACE))
+        forward_chain.append("--append {chain} --jump {jump}".format(
+                chain=CHAIN_FORWARD, jump=CHAIN_TO_IFACE))
+
+        return forward_chain, set([CHAIN_FROM_ENDPOINT, CHAIN_TO_ENDPOINT,
+                                   CHAIN_FROM_IFACE, CHAIN_TO_IFACE])
 
     def raw_prerouting_chain(self, ip_version):
         """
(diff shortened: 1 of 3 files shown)


calico/felix/plugins/fiptgenerator.py
```diff
@@ -488,7 +488,18 @@ def filter_forward_chain(self, ip_version):
                 (CHAIN_FORWARD, iface_match),
             ])
 
-        return forward_chain, set([CHAIN_FROM_ENDPOINT, CHAIN_TO_ENDPOINT])
+        # If we get to this point in the chain, the packet is not from or to
+        # a workload endpoint.  Hence, it must be being forwarded through
+        # this host, potentially via hst endpoints that we want to police.
+        # Send the traffic to the host endpoint chains for processing.
+        # If there are no host endpoints configured, these chains are no-ops.
+        forward_chain.append("--append {chain} --jump {jump}".format(
```
neiljerram: We normally --goto the FROM-IFACE and TO-IFACE chains, rather than --jumping to them, and I think they do use RETURN - so care is needed about whether that will behave as desired.

What's your plan for testing this?  It would be better to have a successful ST before merging.

calico/felix/plugins/fiptgenerator.py
```diff
@@ -488,7 +488,18 @@ def filter_forward_chain(self, ip_version):
                 (CHAIN_FORWARD, iface_match),
             ])
 
-        return forward_chain, set([CHAIN_FROM_ENDPOINT, CHAIN_TO_ENDPOINT])
+        # If we get to this point in the chain, the packet is not from or to
+        # a workload endpoint.  Hence, it must be being forwarded through
+        # this host, potentially via hst endpoints that we want to police.
```
neiljerram: Typo here: hst

fasaxc:
Looks like the FV run hit a random ipset glitch:
```
calico.felix.ipsets 780: Failed to update ipset RefCountedIpsetActor(felix-v4-_71d40657c2b731f), attempting to do a full rewrite RC=1, err=ipset v6.20.1: Error in line 2: Kernel error received: Connection refused
```
Apparently the retry succeeded because no further error was logged.

fasaxc:
@neiljerram I've now written an automated ST in the containers framework; in doing so, I found an issue with how we were using conntrack.  Namely, we were only matching on the conntrack state if we were forwarding to/from a workload.  I fixed that up, and saw an opportunity to tighten up the untracked accept rule (by matching the conntrack state UNTRACKED as well as the MARK).

Please can you re-review?

calico/felix/plugins/fiptgenerator.py
```diff
@@ -434,31 +434,22 @@ def filter_forward_chain(self, ip_version):
         # the raw table.
         forward_chain.append(
             "--append {chain} --jump ACCEPT --match mark "
-            "--mark {mark}/{mark}".format(
+            "--mark {mark}/{mark} "
+            "--match conntrack --ctstate UNTRACKED".format(
                 chain=CHAIN_FORWARD,
                 mark=self.IPTABLES_MARK_ACCEPT)
```
neiljerram: This change is fine, but it will affect some future situations where a user has other funny things in their iptables - so I think it would be good to add a fuller comment to explain the thinking here.  This will allow a future developer to balance what you're thinking now against whatever future needs might arise.

calico/felix/plugins/fiptgenerator.py
```diff
@@ -434,31 +434,22 @@ def filter_forward_chain(self, ip_version):
...
+        forward_chain.extend(self.drop_rules(
+            ip_version, CHAIN_FORWARD,
+            "--match conntrack --ctstate INVALID", None))
+        forward_chain.extend([
+            # First, a pair of conntrack rules, which accept established
+            # flows to/from workload interfaces.
+            "--append %s --match conntrack "
+            "--ctstate RELATED,ESTABLISHED --jump ACCEPT" %
+            (CHAIN_FORWARD,),
+        ])
```
neiljerram: Similarly here, to explain the thinking, which I think is that:

- in principle, the right thing for function is to move these acceptances further down into the per-host-endpoint chains, so that they only apply when host endpoints are in use (and on the relevant reverse flows)

- but we are leaving them here for performance, because moving them down would require a lot more matches before hitting the conntrack rules.

PR #1247, opened 2016-12-23
Use host endpoint ID as map key, instead of pointer to ID

neiljerram:


neiljerram:
Yes, I think so.  Suppose we go through resolveHostEndpoints() twice; the first time should set up chains for a host endpoint for the first time, the second should do some kind of update, with the chains still existing.

So, both times, the same id exists as a key in m.rawHostEndpoints.  That id is copied as part of the

    for id, hostEp := range m.rawHostEndpoints {

loop, and then bestHostEpId is set to a pointer to that copy.  I think that means that bestHostEpId will be a different pointer, each time that resolveHostEndpoints() is called.

Then, on the second time through,

		delete(m.activeHostIdToChains, id)

is accidentally a no-op and so does not delete the intended id from that map.  And hence in the next block of code we queue a delete for the chain:

	// Remove programming for host endpoints that are not now in use.
	for id, chains := range m.activeHostIdToChains {
		log.WithField("id", id).Info("HostEp removed, deleting its chains.")
		m.filterTable.RemoveChains(chains)
	}

even though the chain is actually still wanted.

fasaxc:
Makes sense, I thought that maps followed pointers when hashing but I just checked and that is not the case.

neiljerram:
Looks like the latest test (#738) has now successfully passed the point where we had an issue before.

PR #1259, opened 2017-01-10
Endpoint manager UTs

neiljerram:


neiljerram:
@fasaxc I think there's enough here to be worth a checkpoint review from you.

go/felix/intdataplane/endpoint_mgr_test.go
```diff
@@ -0,0 +1,406 @@
...
+	expectedRemove chan []*iptables.Chain
+
+	lastWorkDispatch []*iptables.Chain
+	lastHostDispatch []*iptables.Chain
+	lastHostEndpoint []*iptables.Chain
+}
+
+func newMockTable() *mockTable {
+	return &mockTable{
+		expectedUpdate: make(chan []*iptables.Chain, 2),
```
fasaxc: EndpointMgr is synchronous so you shouldn't need any channels in your test.  You could implement the mock Table more simply with a `map[string]*Chain` mapping chain name to current contents of the chain.  Then tests can send in a stimulus, then assert the chain contents.

go/felix/intdataplane/endpoint_mgr_test.go
```diff
@@ -0,0 +1,406 @@
...
+				[]string{"cali"},
+				nil,
+			)
+		})
+
+		It("should be constructable", func() {
+			Expect(epMgr).ToNot(BeNil())
+		})
+
+		It("should process host endpoints", func() {
```
fasaxc: I'd probably break this into a 2-3 nested Describe blocks so that you can do multiple asserts at each stage and fork the test scenario.

```
Describe("after sending in one host endpoint", func() {
  BeforeEach(func() {
    ... send in the host endpoint ...
  })
  It("should update the dispatch chain", func() {
    Expect(filterTable.chains["cali-..."]).ToEqual(...)
  })
  It("should render a to-endpoint chain", func() {
    Expect(filterTable.chains["calitw-..."]).ToEqual(...)
  })
  It("should render a from-endpoint chain", func() {
    Expect(filterTable.chains["califw-..."]).ToEqual(...)
  })

  Describe("after sending an interface state update", func() {
    BeforeEach(func() {
      ... send in the interfaceupdate ...
    })
    ...
  })

  // Instead...
  Describe("after deleting the endpoint", func() {
    BeforeEach(func() {
      ... send in the delete ...
    })
    ...
  })
}
```
WDYT?

go/felix/intdataplane/endpoint_mgr_test.go
fasaxc: If you're worried about policing for issues like reprogramming a chain unexpectedly, you can check for that in the UpdateChain method or you can record the changes in addition to the final state.

neiljerram: Thanks, I've made these changes now, and agree that it's much nicer!

go/felix/intdataplane/endpoint_mgr_test.go
neiljerram: TBH, I don't think this is especially helpful, as a transformation from what I have now.  In general I suspect it won't be, for the kind of UT that I'm working on, because I'm testing a long lived object (EndpointMgr) that _doesn't_ keep backtracking to its initial state, or to earlier states; rather it always processes changes with respect to its current state.  So for this kind of object, I think the serial testing as currently written is more realistic.

fasaxc: I don't follow, the sequences of method calls to the object are the same.  The advantage is that it isolates each assertion into a separate test case.

For example, suppose you want to test two sequences of operations with assertions after operation:
```
A->B->C
```
and
```
A->B->D
```
you'd do
```
Describe("A", func() {
  BeforeEach(func() {
    A
  })
  
  It("should do the thing we expect in state A")

  Describe("then B", func() {
    BeforeEach(func() {
      B
    })

    It("should do the thing we expect in A->B", ...)
    It("should do the second thing we expect in A->B", ...)
    It("should do the third thing we expect in A->B", ...)

    Describe("then C", func() {
      BeforeEach(func() {
        C
      })
      It("should do the thing we expect in A->B->C", ...)
      ...
    })

    Describe("then D", func() {
      BeforeEach(func() {
        D
      })
      It("should do the thing we expect in A->B->D", ...)
      ...
    })
  })
})
```

neiljerram: OK, cool, I was worried about the ever-increasing indentation, but I'll try that structure.

neiljerram:
@fasaxc This feels like a good checkpoint for re-review and possibly merging, and I'll move on to workload endpoints in a new PR.  Please could you do the honours?

go/felix/intdataplane/endpoint_mgr_test.go
```diff
@@ -0,0 +1,375 @@
...
+			})
+
+			It("should have empty dispatch chains", func() {
+				filterTable.checkChains(append(wlDispatchEmpty, hostDispatchEmpty...))
+			})
+
+			Describe("with host endpoint matching eth0", func() {
+				JustBeforeEach(configureHostEp("id1", "eth0", []string{}, []string{}))
+				It("should have expected chains", expectChainsFor("eth0"))
+
```
fasaxc: Unless it's covered subtly and I missed it, one case that may be missing here is checking the precedence when you have multiple matches (i.e. checking that the best match is stable)

go/felix/intdataplane/endpoint_mgr_test.go
```diff
@@ -0,0 +1,375 @@
...
+						ExpectedIpv6Addrs: ipv6Addrs,
+					},
+				})
+				epMgr.CompleteDeferredWork()
+			}
+		}
+
+		expectChainsFor := func(name string) func() {
+			return func() {
+				filterTable.checkChains(append(wlDispatchEmpty, hostDispatchForIface(name)...))
```
fasaxc: Bug: append can mutate its first argument.

neiljerram: D'oh - fixed now.

go/felix/intdataplane/endpoint_mgr_test.go
neiljerram: Hmm.  Not completely sure what you mean here.

If by stable you mean depending on the historical order that endpoint and iface updates arrived in... then no, the algorithm now is emphatically not stable in that sense; its choice of the 'best' current host endpoint, for a given iface, only depends on current data.

neiljerram:
@fasaxc I added tests for two resolved interfaces, and fixed the append bug; wasn't exactly sure what your other comment meant - could you take another look now?

go/felix/intdataplane/endpoint_mgr_test.go
fasaxc: I mean the latter.  I.e. it "only depends on current data".  To test that, I think you need a test with multiple candidate matches.  Particular issue in go, which randomises `map` iteration order.

go/felix/rules/endpoints.go
```diff
@@ -45,6 +46,7 @@ func (r *DefaultRuleRenderer) HostDispatchChains(endpoints map[string]proto.Host
 	for ifaceName, _ := range endpoints {
 		names = append(names, ifaceName)
 	}
+	sort.Strings(names)
```
fasaxc: should this be pushed down into `dispatchChains()`?  I bet our test of whether the workload chains have changed is failing right now, making us churn the chains each time.

fasaxc: I think this needs to be pushed down to avoid non-determinism in the workload case too (i.e. the mainline code relied on determinism here too).

neiljerram: Yes, done.

neiljerram: Yes.

PR #1270, opened 2017-01-16
UTs for IPIP manager; IP sets manager; masquerade manager; policy manager

fasaxc:
Included in this change: 

- I spotted that the ipipManager didn't correctly handle transient duplicate IPs so I simplified the logic to have it always rewrite the IP set.  I think that'll be OK since host churn is fairly rare and we're talking about a few thousand IPs at most.
- I spotted that we weren't honouring the config for max IP set size so I plumbed that in.

I kicked the tyres on the tunnel config updates.  Seems to work as expected without flapping the IP if it's already present.

### DIFF
--- a/go/felix/intdataplane/endpoint_mgr_test.go
+++ b/go/felix/intdataplane/endpoint_mgr_test.go
@@ -15,7 +15,6 @@
 package intdataplane
 
 import (
-	log "github.com/Sirupsen/logrus"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	"github.com/projectcalico/felix/go/felix/ipsets"
@@ -23,69 +22,11 @@ import (
 	"github.com/projectcalico/felix/go/felix/proto"
 	"github.com/projectcalico/felix/go/felix/rules"
 	"github.com/projectcalico/felix/go/felix/set"
-	"reflect"
 	"strings"
 )
 
-type mockTable struct {
-	currentChains  map[string]*iptables.Chain
-	expectedChains map[string]*iptables.Chain
-}
-
-func newMockTable() *mockTable {
-	return &mockTable{
-		currentChains:  map[string]*iptables.Chain{},
-		expectedChains: map[string]*iptables.Chain{},
-	}
-}
-
-func logChains(message string, chains []*iptables.Chain) {
-	if chains == nil {
-		log.Debug(message, " with nil chains")
-	} else {
-		log.WithField("chains", chains).Debug(message)
-		for _, chain := range chains {
-			log.WithField("chain", *chain).Debug("")
-		}
-	}
-}
-
-func (t *mockTable) UpdateChains(chains []*iptables.Chain) {
-	logChains("UpdateChains", chains)
-	for _, chain := range chains {
-		t.currentChains[chain.Name] = chain
-	}
-}
-
-func (t *mockTable) RemoveChains(chains []*iptables.Chain) {
-	logChains("RemoveChains", chains)
-	for _, chain := range chains {
-		_, prs := t.currentChains[chain.Name]
-		Expect(prs).To(BeTrue())
-		delete(t.currentChains, chain.Name)
-	}
-}
-
-func (t *mockTable) checkChains(expecteds [][]*iptables.Chain) {
-	t.expectedChains = map[string]*iptables.Chain{}
-	for _, expected := range expecteds {
-		for _, chain := range expected {
-			t.expectedChains[chain.Name] = chain
-		}
-	}
-	t.checkChainsSameAsBefore()
-}
-
-func (t *mockTable) checkChainsSameAsBefore() {
-	log.Debug("Expected chains")
-	for _, chain := range t.expectedChains {
-		log.WithField("chain", *chain).Debug("")
-	}
-	Expect(reflect.DeepEqual(t.currentChains, t.expectedChains)).To(BeTrue())
-}
-
 var wlDispatchEmpty = []*iptables.Chain{
-	&iptables.Chain{
+	{
 		Name: "cali-to-wl-dispatch",
 		Rules: []iptables.Rule{
 			{
@@ -95,7 +36,7 @@ var wlDispatchEmpty = []*iptables.Chain{
 			},
 		},
 	},
-	&iptables.Chain{
+	{
 		Name: "cali-from-wl-dispatch",
 		Rules: []iptables.Rule{
 			{
@@ -108,11 +49,11 @@ var wlDispatchEmpty = []*iptables.Chain{
 }
 
 var hostDispatchEmpty = []*iptables.Chain{
-	&iptables.Chain{
+	{
 		Name:  "cali-to-host-endpoint",
 		Rules: []iptables.Rule{},
 	},
-	&iptables.Chain{
+	{
 		Name:  "cali-from-host-endpoint",
 		Rules: []iptables.Rule{},
 	},
--- a/go/felix/intdataplane/int_dataplane.go
+++ b/go/felix/intdataplane/int_dataplane.go
@@ -437,3 +437,10 @@ func (d *InternalDataplane) loopReportingStatus() {
 		}
 	}
 }
+
+// iptablesTable is a shim interface for iptables.Table.
+type iptablesTable interface {
+	UpdateChain(chain *iptables.Chain)
+	UpdateChains([]*iptables.Chain)
+	RemoveChains([]*iptables.Chain)
+}
--- a/go/felix/intdataplane/int_dataplane.go
+++ b/go/felix/intdataplane/int_dataplane.go
@@ -32,6 +32,8 @@ type Config struct {
 	RuleRendererOverride rules.RuleRenderer
 	IPIPMTU              int
 
+	MaxIPSetSize int
+
 	IptablesRefreshInterval time.Duration
 
 	RulesConfig rules.Config
@@ -140,7 +142,7 @@ func NewIntDataplaneDriver(config Config) *InternalDataplane {
 
 	dp.endpointStatusCombiner = newEndpointStatusCombiner(dp.fromDataplane, !config.DisableIPv6)
 
-	dp.RegisterManager(newIPSetsManager(ipSetRegV4))
+	dp.RegisterManager(newIPSetsManager(ipSetRegV4, config.MaxIPSetSize))
 	dp.RegisterManager(newPolicyManager(filterTableV4, ruleRenderer, 4))
 	dp.RegisterManager(newEndpointManager(
 		filterTableV4,
@@ -176,7 +178,7 @@ func NewIntDataplaneDriver(config Config) *InternalDataplane {
 		routeTableV6 := routetable.New(config.RulesConfig.WorkloadIfacePrefixes, 6)
 		dp.routeTables = append(dp.routeTables, routeTableV6)
 
-		dp.RegisterManager(newIPSetsManager(ipSetRegV6))
+		dp.RegisterManager(newIPSetsManager(ipSetRegV6, config.MaxIPSetSize))
 		dp.RegisterManager(newPolicyManager(filterTableV6, ruleRenderer, 6))
 		dp.RegisterManager(newEndpointManager(
 			filterTableV6,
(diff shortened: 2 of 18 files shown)


go/felix/intdataplane/ipip_mgr_netlink.go
```diff
@@ -0,0 +1,61 @@
...
+
+// ipipDataplane is a shim interface for mocking netlink and os/exec in the IPIP manager.
+type ipipDataplane interface {
+	LinkByName(name string) (Link, error)
+	LinkSetMTU(link Link, mtu int) error
+	LinkSetUp(link Link) error
+	AddrList(link Link, family int) ([]Addr, error)
+	AddrAdd(link Link, addr *Addr) error
+	AddrDel(link Link, addr *Addr) error
+	RunCmd(name string, args ...string) error
```
neiljerram: RunCmd is quite generic - I wonder if the interface should instead have individual methods for what those commands are going to do.

go/felix/intdataplane/ipip_mgr_netlink.go
```diff
@@ -0,0 +1,61 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package intdataplane
+
+import (
+	. "github.com/vishvananda/netlink"
```
neiljerram: I think the code below would be clearer with explicit "netlink."s, and hence would prefer not using a dot import here.

go/felix/intdataplane/ipip_mgr_test.go
```diff
@@ -0,0 +1,310 @@
...
+
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	"errors"
+	"fmt"
+	log "github.com/Sirupsen/logrus"
+	"github.com/projectcalico/felix/go/felix/ipsets"
+	. "github.com/vishvananda/netlink"
```
neiljerram: Except for the ginkgo and gomega ones, I think I generally prefer not using dot imports unless there's a strong reason.

go/felix/intdataplane/ipip_mgr.go
```diff
@@ -104,22 +160,23 @@ func setLinkAddressV4(linkName string, address net.IP) error {
...
 		}
 	}
 
 	if !found && address != nil {
 		logCxt.Info("Address wasn't present, adding it.")
+		mask := net.CIDRMask(32, 32)
 		ipNet := net.IPNet{
-			IP:   address,
-			Mask: net.CIDRMask(32, 32),
+			IP:   address.Mask(mask), // Mask the IP to match ParseCIDR()'s behaviour.
```
fasaxc: That's all, it makes comparisons in the UT easier.

go/felix/intdataplane/ipip_mgr_netlink.go
fasaxc: I wondered that too, but doing that would put more code in the shim, where it's not possible to UT

go/felix/intdataplane/ipip_mgr_test.go
fasaxc: I don't mind making that change.  It does make some of the code a bit verbose/need line wrapping.

go/felix/intdataplane/ipip_mgr.go
neiljerram: Thanks.

go/felix/intdataplane/ipip_mgr_netlink.go
neiljerram: Ah right, yes, that's a good reason not to do that.

go/felix/intdataplane/ipip_mgr_test.go
neiljerram: Thanks - what do you think on this generally though?  (I wonder if my feeling so far might be wrong.)

go/felix/intdataplane/ipip_mgr_test.go
```diff
@@ -0,0 +1,444 @@
...
+			Expect(dataplane.addrs).To(HaveLen(0))
+		})
+	})
+
+	// Cover the error cases.  We pass the error back up the stack, check that that happens
+	// for all calls.
+	expNumCalls := 9
+	It("a successful call should only call into dataplane expected number of times", func() {
+		Expect(dataplane.NumCalls).To(BeNumerically("<=", expNumCalls))
+	})
```
neiljerram: This is a curious one, IIUC.  Am I right that it's <= because we don't know in what order different specs will be run?

go/felix/intdataplane/ipip_mgr_test.go
```diff
@@ -0,0 +1,444 @@
...
+		Describe(fmt.Sprintf("with a failure after %v calls", i), func() {
+			BeforeEach(func() {
+				dataplane.ErrorAtCall = i
+			})
+
+			It("should return the error", func() {
+				Expect(ipipMgr.configureIPIPDevice(1400, ip)).To(Equal(mockFailure))
+			})
+
+			Describe("with an IP to remove", func() {
```
neiljerram: In this case should I expect 1 more expNumCalls, and is there therefore another failure point?  Should the test express that?

go/felix/intdataplane/ipip_mgr_test.go
```diff
@@ -0,0 +1,444 @@
...
+					ipipMgr.OnUpdate(&proto.HostMetadataRemove{
+						Hostname: "host2",
+					})
+					ipipMgr.CompleteDeferredWork()
+				})
+				It("should keep the IP in the IP set", func() {
+					Expect(allHostsSet()).To(Equal(set.From("10.0.0.1")))
+				})
+
+				Describe("after removing iniital copy of IP", func() {
```
neiljerram: 'initial'

go/felix/set/set.go
```diff
@@ -56,6 +70,14 @@ func (set mapSet) Add(item interface{}) {
 	set[item] = emptyValue
 }
 
+func (set mapSet) AddAll(itemArray interface{}) {
+
+	arrVal := reflect.ValueOf(itemArray)
```
neiljerram: Interesting.  Go's type checking is normally compile time, isn't it?  But here I guess it must be at runtime instead?

go/felix/intdataplane/endpoint_mgr_test.go
```diff
@@ -108,11 +49,11 @@ var wlDispatchEmpty = []*iptables.Chain{
 }
 
 var hostDispatchEmpty = []*iptables.Chain{
-	&iptables.Chain{
+	{
 		Name:  "cali-to-host-endpoint",
 		Rules: []iptables.Rule{},
 	},
-	&iptables.Chain{
+	{
```
neiljerram: Thanks, I only recently realized that I could avoid repeating the type here.

go/felix/intdataplane/int_dataplane.go
```diff
@@ -32,6 +32,8 @@ type Config struct {
 	RuleRendererOverride rules.RuleRenderer
 	IPIPMTU              int
 
+	MaxIPSetSize int
```
fasaxc: Kernel requires us to set the maximum possible size when we create the IP set.

go/felix/intdataplane/int_dataplane.go
```diff
@@ -150,7 +154,8 @@ func NewIntDataplaneDriver(config Config) *InternalDataplane {
 	dp.RegisterManager(newMasqManager(ipSetRegV4, natTableV4, ruleRenderer, 1000000, 4))
 	if config.RulesConfig.IPIPEnabled {
 		// Add a manger to keep the all-hosts IP set up to date.
-		dp.RegisterManager(newIPIPManager(ipSetRegV4, 1000000)) // IPv4-only
+		dp.ipipManager = newIPIPManager(ipSetRegV4, 1000000)
```
fasaxc: It's the max IP set size, should be dependent on the parameter.  I'll fix up.

go/felix/intdataplane/ipip_mgr.go
```diff
@@ -129,69 +189,41 @@ func setLinkAddressV4(linkName string, address net.IP) error {
...
+		// the whole IP set whenever we get a change.  To replace this with delta handling
+		// would require reference counting the IPs because it's possible for two hosts
+		// to (at least transiently) share an IP.  That would add occupancy and make the
+		// code more complex.
+		log.Info("All-hosts IP set out-of sync, refreshing it.")
+		members := make([]string, 0, len(m.activeHostnameToIP))
+		for _, ip := range m.activeHostnameToIP {
+			members = append(members, ip)
+		}
+		m.ipsetReg.AddOrReplaceIPSet(m.ipSetMetadata, members)
```
fasaxc: It gets converted to a Set in the Ipset object.  I don't think it's wise to pass Sets around.  since they're designed to be mutated, there's temptation to share them between objects.

go/felix/intdataplane/ipip_mgr_test.go
```diff
@@ -0,0 +1,444 @@
...
+	)
+
+	ip, _, err := net.ParseCIDR("10.0.0.1/32")
+	if err != nil {
+		panic("Failed to parse test IP")
+	}
+	_, ipNet2, err := net.ParseCIDR("10.0.0.2/32")
+	if err != nil {
+		panic("Failed to parse test IP")
+	}
```
fasaxc: I'm thinking of these as constants, it's simply annoying that the `err` requires a bit of extra handling.

go/felix/intdataplane/ipip_mgr_test.go
fasaxc: Oops, no, there's a copy+paste bug there.  It's not actually calling the code under test!

Specs should always be isolated.

go/felix/intdataplane/ipip_mgr_test.go
fasaxc: Yes, but that code (after the failure case I'm trying to hit) is common so I don't think it matters too much

go/felix/set/set.go
fasaxc: Yes, Go's arrays aren't covariant (think that's the word) so you can't pass a `[]foo` to a function that wants a `[]interface{}`.  Only way to handle that is with `reflect`, which I think is OK in such a datastructure.

go/felix/intdataplane/policy_mgr_test.go
```diff
@@ -0,0 +1,137 @@
...
+					},
+				},
+			})
+			policyMgr.CompleteDeferredWork()
+		})
+
+		It("should install the in and out chain", func() {
+			filterTable.checkChains([][]*iptables.Chain{{
+				{Name: "calipi-tier1/pol1"},
+				{Name: "calipo-tier1/pol1"},
```
neiljerram: Why do I not see any rules here?

go/felix/intdataplane/policy_mgr_test.go
```diff
@@ -0,0 +1,137 @@
...
+					},
+				},
+			})
+			policyMgr.CompleteDeferredWork()
+		})
+
+		It("should install the in and out chain", func() {
+			filterTable.checkChains([][]*iptables.Chain{{
+				{Name: "calipi-prof1"},
+				{Name: "calipo-prof1"},
```
neiljerram: (and similarly here)

go/felix/intdataplane/policy_mgr_test.go
```diff
@@ -0,0 +1,137 @@
...
+
+type mockPolRenderer struct {
+}
+
+func (r *mockPolRenderer) PolicyToIptablesChains(policyID *proto.PolicyID, policy *proto.Policy, ipVersion uint8) []*iptables.Chain {
+	inName := rules.PolicyChainName(rules.PolicyInboundPfx, policyID)
+	outName := rules.PolicyChainName(rules.PolicyOutboundPfx, policyID)
+	return []*iptables.Chain{
+		{Name: inName},
+		{Name: outName},
```
neiljerram: Oh I see - guess that's OK then.

go/felix/intdataplane/endpoint_mgr_test.go
fasaxc: My editor was giving it a yellow squiggly line so I cleaned it up.

PR #1274, opened 2017-01-18
Implement tree-based dispatch chains.

fasaxc:


go/felix/stringutils/common_prefix_test.go
```diff
@@ -0,0 +1,55 @@
...
+			input[i] = input[len(input)-i-1]
+			input[len(input)-i-1] = temp
+		}
+		Expect(CommonPrefix(input)).To(Equal(expected))
+
+		By("giving correct result in sorted order")
+		sort.Strings(input)
+		Expect(CommonPrefix(input)).To(Equal(expected))
+	},
+	Entry("Empt array", []string{}, ""),
```
neiljerram: 'Empty'

go/felix/stringutils/common_prefix_test.go
```diff
@@ -0,0 +1,55 @@
...
+	Entry("Empty string", []string{""}, ""),
+	Entry("Empty strings", []string{"", ""}, ""),
+	Entry("Empty string and another string", []string{"", "a"}, ""),
+	Entry("Empty string and another string", []string{"", "abcdef"}, ""),
+	Entry("Empty string and another string", []string{"", "abcdef"}, ""),
```
neiljerram: duplicate entry

go/felix/rules/dispatch.go
```diff
@@ -0,0 +1,217 @@
...
+				})
+				childToEndpointRules = append(childToEndpointRules, Rule{
+					Match: Match().OutInterface(name),
+					Action: GotoAction{
+						Target: EndpointChainName(toEndpointPfx, name),
+					},
+				})
+			}
+			if dropAtEndOfChain {
+				logCxt.Debug("Adding drop rules at end of child chains.")
```
neiljerram: I think worth commenting here that we use Goto instead of Jump, and drop at the end of child chains, so that a non-matching packet does not have to iterate through all of the chain tree.  (It took me a few minutes to see that.)

go/felix/rules/dispatch.go
```diff
@@ -0,0 +1,217 @@
...
+			})
+			rootToEndpointRules = append(rootToEndpointRules, Rule{
+				Match: Match().OutInterface(ifaceMatch),
+				Action: GotoAction{
+					Target: childToChainName,
+				},
+			})
+
+			// ...and child chains.
+			childFromEndpointRules := make([]Rule, 0, len(ifaceNames)+2)
```
neiljerram: Instead of `2`, I guess it would be better to add a (renderer)DropRulesLen() method, to say how many Rule objects are needed for DropRules.
(and similarly above)

go/felix/rules/dispatch_test.go
```diff
@@ -0,0 +1,350 @@
...
+						outboundGotoRule("cali1+", "cali-to-wl-dispatch-1"),
+						outboundGotoRule("cali2+", "cali-to-wl-dispatch-2"),
+						expDropRule,
+					},
+				},
+			}),
+
+		// Duplicate interfaces could occur during transient misconfigurations, while
+		// there's no way to make them "work" since we can't distinguish the dupes, we
+		// should still render something sensible.
```
neiljerram: Hm, I'm not sure the following output _is_ sensible.  Wonder how hard it would be to change the code so that duplicates like this are coalesced?

fasaxc: Used that as a way to output a clear error message.

PR #1276, opened 2017-01-18
Port ChainInsertMode to Go

fasaxc:
- Introduce TableOptions struct for optional Table params. (First commit.)
- Plumb ChainInsertMode through and implement in Table:
  - Use same logic to calculate hashes of insert chains when doing the insert and comparing them in the dataplane.  Makes the comparison a lot simpler.
  - Implement the append itself.
  - Add UT.

### DIFF
--- a/go/felix/iptables/table.go
+++ b/go/felix/iptables/table.go
@@ -167,6 +167,10 @@ type Table struct {
 	iptablesRestoreCmd string
 	iptablesSaveCmd    string
 
+	// insertMode is either "insert" or "append"; whether we insert our rules or append them
+	// to top-level chains.
+	insertMode string
+
 	logCxt *log.Entry
 
 	// Factory for making commands, used by UTs to shim exec.Command().
@@ -178,6 +182,7 @@ type Table struct {
 type TableOptions struct {
 	HistoricChainPrefixes    []string
 	ExtraCleanupRegexPattern string
+	InsertMode               string
 
 	// NewCmdOverride for tests, if non-nil, factory to use instead of the real exec.Command()
 	NewCmdOverride cmdFactory
@@ -218,6 +223,16 @@ func NewTable(
 		dirtyInserts.Add(kernelChain)
 	}
 
+	var insertMode string
+	switch options.InsertMode {
+	case "", "insert":
+		insertMode = "insert"
+	case "append":
+		insertMode = "append"
+	default:
+		log.WithField("insertMode", options.InsertMode).Panic("Unknown insert mode")
+	}
+
 	// Allow override of exec.Command() and time.Sleep() for test purposes.
 	newCmd := newRealCmd
 	if options.NewCmdOverride != nil {
@@ -244,6 +259,7 @@ func NewTable(
 		hashCommentRegexp: hashCommentRegexp,
 		ourChainsRegexp:   ourChainsRegexp,
 		oldInsertRegexp:   oldInsertRegexp,
+		insertMode:        insertMode,
 
 		newCmd: newCmd,
 		sleep:  sleep,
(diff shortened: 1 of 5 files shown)


go/felix/iptables/table.go
```diff
@@ -366,23 +345,29 @@ func (t *Table) loadDataplaneState() {
...
 		if _, ok := t.chainToDataplaneHashes[chainName]; ok {
 			// Chain expected, we'll have checked its contents above.
 			logCxt.Debug("Skipping expected chain")
 			continue
 		}
+		if !t.ourChainsRegexp.MatchString(chainName) {
+			// Non-calico chain, check it doesn't have any rule insertions in it.
+			for _, hash := range dataplaneHashes {
+				if hash != "" {
+					logCxt.Info("Found unexpected insert, marking for cleanup")
```
neiljerram: I'm not sure I understand why Calico inserts are unexpected here.  Is it because this method only runs at start of day?

go/felix/iptables/table.go
```diff
@@ -630,18 +619,29 @@ func (t *Table) applyUpdates() error {
...
-				// Make sure currentHashes ends up the right length.
-				currentHashes = append(currentHashes, "")
 			}
 		}
-		for i := len(rules) - 1; i >= 0; i-- {
-			prefixFrag := t.commentFrag(currentHashes[i])
-			line := rules[i].RenderInsert(chainName, prefixFrag)
-			inputBuf.WriteString(line)
-			inputBuf.WriteString("\n")
+
```
neiljerram: Code just above only removes, and the folllowing code uses reverse order for insert but normal order for append, so perhaps:
- update the comment above to say "remove" instead of "remove/insert"
- add a comment for the insert case below to explain that we use reverse order because (I guess) we always insert at position 1.

go/felix/iptables/table_test.go
```diff
@@ -357,6 +374,13 @@ var _ = Describe("Table with a dirty dataplane", func() {
 				// This rule will get rewritten because its hash is incorrect.
 				"-m comment --comment \"cali:1234567890ksamdl\" --jump DROP",
 			},
+			"non-calico-insert": {
+				// This rule will get cleaned up because we don't insert any rules
+				// into this chain.
+				"--jump ACCEPT",
+				"-m comment --comment \"cali:hecdSCslEjdBPfds\" --jump DROP",
```
neiljerram: How is this a non-Calico insert if it has a Calico hash comment?  (Probably just needs a tad more commenting.)

go/felix/iptables/table_test.go
```diff
@@ -436,19 +472,23 @@ var _ = Describe("Table with a dirty dataplane", func() {
...
-			Expect(dataplane.Chains).To(Equal(map[string][]string{
+			expChains := map[string][]string{
 				"FORWARD": {
-					"-m comment --comment \"cali:hecdSCslEjdBPBPo\" --jump DROP",
-					"-m comment --comment \"cali:plvr29-ZiKUwbzDV\" --jump ACCEPT",
 					"--jump RETURN",
 					"--jump ACCEPT",
 					"--jump foo-bar",
+					"-m comment --comment \"cali:hecdSCslEjdBPBPo\" --jump DROP",
+					"-m comment --comment \"cali:plvr29-ZiKUwbzDV\" --jump ACCEPT",
```
neiljerram: Is there a reason for setting the FORWARD rules here, given that they get overwritten immediately below?  I think it would be clearer to omit them here.

go/felix/iptables/table.go
fasaxc: Yes, it's to catch rules that shouldn't be there at start of day.  I'll improve the comment.

go/felix/iptables/table.go
```diff
@@ -393,6 +378,28 @@ func (t *Table) loadDataplaneState() {
...
+// given our inserts.  If we're in append mode, that consists of numNonCalicoRules empty strings
+// followed by our hashes; in insert mode, the opposite way round.  To avoid recalculation, it
+// returns the rule hashes as a second output.
+func (t *Table) expectedHashesForInsertChain(
+	chainName string,
+	numNonCalicoRules int,
+) (allHashes, ruleHashes []string) {
+	insertedRules := t.chainToInsertedRules[chainName]
+	allHashes = make([]string, len(insertedRules)+numNonCalicoRules)
+	ruleHashes = calculateRuleInsertHashes(chainName, insertedRules)
```
fasaxc: Sounds good

go/felix/iptables/table_test.go
fasaxc: No, meant to snip those out.

go/felix/iptables/table_test.go
fasaxc: It was meant to mean  "non-calico chain with an insert"; will improve the name

fasaxc:
@neiljerram Think I've addressed all of those comments, please can you take another look?

fasaxc:
retest this please

fasaxc:
retest this please

fasaxc:
retest this please

PR #1278, opened 2017-01-19
Add support for floating IPs to the golang dataplane driver

neiljerram:


go/felix/intdataplane/floating_ip_mgr.go
```diff
@@ -0,0 +1,163 @@
...
+// There generally _isn't_ a corresponding SNAT for the other direction; in other words, packets
+// sent _from_ a workload don't have their source address changing to a floating IP, if the workload
+// has floating IPs associated with it.  This is because people don't expect floating IPs to do
+// that, and its faster in the datapath to avoid NATs if we can, especially when sending to another
+// workload nearby.  (And for sending outbound to the Internet, there will be an SNAT if needed at
+// the data center's border gateway.)  However SNAT is needed in the loopback case where a workload
+// sends to a floating IP that maps back to itself; for this case the datapath processing must be as
+// follows:
+//
+// 1. Workload sends to <floating IP>, so packet has SRC=<workload IP> DST=<floating IP>
```
fasaxc: I think the lack of a `.` at the end of these lines will render them as headings in godoc, which is probably not what you want.

go/felix/intdataplane/endpoint_mgr.go
```diff
@@ -207,8 +207,24 @@ func (m *endpointManager) resolveWorkloadEndpoints() error {
 			var ipStrings []string
```
fasaxc: How about defining an `extraIPs` slice here, then adding the NAT IPs to it below, then the copy logic can be shared?

fasaxc: Alternatively, this code might be simpler if we make a utility function that did the extraction from the proto endpoint and converted into a simpler struct holding just the information we need.

go/felix/intdataplane/floating_ip_mgr.go
```diff
@@ -0,0 +1,163 @@
...
+				log.Debug("Wanted reverse mapping")
+				snats[intIp] = extIp
+			}
+		}
+		// Render chains for those NATs.
+		dnatChains := m.ruleRenderer.DNATsToIptablesChains(dnats)
+		snatChains := m.ruleRenderer.SNATsToIptablesChains(snats)
+		// Update iptables if they have changed.
+		if !reflect.DeepEqual(m.activeDnatChains, dnatChains) {
+			m.natTable.UpdateChains(dnatChains)
```
fasaxc: Might be best to remove the old chains here first, in case we add a tree or something to the NAT chains later.

go/felix/intdataplane/floating_ip_mgr.go
```diff
@@ -0,0 +1,163 @@
...
+// cali-fip-snat from cali-POSTROUTING.
+type floatingIpManager struct {
+	ipVersion uint8
+
+	// Our dependencies.
+	natTable     iptablesTable
+	ruleRenderer rules.RuleRenderer
+
+	// Internal state.
+	activeDnatChains []*iptables.Chain
```
fasaxc: Nit: `s/Dnat/DNAT/` for go's naming convention (and a couple more below).

go/felix/rules/nat.go
```diff
@@ -36,3 +39,47 @@ func (r *DefaultRuleRenderer) NATOutgoingChain(natOutgoingActive bool, ipVersion
 		Rules: rules,
 	}
 }
+
+func (r *DefaultRuleRenderer) DNATsToIptablesChains(dnats map[string]string) []*iptables.Chain {
+	// Extract and sort map keys so we can program rules in a determined order.
+	sortedExtIps := make([]string, 0, len(dnats))
+	for extIp, _ := range dnats {
```
fasaxc: `, _` isn't needed here, you only need the `_`  at the start if you want the second value of the "tuple"

go/felix/rules/rule_defs.go
```diff
@@ -36,13 +36,17 @@ const (
...
 	ChainNATPostrouting = ChainNamePrefix + "-POSTROUTING"
+	ChainNATOutput      = ChainNamePrefix + "-OUTPUT"
 	ChainNATOutgoing    = ChainNamePrefix + "-nat-outgoing"
 
 	IPSetIDNATOutgoingAllPools  = "all-ipam-pools"
 	IPSetIDNATOutgoingMasqPools = "masq-ipam-pools"
 
 	IPSetIDAllHostIPs = "all-hosts"
 
+	ChainFipDnat = ChainNamePrefix + "-fip-dnat"
```
fasaxc: Nit: acronyms should be upper case in go names

go/felix/rules/rule_defs.go
```diff
@@ -101,6 +105,9 @@ type RuleRenderer interface {
 	ProtoRuleToIptablesRules(pRule *proto.Rule, ipVersion uint8) []iptables.Rule
 
 	NATOutgoingChain(active bool, ipVersion uint8) *iptables.Chain
+
+	DNATsToIptablesChains(dnats map[string]string) []*iptables.Chain
```
fasaxc: Any way we can give `dnats` a better name (especially since it's the other way around to snats)? `extIPtoIntIP` maybe?

go/felix/intdataplane/endpoint_mgr.go
neiljerram: I've done something like the former of these - do you think that more is still advisable?

go/felix/rules/rule_defs.go
neiljerram: As discussed, leaving this as is for now.

fasaxc:
retest this please

PR #1279, opened 2017-01-19
Add host endpoint status reports.

fasaxc:


go/felix/intdataplane/endpoint_mgr.go
```diff
@@ -173,70 +194,181 @@ func (m *endpointManager) OnUpdate(protoBufMsg interface{}) {
...
+	m.resolveWorkloadEndpoints()
+
+	if m.hostEndpointsDirty {
+		log.Debug("Host endpoints updated, resolving them.")
+		m.resolveHostEndpoints()
+		m.hostEndpointsDirty = false
+	}
+
+	// Since the pairings of endpoints and interfaces may have changed, rescan the interface
+	// state updates and mark endpoint statuses for update.
```
neiljerram: This is safe and defensive, but is it necessary?  I would expect the resolve calls above already to add any updated endpoints to the dirty set - in fact they _must_ do so (or else we wouldn't get expected status updates) because we generally _won't_ have endpoint updates and iface updates in the same CompleteDeferredWork batch.

So are there any cases where this code adds an ep ID into the dirty set that isn't already in it?  (and that needs to be in it)

go/felix/intdataplane/endpoint_mgr.go
```diff
@@ -173,70 +194,181 @@ func (m *endpointManager) OnUpdate(protoBufMsg interface{}) {
...
 
-	if m.dirtyHostEndpoints {
-		if err := m.resolveHostEndpoints(); err != nil {
-			log.WithError(err).Warn("Failed to resolve host endpoints")
-			return err
+func (m *endpointManager) updateEndpointStatuses() {
+	log.WithField("dirtyEndpoints", m.epIDsToUpdateStatus).Debug("Reporting endpoint status.")
+	m.epIDsToUpdateStatus.Iter(func(item interface{}) error {
+		switch id := item.(type) {
+		case proto.WorkloadEndpointID:
```
neiljerram: How can `id` here be both the type and the full value?  Some golang magic that I don't understand yet?

go/felix/intdataplane/endpoint_mgr.go
```diff
@@ -173,70 +194,181 @@ func (m *endpointManager) OnUpdate(protoBufMsg interface{}) {
...
+	}
+	logCxt = logCxt.WithFields(log.Fields{
+		"known":   known,
+		"failed":  failed,
+		"operUp":  operUp,
+		"adminUp": adminUp,
+		"status":  status,
+	})
+	logCxt.Info("Re-evaluated workload endpoint status")
+	return status
```
neiljerram: Looks like status will be "" if known is false.  Is that the intention?  (Probably worth a comment if so.)

go/felix/intdataplane/endpoint_mgr.go
```diff
@@ -439,13 +534,34 @@ func (m *endpointManager) resolveHostEndpoints() error {
...
+				newHostEpIDToIfaceNames[bestHostEpId], ifaceName)
+			m.epIDsToUpdateStatus.Add(bestHostEpId)
+		}
+
+		oldID, wasKnown := m.activeIfaceNameToHostEpID[ifaceName]
+		newID, isKnown := newIfaceNameToHostEpID[ifaceName]
+		if oldID != newID {
+			logCxt := ifaceCxt.WithFields(log.Fields{
+				"oldEndpoint": m.activeIfaceNameToHostEpID[ifaceName],
+				"newEndpoint": newIfaceNameToHostEpID[ifaceName],
```
neiljerram: Use oldID and newID here.

go/felix/intdataplane/endpoint_mgr.go
```diff
@@ -56,32 +57,48 @@ type endpointManager struct {
...
+	// rawHostEndpoints contains the raw (i.e. not resolved to interface) host endpoints.
+	rawHostEndpoints map[proto.HostEndpointID]*proto.HostEndpoint
+	// hostEndpointsDirty is set to true when host endpoints are updated.
+	hostEndpointsDirty bool
+	// activeHostIfaceToChains maps host interface name to the chains that we've programmed.
+	activeHostIfaceToChains map[string][]*iptables.Chain
+	// activeHostDispatchChains contains the dispatch chains that we've programmed for host
+	// endpoints.
 	activeHostDispatchChains []*iptables.Chain
+	// activeHostEpIDToIfaceNames records which interfaces we resolved each host endpoint to.
```
fasaxc: Seemed like the most sensible thing to do if an endpoint resolves to more than one interface.

go/felix/intdataplane/status_combiner.go
```diff
@@ -84,16 +83,33 @@ func (e *endpointStatusCombiner) Apply() {
 		}
 		if statusToReport == "" {
 			logCxt.Info("Reporting endpoint removed.")
-			e.fromDataplane <- &proto.WorkloadEndpointStatusRemove{
-				Id: &id,
+			switch id := id.(type) {
```
fasaxc: It's syntactic sugar, the new `id` variable shadows the outer one and it takes on the type of the particular `case` that it's in.

go/felix/intdataplane/endpoint_mgr.go
fasaxc: That's just the syntax for a "type switch".  It switches on the type but the `id` variable refers to the (casted) value.

go/felix/intdataplane/endpoint_mgr.go
fasaxc: Yes, intentional.

go/felix/intdataplane/endpoint_mgr.go
fasaxc: I think you're right, it's over-defensive.  I don't think it can add anything new.  I'll remove it.

go/felix/intdataplane/endpoint_mgr.go
neiljerram: But it's an error, isn't it?  (I think it's only possible if the same IP address is assigned to multiple interfaces.)

neiljerram: I guess if it is an error, it's one that will affect other parts of the OS much more, and that there is a sane way (as here) for us to handle it - so OK.

PR #1284, opened 2017-01-24
Add untracked support to golang dataplane driver.

fasaxc:
- Pin libcalico-go to branch with untracked support.
- Plumb flag through to policy manager, which inserts the policy into the filter or raw table as appropriate.
- Split policies into tracked/untracked as we copy them to the protobuf API structs.
- Implement top-level raw cali-PREROUTING and cali-OUTPUT chains.
- Includes implementing RPF check in the PREROUTING chain.


fasaxc:
TODO:

- [x] UT for new chains.
- [ ] UT for endpoint manager
- [ ] Make sure untracked CI test enabled and passing.

fasaxc:
retest this please

go/felix/rules/static.go
```diff
@@ -429,6 +450,87 @@ func (r *DefaultRuleRenderer) StaticNATOutputChains(ipVersion uint8) []*Chain {
 	}}
 }
 
+func (r *DefaultRuleRenderer) StaticRawTableChains(ipVersion uint8) []*Chain {
+	return []*Chain{
+		r.failsafeInChain(),
+		r.failsafeOutChain(),
```
neiljerram: Why do we put failsafes in the raw table?

I just checked that we did this in the 1.4 version too, but I can't remember why.  I'm thinking they are perhaps not needed there because we don't make any DROP decisions in the raw table.

go/felix/rules/static.go
```diff
@@ -429,6 +450,87 @@ func (r *DefaultRuleRenderer) StaticNATOutputChains(ipVersion uint8) []*Chain {
...
+			// Then, jump to the untracked policy chains.
+			{Action: JumpAction{Target: ChainDispatchToHostEndpoint}},
+			// Then, if the packet was marked as allowed, accept it.  Packets also
+			// return here without the mark bit set if the interface wasn't one that
+			// we're policing.
+			{Match: Match().MarkSet(r.IptablesMarkAccept),
+				Action: AcceptAction{}},
+		},
+	}
+}
```
neiljerram: There's no workload prefix check here; in theory I think that means someone could configure a host endpoint matching a workload i/f or address, with untracked policy, and then a local send from the compute host to that workload could be denied by that policy.  Does that matter?

neiljerram: Probably more important, wonder about the interaction between untracked and gateway protection.  For untracked we haven't hooked the forwarding chains, so implementation-wise I think that means we don't do untracked gateway protection.  Is that also what makes sense requirements-wise?

(I think probably yes, as it would be crazy to act as a gateway without conntrack help - but worth checking.)

go/felix/intdataplane/policy_mgr.go
```diff
@@ -47,13 +49,16 @@ func (m *policyManager) OnUpdate(msg interface{}) {
 	case *proto.ActivePolicyUpdate:
 		log.WithField("id", msg.Id).Debug("Updating policy chains")
 		chains := m.ruleRenderer.PolicyToIptablesChains(msg.Id, msg.Policy, m.ipVersion)
+		m.rawTable.UpdateChains(chains)
```
fasaxc: Yes, that was the quick fix I described last night.  Need to revisit to work out if there's a better way...

go/felix/iptables/table.go
```diff
@@ -531,6 +533,14 @@ func (t *Table) Apply() {
...
 				continue
 			} else {
+				t.logCxt.WithError(err).Error("Failed to program iptables, loading diags before panic.")
+				cmd := t.newCmd(t.iptablesSaveCmd, "-t", t.Name)
+				output, err2 := cmd.Output()
+				if err2 != nil {
+					t.logCxt.WithError(err2).Error("Failed to load iptables state")
+				} else {
+					t.logCxt.WithField("iptablesState", string(output)).Error("Current state of iptables")
+				}
```
fasaxc: I believe it'll use `\n`.  That's what we get elsewhere.  Is that nicer or worse than the python logging?  It has the advantage of being easier to machine parse.

go/felix/proto/felixbackend.proto
```diff
@@ -258,6 +259,7 @@ message HostEndpoint {
   string name = 1;
   repeated string profile_ids = 2;
   repeated TierInfo tiers = 3;
+  repeated TierInfo untracked_tiers = 6;
```
fasaxc: Yes.  That's the correct way to evolve a protobuf schema.  You shouldn't change the IDs assigned to existing fields if you want to maintain back-compat.  However, you can put the new field in any position in the definition in order to maximise readability.

go/felix/rules/static.go
fasaxc: You can have explicit DROPs in your policy even if there aren't any implicit "end of xyz" DROPs?

go/felix/rules/static.go
fasaxc: I think I can live with someone being able to do that.  Policing it at the config stage would be tricky (since the interface prefix could be changed later).  Inserting the check into the iptables chain adds latency to each packet just for that case, I think, which doesn't feel great.

There is no raw FORWARD table, the raw table operates before the routing decision so untracked policy does affect through traffic, but only at ingress.

go/felix/intdataplane/endpoint_mgr.go
```diff
@@ -524,17 +534,23 @@ func (m *endpointManager) resolveHostEndpoints() {
...
 				"ifaceName":    ifaceName,
 				"bestHostEpId": bestHostEpId,
-			}).Debug("Got HostEp for interface")
+			})
+			logCxt.Debug("Got HostEp for interface")
 			newIfaceNameToHostEpID[ifaceName] = bestHostEpId
+			if len(bestHostEp.UntrackedTiers) > 0 {
+				logCxt.Debug("Endpoint has untracked policies.")
+				newUntrackedIfaceNameToHostEpID[ifaceName] = bestHostEpId
+			}
```
fasaxc: No, the default in the tracked policy is to drop or fall through to profiles (and we need to render the failsafe policy too).  Will add a comment.

go/felix/rules/endpoints.go
```diff
@@ -31,9 +31,47 @@ func (r *DefaultRuleRenderer) WorkloadEndpointToIptablesChains(epID *proto.Workl
...
+		endpoint.Tiers,
+		endpoint.ProfileIds,
+		ifaceName,
+		PolicyOutboundPfx,
+		PolicyInboundPfx,
+		HostToEndpointPfx,
+		HostFromEndpointPfx,
+		ChainFailsafeOut,
+		ChainFailsafeIn,
+		untrackedDisabled,
```
fasaxc: Swapped for a `chainType` enum, which makes it more clear.  I wanted to avoid a string of `bool` arguments, which can be hard to marry up.

go/felix/rules/static.go
neiljerram: Hmm, not sure; and sounds like you're not sure too.  Should we leave some kind of reminder to come back to this?

go/felix/iptables/table.go
neiljerram: Definitely worse if I'm reading the logging and trying to understand why the programming failed.

But anyway, it's an orthogonal detail that we can come back to (if/when we care enough).

go/felix/intdataplane/policy_mgr.go
neiljerram: Yep - sorry, understood those things in the wrong order!

fasaxc:
Added a trello note to revisit that.

PR #1288, opened 2017-01-25
Felix 2.0.2 Deb/RPM packaging

neiljerram:


utils/make-packages.sh
```diff
@@ -55,7 +65,10 @@ EOF
 			cat <<EOF
   * Felix ${version} (from Git commit ${sha}).
 EOF
-			git show ${version} --format=oneline -s | head -n -1 | tail -n +5 | sed 's/^/    /'
+			for tag in ${tags}; do
+			    echo "    [Changes recorded in $tag tag]"
```
fasaxc: Slightly concerned about the square brackets, could they end up being some sort of markup?  What format are those text blocks in, are they just plain text?

neiljerram: I used that format because I've seen it before in debian changelogs (for very similar reasons).  So pretty sure it's good in that context.

Less sure about the RPM spec - but there the changelog is separated off at the end and so seems less central anyway.  And in any case, there's no apparent problem in the RPM generation.

Testing-wise, we've now had a smoke pass with Debs built like this.  Don't have an RPM pass yet because I forgot to sign the packages - so let's at least wait until we get that RPM pass.

neiljerram:
FYI we've had smoke RPM passes now, so it appears that the RPM changelog formatting is fine.

full-fv runs are now in progress for these packages, and not far off done, so I'll just wait for those before merging this PR.

neiljerram:
Just the expected libvirt 2 - related failures in full-fv run, so now merging this.

PR #1290, opened 2017-01-26
[WiP] Remove python files and builds.

### LINKED ISSUE #1266: python/version.py not a proper make target and not cleaned up
`python/version.py` is generated as a side effect of a set-up command in the Makefile and it isn't cleaned up by `make clean`.

That at least causes issues on Mac.  We should 

- [ ] move `python/version.py` into a proper target 
- [ ] containerise it so it doesn't rely on the environment
- [ ] add it to make clean.

fasaxc:
Remove the python version of the dataplane driver.

Fixes #1266


neiljerram:
Can also remove `utils/gen-version.sh`

calc/async_decoupler.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2016 Tigera, Inc. All rights reserved.
+// Copyright (c) 2016-2017 Tigera, Inc. All rights reserved.
```
neiljerram: Is it golang-conventional that we now have packages - calc, proto etc. - directly at the top level of the Git repo?  (Or should there be a top-level dir named 'src', or something similar?)

Makefile
```diff
@@ -1,122 +1,99 @@
...
-
-# Figure out what git commit we have checked out.  We'll bake that into the
-# executable.
-GIT_COMMIT:=$(shell git rev-parse HEAD)
-GIT_COMMIT_SHORT:=$(shell git rev-parse --short HEAD)
-GIT_DESCRIPTION:=$(shell git describe --tags)
+# Figure out version information.  To support builds from release tarballs, we default to
+# <unknown> if this isn't a git checkout.
+GIT_COMMIT:=$(shell git rev-parse HEAD || echo '<unknown>')
+BUILD_ID:=$(shell git rev-parse HEAD || uuidgen | sed 's/-//g')
```
neiljerram: Not sure about operator precedence here.  In the Git case, does  the output go through sed?

Makefile
```diff
@@ -449,6 +351,10 @@ release: clean
 ifndef VERSION
 	$(error VERSION is undefined - run using make release VERSION=X.Y.Z)
 endif
+ifeq ($(GIT_COMMIT),<unknown>)
+	$(error git commit ID couldn't be determined, releases must be done from a git working copy)
+endif
+	git rev-parse HEAD
```
neiljerram: Last line left in by accident?  (Or else what is its point?)

neiljerram:
OK, that's a complete review, as the only thing you've added since is the removal of gen-version.sh.

Makefile
```diff
@@ -476,7 +391,7 @@ release-once-tagged:
 	@echo "- Copy the tag content (release notes) shown on that page"
 	@echo "- Go to https://github.com/projectcalico/felix/releases/new?tag=$(VERSION)"
 	@echo "- Paste the copied tag content into the large textbox"
-	@echo "- Attach the PyInstaller bundle"
+	@echo "- Attach the binary"
```
fasaxc: I think we do need to attach something; we need it to maintain the bare-metal install guide (which doesn't use debs or RPMs).  The bundle didn't have source code inside, it was pre-compiled python.

github automatically adds a source tarball to the release, which is annoying because that tarball doesn't know what its version number is.

debian/control
```diff
@@ -2,15 +2,13 @@ Source: felix
...
-XS-Python-Version: >= 2.7
+Build-Depends: debhelper (>= 8.0.0), dh-systemd
 Standards-Version: 3.9.4
 
 Package: calico-common
 Architecture: all
 Depends:
- ${misc:Depends},
- ${shlibs:Depends}
+ ${misc:Depends}
```
fasaxc: I think it means "shared library deps"; I don't think we were using it (and debhelper was throwing a warning saying it didn't know what it was).

debian/rules
```diff
@@ -6,23 +6,15 @@ export CALICODEPS = none
...
-	install -m 755 etc/*.sh debian/tmp/usr/bin
-	install -d debian/tmp/opt/calico-felix
-	cd pyi && find . -type d | xargs -I DIR install -d ../debian/tmp/opt/calico-felix/DIR
-	cd pyi && find . -type f | grep -v -E 'calico-iptables-plugin|calico-felix' | xargs -I FILE install -m 644 FILE ../debian/tmp/opt/calico-felix/FILE
-	install -m 755 pyi/calico-iptables-plugin debian/tmp/opt/calico-felix/calico-iptables-plugin
-	install -m 755 pyi/calico-felix debian/tmp/opt/calico-felix/calico-felix
-	cd pyi && find . -type l | xargs -I FILE install FILE ../debian/tmp/opt/calico-felix/FILE
-	cd debian/tmp/usr/bin && ln -s ../../opt/calico-felix/calico-felix ./calico-felix
-	cd debian/tmp/usr/bin && ln -fs ../../opt/calico-felix/calico-iptables-plugin ./calico-iptables-plugin
+	install -m 755 bin/* debian/tmp/usr/bin/
```
fasaxc: I've changed it to install `calico-felix` directly to `/usr/bin`, since there's no longer a need for a /opt/calico-felix directory to hold the pyinstaller bundle.

debian/build-debs
```diff
@@ -16,10 +16,11 @@ deb_dir=dist/$STREAM
 unpacked_src_dir=${deb_dir}/calico-${DEB_VERSION}
 
 rm -rf ${deb_dir}
-mkdir -p ${unpacked_src_dir}
-cp -r dist/calico-felix ${unpacked_src_dir}/pyi
+mkdir -p ${unpacked_src_dir}/bin
+cp bin/calico-felix ${unpacked_src_dir}/bin
+cp etc/calico-diags ${unpacked_src_dir}/bin
+cp etc/*.sh ${unpacked_src_dir}/bin
```
fasaxc: Maybe, but not under this PR.

rpm/felix.spec
```diff
@@ -138,7 +125,6 @@ rm -rf $RPM_BUILD_ROOT
 %files -n calico-common
 %defattr(-,root,root,-)
 /usr/bin/calico-diags
-/usr/bin/calico-cleanup
```
fasaxc: That was a python tool so it was removed with the clear-out.  Could put it back but I'd want to write a go version rather than maintain the python one.

calc/async_decoupler.go
fasaxc: Each project seems to have its own convention here; I think our other projects do it this way.

Makefile
neiljerram: OK

Makefile
fasaxc: Good spot; I just checked and I don't think it does.

calc/async_decoupler.go
neiljerram: https://golang.org/doc/code.html supports our new arrangement, so I'm happy.

.gitignore
```diff
@@ -1,30 +1,46 @@
...
+
+# Patterns to match files from older releases (prevents lots of
+# untracked files from showing up when switching from a maintenance branch.
+
+# Python builds/UT data.
+.hypothesis/
+.coverage
 *_pb2.py
-.glide
+/env*
```
neiljerram: This one looks dangerous, e.g. would accidentally ignore a new 'environ' go package.  Can it be more specific?

fasaxc: Changed to `/env/`, which is the typical name for our virtual env.

fasaxc:
Great, I'm going to merge this now because it's blocking all the other mop-up I'd like to start on.

PR #1301, opened 2017-01-30
Switch to calico/go-build container

### LINKED ISSUE #1216: make release doesn't work on a mac
Make release fails on a mac for at least this reason:

```
# Check that the executable is correctly statically linked.
ldd bin/calico-felix | grep -q "not a dynamic executable"
/bin/sh: ldd: command not found
```

We should be containerizing all the things.

fasaxc:
- Switch to the shared `calico/go-build` container.
- Run make targets inside that container (which includes `bash`).
- Remove use of EDITOR from release script.  Allows for release script to be run inside container, which fixes `make release` on Mac.
- Moving to `calico/go-build` seems to fix the issue that made the automated license check PR fail so I've incorporated that here.

Fixes #1216


Makefile
```diff
@@ -7,7 +7,7 @@
 #                                                                           |
 #                                                    +-------+              v
 #                                                    | Felix |   +---------------------+
-#                                                    |  Go   |   | calico-build/golang |
+#                                                    |  Go   |   | calico/go-build     |
```
fasaxc: Agree, that would be better but this container is already in use so it's not easy to change.

Makefile
```diff
@@ -149,6 +140,24 @@ calico/felix: bin/calico-felix
 DOCKER_RUN_RM:=docker run --rm --user $(MY_UID):$(MY_GID) -v $${PWD}:/code
 DOCKER_RUN_RM_ROOT:=docker run --rm -v $${PWD}:/code
 
+DOCKER_GO_BUILD_PFX := mkdir -p .go-pkg-cache && \
+                       docker run --rm \
+                                  --net=host \
+                                  -e LOCAL_USER_ID=$(MY_UID) \
+                                  -v $${PWD}:/code \
+                                  -v $${PWD}:/go/src/github.com/projectcalico/felix:rw \
```
fasaxc: > Why do we map the repo into the container under two places?

Historic artefact, I don't think it's needed any more.  I'll try to clean it up...

> make the path inside the container exactly the same as that outside.

I think that'd take quite a bit of work to make work.  (The build container already has a gopath set up that we inherit so we'd have to do more work in building the container, then do a bunch of symlinking or something).  Not sure it's worth it.

Makefile
```diff
@@ -149,6 +140,24 @@ calico/felix: bin/calico-felix
...
+                                  -v $${PWD}/.go-pkg-cache:/go/pkg/:rw \
```
fasaxc: I don't think so, will clean up

Makefile
neiljerram: OK

Makefile
```diff
@@ -249,52 +232,34 @@ dist/calico-felix/calico-felix: bin/calico-felix
...
-	$(MAKE) calico-build/golang
-	mkdir -p .go-pkg-cache
-	$(DOCKER_RUN_RM) \
-	    --net=host \
-	    -v $${PWD}:/go/src/github.com/projectcalico/felix:rw \
-	    -v $${PWD}/.go-pkg-cache:/go/pkg/:rw \
-	    -w /go/src/github.com/projectcalico/felix \
-	    calico-build/golang \
-	    ./utils/run-coverage
+	$(DOCKER_GO_BUILD) ./utils/run-coverage
```
fasaxc: Moved

Makefile
```diff
@@ -249,52 +228,34 @@ dist/calico-felix/calico-felix: bin/calico-felix
...
+
+.PHONY: check-licenses
+check-licenses: licensecheck/dependency-licenses.txt
+	@echo Checking dependency licenses
+	$(DOCKER_GO_BUILD) ginkgo licensecheck
 
 .PHONY: ut-no-cover
-ut-no-cover: vendor/.up-to-date $(GO_FILES)
+ut-no-cover: vendor/.up-to-date $(GO_FILES) licensecheck/dependency-licenses.txt
```
neiljerram: I believe dependency-licenses.txt should not now be a dependency of the UT targets - right?

PR #1304, opened 2017-01-31
Add extra prometheus metrics

### LINKED ISSUE #1291: Add more prometheus stats for scale testing
- [x] Current number of each resource being tracked on that node.
- [x] Number of policies which apply to endpoints on this node.
- [x] Error counts
- [x] Number of dataplane programming errors
- [x] Number of retries in various places.
- [x] Number of active selectors
- [x] Number of active ipsets
- [x] Number of iptables rules in place.

These are tricky to get:

- [ ] Latest index applied to the dataplane
- [ ] Size of largest ipset
- [ ] Etcd poll index

fasaxc:
- [x] Current number of endpoints/policies being tracked on eahc
- [x] Number of policies which apply to endpoints on this node.
- [x] Error counts
- [x] Number of dataplane programming errors
- [x] Number of retries in various places.
- [x] Number of active selectors
- [x] Number of active ipsets
- [x] Number of iptables rules in place.
- [x] Number of updates by type
- [x] Histograms for time taken to process updates in the graph and dataplane.

Fixes #1291

### DIFF
--- a/calc/async_calc_graph.go
+++ b/calc/async_calc_graph.go
@@ -21,6 +21,7 @@ import (
 	"github.com/projectcalico/felix/proto"
 	"github.com/projectcalico/libcalico-go/lib/backend/api"
 	"github.com/prometheus/client_golang/prometheus"
+	"reflect"
 	"time"
 )
 
@@ -43,11 +44,26 @@ var (
 		api.ResyncInProgress: 2,
 		api.InSync:           3,
 	}
+	countUpdatesProcessed = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Name: "felix_calc_graph_updates_processed",
+		Help: "Number of datastore updates processed by the calculation graph.",
+	}, []string{"type"})
+	countOutputEvents = prometheus.NewCounter(prometheus.CounterOpts{
+		Name: "felix_calc_graph_output_events",
+		Help: "Number of events emitted by the calculation graph.",
+	})
+	histUpdateTime = prometheus.NewHistogram(prometheus.HistogramOpts{
+		Name: "felix_calc_graph_update_time_seconds",
+		Help: "Seconds to update calculation graph for each datastore OnUpdate call.",
+	})
 )
 
 func init() {
 	prometheus.MustRegister(dataplaneStatusGauge)
 	prometheus.MustRegister(resyncsStarted)
+	prometheus.MustRegister(countUpdatesProcessed)
+	prometheus.MustRegister(countOutputEvents)
+	prometheus.MustRegister(histUpdateTime)
 }
 
 type AsyncCalcGraph struct {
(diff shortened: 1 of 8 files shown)


calc/async_calc_graph.go
```diff
@@ -99,7 +115,18 @@ func (acg *AsyncCalcGraph) loop() {
 			case []api.Update:
 				// Update; send it to the dispatcher.
 				log.Debug("Pulled []KVPair off channel")
+				updStartTime := time.Now()
 				acg.Dispatcher.OnUpdates(update)
+				updEndTime := time.Now()
+				if updEndTime.After(updStartTime) {
```
neiljerram: Worth commenting that this is to do with protecting against time jumping backwards.

What if it jumps forwards, though?

calc/async_calc_graph.go
```diff
@@ -99,7 +115,18 @@ func (acg *AsyncCalcGraph) loop() {
...
+					histUpdateTime.Observe(updEndTime.Sub(updStartTime).Seconds())
+				}
+				// Record stats for the number of messages processed.
+				for _, upd := range update {
+					typeName := reflect.TypeOf(upd.Key).Name()
+					count := countUpdatesProcessed.WithLabelValues(typeName)
+					count.Add(float64(len(update)))
```
neiljerram: Should that be `upd` rather than `update`?

intdataplane/int_dataplane.go
```diff
@@ -406,11 +430,24 @@ func (d *InternalDataplane) loopUpdatingDataplane() {
 		}
 
 		if datastoreInSync && d.dataplaneNeedsSync {
+			applyStart := time.Now()
 			d.apply()
+			applyEnd := time.Now()
+			if applyEnd.After(applyStart) {
```
neiljerram: Same here as in the other similar case.

iptables/table.go
```diff
@@ -263,6 +308,10 @@ func NewTable(
 
 		newCmd: newCmd,
 		sleep:  sleep,
+
+		gaugeNumChains:        gaugeNumChains.WithLabelValues(fmt.Sprintf("%d", ipVersion), name),
```
neiljerram: Would `string(ipVersion)` work here too?  (As you used above)  I think that's nicer.

iptables/table.go
```diff
@@ -289,7 +341,14 @@ func (t *Table) UpdateChains(chains []*Chain) {
 
 func (t *Table) UpdateChain(chain *Chain) {
 	t.logCxt.WithField("chainName", chain.Name).Info("Queueing update of chain.")
+	oldChain := t.chainNameToChain[chain.Name]
+	oldNumRules := 0
+	if oldChain != nil {
```
neiljerram: Perhaps be more idiomatic by moving the oldChain assignment into the previous line?

iptables/table.go
fasaxc: Unfortunately not, the other one is already a typedeffed string, I don't think you can stringify a number like that.

fasaxc:
retest this please

PR #1307, opened 2017-01-31
Add datamodel overview to API doc.

fasaxc:


proto/doc.go
```diff
@@ -39,7 +39,73 @@
 //	                    |
 //	         ======= Dataplane ==========
 //
-// Overview
+//
+// Data-model Overview
```
neiljerram: I think "data model".

proto/doc.go
```diff
@@ -39,7 +39,73 @@
...
+//
+// The data-model used on the dataplane driver API uses similar concepts to
+// the main "datastore" data-model (such as host/workload endpoints, policies
+// and profiles).  However, the calculation engine does some pre-calculation
+// that simplifies the job of the dataplane driver:
```
neiljerram: And it uses different exact terminology in various places (e.g. untracked vs doNotTrack).  Is that worth mentioning as an overall point?

proto/doc.go
```diff
@@ -39,7 +39,73 @@
...
+//
+// Rules in the datastore data-model can contain selectors, such as
+// "role == webserver", that refer to dynamic groups of endpoints.  The calculation
+// engine pre-renders those selectors (and the older tag mechanism) to sets of
```
neiljerram: 'compute' would be clearer to me than 'render'.  (+ another occurrence following)

proto/doc.go
```diff
@@ -39,7 +39,73 @@
...
+// IP addresses.  Only the sets of IPs are sent over the API so the dataplane
+// driver doesn't need to render selectors itself, it only needs a way to
+// represent a set of IP addresses on which to match.
```
neiljerram: 'program that set of IP addresses into the dataplane'?

proto/doc.go
```diff
@@ -39,7 +39,73 @@
...
+//
+// Policies in the datastore data-model need to be filtered and sorted. The
+// calculation engine takes care of that too.  When it sends an endpoint to the
+// datastore driver, it adds the complete list of policies to apply to the
```
neiljerram: 'that apply to'

PR #1308, opened 2017-01-31
Fix incorrect quotes in find calls.

fasaxc:
Should be single quotes to avoid expansion.


fasaxc:
CI failure was `test_floating_ip`, which is unrelated.

neiljerram:
Actually double quotes are already fine here, so closing this (with @fasaxc agreement).

PR #1313, opened 2017-02-01
Implement periodic refresh of route table.

### LINKED ISSUE #1185: Routes to devices are not restored automatically
Routes to "cali*" interfaces are not monitored so they could be removed accidentally which leads to unreachable workload (pod).

**Steps to reproduce:**

on a k8s+calico node remove any route to endpoint:
```
$ ip ro | grep cali
10.233.120.130 dev calia44f8339e4a  scope link 
$ ip ro del 10.233.120.130
```

**Expected result:**
route recovers

**Actual result:**
route is missing

**Version**
calico/node:v1.0.0-beta

**Workaround:**
On a problem node execute:
```
docker exec -t -i calico-node killall -HUP calico-felix
```

fasaxc:
Fixes #1185

### DIFF
--- a/routetable/route_table.go
+++ b/routetable/route_table.go
@@ -122,6 +122,11 @@ func (r *RouteTable) SetRoutes(ifaceName string, targets []Target) {
 	r.dirtyIfaces.Add(ifaceName)
 }
 
+func (r *RouteTable) QueueResync() {
+	r.logCxt.Info("Queueing a resync.")
+	r.inSync = false
+}
+
 func (r *RouteTable) Apply() error {
 	if !r.inSync {
 		links, err := r.dataplane.LinkList()
--- a/routetable/route_table_test.go
+++ b/routetable/route_table_test.go
@@ -71,7 +71,7 @@ var _ = Describe("RouteTable", func() {
 
 	Describe("with some interfaces", func() {
 		var cali1, cali2, cali3, eth0 *mockLink
-		var gatewayRoute, cali1Route, cali3Route netlink.Route
+		var gatewayRoute, cali1Route, cali1Route2, cali3Route netlink.Route
 		BeforeEach(func() {
 			eth0 = dataplane.addIface(0, "eth0", true, true)
 			cali1 = dataplane.addIface(1, "cali1", true, true)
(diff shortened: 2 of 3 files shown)


routetable/route_table.go
```diff
@@ -122,6 +122,11 @@ func (r *RouteTable) SetRoutes(ifaceName string, targets []Target) {
 	r.dirtyIfaces.Add(ifaceName)
 }
 
+func (r *RouteTable) QueueResync() {
+	r.logCxt.Info("Queueing a resync.")
+	r.inSync = false
```
neiljerram: How soon after this will Apply() be called?  It would be a shame if there was still a significant delay before a missing or superfluous route was corrected.

fasaxc: Immediately, but `QueueResync` will only be called once every 30s.  

A community member put up a PR to monitor routes, which would improve that further: https://github.com/projectcalico/felix/pull/1316  but I see value in a periodic resync too; it's to clear to me that the netlink API is race-proof so I'd like to put this patch in too.

routetable/route_table_test.go
```diff
@@ -180,6 +180,31 @@ var _ = Describe("RouteTable", func() {
...
+						Expect(dataplane.routeKeyToRoute).To(HaveLen(5))
+						Expect(dataplane.routeKeyToRoute).To(ContainElement(cali1Route2))
+					})
+					It("after a QueueResync() should remove the route", func() {
+						rt.QueueResync()
+						rt.Apply()
+						Expect(dataplane.routeKeyToRoute).To(HaveLen(4))
+						Expect(dataplane.routeKeyToRoute).NotTo(ContainElement(cali1Route2))
+					})
+				})
```
fasaxc: Done.

routetable/route_table.go
neiljerram: 30s seems a bit long, but I guess acceptable considering that we're dealing with Enemy Action :)

neiljerram: (+ agree that netlink monitoring would be a big increase in complexity here, so I'm inclined against it unless we really need it)

PR #1318, opened 2017-02-02
Explicitly ACCEPT packets that are allowed by host policy.

### LINKED ISSUE #1317: How to allow traffic originating from mesos-containers to 'static' applications?
Hi,

I've got a rather lengthy question about allow traffic originating from mesos-containers (running docker with the libnetwork-plugin) connecting to (legacy) statically configured services.

What we did to (try to) achieve this is:
- the mesos-cluster is running the full calico-stack
- the static host is running calico-felix.

On the static host, we create a hostendpoint with configuration for the IP-address, and set a couple of labels:

```- apiVersion: v1
  kind: hostEndpoint
  metadata:
    labels:
      application: tetris
      depends_on_xpr-pcs-app: "true"
      environment: xpr
      role: tetris-app
    name: eth0
    node: xpr-tetris-app-001.bolcom.net
  spec:
    expectedIPs:
    - 10.98.194.221
```

We have a general policy that egress traffic is allowed, so the outbound traffic from the container is allowed:
```- apiVersion: v1
  kind: policy
  metadata:
    name: allow_outgoing_generic
  spec:
    egress:
    - action: allow
      destination: {}
      source: {}
    selector: ""
```

For allowing traffic to this specific (static) hostendpoint, we created a policy:

```- apiVersion: v1
  kind: policy
  metadata:
    name: xpr-tetris-app
  spec:
    ingress:
    - action: allow
      destination:
        ports:
        - 8009
        - 8080
      protocol: tcp
      source:
        selector: has(depends_on_xpr-tetris-app)
    order: 100
    selector: environment == 'xpr' && application == 'tetris' && role == 'tetris-app'
```

On the containers where we want to connect to this hostendpoint, we have set this label:

```- apiVersion: v1
  kind: workloadEndpoint
  metadata:
    labels:
      application: frieps
      depends_on_xpr-int-lb: "true"
      depends_on_xpr-mesos-master: "true"
      depends_on_xpr-tetris-app: "true"
      environment: xpr
    name: 0333d4e9db950c4cf6cf101d8d8d9980416f1c2622d548b6edf3813907248520
    node: xpr-mesos-slave-003.bolcom.net
    orchestrator: libnetwork
    workload: libnetwork
  spec:
    interfaceName: cali0333d4e9db9
    ipNetworks:
    - 10.102.0.33/32
    mac: ee:ee:ee:ee:ee:ee
    profiles:
    - xpr-mesos-cluster
```

As you can see, the depends_on_xpr-tetris-app label has been set, which theoretically allows traffic from this container/workload to the hostendpoint.

tcpdump shows the traffic as successfully arriving on the destination host, and it's being processed by the calico-felix iptables rules. But although it hits the `-A felix-from-eth1 -m mark --mark 0x1000000/0x1000000 -m comment --comment "Return if policy accepted" -j RETURN` rule, eventually the traffic never gets accepted.
We've tested with the DefaultEndpointToHostAction configuration value to different settings, but none of them configures it the way we would like to see, which would be:

- allow traffic if the correct label is set
- don't allow traffic if that's not the case
- respect our general iptables-rulesets (we use the `ChainInsertMode = append` to put the felix rules to the end of the chains), which includes a default policy on the INPUT chain of 'DROP' in stead of 'ACCEPT'.

Are you able to help us out in whether we are doing something wrong, unsupported, or hit a bug? We tried with both the new 2.1.0-rc1, and the older 2.0.0-versions of felix.

I'm available on slack as @wietse for helping to give more information if needed - but this story was quite long, so I decided it would be best to type here first :)

fasaxc:
Makes Calico compatible with using DROP as the iptables policy.

Fixes #1317


wmuizelaar:
Confirmed as working.

PR #1320, opened 2017-02-02
Plumb through Ipv6Support flag and ReportingIntervalSecs.

fasaxc:


wmuizelaar:
Confirmed as working.

Issue #1325, opened 2017-02-03
Felix and kube-proxy can clobber each other's iptables updates


fasaxc:
Seen in a scale run with kube-proxy 1.5.2, this script reproduces the issue:
```
#!/usr/bin/env bash

failed=false

function felix () {
    while true; do
        echo -n a
        iptables-restore --noflush --verbose <<EOF
*filter
:felix -
--append felix -m comment --comment "FELIX-A"
COMMIT
EOF
        echo -n A
        echo -n b
        iptables-restore --noflush --verbose <<EOF
*filter
:felix -
--append felix -m comment --comment "FELIX-B"
COMMIT
EOF
        echo -n B
        if iptables-save | grep -q FELIX-A; then
          echo
          echo "FELIX-B update was clobbered"
          failed=true
          return 1
        fi
        echo -n C
    done
}

function kube () {
    while ! ${failed}; do
       echo -n k
       iptables-restore --noflush --verbose <<EOF
*filter
:kube -
COMMIT
EOF
        echo -n K
    done
}


kube &
kube_pid=$!
felix; kill $kube_pid

```

When I run that locally, I see output like this:

```
# ./ipt-clash-repro.sh 
akKkAbKkBCaKkAbKkBKkCaKkAbKkBKkCaKkAbKkKkKkBKkCaKkAbBKk
FELIX-B update was clobbered
```

Confusingly, however, this script, which is identical apart from adding an extra update to the FELIX-B iptables-restore call, exhibits compare-and-swap behaviour.  The update is never clobbered but the kube loop's updates fail if there's a concurrent modification:

```
#!/usr/bin/env bash

failed=false

function felix () {
    while true; do
        echo -n a
        iptables-restore --noflush --verbose <<EOF
*filter
:felix -
--append felix -m comment --comment "FELIX-A"
COMMIT
EOF
        echo -n A
        echo -n b
        iptables-restore --noflush --verbose <<EOF
*filter
:felix -
--append felix -m comment --comment "FELIX-B"
--append felix -m comment --comment "extra update"
COMMIT
EOF
        echo -n B
        if iptables-save | grep -q FELIX-A; then
          echo
          echo "FELIX-B update was clobbered"
          failed=true
          return 1
        fi
        echo -n C
    done
}

function kube () {
    while ! ${failed}; do
       echo -n k
       iptables-restore --noflush --verbose <<EOF
*filter
:kube -
COMMIT
EOF
        echo -n K
    done
}


kube &
[...]

PR #1326, opened 2017-02-03
Aggressively re-check iptables after an update.

fasaxc:
- Add ability for int_dataplane to reschedule on a timer if requested by `iptables.Table`.
- After an iptables update, reschedule at exponential intervals to check that the update is still present.
- Reload dataplane state before each update.  (This is what kube-proxy does too, it probably doesn't slow us down that much.)

### DIFF
--- a/intdataplane/int_dataplane.go
+++ b/intdataplane/int_dataplane.go
@@ -607,10 +607,10 @@ func (d *InternalDataplane) apply() {
 		d.cleanupPending = false
 	}
 
-	// Set up any needed rescheduling kick.  Any pending kick will have been recalculated above
-	// so we always cancel and reschedule.
+	// Set up any needed rescheduling kick.
 	if d.reschedC != nil {
-		// We have an active rescheduling timer, stop it so we can reset it below if needed.
+		// We have an active rescheduling timer, stop it so we can restart it with a
+		// different timeout below if it is still needed.
 		// This snippet comes from the docs for Timer.Stop().
 		if !d.reschedTimer.Stop() {
 			// Timer had already popped, drain its channel.
--- a/intdataplane/int_dataplane.go
+++ b/intdataplane/int_dataplane.go
@@ -130,6 +130,9 @@ type InternalDataplane struct {
 	forceDataplaneRefresh bool
 	cleanupPending        bool
 
+	reschedTimer *time.Timer
+	reschedC     <-chan time.Time
+
 	config Config
 }
 
@@ -162,6 +165,7 @@ func NewIntDataplaneDriver(config Config) *InternalDataplane {
 			HistoricChainPrefixes:    rules.AllHistoricChainNamePrefixes,
 			ExtraCleanupRegexPattern: rules.HistoricInsertedNATRuleRegex,
 			InsertMode:               config.IptablesInsertMode,
+			RefreshInterval:          config.IptablesRefreshInterval,
 		},
 	)
 	rawTableV4 := iptables.NewTable(
@@ -171,6 +175,7 @@ func NewIntDataplaneDriver(config Config) *InternalDataplane {
 		iptables.TableOptions{
 			HistoricChainPrefixes: rules.AllHistoricChainNamePrefixes,
 			InsertMode:            config.IptablesInsertMode,
+			RefreshInterval:       config.IptablesRefreshInterval,
 		})
 	filterTableV4 := iptables.NewTable(
 		"filter",
@@ -179,6 +184,7 @@ func NewIntDataplaneDriver(config Config) *InternalDataplane {
 		iptables.TableOptions{
 			HistoricChainPrefixes: rules.AllHistoricChainNamePrefixes,
 			InsertMode:            config.IptablesInsertMode,
+			RefreshInterval:       config.IptablesRefreshInterval,
 		})
 	ipSetsConfigV4 := config.RulesConfig.IPSetConfigV4
 	ipSetRegV4 := ipsets.NewRegistry(ipSetsConfigV4)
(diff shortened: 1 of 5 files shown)


fasaxc:
Should help with #1325

iptables/table.go
```diff
@@ -214,6 +214,12 @@ type Table struct {
 	// to top-level chains.
 	insertMode string
 
+	// Record when we did our most recent updates and refreshes of the table.  We use these to
+	// calculate the next time we should force a refresh.
+	lastUpdateTime  time.Time
+	lastRefreshTime time.Time
```
neiljerram: Can we rename 'refresh' to 'read' and 'update' to 'write'?  Until I had read more of the code below, I couldn't tell the difference between 'update' and 'refresh'.

intdataplane/int_dataplane.go
```diff
@@ -558,14 +572,15 @@ func (d *InternalDataplane) apply() {
 			// Queue a resync on the next Apply().
 			r.QueueResync()
 		}
-		for _, t := range d.allIptablesTables {
-			t.InvalidateDataplaneCache()
-		}
```
neiljerram: Why should we not still invalidate all the tables here?

intdataplane/int_dataplane.go
```diff
@@ -591,6 +606,31 @@ func (d *InternalDataplane) apply() {
 		}
 		d.cleanupPending = false
 	}
+
+	// Set up any needed rescheduling kick.  Any pending kick will have been recalculated above
+	// so we always cancel and reschedule.
```
neiljerram: I don't understand what the last sentence here is trying to tell me.  What would be the alternative to cancelling and rescheduling, and why does choosing or not choosing that depend on having calculated the reschedDelay above?

intdataplane/int_dataplane.go
```diff
@@ -591,6 +606,31 @@ func (d *InternalDataplane) apply() {
...
+		}
+		// Nil out our copy of the channel to record that the timer is inactive.
+		d.reschedC = nil
+	}
+	if reschedDelay != 0 {
+		// We need to reschedule.
+		log.WithField("delay", reschedDelay).Debug("Asked to reschedule.")
+		if d.reschedTimer == nil {
+			// First time, create the timer.
+			d.reschedTimer = time.NewTimer(reschedDelay)
```
neiljerram: Is it possible to create a timer without actually starting it?  If so, I think it would be nicer to do this in the dataplane constructor, instead of JIT here.

iptables/table.go
neiljerram: (Or perhaps just 'refresh' to 'read' or 'check', if you don't want to churn pre-existing uses of 'update'.)

iptables/table.go
```diff
@@ -614,6 +699,43 @@ func (t *Table) Apply() {
 	}
 
 	t.gaugeNumChains.Set(float64(len(t.chainNameToChain)))
+
+	// Check whether we need to be rescheduled.
```
neiljerram: and how soon

iptables/table.go
```diff
@@ -560,11 +596,60 @@ func (t *Table) getHashesFromBuffer(buf *bytes.Buffer) map[string][]string {
...
+		200 * time.Millisecond,
+		100 * time.Millisecond,
+		50 * time.Millisecond,
+	}
+)
+
+func (t *Table) Apply() (rescheduleAfter time.Duration) {
+	if t.inSyncWithDataPlane {
+		// We _think_ we're in sync, check if there are any reasons to think we might
+		// not be in sync.
```
neiljerram: I'm thinking there must be a simpler alternative to this code - which is quite complex, and more or less a duplicate of that at the end of Apply.

Idea 1: can the IntDataplane invalidate before calling Apply, when it knows it's handling a resched kick?  But I think no, it can't, because IntDataplane schedules a kick for all tables at the minimum of the needed resched kick for all of them.  So if we did that, we'd be refreshing tables that didn't need it.

Idea 2: when it runs the code at the end of Apply, can the Table store off the time when it expects a refresh kick?  Then when Apply is called, if now >= that time, it invalidates.

iptables/table_test.go
```diff
@@ -348,6 +390,138 @@ var _ = Describe("Table with an empty dataplane", func() {
...
+			rules.RuleHashPrefix,
+			options,
+		)
+		table.SetRuleInsertions("FORWARD", []Rule{
+			{Action: DropAction{}},
+		})
+		table.Apply()
+	})
+
+	advanceAndReset := func(amount time.Duration) func() {
```
neiljerram: resetAndAdvance?

iptables/table_test.go
```diff
@@ -348,6 +390,138 @@ var _ = Describe("Table with an empty dataplane", func() {
...
+		return func() {
+			requestedDelay = -1 * time.Second
```
neiljerram: delete?  I'm not sure how this line can be doing anything.

iptables/table.go
fasaxc: Sounds good, read and write are clearer

iptables/table.go
fasaxc: Yes, I wasn't too happy with it but also struggling to see a great way to simplify; I'll noodle it some more

intdataplane/int_dataplane.go
fasaxc: Only because the table now handles that via the rescheduling mechanism.

intdataplane/int_dataplane.go
fasaxc: I was trying to explain that we don't need a queue of ticks because the iptables.Table objects will recalculate any rescheduling that they need each time.

intdataplane/int_dataplane.go
fasaxc: No, I couldn't seem to find a way to do that.

iptables/table_test.go
fasaxc: Requirement was removed while refactoring

iptables/table.go
fasaxc: Went with storing off the interval in a field and simply doubling it each time.  Seems to be simplest option.

fasaxc:
@neiljerram Markups done, back to you for re-review

intdataplane/int_dataplane.go
neiljerram: As discussed - the point here is that we still need the dataplane-level force-refresh concept for the route table.

fasaxc:
retest this please

PR #1329, opened 2017-02-06
Handle interfaces being renamed in interface monitor.

### LINKED ISSUE #1328: Host endpoint resolution confused by interface renaming
Investigating the flaky test https://github.com/projectcalico/calicoctl/issues/1492, turns out to be a real failure:

- The test adds a second veth to the gateway `DockerHost` container.
- When adding a veth to the container, felix sees the veth appear with name `veth12345`, then libcalico-go renames it to `eth1` but felix seems to continue thinking the veth has the old name.
- Endpoint never resolves.

fasaxc:
Fixes #1328 by simulating a delete for the old name before adding the new.


fasaxc:
retest this please

PR #1337, opened 2017-02-13
Cleanup unused policy rendering code

fasaxc:
Remove unhittable policy rendering code and clean up naming/interfaces.

### DIFF
--- a/rules/endpoints_test.go
+++ b/rules/endpoints_test.go
@@ -22,17 +22,16 @@ import (
 
 	"github.com/projectcalico/felix/ipsets"
 	. "github.com/projectcalico/felix/iptables"
-	"github.com/projectcalico/felix/proto"
 )
 
 var _ = Describe("Endpoints", func() {
 	var rrConfigNormal = Config{
-		IPIPEnabled:          true,
-		IPIPTunnelAddress:    nil,
-		IPSetConfigV4:        ipsets.NewIPVersionConfig(ipsets.IPFamilyV4, "cali", nil, nil),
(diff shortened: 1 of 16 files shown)


rules/endpoints_test.go
```diff
@@ -288,10 +226,10 @@ var _ = Describe("Endpoints", func() {
 
 					{Action: ClearMarkAction{Mark: 0x8}},
 
-					{Comment: "Start of tier tier2",
+					{Comment: "Start of policies",
 						Action: ClearMarkAction{Mark: 0x10}},
 					{Match: Match().MarkClear(0x10),
-						Action: JumpAction{Target: "cali-po-tier2/c"}},
+						Action: JumpAction{Target: "cali-po-/c"}},
```
neiljerram: That looks a little odd now; should it be simplified to cali-po-<name> ?

rules/policy.go
```diff
@@ -158,9 +158,9 @@ func (r *DefaultRuleRenderer) CalculateActions(match iptables.MatchCriteria, pRu
 		mark = r.IptablesMarkAccept
 		actions = append(actions, iptables.ReturnAction{})
 	case "next-tier":
```
neiljerram: Add a case to allow "pass" as well, as a synonym for "next-tier"?

rules/endpoints_test.go
fasaxc: The leading slash distinguishes between policies and profiles with the same name.  Was just trying to keep as much room for the name.

neiljerram: Ah, OK.

PR #1339, opened 2017-02-15
Propagate datastore type from Felix config to CalicoAPIConfig

neiljerram:
This seems to be needed when starting calico/felix with
FELIX_DATASTORETYPE=kubernetes so as to select the k8s datastore
backend.  Without this, libcalico-go/lib/backend.NewClient() still
constructs and returns an etcd v2 client.

### DIFF
--- a/config/config_params.go
+++ b/config/config_params.go
@@ -317,6 +317,7 @@ func (config *Config) DatastoreConfig() api.CalicoAPIConfig {
 		if err != nil {
 			log.WithError(err).Panic("Failed to create empty config")
 		}
+		cfg.Spec.DatastoreType = api.Kubernetes
 		return *cfg
 	} else {
 		var etcdEndpoints string


config/config_params.go
```diff
@@ -317,6 +317,7 @@ func (config *Config) DatastoreConfig() api.CalicoAPIConfig {
 		if err != nil {
```
caseydavenport: Hmm.. this code is assuming that `DATASTORE_TYPE=kubernetes` is set.  I suppose the issue is that `config.DatastoreType` can be `kubernetes` without that environment variable set.

Changing the `cfg.Spec.DatastoreType` below is only half of the problem - there's still no way to configure things like where the k8s API is from say, a Felix config file.

neiljerram: Well, when I was playing with the k8s backend (prior to our chat earlier today), my complete invocation was:

	docker run --rm \
	-e CALICO_K8S_API_ENDPOINT=scale-test \
	-e FELIX_DATASTORETYPE=kubernetes \
	-e FELIX_LOGSEVERITYSCREEN=debug \
	calico/felix

In other words, if FELIX_DATASTORETYPE works to select the k8s datastore, you can then also use variables like CALICO_K8S_API_ENDPOINT to configure fields of the KubeConfig.  With the fix that I've proposed, that all seems to hang together and make sense.

I haven't yet noticed any doc or code based on DATASTORE_TYPE (without FELIX_ or CALICO_ prefix) - are you sure about that?

caseydavenport: Yeah, the environment variables all work with the `CALICO_` prefix, or without (with the prefix takes precedence).  Here's an example: http://docs.projectcalico.org/v2.0/reference/calicoctl/setup/kubernetes

Note:

> Note that all environment variables may also be prefixed with “CALICO_”, for example “CALICO_DATASTORE_TYPE” and “CALICO_KUBECONFIG” etc. may also be used. This is useful if the non-prefixed names clash with existing environment variables defined on your system


In `calico/node`, we pass both essentially to work around the same thing you described above: https://github.com/projectcalico/calicoctl/blob/master/calico_node/filesystem/etc/service/available/felix/run#L15

neiljerram: Ah, I see now; and also

	DatastoreType DatastoreType `json:"datastoreType" envconfig:"DATASTORE_TYPE" default:"etcdv2"`

OK, so this PR is wrong then, I'll close it.  Instead, in the longer term, I guess it would be nice if we could remove the need for setting FELIX_DATASTORETYPE as well as DATASTORE_TYPE or CALICO_DATASTORE_TYPE.

PR #1341, opened 2017-02-15
Generalize and explain datastore config construction

neiljerram:
Following discussion in PR #1339, this change clarifies how and why datastore
config is constructed, and generalizes the code so that it's ready for
other future datastore types.  (Specifically the test one that I am
also working on.)

### DIFF
--- a/config/config_params.go
+++ b/config/config_params.go
@@ -310,15 +310,11 @@ func (config *Config) EndpointReportingDelay() time.Duration {
 }
 
 func (config *Config) DatastoreConfig() api.CalicoAPIConfig {
-	if config.DatastoreType == "kubernetes" {
-		// Create a new Client.  The client will be configured
-		// based on the provided environment.
-		cfg, err := client.LoadClientConfig("")
-		if err != nil {
-			log.WithError(err).Panic("Failed to create empty config")
-		}
-		return *cfg
-	} else {
+	// Special case for etcdv2 datastore, where we want to honour established Felix-specific
+	// config mechanisms.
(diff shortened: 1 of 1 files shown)


config/config_params.go
```diff
@@ -338,6 +334,22 @@ func (config *Config) DatastoreConfig() api.CalicoAPIConfig {
...
 	}
+
+	// Build CalicoAPIConfig from the environment.  This means that any XxxYyy field in
+	// CalicoAPIConfigSpec can be set by a corresponding XXX_YYY or CALICO_XXX_YYY environment
+	// variable, and that the datastore type can be set by a DATASTORE_TYPE or
+	// CALICO_DATASTORE_TYPE variable.  (Except in the etcdv2 case which is handled specially
+	// above.)
+	cfg, err := client.LoadClientConfigFromEnvironment()
+	if err != nil {
+		log.WithError(err).Panic("Failed to create datastore config")
```
fasaxc: Wonder if we should return an error and then shut down gracefully, presumably we'd hit this case because the config is invalid.

The panic was reasonable before because we were trying to create an empty config with no external input.  If that failed, it'd be a programming error.

neiljerram: I don't think the situation is actually any different from what it was before; I just simplified the code a bit.  The previous code `client.LoadClientConfig("")` is equivalent to `client.LoadClientConfigFromEnvironment()` (because no file is specified), and I think the latter would be clearer.

fasaxc: Sorry, was getting confused with another function that's similar.

PR #1350, opened 2017-02-17
Add log to indicate end of initial sync.

### LINKED ISSUE #1345: Emit clear log when Felix is shutting down, done re-syncing.
For example, when running as a Kubernetes pod, it would be useful to get timestamps for when the kubelet killed the Calico pod on a given node, and then when the kubelet started the Pod again.

Basically, we want to know the difference in time between:
- Felix gets shutdown signal.
- Felix comes back up, finishes syncing dataplane for first time.

This will help debug when Felix isn't running and thus policy isn't being updated for that node.

fasaxc:
Fixes #1345

### DIFF
--- a/intdataplane/int_dataplane.go
+++ b/intdataplane/int_dataplane.go
@@ -48,12 +48,15 @@ var (
 		Name: "felix_int_dataplane_apply_time_seconds",
 		Help: "Time in seconds that it took to apply a dataplane update.",
 	})
+
+	processStartTime time.Time
 )
 
 func init() {
 	prometheus.MustRegister(countDataplaneSyncErrors)
 	prometheus.MustRegister(histApplyTime)
 	prometheus.MustRegister(countMessages)
+	processStartTime = time.Now()
 }
 
 type Config struct {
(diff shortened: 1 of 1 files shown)


intdataplane/int_dataplane.go
```diff
@@ -463,16 +468,24 @@ func (d *InternalDataplane) loopUpdatingDataplane() {
...
+			applyTime := time.Since(applyStart)
+			if applyTime > 0 {
 				// Avoid a negative interval in case the clock jumps.
-				histApplyTime.Observe(applyEnd.Sub(applyStart).Seconds())
+				histApplyTime.Observe(applyTime.Seconds())
 			}
 			if d.dataplaneNeedsSync {
 				countDataplaneSyncErrors.Inc()
 			}
+			log.WithField("msecToApply", applyTime.Seconds()/1000.0).Info(
```
fasaxc: Note: I'm forcing these to msec/seconds to make Casey's parsing easier.  Without that, they'll pretty-print in an awkward-to-parse way.

neiljerram: So shouldn't that be *1000.0, instead of /1000.0 ?

fasaxc: Yes, oops!

PR #1352, opened 2017-02-17
Ignore removal of non-Calico chain.

### LINKED ISSUE #1349: Under churn with k8s, golang DP driver thinks KUBE-SEP chain out-of-sync
Under churn tests, we're seeing logs like this:
```
2017-02-16 22:24:01.608 [WARNING][32] table.go 444: Detected out-of-sync inserts, marking for resync actualRuleIDs=[]string(nil) chainName="KUBE-SEP-WE2O5CDAE2MDKF6F" expectedRuleIDs=[]string{} ipVersion=0x4 table="nat"
```
_Probably_ benign since the later stage of processing will spot that there's nothing to do but we should squash this false positive.

fasaxc:
Fixes #1349

### DIFF
--- a/iptables/table_test.go
+++ b/iptables/table_test.go
@@ -921,3 +921,55 @@ func describeDirtyDataplaneTests(appendMode bool) {
 		})
 	})
 }
+
+var _ = Describe("Table with inserts and a non-Calico chain", func() {
+	var dataplane *mockDataplane
+	var table *Table
+	BeforeEach(func() {
+		dataplane = newMockDataplane("filter", map[string][]string{
+			"FORWARD":    {},
+			"non-calico": {"-m comment \"foo\""},
+		})
+		table = NewTable(
+			"filter",
+			6,
+			rules.RuleHashPrefix,
+			TableOptions{
+				HistoricChainPrefixes: rules.AllHistoricChainNamePrefixes,
+				NewCmdOverride:        dataplane.newCmd,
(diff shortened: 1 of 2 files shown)


iptables/table_test.go
```diff
@@ -921,3 +921,51 @@ func describeDirtyDataplaneTests(appendMode bool) {
...
+			dataplane.Chains = map[string][]string{
+				"FORWARD": {"-m comment --comment \"cali:hecdSCslEjdBPBPo\" --jump DROP"},
+			}
+			table.Apply()
+		})
+
+		It("should ignore the deletion", func() {
+			Expect(dataplane.Chains).To(Equal(map[string][]string{
+				"FORWARD": {"-m comment --comment \"cali:hecdSCslEjdBPBPo\" --jump DROP"},
+			}))
```
neiljerram: Shouldn't we also test here that there isn't any refresh processing?

fasaxc: I couldn't see a way to verify that without adding a test-only flag of some sort.  Any ideas?

neiljerram: I thought that was what the infrastructure like

			dataplane.ResetCmds()

and

		Expect(dataplane.CmdNames).To(BeEmpty())

was for.  No?

fasaxc: Ah yep, I was thinking of how to make sure the old warning didn't get hit.

PR #1353, opened 2017-02-17
Decouple stats collector from usage reporter.

### LINKED ISSUE #1348: Spammy stats channel blocked warnings
```
10326 2017-02-16 20:59:04.664 [WARNING][32] stats_collector.go 154: Failed to report stats error=Stats channel blocked
```
Under heavy churn, we see the above warning in the logs, which seems to indicate that the usage reporting goroutine can't keep up.  That's fine, we'll just drop the update but it'd be good to understand it a bit better and make sure we're not losing all updates.

fasaxc:
Fixes #1348.

### DIFF
--- a/felix.go
+++ b/felix.go
@@ -15,7 +15,6 @@
 package main
 
 import (
-	"errors"
 	"fmt"
 	"net"
 	"net/http"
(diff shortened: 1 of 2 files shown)


felix.go
```diff
@@ -269,24 +268,46 @@ configRetry:
 	asyncCalcGraph := calc.NewAsyncCalcGraph(configParams, dpConnector.ToDataplane)
 
 	if configParams.UsageReportingEnabled {
-		// Usage reporting enabled, add stats collector to graph and
-		// start the usage reporting thread.
-		statsChan := make(chan calc.StatsUpdate, 1)
+		// Usage reporting enabled, add stats collector to graph.  When it detects an update
+		// to the stats, it makes a callback, which we use to send an update on a channel.
+		// We use a buffered channel here to avoid blocking the calculation graph.
+		statsChanIn := make(chan calc.StatsUpdate, 1)
```
neiljerram: Is there really still a reason for this 1-size buffer?

fasaxc: My thinking was that it means that the calc graph won't have to wait for the buffer goroutine to complete its loop (so we get a pipelining effect).  Could well believe it makes no real difference in practice though!

neiljerram: Fair enough.  Happy to check my understanding, but also happy to approve the change as is.

PR #1356, opened 2017-02-20
Throttle dataplane updates and opportunistically batch.

fasaxc:
Improves throughput when under heavy load.


fasaxc:
Note: on its own, this exposes a bug in the IP sets logic where it doesn't correctly handle the sequence: <IP not in set>, add IP, remove IP, add IP, apply.  (It tries to render a del for the IP, causing a failure and full rewrite of the IP set.)

PR #1357, opened 2017-02-20
Implement bulk updates to IP sets

### LINKED ISSUE #1347: ipsets restore sometimes loses an update
Seen in a simulated scale run with lots of churn.  From time to time, we hit an error where ipset claims that an IP can't be deleted from the set because it's not present.  I created a script that replayed the ipset interactions and it fails only some of the time, indicating that ipset or the kernel is at fault.

fasaxc:
-  Improve IP sets update performance by batching multiple set updates into single ipset restore calls.  
- Avoid buffering large amounts of input/output by doing streaming reads/writes from pipes to the child process.
- Replace `AttemptCleanup()` start-of-day cleanup routine with a resync mechanism that reads back the contents of the dataplane.  Works around / Fixes #1347. 

Design:

- Move write function to `Registry`, renamed as `IPSets`.
- Reduce IPSet struct to a data struct rather than a fully fledged object to avoid having an intricate coupling between the two objects.
- To support comparison during resync, convert IPs/CIDRs to canonical form and store them as ip.Addr/ip.CIDR objects instead.
- Remove IP set existence cache since logic is now rolled into resync.

TODO:

- [x] Reinstate UTs.


fasaxc:
retest this please

ipsets/ipsets.go
```diff
@@ -0,0 +1,810 @@
...
+		// - membersInDataplane non-nil
+		// - pendingAdds/Deletions hold the deltas.
+		if ipSet.pendingAdds.Len() == 0 && ipSet.pendingDeletions.Len() == 0 {
+			// We hit this case if an IP is added, then removed before we actually
+			// write it, nothing to do.
+			logCxt.Debug("Skipping delta write, IP set not dirty.")
+			return nil
+		}
+		logCxt.Info("Calculating deltas to IP set")
+		return s.writeDeltas(ipSet, w)
```
neiljerram: Pass down (and use) the logCxt here?

ipsets/ipsets.go
```diff
@@ -0,0 +1,810 @@
...
+	} else {
+		// In full-rewrite mode.
+		// - pendingReplace is non-nil
+		// - membersInDataplane nil
+		// - pendingAdds/Deletions empty.
+		logCxt.Info("Doing full IP set rewrite")
+		return s.writeFullRewrite(ipSet, w)
```
neiljerram: (similarly here)

ipsets/ipsets.go
```diff
@@ -0,0 +1,810 @@
...
+			return set.StopIteration
+		}
+		countNumIPSetLinesExecuted.Inc()
+		return nil
+	})
+	return
+}
+
+// ApplyDeletions tries to delete any IP sets that are no longer needed.
+// Failures are ignored, deletions will be retried the next time AttemptCleanup() is called.
```
neiljerram: s/AttemptCleanup/ApplyDeletions/

ipsets/ipsets.go
```diff
@@ -0,0 +1,810 @@
...
+		logCxt.Info("Deleting IP set (if it exists)")
+		if s.existingIPSetNames.Contains(setName) {
+			if err := s.deleteIPSet(setName); err != nil {
+				logCxt.WithError(err).Warning("Failed to delete IP set.")
+			}
+		}
+		return set.RemoveItem
+	})
+
+	s.gaugeNumIpsets.Set(float64(len(s.ipSetIDToIPSet)))
```
neiljerram: Presumably `s.ipSetIDToIPSet` should be decremented in this code path, but I don't see where that is happening.

fasaxc: I was just using this as a point in the code where we know we've got no updates in flight.  Added a comment.

fasaxc:
@neiljerram Think I've addressed all your comments now.  Back to you for re-review.

neiljerram:
retest this please

fasaxc:
retest this please

PR #1358, opened 2017-02-21
Fix IP address parsing to use net.IP.To4() instead of len().

### LINKED ISSUE #1359: Golang felix incorrectly stores IPv4 addresses as IPv6
We were using `len()` on a net.IP to tell if it was an IPv4 or IPv6 address but that isn't sound because Go can represent an IPv4 address using 16 bytes.

Since we then stringify the IPs later int he pipeline it's not clear that there's an impact but this needs to be fixed for some upcoming work.

fasaxc:
len(ip) can be 16, even for an IPv4 address due to Go's internal representation of IPs.

Fixed up some bad function naming while I was in there.

Fixes #1359


fasaxc:
retest this please

PR #1366, opened 2017-03-02
Remove tag index and reimplement on top of tag->label inheritance.

fasaxc:
We already implement inheritance between tags (which are slowly being deprecated) and labels so that a tag `foo` is mapped to label `foo=""`.  This PR converts matches on tags to selectors of the form `has(labelName)`.  By combining the tag and selector indexing logic, we save a lot of occupancy at high scale (>10% in my tests).

Those selectors result in identical IP sets unless the user was using the same name for a tag and a label and not applying them to the same endpoints.  Since tags are being deprecated, that should be acceptable.

TODO:

- [x] Document that tags are being deprecated: https://github.com/projectcalico/calico/issues/589


calc/rule_scanner.go
```diff
@@ -16,21 +16,23 @@ package calc
...
 	"github.com/projectcalico/libcalico-go/lib/selector"
 )
 
-// RuleScanner calculates which selectors and tags are in use by the active rules (calculated
-// by the ActiveRulesCalculator).  I.e. which tags and selectors are in the rules and need to
-// be rendered to the dataplane (as IP sets, for example).
+// RuleScanner scans the rules sent to it by the ActiveRulesCalculator, looking for tags and
+// selectors. It calculates the set of active tags and selectors and emits events when they become
+// active/inactive.  To avoid the occupancy overhead of doing two types of indexing, it maps tags
+// to selectors of the form "has(tag-name)"; those are broadly equivalent due to the inheritance
```
neiljerram: 'broadly' doesn't sound very reassuring here.  Can this say something more definitive?

calc/rule_scanner.go
```diff
@@ -271,78 +256,80 @@ func ruleToParsedRule(rule *model.Rule) (parsedRule *ParsedRule, allTagOrSels []
...
-}
-
-func tagOrSelFromSel(sel string) (tos tagOrSel, err error) {
-	parsedSel, err := selector.Parse(sel)
-	if err == nil {
-		tos = tagOrSel{selector: parsedSel, uid: parsedSel.UniqueId()}
-	}
-	return
+func selFromTag(tag string) (selector.Selector, error) {
+	return selector.Parse(fmt.Sprintf("has(%s)", tag))
```
neiljerram: Can we really get meaningful errors from `selector.Parse("has(<tag>)")` ?

If the answer is no, it feels like we should be placing an error firewall here, with a panic, and then not passing any possible error upwards (and so re-simplifying the callers above).

If the answer is yes, can we justify why prematurely returning from `extractTagsAndSelectors` will be the Right Thing?  What will be the detailed effect of that?

calc/active_rules_calculator.go
```diff
@@ -282,3 +281,47 @@ func (arc *ActiveRulesCalculator) sendPolicyUpdate(policyKey model.PolicyKey) {
...
+	}
+}
+
+func (idx EndpointKeyToProfileIDMap) Update(
+	key model.Key,
+	profileIDs []string,
+) (
+	removedIDs, addedIDs map[string]bool,
+) {
+	oldIDs := idx.endpointKeyToProfileIDs[key]
```
fasaxc: If key isn't in the map, oldIDs will be a nil slice, which is safe to iterate over (it simply executes the loop 0 times).  It's also safe to delete something that's not in the map.

calc/rule_scanner.go
fasaxc: We panic further up the stack.  This was some of my early Go code.  I'll clean it up...

fasaxc:
@neiljerram Markups done, back to you for re-review.

PR #1367, opened 2017-03-06
Scale testing Felix with the k8s datastore driver

neiljerram:


neiljerram:
@fasaxc I think this is now in a good enough state to be worth your review.  Specifically:

- it doesn't mess up the existing pull-request-build process (unless I've messed up in the rebase that I just did)

- it works to run the complete FV suite (with `make -r k8s-fv-test`)

- I believe it provides a reproducible prometheus/grafana setup

- it has the beginnings of a useful set of tests, so I think could be worth banking at this stage.

Please let me know what you think.

neiljerram:
retest this please

Makefile
```diff
@@ -142,6 +150,83 @@ calico/felix: bin/calico-felix
...
+.PHONY: k8s-fv-test run-k8s-apiserver stop-k8s-apiserver run-etcd stop-etcd
+k8s-fv-test: calico/felix run-k8s-apiserver k8sfv/k8sfv.test
+	@-docker rm -f $(FELIX_K8S)
+	sleep 1
+	docker run --detach --privileged --name=$(FELIX_K8S) \
+	-e FELIX_LOGSEVERITYSCREEN=info \
+	-e FELIX_DATASTORETYPE=kubernetes \
+	-e FELIX_PROMETHEUSMETRICSENABLED=true \
+	-e K8S_API_ENDPOINT=https://$(LOCAL_IP_ENV):6443 \
+	-e K8S_INSECURE_SKIP_TLS_VERIFY=true \
```
fasaxc: Good chance that TLS slows us down, maybe we need to turn it on?  It's the norm in k8s-land.

Makefile
```diff
@@ -142,6 +150,83 @@ calico/felix: bin/calico-felix
 	cp bin/calico-felix docker-image/bin/
 	docker build -t calico/felix docker-image
 
+# Targets for Felix testing with the k8s backend and a k8s API server,
+# with k8s model resources being injected by a separate test client.
+LOCAL_IP_ENV?=$(shell ip route get 8.8.8.8 | head -1 | awk '{print $$7}')
```
fasaxc: I'm really keen that we don't run etcd on the host's IP.  It's a real pain for me and it'll clash on jenkins etc.

Makefile
```diff
@@ -142,6 +150,83 @@ calico/felix: bin/calico-felix
...
+	@-docker rm -f k8sfv-prometheus
+	sleep 2
+
+run-grafana: stop-grafana run-prometheus
+	docker run --detach --rm --name k8sfv-grafana --net=host \
+	grafana/grafana
+	# Wait for it to get going.
+	sleep 5
+	# Configure prometheus data source.
+	curl 'http://admin:admin@127.0.0.1:3000/api/datasources' -X POST -H 'Content-Type: application/json;charset=UTF-8' --data-binary '{"name":"my-prom","type":"prometheus","url":"http://localhost:9090","access":"direct","isDefault":true,"database":"mydb","user":"admin","password":"admin"}'
```
fasaxc: Seems crazy to need to use curl to configure grafana, I thought it had a config file and/or discovery mechanism.  I know the scale-test-operator repo auto-configures it, for example.

Makefile
```diff
@@ -142,6 +150,83 @@ calico/felix: bin/calico-felix
...
+	# Configure dashboard for observing the FV test.
+	{ \
```
fasaxc: I'm pretty sure you can drop dashboards into a directory that grafana watches, that feels cleaner than using curl.  We do that in the calico-k8s-cluster set-up scripts.

Makefile
neiljerram: But this way works, and isn't that crazy, surely?

neiljerram: I believe we still need this for automatic data source configuration...  I don't see anything in calico-k8s-cluster that does this, and scale-test-operator uses curl as we do here.

Makefile
neiljerram: I'll look into this, as I agree it would be nicer.

neiljerram: Yes, I have something now that seems much better.

Makefile
neiljerram: Think this is one to track under futures.  The issue is working out what creds the API server is using, and how I can get at those.  I did spend a little time on that at the start of this work, without success yet.

neiljerram:
@fasaxc, ready for your eyes again.  Unless there's some very easy solution to the markup about using TLS, I think this is good for merging now.

k8sfv/pod.go
```diff
@@ -0,0 +1,230 @@
...
+func cleanupAllPods(clientset *kubernetes.Clientset, nsPrefix string) {
+	log.WithField("nsPrefix", nsPrefix).Info("Cleaning up all pods...")
+	nsList, err := clientset.Namespaces().List(v1.ListOptions{})
+	if err != nil {
+		panic(err)
+	}
+	log.WithField("count", len(nsList.Items)).Info("Namespaces present")
+	podsDeleted := 0
+	admission := make(chan int, 10)
+	waiter := make(chan int, len(nsList.Items))
```
fasaxc: Good use case for a WaitGroup?

k8sfv/pod.go
```diff
@@ -0,0 +1,230 @@
...
+type localNetworking struct {
+	podIf     netlink.Link
+	hostIf    netlink.Link
+	namespace ns.NetNS
+}
+
+var localNetworkingMap = map[string]*localNetworking{}
+var localNetworkingMutex = sync.Mutex{}
+
+func createPod(clientset *kubernetes.Clientset, d deployment, nsName string, spec podSpec) {
```
fasaxc: Thinking ahead to making this test more aggressive (and hopefully flush out more issues like the one we hit in the scale tests), you might want to make this function thread-safe so you can create lots of pods in parallel?

k8sfv/pod.go
```diff
@@ -0,0 +1,230 @@
...
+	}
+	log.WithField("pod_in", pod_in).Debug("Pod defined")
+	pod_out, err := clientset.Pods(nsName).Create(pod_in)
+	if err != nil {
+		panic(err)
+	}
+	log.WithField("pod_out", pod_out).Debug("Created pod")
+	pod_in = pod_out
+	pod_in.Status.PodIP = ip
+	pod_out, err = clientset.Pods(nsName).UpdateStatus(pod_in)
```
fasaxc: For an accurate simulation, are we doing the same number of status updates as happen with a real pod?  Worth checking with @caseydavenport

Makefile
```diff
@@ -142,6 +150,89 @@ calico/felix: bin/calico-felix
 	cp bin/calico-felix docker-image/bin/
 	docker build -t calico/felix docker-image
 
+# Targets for Felix testing with the k8s backend and a k8s API server,
+# with k8s model resources being injected by a separate test client.
+GET_CONTAINER_IP := docker inspect --format='{{range .NetworkSettings.Networks}}{{.IPAddress}}{{end}}'
+K8S_VERSION=1.5.3
+FELIX_K8S=felix-k8s
```
neiljerram: Nice idea.

Makefile
```diff
@@ -142,6 +150,89 @@ calico/felix: bin/calico-felix
...
+.PHONY: k8s-fv-test run-k8s-apiserver stop-k8s-apiserver run-etcd stop-etcd
+k8s-fv-test: calico/felix run-k8s-apiserver k8sfv/k8sfv.test
+	@-docker rm -f $(FELIX_K8S)
+	sleep 1
+	K8S_IP=`$(GET_CONTAINER_IP) calico-k8s-apiserver` && \
+	docker run --detach --privileged --name=$(FELIX_K8S) \
```
neiljerram: No, because:

- Sometimes I screw some other piece of config up, that causes this container to exit shortly after start up, and when that happens I want to be able to look at its logs.

- We have another mechanism to avoid leaving lots of zombie containers around: using a specific name, and doing `docker rm` on that name just before starting a new one.

Makefile
```diff
@@ -142,6 +150,89 @@ calico/felix: bin/calico-felix
...
+	sleep 5
+	# Configure prometheus data source.
+	PROMETHEUS_IP=`$(GET_CONTAINER_IP) k8sfv-prometheus` && \
+	sed "s/__PROMETHEUS_IP__/$${PROMETHEUS_IP}/" < $(K8SFV_DIR)/grafana-datasources/my-prom.json.in | \
+	curl 'http://admin:admin@127.0.0.1:3000/api/datasources' -X POST \
+	    -H 'Content-Type: application/json;charset=UTF-8' --data-binary @-
+
+stop-grafana:
+	@-docker rm -f k8sfv-grafana
+	sleep 2
```
neiljerram: Paranoia about whether the new container will be able to rebind to the relevant port.

k8sfv/pod.go
neiljerram: Yes, agree this is worth doing now.

I think it's just that the setting of `localNetworkingMap[nsName+"."+name]` needs mutex protection, right?

k8sfv/pod.go
neiljerram: I would guess not as many; will add this to my futures list.

k8sfv/pod.go
neiljerram: Cool, yes.  Is there anything in the stdlib for what I'm using `admission` for?

k8sfv/pod.go
caseydavenport: IIRC we get the following updates from the API:

- Pod gets created in the API.  Missing both `Pod.Status.PodIP` and `Pod.Spec.NodeName`
- Pod gets scheduled to a Node.  Missing `Pod.Status.PodIP` but now has a `Pod.Spec.NodeName`
- Pod goes through CNI on the node.  It now has a `Pod.Status.PodIP` and a `Pod.Spec.NodeName`

There might be one more event in there.  Will have to double check, but that's pretty close.

caseydavenport: Theoretically there could be many more (before / after any of the above events) depending on which other third-party controllers / plugins are running on the cluster, but those all come from Kubernetes core itself.

neiljerram:
retest this please

neiljerram:
retest this please

PR #1368, opened 2017-03-06
Add config option to do memory profile

fasaxc:
The resulting memory profile can be cracked with `go tool pprof bin/calico-felix mem.pprof`.  I've found it very useful in my occupancy optimisation work.

The heap dump is less useful, in that there's no tool to crack it right now but there may be one in future.

### DIFF
--- a/felix.go
+++ b/felix.go
@@ -29,6 +29,9 @@ import (
 	"github.com/docopt/docopt-go"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 
+	"runtime"
+	"runtime/pprof"
+
 	"github.com/projectcalico/felix/buildinfo"
 	"github.com/projectcalico/felix/calc"
 	"github.com/projectcalico/felix/config"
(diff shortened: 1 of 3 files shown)


neiljerram:
What appears to me to be the begged question here: why should these options be treated specially, instead of being declared in the normal way, that would allow them to be set by the datastore, or by the environment, or by felix.cfg?

fasaxc:
@neiljerram I've switched to a config param and removed the heap dump option for now since there's no tool to crack it.

felix.go
```diff
@@ -172,6 +175,17 @@ configRetry:
...
 		"Successfully loaded configuration.")
 
+	// If asked to do so, dump a memory profile after the first sync.
+	var memProfFile *os.File
+	memProfFileName := configParams.DebugMemoryProfilePath
+	if memProfFileName != "" {
+		log.WithField("file", memProfFileName).Info("Asked to create a memory profile.")
+		memProfFile, err = os.Create(memProfFileName)
+		if err != nil {
+			log.WithError(err).Fatal("Could not create memory profile file")
```
neiljerram: Set `memProfFile` to nil here, to ensure that the code below doesn't trigger?

PR #1373, opened 2017-03-07
Compute label union on the fly to reduce occupancy.

fasaxc:
- Previously we had multiple maps indexed on the same key in the inheritance index.  Combine those by introducing a struct for the value.
- Previously, the label indexing code has its own `map[...]bool`-based set, swap in set.Set instead, which avoids storing an extra bool (and cleans up the code).
- Previously, we pre-calculated the inherited labels, which meant storing them twice.  Switch to calculating the value of a label on-the-fly.  (Relies on https://github.com/projectcalico/libcalico-go/pull/360)

Saves about 14% occupancy in my tests.

### DIFF
--- a/glide.yaml
+++ b/glide.yaml
@@ -21,8 +21,7 @@ import:
 - package: github.com/go-ini/ini
   version: ^1.21.0
 - package: github.com/projectcalico/libcalico-go
-  repo: https://github.com/fasaxc/libcalico-go.git
-  version: 3428e33dd83712164f8b46b576c91098f7cf9b36
+  version: 70c9896a09b3dd8b4a3398f6eade318e07d464b9
   subpackages:
   - lib
 - package: github.com/Sirupsen/logrus
--- a/glide.yaml
+++ b/glide.yaml
@@ -21,7 +21,8 @@ import:
 - package: github.com/go-ini/ini
   version: ^1.21.0
 - package: github.com/projectcalico/libcalico-go
-  version: v1.1.2
+  repo: https://github.com/fasaxc/libcalico-go.git
+  version: 3428e33dd83712164f8b46b576c91098f7cf9b36
   subpackages:
   - lib
 - package: github.com/Sirupsen/logrus
--- a/labelindex/label_index.go
+++ b/labelindex/label_index.go
@@ -17,13 +17,15 @@ package labelindex
 import (
 	log "github.com/Sirupsen/logrus"
 
+	"github.com/projectcalico/felix/set"
 	"github.com/projectcalico/libcalico-go/lib/selector"
+	"github.com/projectcalico/libcalico-go/lib/selector/parser"
 )
 
 type Index interface {
 	UpdateSelector(id interface{}, sel selector.Selector)
 	DeleteSelector(id interface{})
-	UpdateLabels(id interface{}, labels map[string]string)
+	UpdateLabels(id interface{}, labels parser.Labels)
 	DeleteLabels(id interface{})
 }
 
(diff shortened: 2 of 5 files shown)


fasaxc:
retest this please

neiljerram:
(As there are 3 independent changes here, according to the commit message, I think it would have eased review to present them as 3 separate PRs, or perhaps 3 separate commits in a PR.)

labelindex/label_inheritance_index.go
```diff
@@ -110,105 +140,150 @@ func (idx *InheritIndex) DeleteSelector(id interface{}) {
...
+	parent := idx.getOrCreateParent(parentID)
+	parent.labels = labels
 	idx.flushChildren(parentID)
 }
 
 func (idx *InheritIndex) DeleteParentLabels(parentID string) {
-	delete(idx.labelsByParentID, parentID)
+	parent := idx.getOrCreateParent(parentID)
+	parent.labels = nil
+	idx.discardParentIfEmpty(parentID)
```
neiljerram: For this case it seems a bit wasteful to create the parent if it didn't already exist, only to then delete it again.

labelindex/label_inheritance_index.go
```diff
@@ -110,105 +140,150 @@ func (idx *InheritIndex) DeleteSelector(id interface{}) {
...
 }
 
 func (idx *InheritIndex) DeleteParentTags(parentID string) {
-	delete(idx.tagsByParentID, parentID)
+	parentData := idx.parentDataByParentID[parentID]
+	if parentData == nil {
+		return
+	}
+	parentData.tags = nil
+	idx.parentDataByParentID[parentID] = parentData
```
neiljerram: This line is superfluous, isn't it?  If it isn't, I think there are quite a few other cases that need it.  If it is, I think better to remove, for overall consistency.

labelindex/label_inheritance_index.go
```diff
@@ -110,105 +140,150 @@ func (idx *InheritIndex) DeleteSelector(id interface{}) {
...
+		itemData, ok := idx.itemDataByID[itemID]
 		if !ok {
 			// Item deleted.
 			log.Debugf("Flushing delete of item %v", itemID)
 			idx.index.DeleteLabels(itemID)
 		} else {
 			// Item updated/created, re-evaluate labels.
 			log.Debugf("Flushing update of item %v", itemID)
-			combinedLabels := make(map[string]string)
-			parentIDs := idx.parentIDsByItemID[itemID]
```
neiljerram: (Later: I don't think so, actually, as I believe all this code is taking effect _before_ the conversion from tags to 'has(<tag>)'.  But will leave these comments to double check with you.)

labelindex/label_inheritance_index.go
fasaxc: Ah yes, leaked in from an earlier draft (where parentData wasn't a pointer)

labelindex/label_inheritance_index.go
fasaxc: No, this is the core change of this PR.  We no longer calculate an explicit copy of the combined labels, instead we implement the `Labels` interface on `itemData`.

The actual inheritance result isn't changed by this PR.  Its still:

- explicit item labels
- profile labels
- implicit labels derived from tags

(Earliest label with a particular name wins.)

glide.yaml
```diff
@@ -21,7 +21,8 @@ import:
 - package: github.com/go-ini/ini
   version: ^1.21.0
 - package: github.com/projectcalico/libcalico-go
-  version: v1.1.1
+  repo: https://github.com/fasaxc/libcalico-go.git
+  version: 3428e33dd83712164f8b46b576c91098f7cf9b36
```
fasaxc: As discussed, was worried that glide didn't freshen the branch.  It's now been merged so I an pin to master now.

labelindex/label_index.go
```diff
@@ -17,38 +17,40 @@ package labelindex
...
+	selIdsByLabelId map[interface{}]set.Set
+	labelIdsBySelId map[interface{}]set.Set
 
 	// Callback functions
 	OnMatchStarted MatchCallback
 	OnMatchStopped MatchCallback
 }
 
-func NewIndex(onMatchStarted, onMatchStopped MatchCallback) Index {
+func NewIndex(onMatchStarted, onMatchStopped MatchCallback) *linearScanIndex {
```
fasaxc: Cleaning that up in my follow on PR...  #1374

labelindex/label_inheritance_index.go
```diff
@@ -18,32 +18,62 @@ import (
...
 	"github.com/projectcalico/libcalico-go/lib/backend/model"
 	"github.com/projectcalico/libcalico-go/lib/selector"
 )
 
+type itemData struct {
+	labels  map[string]string
+	parents []*parentData
+}
+
+// Get implements the Labels interface for itemData.  Combines the item's own labels with that
```
fasaxc: Yes, I've added comments.

labelindex/label_inheritance_index.go
```diff
@@ -18,32 +18,62 @@ import (
...
+	if value, present = itemData.labels[labelName]; present {
+		return
+	}
+	for _, parent := range itemData.parents {
+		if value, present = parent.labels[labelName]; present {
+			return
+		}
+		for _, tag := range parent.tags {
+			if tag == labelName {
+				present = true
```
fasaxc: I think we can rely on the go convention of returning the zero value with an error.

labelindex/label_inheritance_index.go
fasaxc: Agreed.

fasaxc:
@neiljerram think I've addressed your comments.

fasaxc:
retest this please

PR #1376, opened 2017-03-08
Ensure the IP forwarding is enabled on the interfaces we control.

### LINKED ISSUE #187: Felix should ensure IP forwarding is enabled for all endpoints
Forwarding must be enabled on all endpoint interfaces, or they won't work. In other words, `/proc/sys/net/ipv4/conf/<interface ID>/forwarding` should be 1 (and the same for IPv6). This is usually the default, based on `net.ipv4.ip_forward` and `net.ipv6.conf.all.forwarding` in `/etc/sysctl.conf`, but if it is not then endpoints don't work properly.

Felix should probably check the values for each tap interface, just as it sets the value of `proxy_arp` for them, so that they are guaranteed to work.

fasaxc:
Fixes #187 
Fixes https://github.com/projectcalico/cni-plugin/issues/283

### DIFF
--- a/intdataplane/endpoint_mgr.go
+++ b/intdataplane/endpoint_mgr.go
@@ -691,24 +691,62 @@ func (m *endpointManager) configureInterface(name string) error {
 	log.WithField("ifaceName", name).Info(
 		"Applying /proc/sys configuration to interface.")
 	if m.ipVersion == 4 {
+		// Enable strict reverse-path filtering.  This prevents a workload from spoofing its
+		// IP address.  Non-privileged containers have additional anti-spoofing protection
+		// but VM workloads, for example, can easily spoof their IP.
 		err := m.writeProcSys(fmt.Sprintf("/proc/sys/net/ipv4/conf/%s/rp_filter", name), "1")
 		if err != nil {
 			return err
 		}
+		// Enable routing to localhost.  This is required to allow for NAT to the local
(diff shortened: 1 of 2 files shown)


intdataplane/endpoint_mgr.go
```diff
@@ -691,27 +691,50 @@ func (m *endpointManager) configureInterface(name string) error {
...
 		if err != nil {
 			return err
 		}
+		// Normally, the kernel has a delay before responding to proxy ARP but we know
+		// that's not needed in a Calico network so we disable it.
 		err = m.writeProcSys(fmt.Sprintf("/proc/sys/net/ipv4/neigh/%s/proxy_delay", name), "0")
 		if err != nil {
 			return err
 		}
+		// Enable IP forwarding on the interface, which is how Calico routes packets.
```
neiljerram: I just researched this and established that this setting will allow routing of packets _from_ the Calico interface.  Routing of packets _to_ a Calico interface will still require

- either the global forwarding settings - e.g. as we do in networking-calico package postinst:

        echo "net.ipv4.ip_forward=1" >> /etc/sysctl.conf
        echo "net.ipv6.conf.all.forwarding=1" >> /etc/sysctl.conf

- or settings like this one on each fabric-facing interface that the operator wants to forward from.

Maybe worth adding a comment near here to record that?

PR #1380, opened 2017-03-09
Reduce log spam from unconditional rewriting of dispatch chains.

fasaxc:
Only update the dispatch chains that have actually changed.

We were getting a lot of spam from the "chain removed/updated" logs in the scale runs (and doing unnecessary work in the iptables layer to figure out that nothing had changed).

### DIFF
--- a/intdataplane/endpoint_mgr.go
+++ b/intdataplane/endpoint_mgr.go
@@ -66,11 +66,11 @@ type endpointManager struct {
 	pendingIfaceUpdates map[string]ifacemonitor.State
 
 	// Active state, updated in CompleteDeferredWork.
-	activeWlEndpoints     map[proto.WorkloadEndpointID]*proto.WorkloadEndpoint
-	activeWlIfaceNameToID map[string]proto.WorkloadEndpointID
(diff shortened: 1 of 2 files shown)


intdataplane/endpoint_mgr.go
```diff
@@ -358,9 +367,10 @@ func (m *endpointManager) calculateHostEndpointStatus(id proto.HostEndpointID) (
 }
 
 func (m *endpointManager) resolveWorkloadEndpoints() {
-	// Optimisation, only recalculate the dispatch chains if we've never done so or if the
-	// workloads have changed in some way.
-	needToCheckDispatchChains := m.activeDispatchChains == nil || len(m.pendingWlEpUpdates) > 0
+	if len(m.pendingWlEpUpdates) > 0 {
+		// We're about to make endpoint updates, make sure we re-check the dispatch chains.
```
neiljerram: nit: I think it's just "recheck".

neiljerram: (I'm happy to leave this.  If you do want to fix it, I don't think we need another test cycle before merging!)

PR #1382, opened 2017-03-09
Truncate long API messages when logging at INFO level.

fasaxc:


intdataplane/int_dataplane.go
```diff
@@ -750,3 +752,39 @@ type iptablesTable interface {
...
 }
+
+// msgStringer wraps an API message to customise how we stringify it.  For example, it truncates
+// the lists of members in the (potentially very large) IPSetsUpdate messages.
+type msgStringer struct {
+	msg interface{}
+}
+
+func (m msgStringer) String() string {
+	if log.GetLevel() < log.DebugLevel && m.msg != nil {
```
neiljerram: Does `log.GetLevel()` give the level of the log that is currently being emitted, or the configured logging level?

If it's the latter, is that what you wanted?  I.e., if configured level is DEBUG, then we won't do any message truncation, even in INFO logs.  Seems fine to me FWIW - just checking.

intdataplane/int_dataplane.go
```diff
@@ -750,3 +752,39 @@ type iptablesTable interface {
 	RemoveChains([]*iptables.Chain)
 	RemoveChainByName(name string)
 }
+
+// msgStringer wraps an API message to customise how we stringify it.  For example, it truncates
+// the lists of members in the (potentially very large) IPSetsUpdate messages.
+type msgStringer struct {
+	msg interface{}
+}
```
neiljerram: Is this way of doing things better than just defining and calling a toString() function?   (Genuine question - it quite possibly is, but I haven't got my head around the pros/cons yet!)

intdataplane/int_dataplane.go
```diff
@@ -750,3 +752,39 @@ type iptablesTable interface {
...
+				return fmt.Sprintf("%v", msg)
+			}
+			addedNum := truncateAt
+			removedNum := truncateAt
+			if len(msg.AddedMembers) < addedNum {
+				addedNum = len(msg.AddedMembers)
+			}
+			if len(msg.RemovedMembers) < removedNum {
+				removedNum = len(msg.RemovedMembers)
+			}
```
neiljerram: If these calculations of the `X` in `<slice>[:X]` are needed here, why aren't they also needed in the IPSetUpdate case just above?

(If they are needed, I guess that's because `a[:X]` will panic if `len(a) < X`.  Is that correct?)

intdataplane/int_dataplane.go
fasaxc: Yes, it's the global setting.  I was thinking if you turn on DEBUG, then expand the INFO message with the full information.

intdataplane/int_dataplane.go
fasaxc: The advantage is that we defer the call to `String()` until after the logging framework has decided to log.

intdataplane/int_dataplane.go
fasaxc: Yes, Go panics in that case, which makes this a bit messy

intdataplane/int_dataplane.go
neiljerram: Ah, very cunning.

intdataplane/int_dataplane.go
neiljerram: Cool, and I see now why the IPSetUpdate code is OK.

PR #1383, opened 2017-03-09
Do dataplane updates in parallel for IPv4/6.

fasaxc:
Roughly halves the time spent in apply().

### DIFF
--- a/intdataplane/int_dataplane.go
+++ b/intdataplane/int_dataplane.go
@@ -20,6 +20,7 @@ import (
 	"reflect"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	log "github.com/Sirupsen/logrus"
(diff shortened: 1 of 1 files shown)


intdataplane/int_dataplane.go
```diff
@@ -666,32 +667,64 @@ func (d *InternalDataplane) apply() {
...
 	// iptables.
-	for _, w := range d.ipSets {
-		w.ApplyUpdates()
+	var ipSetsWG sync.WaitGroup
+	for _, ipSets := range d.ipSets {
+		ipSetsWG.Add(1)
+		go func(ipSets *ipsets.IPSets) {
+			ipSets.ApplyUpdates()
+			ipSetsWG.Done()
+		}(ipSets)
```
neiljerram: Out of interest, is passing ipSets through as an arg, as here, equivalent to adding `ipSets := ipSets` at the start of the loop?  Or are there other factors?

intdataplane/int_dataplane.go
```diff
@@ -666,32 +667,64 @@ func (d *InternalDataplane) apply() {
...
 	}
 
+	// Update the routing table in parallel with the other updates.  We'll wait for it to finish
+	// before we return.
+	var routesWG sync.WaitGroup
+	for _, r := range d.routeTables {
+		routesWG.Add(1)
+		go func(r *routetable.RouteTable) {
```
neiljerram: I believe that our logs don't currently have a goroutine ID in them - but given this increased use of goroutines, I think it would be helpful if they did have a goroutine ID.  Could that be added?

intdataplane/int_dataplane.go
```diff
@@ -666,32 +667,64 @@ func (d *InternalDataplane) apply() {
...
+	iptablesWG.Wait()
 
 	// Now clean up any left-over IP sets.
-	for _, w := range d.ipSets {
-		w.ApplyDeletions()
+	for _, ipSets := range d.ipSets {
+		ipSetsWG.Add(1)
+		go func(s *ipsets.IPSets) {
+			s.ApplyDeletions()
+			ipSetsWG.Done()
```
neiljerram: Presumably the Done() method is atomic?  (It wouldn't be very sensible if it wasn't!)

intdataplane/int_dataplane.go
fasaxc: Yes, I think that'd be the same.  I'm pretty sure it doesn't work without one of those tweaks though?

intdataplane/int_dataplane.go
fasaxc: I looked into adding that when I added the file and line number to the logs but the runtime (deliberately) doesn't expose an ID for the current goroutine.  We'd have to roll our own, which we're already doing by having a logCxt attached the the IPSets/Tables objects.

intdataplane/int_dataplane.go
fasaxc: Yes it is.  I also checked that it's safe to re-use a waitgroup (as I do with ipSetsWG) and the docs are explicit that it's OK.

intdataplane/int_dataplane.go
neiljerram: Yes, I think so.  (So golang apparently uses the same variable location for all iterations of a loop.)

intdataplane/int_dataplane.go
neiljerram: OK, fair enough; log cxt sounds good.

intdataplane/int_dataplane.go
neiljerram: Cool.

PR #1384, opened 2017-03-09
Improve stats: add route table stats, swap histograms for summaries.

fasaxc:
Switch to monotime for calculating time deltas.  Removes need to check for time discontinuities.  Nice thing about that switch is the type system ensures that you can't get confused between a monotime timestamp (which is a `time.Duration`) and a `time.Time`.

### DIFF
--- a/routetable/route_table.go
+++ b/routetable/route_table.go
@@ -28,6 +28,8 @@ import (
 	"github.com/projectcalico/felix/ifacemonitor"
 	"github.com/projectcalico/felix/ip"
 	"github.com/projectcalico/felix/set"
+	"github.com/prometheus/client_golang/prometheus"
+	"time"
 )
 
 var (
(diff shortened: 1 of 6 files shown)


routetable/route_table.go
```diff
@@ -148,6 +165,8 @@ func (r *RouteTable) Apply() error {
 			}
 		}
 		r.inSync = true
+
+		listIfaceTime.Observe(time.Since(listStartTime).Seconds())
```
neiljerram: No longer worried about backwards time jumps?

(Perhaps you're happy now that they don't actually occur in this interface.  That would be great if so!)

fasaxc: I've switch to the monotime package in the most recent commit.

fasaxc:
@neiljerram Back to you to review switch to `monotime`.

fasaxc:
retest this please

PR #1389, opened 2017-03-15
Move log writing to background threads.

fasaxc:
- Replace separate hooks with a single BackgroundHook that queues logs to different destinations.
- Drop logs if the queue channels are full (but record stats).
- Handle fatal/panic logs specially, blocking until the log is written.
- Adjust felix's shutdown logic to use log.Fatal instead of os.Exit() directly so that its logs get flushed on exit.
- While making the change, I spotted that Felix was logging to stderr during startup then switching to stdout, I've changed that to log to stderr only.

### DIFF
--- a/k8sfv/leastsquares_test.go
+++ b/k8sfv/leastsquares_test.go
@@ -17,6 +17,7 @@ package main
 import (
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+
 	"github.com/projectcalico/felix/k8sfv/leastsquares"
 )
 
--- a/k8sfv/scale_test.go
+++ b/k8sfv/scale_test.go
@@ -21,8 +21,9 @@ import (
 	log "github.com/Sirupsen/logrus"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
-	"github.com/projectcalico/felix/k8sfv/leastsquares"
 	"k8s.io/client-go/kubernetes"
+
+	"github.com/projectcalico/felix/k8sfv/leastsquares"
 )
 
 var _ = Describe("testing the test code", func() {
--- a/logutils/logutils.go
+++ b/logutils/logutils.go
@@ -25,7 +25,6 @@ import (
 	"sort"
 	"strings"
 	"sync"
-	"sync/atomic"
 	"time"
 
 	log "github.com/Sirupsen/logrus"
@@ -52,6 +51,8 @@ func init() {
 		counterLogErrors,
 	)
 }
+
+const logQueueSize = 100
 
 // ConfigureEarlyLogging installs our logging adapters, and enables early logging to screen
 // if it is enabled by either the FELIX_EARLYLOGSEVERITYSCREEN or FELIX_LOGSEVERITYSCREEN
@@ -110,9 +111,10 @@ func ConfigureLogging(configParams *config.Config) {
 	log.SetLevel(mostVerboseLevel)
 
 	// Screen target.
-	var dests []Destination
+	var dests []*Destination
 	if configParams.LogSeverityScreen != "" {
-		screenDest := NewStreamDestination(logLevelScreen, os.Stderr)
+		screenDest := NewStreamDestination(
+			logLevelScreen, os.Stderr, make(chan QueuedLog, logQueueSize))
 		dests = append(dests, screenDest)
 	}
 
@@ -124,7 +126,8 @@ func ConfigureLogging(configParams *config.Config) {
 		var rotAwareFile io.Writer
 		rotAwareFile, fileOpenErr = rfw.Open(configParams.LogFilePath, 0644)
 		if fileDirErr == nil && fileOpenErr == nil {
-			fileDest := NewStreamDestination(logLevelFile, rotAwareFile)
+			fileDest := NewStreamDestination(
+				logLevelFile, rotAwareFile, make(chan QueuedLog, logQueueSize))
 			dests = append(dests, fileDest)
 		}
 	}
@@ -144,7 +147,8 @@ func ConfigureLogging(configParams *config.Config) {
 		tag := "calico-felix"
 		w, sysErr := syslog.Dial(net, addr, priority, tag)
 		if sysErr == nil {
-			syslogDest := NewSyslogDestination(logLevelSyslog, w)
+			syslogDest := NewSyslogDestination(
+				logLevelSyslog, w, make(chan QueuedLog, logQueueSize))
 			dests = append(dests, syslogDest)
 		}
 	}
@@ -311,6 +315,10 @@ type QueuedLog struct {
 	Message       []byte
 	SyslogMessage string
 	WaitGroup     *sync.WaitGroup
+
+	// NumSkippedLogs contains the number of logs that were skipped before this log (due to the
+	// queue being blocked).
+	NumSkippedLogs uint
 }
 
 func (ql QueuedLog) OnLogDone() {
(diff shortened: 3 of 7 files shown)


fasaxc:
retest this please

fasaxc:
CI failure was due to new shutdown log.

logutils/logutils.go
```diff
@@ -168,32 +193,57 @@ func filterLevels(maxLevel log.Level) []log.Level {
...
 // number.
 //
 //    2017-01-05 09:17:48.238 [INFO][85386] endpoint_mgr.go 434: Skipping configuration of
 //    interface because it is oper down. ifaceName="cali1234"
 type Formatter struct{}
 
 func (f *Formatter) Format(entry *log.Entry) ([]byte, error) {
+	stamp := entry.Time.Format("2006-01-02 15:04:05.000")
+	levelStr := strings.ToUpper(entry.Level.String())
+	pid := os.Getpid()
```
fasaxc: I have often found it useful, and I think calico-test uses it (or at least expects it) so I'd prefer to leave it in.

fasaxc:
@neiljerram I think we have to format the log synchronously because the items in `log.Fields{}` could be mutable (or unsafe to share to another thread). That means that we need to queue the pre-formatted log to the background thread.  That has the knock-on effect that it doesn't make sense to re-use logrus Hooks so we end up where the code is with its own parallel system.

logutils/logutils.go
```diff
@@ -168,32 +193,57 @@ func filterLevels(maxLevel log.Level) []log.Level {
...
+	fileName := entry.Data["__file__"]
+	lineNo := entry.Data["__line__"]
+	b := entry.Buffer
+	if b == nil {
+		b = &bytes.Buffer{}
+	}
+	fmt.Fprintf(b, "%s [%s][%d] %v %v: %v", stamp, levelStr, pid, fileName, lineNo, entry.Message)
```
fasaxc: Again, this would have knock-on effects in calico-test, for example so I'd rather leave as-is.

logutils/logutils_test.go
```diff
@@ -64,3 +71,241 @@ var _ = Describe("Logutils", func() {
...
+			Message:       []byte("Message"),
+			SyslogMessage: "syslog message",
+		}
+		b := make([]byte, 1024)
+		n, err := pr.Read(b)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(string(b[:n])).To(Equal("Message"))
+	})
+
+	It("should log dropped logs after time advances far enough", func() {
```
fasaxc: I'll add a comment.

logutils/logutils.go
```diff
@@ -207,13 +257,12 @@ func (f *Formatter) Format(entry *log.Entry) ([]byte, error) {
 			stringifiedValue = stringer.String()
 		} else {
 			// No string method, use %#v to get a more thorough dump.
-			stringifiedValue = fmt.Sprintf("%#v", value)
+			fmt.Fprintf(b, " %v=%#v", key, value)
```
fasaxc: On this branch, value doesn't have a String() method, I'm explicitly wanting the `%#v` format.

logutils/logutils_test.go
fasaxc: Actually, why don't I just simplify the dropped logs logging and remove the time element.

fasaxc:
@neiljerram Think I've addressed all your comments.  Please can you re-review?

fasaxc:
@neiljerram I think the next action is on you; you wanted to re-test and check occupancy/CPU usage with your new rig?

logutils/logutils.go
neiljerram: Agree that calico-test would make this fiddly.
(I don't really see how it can be useful to have the PID on every log line, though.  Would you like to say more about that?)

logutils/logutils.go
neiljerram: OK.  I can imagine that we might be motivated in future to handle the variation in calico-test (which I think would not be hard) - but fine to leave as is for now.

logutils/logutils.go
```diff
@@ -255,42 +306,243 @@ func shouldSkipFrame(frame runtime.Frame) bool {
...
+		_, err := d.writer.Write(ql.Message)
+		if err != nil {
+			counterLogErrors.Inc()
+			fmt.Fprintf(os.Stderr, "Failed to write to log: %v", err)
+		}
+		ql.OnLogDone()
+	}
+}
+
+type monotimeIface interface {
```
neiljerram: This interface is now unused, so please remove.

logutils/logutils.go
```diff
@@ -255,42 +306,243 @@ func shouldSkipFrame(frame runtime.Frame) bool {
...
+	atomic.AddUint64(&d.numDroppedLogs, 1)
+}
+
+func (d *StreamDestination) LoopWritingLogs() {
+	var numSeenDroppedLogs uint64
+	for ql := range d.channel {
+		currentNumDroppedLogs := atomic.LoadUint64(&d.numDroppedLogs)
+		if currentNumDroppedLogs > numSeenDroppedLogs {
+			fmt.Fprintf(d.writer, "... dropped %d logs ...\n",
+				currentNumDroppedLogs-numSeenDroppedLogs)
```
neiljerram: I don't think this logic works; e.g.:
- writer writes logs A, B, C, then finds channel blocked so calls OnLogDropped
- reader runs and sees an increment in the dropped log count, so logs:
     1 log dropped
     A
     B
     C
which is the wrong ordering.

I think the count needs to be per-destination on the writer side, per the logic that I proposed in my last review:

    enqueueLog(entry) {
    tryAgain:
        select {
            case channel writable:
                if droppedLogs != 0 {
                    channel <- {WARNING entry to report droppedLogs}
                    droppedLogs = 0
                    goto tryAgain
                } else {
                    channel <- entry
                }
            default:
                droppedLogs++
        }
    }

This also means less code, as the complexity is then once on the writer side instead of per-destination on the receiving side.

logutils/logutils.go
```diff
@@ -255,42 +306,243 @@ func shouldSkipFrame(frame runtime.Frame) bool {
...
+
+	for _, dest := range h.destinations {
+		if ql.Level > dest.Level() {
+			continue
+		}
+		if waitGroup != nil {
+			waitGroup.Add(1)
+		}
+		select {
+		case dest.Channel() <- ql:
```
neiljerram: I'd prefer putting the

		if waitGroup != nil {
			waitGroup.Add(1)
		}

here and deleting the

			if waitGroup != nil {
				waitGroup.Done()
			}

below.

I think the code will work as you have it, but it raises worry (unjustified, but) about whether Add, Done, Add, Done, Wait would DTRT.  And my rearrangement means less code.

neiljerram:
@fasaxc Overall I'm feeling a lot more comfortable now with the remaining complexity of this change - and if you agree with some of my new comments, then the complexity should reduce even further.

In terms of occupancy and performance, I'm happy if you are - but perhaps it might be worth running another test, once the code is in a form ready to merge, to check that the occupancy difference is within the general statistical variation that you've been observing.

logutils/logutils.go
fasaxc: It's particularly useful if you have disjoint chunks of logs, for example where a user has sent some snippets or if you're missing intermediate logs for some reason.

logutils/logutils.go
fasaxc: Yes, I think that's a fair point that the log could be confusingly out-of-order.  I'll take another pass at it.

logutils/logutils.go
fasaxc: Sadly, that's not safe, you have to increment it before sending it to the background thread (or the background thread could get scheduled immediately and decrement it before foreground increments, which fails).

fasaxc: I'll add an explanatory comment.

neiljerram: D'oh, of course - sorry for not seeing that.

k8sfv/leastsquares_test.go
```diff
@@ -17,6 +17,7 @@ package main
 import (
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+
```
neiljerram: No issue here, but what tool are you running to put these imports into the preferred organization?  I have configuration to run `goimports` whenever I save a file, but it appears that that is not quite sufficient.

k8sfv/scale_test.go
```diff
@@ -21,8 +21,9 @@ import (
 	log "github.com/Sirupsen/logrus"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
-	"github.com/projectcalico/felix/k8sfv/leastsquares"
 	"k8s.io/client-go/kubernetes"
+
+	"github.com/projectcalico/felix/k8sfv/leastsquares"
```
neiljerram: (similarly here)

k8sfv/leastsquares_test.go
fasaxc: I do `make go-fmt`, which just runs `goimports` in the build container.  Maybe a version mismatch?

fasaxc:
Retested occupancy and perf.  Any occupancy change seems to be in the noise; resync perf is slightly improved, if anything.

PR #1390, opened 2017-03-15
vendor: Update libcalico-go to v1.1.3

tomdee:


fasaxc:
@tomdee Did you use `make update-vendor` to rebuild glide.lock?  Looks like you managed to pull in a mercurial dependency but mercurial wasn't in the build container.  I've pushed a v0.5 of the build container that does have it.

PR #1397, opened 2017-03-27
Squash warnings about missing profiles during resync.

### LINKED ISSUE #1387: Spurious warning during resync period
Depending on the order that updates arrive from the Syncer during a resync, we may log this warning for a profile that does exist but hasn't arrived yet:
```
Mar 13 16:46:11 smc-felix-scale-test calico-felix[28936]: WARNING active_rules_calculator.go 260: Profile not known or invalid, generating dummy profile that drops all traffic. profileID="prof-093
```
We should defer such warnings until the end of the resync period.

fasaxc:
It's normal for profiles to be missing if we see the endpoint before the profile.

Fixes #1387.

### DIFF
--- a/calc/active_rules_calculator.go
+++ b/calc/active_rules_calculator.go
@@ -22,6 +22,7 @@ import (
 	"github.com/projectcalico/felix/dispatcher"
 	"github.com/projectcalico/felix/labelindex"
 	"github.com/projectcalico/felix/multidict"
+	"github.com/projectcalico/felix/set"
 	"github.com/projectcalico/libcalico-go/lib/backend/api"
 	"github.com/projectcalico/libcalico-go/lib/backend/model"
 	"github.com/projectcalico/libcalico-go/lib/selector"
@@ -70,6 +71,12 @@ type ActiveRulesCalculator struct {
 	// Cache of profile IDs by local endpoint.
 	endpointKeyToProfileIDs *EndpointKeyToProfileIDMap
 
+	// True if we've got the in-sync message from the datastore.
+	datastoreInSync bool
+	// Set containing the names of any profiles that were missing during the resync.  Used to
+	// log out those profiles at the end of the resync.
+	missingProfiles set.Set
+
 	// Callback objects.
 	RuleScanner         ruleScanner
 	PolicyMatchListener PolicyMatchListener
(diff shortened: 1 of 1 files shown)


calc/active_rules_calculator.go
```diff
@@ -184,6 +193,23 @@ func (arc *ActiveRulesCalculator) OnUpdate(update api.Update) (_ bool) {
...
+			// this until now because we may hear about profiles or endpoints first.
+			arc.missingProfiles.Iter(func(item interface{}) error {
+				log.WithField("profileID", item).Warning(
+					"End of resync: local endpoints refer to missing " +
+						"or invalid profile, profile's rules replaced " +
+						"with drop rules.")
+				return set.RemoveItem
+			})
+		}
+	}
```
neiljerram: Isn't it possible for Felix to get out of sync again, after an initial in-sync?  If so, we should handle the possibility here of setting arc.datastoreInSync back to False.

calc/active_rules_calculator.go
```diff
@@ -184,6 +193,23 @@ func (arc *ActiveRulesCalculator) OnUpdate(update api.Update) (_ bool) {
...
+func (arc *ActiveRulesCalculator) OnStatusUpdate(status api.SyncStatus) {
+	if status == api.InSync && !arc.datastoreInSync {
+		arc.datastoreInSync = true
+		if arc.missingProfiles.Len() > 0 {
+			// Log out any profiles that were missing during the resync.  We defer
+			// this until now because we may hear about profiles or endpoints first.
+			arc.missingProfiles.Iter(func(item interface{}) error {
+				log.WithField("profileID", item).Warning(
+					"End of resync: local endpoints refer to missing " +
+						"or invalid profile, profile's rules replaced " +
```
neiljerram: Hmm, slight awkwardness here about singular vs. plural.  How about:

    "or invalid profile(s); rules replaced " +

calc/active_rules_calculator.go
fasaxc: We don't hold up further dataplane updates during subsequent resyncs so the normal warning is appropriate, I think.  Might be good to do that corking while we resync but that's a more far-reaching change.

neiljerram:
Cool, thanks for considering those points.

PR #1398, opened 2017-03-27
Rev libcalico-go to v1.1.4

fasaxc:
Fixes a KDD issue.


caseydavenport:
@fasaxc can we ensure that this gets a run through @lwr20 e2e rig before merging?

I've so far only done UT level testing of the libcalico fix.

PR #1401, opened 2017-04-03
Tune GC to reduce occupancy.

fasaxc:
@neiljerram This was a change that Matt and I prototyped on Friday.  It seems to save about 30% occupancy for a small increase in time spent in the GC.  Might be good to see what impact it has on your KDD tests too.

### DIFF
--- a/felix.go
+++ b/felix.go
@@ -23,7 +23,7 @@ import (
 	"os/exec"
 	"os/signal"
 	"reflect"
-	"runtime"
+	"runtime/debug"
 	"runtime/pprof"
 	"syscall"
 	"time"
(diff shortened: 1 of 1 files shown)


felix.go
```diff
@@ -59,6 +59,12 @@ Options:
   --version                    Print the version and exit.
 `
 
+const (
+	// Our default value for GOGC if it is not set.  We reduce this from the default of 100
+	// to trade off a little CPU for improved occupancy.
```
neiljerram: Can you write more about what this setting means?

fasaxc:
@neiljerram I've updated the comment, please can you rereview?

PR #1403, opened 2017-04-03
k8sfv: memory leak and CI integration

neiljerram:


k8sfv/leastsquares/least_squares_method.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+
+	base := (n*sumXX - sumX*sumX)
+	a = (n*sumXY - sumX*sumY) / base
+	b = (sumXX*sumY - sumXY*sumX) / base
+
+	return a, b
+}
+
+/*
```
fasaxc: Commented out code?

k8sfv/metric.go
```diff
@@ -0,0 +1,42 @@
...
+	"bufio"
+	"flag"
+	"net/http"
+	"strings"
+
+	log "github.com/Sirupsen/logrus"
+)
+
+func getFelixMetric(name string) (metric string) {
+	resp, err := http.Get("http://" + flag.Arg(1) + ":9091/metrics")
```
fasaxc: `flag.Arg(1)` is a bit magic; where does that come from?

k8sfv/metric.go
```diff
@@ -0,0 +1,42 @@
...
+
+	scanner := bufio.NewScanner(resp.Body)
+	for scanner.Scan() {
+		line := scanner.Text()
+		log.WithField("line", line).Debug("Line")
+		if strings.HasPrefix(line, name) {
+			metric = strings.TrimSpace(strings.TrimPrefix(line, name))
+			break
+		}
+	}
```
fasaxc: I think you're supposed to check `scanner.Err()` here

k8sfv/scale_test.go
```diff
@@ -65,12 +67,39 @@ var _ = Context("with a k8s clientset", func() {
 
 		It("should not leak memory", func() {
 			addNamespaces(clientset, nsPrefix)
+			occupancyMeasurements := []leastsquares.Point{}
 			for ii := 0; ii < 10; ii++ {
 				addEndpoints(clientset, nsPrefix, d, 10000)
-				time.Sleep(20 * time.Second)
+				time.Sleep(10 * time.Second)
+				bytes, err := strconv.ParseFloat(
```
fasaxc: How about a `getFloatMetric` utility function?

k8sfv/leastsquares/least_squares_method.go
```diff
@@ -0,0 +1,66 @@
...
+// Code copied from https://gist.github.com/ota42y/db4ff0298d9c945cd261, with some modifications.
+
+package leastsquares
+
+type Point struct {
+	X float64
+	Y float64
+}
+
+func LeastSquaresMethod(points *[]Point) (a float64, b float64) {
```
fasaxc: I don't think you need to pass points by pointer, a slice is already a pointer-ish type.

k8sfv/scale_test.go
```diff
@@ -65,12 +67,39 @@ var _ = Context("with a k8s clientset", func() {
 
 		It("should not leak memory", func() {
 			addNamespaces(clientset, nsPrefix)
+			occupancyMeasurements := []leastsquares.Point{}
 			for ii := 0; ii < 10; ii++ {
```
fasaxc: You should probably discard the first couple of measurements since the first runs have the advantage of running in a clean, unfragmented heap.  Also, we should add some way to trigger a GC, such as a SIGUSR.

neiljerram:
Thanks for your review.  I'm afraid the k8sfv code is generally a bit woolly and misshapen at the moment, so it's good, as a start, to work on the things that you've called out.

k8sfv/leastsquares/least_squares_method.go
neiljerram: Removed, thanks.

k8sfv/metric.go
neiljerram: It's provided by the Makefile on the ginkgo command line.

I've used a BeforeSuite to refactor and make this much clearer (I hope).

k8sfv/scale_test.go
neiljerram: :+1: for discarding the first couple of measurements.

Is the GC trigger needed for this PR, do you think?  Does SIGUSR already have that effect, or do you mean that we'd need to add support to Felix for that?

neiljerram:
Apart from your GC trigger idea (on which I've commented separately), I believe I've addressed all your markups.

neiljerram:
retest this please

k8sfv/leastsquares/least_squares_method.go
```diff
@@ -0,0 +1,48 @@
...
+
+type Point struct {
+	X float64
+	Y float64
+}
+
+func LeastSquaresMethod(points []Point) (a float64, b float64) {
+	// Calculate a and b such that y = ax + b is the best fit to the given points.
+	//
+	// http://ja.wikipedia.org/wiki/%E6%9C%80%E5%B0%8F%E4%BA%8C%E4%B9%97%E6%B3%95
```
fasaxc: Didn't know you spoke Japanese

PR #1408, opened 2017-04-06
Skip k8sfv when running ut-no-cover and ut-watch Make targets

neiljerram:


fasaxc:
LGTM, back to you to merge when ready

PR #1410, opened 2017-04-06
k8sfv: allow parallel runs by using unique container prefix

neiljerram:


Makefile
```diff
@@ -169,30 +170,33 @@ k8s-fv-test: calico/felix run-k8s-apiserver k8sfv/k8sfv.test
...
 	docker run --detach \
-	  --name k8sfv-apiserver \
+	  --name $(K8SFV_PREFIX)-apiserver \
 	gcr.io/google_containers/hyperkube-amd64:v$(K8S_VERSION) \
 		  /hyperkube apiserver --etcd-servers=http://$${ETCD_IP}:2379 \
 		  --service-cluster-ip-range=10.101.0.0/16 -v=10
 
 stop-k8s-apiserver: stop-etcd
-	@-docker rm -f k8sfv-apiserver
+	@-docker rm -f $(K8SFV_PREFIX)-apiserver
```
fasaxc: Won't think now stop a randomly-named API server?  That doesn't feel so useful.  Maybe we should pass in the prefix as a variable or default to the jenkins job number if defined?

neiljerram:
Just rebased this on top of other k8sfv that is now in master - conflicts needed resolving.

Makefile
neiljerram: The name still includes a 'k8sfv-' prefix, so we can't be deleting any API server container at all.

Also, using `mktemp` actually creates a file with that name under /tmp, and a subsequent `mktemp` (without an intervening reboot) can therefore be _guaranteed_ to use a different name.

Together, I think those points give us a watertight situation.  Unless some other unrelated process coincidentally uses the 'k8sfv-' prefix, and coincidentally chooses a random 4 character string that matches an existing /tmp/k8sfv-* file, the logic here can't accidentally stomp on the wrong API server.

fasaxc:
As discussed, we think this is getting a bit complex for make so let's either push into a shell script or evaluate whether we could do go-based set up/tear down

neiljerram:
Superseded by #1434

PR #1412, opened 2017-04-07
Add support for failsafe UDP ports and allow DNS/DHCP by default.

### LINKED ISSUE #1064: Felix should have FailsafeIn/OutboundHostPorts equivalents for UDP as well
The current failsafe ports options only apply to TCP.

There should be equivalent function available for UDP (e.g. DNS) as well.

I can see two obvious approaches:
1. Have the existing options cover both TCP and UDP ports.
2. Change the names, and add extra options (e.g. FailsafeInboundHostUDPPorts).

What do people think on 1 vs 2 vs <something else>?

### LINKED ISSUE #1407: Make it harder to cut felix off from etcd with incorrect policy
We already have failsafe rules to allow traffic to etcd over TCP but a prereq for that is being able to resolve the hostname of etcd in the first place.  We should whitelist DNS, at least to the configured DNS servers.

This has caught out at least one customer and it leaves the host in a bad state.

fasaxc:
Fixes #1064
Fixes #1407

### DIFF
--- a/config/config_params.go
+++ b/config/config_params.go
@@ -140,8 +140,8 @@ type Config struct {
 	PrometheusMetricsEnabled bool `config:"bool;false"`
 	PrometheusMetricsPort    int  `config:"int(0,65535);9091"`
 
-	FailsafeInboundHostPorts  []uint16 `config:"port-list;22;die-on-fail"`
-	FailsafeOutboundHostPorts []uint16 `config:"port-list;2379,2380,4001,7001;die-on-fail"`
+	FailsafeInboundHostPorts  []ProtoPort `config:"port-list;tcp:22,udp:68;die-on-fail"`
+	FailsafeOutboundHostPorts []ProtoPort `config:"port-list;tcp:2379,tcp:2380,tcp:4001,tcp:7001,udp:53,udp:67;die-on-fail"`
 
 	UsageReportingEnabled bool   `config:"bool;true"`
 	ClusterGUID           string `config:"string;baddecaf"`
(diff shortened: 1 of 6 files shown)


fasaxc:
@matthewdupre I think your customer asked for this a while agao and it just tripped them up again.  I chose to extend the syntax rather than adding a new field since it kept the existing field naming.  You can now specify "tcp:1234,udp:456".  For back-compat, "1234,5678" is accepted too (defaulting to TCP).

config/config_params.go
```diff
@@ -140,8 +140,8 @@ type Config struct {
 	PrometheusMetricsEnabled bool `config:"bool;false"`
 	PrometheusMetricsPort    int  `config:"int(0,65535);9091"`
 
-	FailsafeInboundHostPorts  []uint16 `config:"port-list;22;die-on-fail"`
-	FailsafeOutboundHostPorts []uint16 `config:"port-list;2379,2380,4001,7001;die-on-fail"`
+	FailsafeInboundHostPorts  []ProtoPort `config:"port-list;tcp:22,udp:68;die-on-fail"`
+	FailsafeOutboundHostPorts []ProtoPort `config:"port-list;tcp:2379,tcp:2380,tcp:4001,tcp:7001,udp:53,udp:67;die-on-fail"`
```
matthewdupre: Not a big fan of this format.  We also need to update the docs - can you link me to that PR when it's ready?  It should probably explain the reason for each port being included as well.

matthewdupre: I'd sort of prefer to see 2 new fields (TCP / UDP), with this left deprecated and undocumented.

Sorting out the defaulting behaviour for that might be a touch awkward though.  Possible, but a bit fiddly.

fasaxc: Sorry for making you sad but we discussed and you said you can live-with so I plan to leave as-is.

fasaxc:
I've addressed your comments, think they were simple enough for me to just merge...

CI failure was an intermittent "VM in state error"

PR #1416, opened 2017-04-10
Buffer the signal channel.

fasaxc:
@neiljerram For context, I'm revving the go-build container, which pulls in a newer linter with more checks.

Fix that the signal channel wasn't buffered (which presumably can block a system goroutine).  Not really an issue in this code, I think, but easier to fix than to suppress the false positive in the linter.


fasaxc:
retest this please

PR #1419, opened 2017-04-10
Rev libcalico-go to version without label cache.

fasaxc:
Add code to squash identical labels earlier in the pipeline.

Reduces occupancy.


neiljerram:
I think the conflicts that are now present should be resolved by imposing your versions.

PR #1424, opened 2017-04-12
Move conntrack rules to per-interface chains

### LINKED ISSUE #1423: conntrack rules apply to non-calico interfaces
Our conntrack state rules are in a high-level chain before we've detected that this is a Calico-controlled interface.  That's caused us to drop packets that we shouldn't be touching in some use cases.

fasaxc:
Prevents us from accepting traffic that is coming from interfaces that we don't own.

I also added an option to turn off the INVALID check since it can cause problems for users who are trying something off the beaten path.

Fixes #1423

### DIFF
--- a/config/config_params.go
+++ b/config/config_params.go
@@ -137,6 +137,8 @@ type Config struct {
 
 	IptablesMarkMask uint32 `config:"mark-bitmask;0xff000000;non-zero,die-on-fail"`
 
+	DisableConntrackInvalidCheck bool `config:"bool;false"`
+
 	PrometheusMetricsEnabled bool `config:"bool;false"`
 	PrometheusMetricsPort    int  `config:"int(0,65535);9091"`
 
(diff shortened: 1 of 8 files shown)


config/config_params.go
```diff
@@ -137,6 +137,8 @@ type Config struct {
 
 	IptablesMarkMask uint32 `config:"mark-bitmask;0xff000000;non-zero,die-on-fail"`
 
+	DisableConntrackInvalidCheck bool `config:"bool;false"`
```
matthewdupre: Do you think we should document this one?

rules/endpoints.go
```diff
@@ -308,6 +315,25 @@ func (r *DefaultRuleRenderer) endpointToIptablesChains(
...
+		// connection.
+		rules = append(rules, Rule{
+			Match:  Match().ConntrackState("INVALID"),
+			Action: DropAction{},
+		})
+	}
+	// Allow return packets for established connections.
+	rules = append(rules,
+		Rule{
+			Match:  Match().ConntrackState("RELATED,ESTABLISHED"),
```
matthewdupre: Let's move this one above the INVALID check for performance.

config/config_params.go
matthewdupre: (I'm happy with no.)

matthewdupre:
Thanks - we should also cherry-pick/merge this into master.

fasaxc:
retest this please

PR #1429, opened 2017-04-13
Run `docker build` with `--pull`.

### LINKED ISSUE #1428: `make release` doesn't pull latest alpine image
Noticed that the 2.2.0-rc1 image had a couple of hits from Quay's security scanner.  Cause seems to be that I had a cached version of the alpine:3.4 image that was missing some updates.

fasaxc:
This ensures that we get the latest patch release of the base image.

Fixes #1428

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -148,7 +148,7 @@ calico/felix: bin/calico-felix
 	rm -rf docker-image/bin
 	mkdir -p docker-image/bin
 	cp bin/calico-felix docker-image/bin/
-	docker build -t calico/felix docker-image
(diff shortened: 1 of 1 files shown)


fasaxc:
@tomdee does this seem like the right fix to you?

tomdee:
Yep, seems reasonable. We don't have cached images on semaphore so I'd be a little surprised if this makes any difference to the quay security scanner.

fasaxc:
Lack of semaphore cache is how I spotted this; it was the image I built on my dev machine that showed the issue.

PR #1430, opened 2017-04-17
Bumping libcalico-go to rev with KDD updates

heschlie:
Bumping libcalico-go to latest version which contains IPIP fixes for KDD.

### DIFF
--- a/glide.yaml
+++ b/glide.yaml
@@ -34,6 +34,7 @@ import:
 - package: github.com/vishvananda/netlink
 - package: github.com/gavv/monotime
 - package: github.com/onsi/ginkgo
-  version: f40a49d81e5c12e90400620b6242fb29a8e7c9d9
+  version: f40a49d81e5c12e90400620b6242fb29a8e7c9
+testImport:
 - package: github.com/onsi/gomega
   version: 9b8c753e8dfb382618ba8fa19b4197b5dcb0434c
--- a/glide.yaml
+++ b/glide.yaml
@@ -21,7 +21,7 @@ import:
 - package: github.com/go-ini/ini
   version: ^1.21.0
 - package: github.com/projectcalico/libcalico-go
-  version: 1.2.0
+  version: v1.2.1-rc1
   subpackages:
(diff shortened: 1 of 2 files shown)


glide.yaml
```diff
@@ -21,7 +21,7 @@ import:
 - package: github.com/go-ini/ini
   version: ^1.21.0
 - package: github.com/projectcalico/libcalico-go
-  version: 1.2.0
+  version: 9c7017a9f079eb344a01cc21b72cb4c36ab2270f
```
robbrockbank: Is there not a libcalico-go RC we can pin to?

heschlie: There is not but I can create one if we prefer it

robbrockbank: I'm not sure what we've done historically.  Do we normally release libcalico-go RCs?  If so I suspect that would be the preferred approach.  But happy to go with whatever we've done historically.

heschlie: I do see some RCs, I'll post up a new RC for the changes

heschlie:
The build failure seems to be in a 3rd party package, when I rev libcalico-go back down it still fails, trying to sort out which one it is.

heschlie:
The tests failing were related to some change in Ginkgo, I pinned to the passing version for now, will raise issue to have the test looked at.

glide.yaml
```diff
@@ -33,3 +33,7 @@ import:
   version: ^0.3.0
 - package: github.com/vishvananda/netlink
 - package: github.com/gavv/monotime
+- package: github.com/onsi/ginkgo
```
robbrockbank: Shouldn't these be added as a testImport section?:

```
testImport:
```

robbrockbank:
One comment in line.

glide.yaml
heschlie: I did not know that was a thing, updating and testing now...

heschlie: As we spoke, I've moved gomega into `testImport` but left ginkgo out due to conflicts.

PR #1432, opened 2017-04-18
Remove ginkgo pin.

fasaxc:
@heschlie I removed the pin and did a `make update-vendor` and it picked the same revision that you'd pinned it to; did you have a stale glide cache or something?


heschlie:
@fasaxc I am not sure, but that is quite possible.

LGTM

PR #1434, opened 2017-04-19
k8sfv test improvements for ongoing CI

neiljerram:
This PR contains the test-side changes that we need for useful ongoing CI.  By 'test-side' I mean everything that is required for the test and runs on the machine where the test is executed: the test program, Felix, API server and etcd.  The other pieces - Prometheus, push gateway and Grafana - are 'server-side' and in some senses more optional.

This PR has options to allow k8sfv to connect to those server-side pieces.  I'm now going to work on setting up those pieces for our nightly CI, and then on making it trivial for a developer to set up those pieces for themselves.

### DIFF
--- a/k8sfv/run-test
+++ b/k8sfv/run-test
@@ -23,9 +23,16 @@
 # should set this to the name of that branch, e.g. 'master'.
 : ${CODE_LEVEL:=dev}
 #
-# Tell this script to wait before cleaning up the test containers -
-# e.g. to allow investigation of Felix logs.
-: ${SLEEP_BEFORE_CLEANUP:=0}
+# Whether to clean up the etcd, API server and Felix containers that
+# this test creates.  The default 'false' is intended to be what
+# development work typically needs: containers are not cleaned up,
+# because (1) that allows the developer to look at them when a test
+# has completed, and (2) the containers from a previous run will
+# always be cleaned up when this script is run again - so we are never
+# going to leak an increasing number of containers.  CI should set
+# CLEANUP to 'true', so that the containers are cleaned up regardless
+# of test outcome.
+: ${CLEANUP:=false}
 #
 # Felix logging level.
 : ${FELIX_LOG_LEVEL:=info}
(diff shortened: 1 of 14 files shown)


k8sfv/run-test
```diff
@@ -0,0 +1,91 @@
...
+# Arrange to always clean up the containers that we create.
+function cleanup {
+    # Clean up the containers that this script generates.
+    docker rm -f ${prefix}-felix || true
+    docker rm -f ${prefix}-apiserver || true
+    docker rm -f ${prefix}-etcd || true
+    # List all remaining containers.
+    docker ps -a
+}
+trap cleanup EXIT SIGINT SIGQUIT
```
fasaxc: Nice; I suggest making it configurable whether we run cleanup.  I've certainly found it useful to have the option to leave the rig running when I'm debugging the other FV frameworks.  WDYT?

k8sfv/run-test
```diff
@@ -0,0 +1,91 @@
...
+# Wait until the newly launched API server can respond to a request,
+# before continuing.
+while ! curl -k ${k8s_api_endpoint}/apis/extensions/v1beta1/thirdpartyresources; do
+    sleep 2
+done
+
+# Run Felix.
+mkdir -p k8sfv/output
+docker run --detach --privileged --name ${prefix}-felix \
+       -e FELIX_LOGSEVERITYSCREEN=info \
```
fasaxc: Worth making the log level configurable?

k8sfv/run-test
neiljerram: I've coincidentally just added sthg for this in parallel - please see SLEEP_BEFORE_CLEANUP in the commits I'm just about to add.

neiljerram:
@fasaxc Would you like to review the deltas since your previous review above?

FYI, also:
- we've now had a clean-looking CI run, with this latest code, at http://localhost:8080/job/k8sfv/11/console (and in which even the leak test passed!)
- I'm working as you know on the ongoing Grafana/Prometheus/push gateway setup.

PR #1437, opened 2017-04-21
k8sfv monitoring work - strawman for review

neiljerram:
I think this might be approaching done - but I also think that another pair of eyes is likely to see lots of rough edges and further work that could or should be done now, and that it will likely be more efficient to interleave that work with the remaining pieces that I have in mind.  Hence putting it up for a review.

The possible things that I still have on my list are:
- [ ] Get the Prometheus server to target Felix(es?) as well as the push gateway, so that the monitoring also shows Felix metrics directly.
- [ ] Decide whether we also want to keep a separate non-GKE process for developers to set up similar monitoring.  If we do, develop that process; if we don't, delete the remaining Prometheus and Grafana bits from the Felix Makefile and k8sfv directory.
- [ ] Incorporate some improvements that I have locally to the 'k8s scale testing' dashboard.  (Although these are only relevant if we're going to be showing Felix metrics.)
- [ ] Something to ensure that the Prometheus database is preserved even if other bits of the GKE cluster go down.


neiljerram:
@fasaxc - Would appreciate your review.  Please do see my overall comments and positioning above.

neiljerram:
retest this please

neiljerram:
Thanks - I will squash down imminently.  Regarding the `.org`, I actually think we should consider it, as
- it is Github-supported and makes really nice output (https://github.com/neiljerram/felix/blob/d9e22156af5470d69d53be6e73165a20260d0f35/k8sfv/monitoring.org)
- for me, it fits really nicely into my workflow, and is much more pleasant to edit than .md
- for others, it's just pattern matching / intuition as usual, and I really doubt would be an issue
- across the project as a whole, we're often introducing new languages - e.g. the templating that we now use in docs - and I generally think that's a good thing!
So possibly worth another thought?   Anyway, I'll do the squashing now...

neiljerram:
@fasaxc Following discussion, I've done the markdown conversion now - actually quite straightforward.

neiljerram:
retest this please

PR #1414, opened 2017-04-07
Fix felix_cluster_* metrics when usage reporting disabled

neiljerram:
Fixes #1413

PR #1440, opened 2017-04-24
Fix felix_cluster_* metrics when usage reporting disabled

neiljerram:
If we add this fix to 2.1.x-series, we should then be able to run the k8sfv test against 2.1.x.  Specifically, this fix is needed for the 'should delete a pod whose IP has been cleared' test.

### DIFF
--- a/felix.go
+++ b/felix.go
@@ -314,6 +314,13 @@ configRetry:
 			configParams.ClusterType,
 			statsChanOut,
 		)
+	} else {
+		// Usage reporting disabled, but we still want a stats collector for the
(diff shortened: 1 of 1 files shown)


neiljerram:
(But of course it is a judgment call whether there would be any undue risk in adding this to 2.1.x.)

neiljerram:
(This is a cherry-pick of the fix that was made in master by #1414 )

fasaxc:
LGTM.  It was a bug that we didn't expose those stats if that config  option was set.

PR #1453, opened 2017-05-10
Reconnect to the datastore after initial config load.

### LINKED ISSUE #1452: Felix doesn't respect IpInIpEnabled when configured via KDD
When the KDD is configured to use IPIP + BGP, Felix sets the `K8sDisableNodeWatch` option based on the FELIX_IPINIPENABLED environment variable.  However, once config comes in over the Syncer which says IPIP is actually enabled (or disabled), Felix doesn't restart the syncer with the correct value for K8sDisableNodeWatch.

Workaround is in https://github.com/projectcalico/calico/pull/772/files.

1. Download KDD BGP Calico manifest on Kubernetes - http://docs.projectcalico.org/master/getting-started/kubernetes/installation/hosted/kubernetes-datastore/calico-networking/1.6/calico.yaml
2. Ensure FELIX_IPINIPENABLED is unset in the environment
3. Notice that IPIP doesn't work.

fasaxc:
Ensures that we reconfigure the Syncer correctly if the IPIP flag is set in the datastore.

Fixes #1452

### DIFF
--- a/felix.go
+++ b/felix.go
@@ -186,6 +186,19 @@ configRetry:
 			time.Sleep(1 * time.Second)
 			continue configRetry
 		}
+
+		// We now have some config flags that affect how we configure the syncer.
+		// After loading the config from the datastore, reconnect, possibly with new
+		// config.  We don't need to re-load the configuration _again_ because the
+		// calculation graph will spot if the config has changed since we were initialised.
+		datastoreConfig = configParams.DatastoreConfig()
+		datastore, err = backend.NewClient(datastoreConfig)
+		if err != nil {
+			log.WithError(err).Error("Failed to (re)connect to datastore")
+			time.Sleep(1 * time.Second)
+			continue configRetry
+		}
+
 		break configRetry
 	}
 


neiljerram:
FYI I'm starting to look at this now.

felix.go
```diff
@@ -186,6 +186,19 @@ configRetry:
 			time.Sleep(1 * time.Second)
 			continue configRetry
 		}
+
+		// We now have some config flags that affect how we configure the syncer.
+		// After loading the config from the datastore, reconnect, possibly with new
+		// config.  We don't need to re-load the configuration _again_ because the
+		// calculation graph will spot if the config has changed since we were initialised.
+		datastoreConfig = configParams.DatastoreConfig()
```
neiljerram: I'm not fully persuaded by the last sentence here.  ISTM that there is a window - between the `loadConfigFromDatastore` above and creating the second datastore client - where the config could change and Felix wouldn't get any update about it.  If that's right, the window can be closed by repeating the `loadConfigFromDatastore` and `UpdateFrom` lines here.

In any case I'll merge this PR, as it's a tiny window - but please do consider and respond to this comment.

fasaxc: You're correct that the config may have changed between loading it above and this line.  However, we reload the config again as part of the snapshot processing.  At the end of the snapshot, we compare the snapshot-loaded config against the values above.

neiljerram: Ah right, I wasn't sure whether there was any config snapshot processing.  (I looked a bit, but not long enough to be sure either way.)

I thought that the `loadConfigFromDatastore` above might be there _because_ of the lack of any config snapshot processing; but I've just realized that it's needed for a different reason: because we need to merge datastore config before entering the config-change-makes-Felix-restart loop, or else Felix would just restart forever.

fasaxc: Yep, that's the reason.

PR #1460, opened 2017-05-18
Add config options for disabling Golang and process metrics

obourdon:
Allow new runtime parameter to include or not Golang GC metrics

Default is kept unchanged aka keep Golang metrics

### DIFF
--- a/config/config_params.go
+++ b/config/config_params.go
@@ -139,8 +139,10 @@ type Config struct {
 
 	DisableConntrackInvalidCheck bool `config:"bool;false"`
 
-	PrometheusMetricsEnabled bool `config:"bool;false"`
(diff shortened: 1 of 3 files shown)


marvin-tigera:

fasaxc:
ok to test

fasaxc:
I'm a little wary of the need to remove the pre-existing collectors.  What's your use-case for wanting to remove them?

fasaxc:
Thanks for the explanation on slack.  Think I can let this go in since the new behaviour is disabled by default.

obourdon:
@fasaxc thanks for your understanding. Just for reference I copy paste what I mentioned on slack here:

**_This is a minor change but I think that it might be interesting for some cases where we do not want to get too many metrics if number of nodes is very high and by default behaviour is kept similar as it is today._**

PR #1461, opened 2017-05-22
Measure occupancy with large numbers of allowed/denied packets

neiljerram:
This commit adds a scenario where we generate large numbers of packets
between 2 pods, and in alternate cycles those packets are either allowed
or denied.  We measure any associated occupancy impact.

### DIFF
--- a/k8sfv/pod.go
+++ b/k8sfv/pod.go
@@ -15,8 +15,10 @@
 package main
 
 import (
+	"bytes"
 	"fmt"
 	"math/rand"
+	"os/exec"
 	"strings"
 	"sync"
 
@@ -123,8 +125,21 @@ func createPod(clientset *kubernetes.Clientset, d deployment, nsName string, spe
 		panicIfError(err)
 		err = netlink.LinkSetNsFd(podIf, int(podNamespace.Fd()))
 		panicIfError(err)
-		err = podNamespace.Do(func(_ ns.NetNS) error {
-			return netlink.LinkSetUp(podIf)
+		err = podNamespace.Do(func(_ ns.NetNS) (err error) {
+			err = netlink.LinkSetUp(podIf)
+			if err != nil {
+				return
+			}
+			err = runCommand("ip", "a", "a", ip+"/32", "dev", veth.PeerName)
+			if err != nil {
+				return
+			}
+			err = runCommand("ip", "r", "a", "169.254.169.254/32", "dev", veth.PeerName)
+			if err != nil {
+				return
+			}
+			err = runCommand("ip", "r", "a", "default", "via", "169.254.169.254", "dev", veth.PeerName)
+			return
 		})
 		panicIfError(err)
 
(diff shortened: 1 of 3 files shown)


k8sfv/pod.go
```diff
@@ -123,8 +125,21 @@ func createPod(clientset *kubernetes.Clientset, d deployment, nsName string, spe
...
 		err = netlink.LinkSetNsFd(podIf, int(podNamespace.Fd()))
 		panicIfError(err)
-		err = podNamespace.Do(func(_ ns.NetNS) error {
-			return netlink.LinkSetUp(podIf)
+		err = podNamespace.Do(func(_ ns.NetNS) (err error) {
+			err = netlink.LinkSetUp(podIf)
+			if err != nil {
+				return
+			}
+			err = runCommand("ip", "a", "a", ip+"/32", "dev", veth.PeerName)
```
fasaxc: I'd prefer if you'd spell out the `"ip", "addr", "add"`.  _I_ know what it means because I'm familiar with that command but it'll be hard to parse for a newbie.

k8sfv/pod.go
```diff
@@ -226,3 +241,31 @@ func cleanupAllPods(clientset *kubernetes.Clientset, nsPrefix string) {
 var zeroGracePeriod int64 = 0
 
 var deleteImmediately = &v1.DeleteOptions{GracePeriodSeconds: &zeroGracePeriod}
+
+func runCommand(command string, args ...string) error {
```
fasaxc: Could you use `Cmd.CombinedOutput()`, or are you keen to keep stdout/err separate?

k8sfv/scale_test.go
```diff
@@ -160,6 +164,49 @@ var _ = Context("with a k8s clientset", func() {
...
+				}
+				pod1 := createPod(clientset, d, nsName, podSpec{})
+				pod2 := createPod(clientset, d, nsName, podSpec{})
+				for ii := 0; ii < 5; ii++ {
+					// Run an nmap scan between the pods.
+					runNmap(pod1, pod2)
+					sumCalicoDeniedPackets(felixIP)
+
+					// Get Felix to GC and dump heap memory profile.
+					exec.Command("pkill", "-USR1", "calico-felix").Run()
```
fasaxc: This feels ripe for making into a utility function.

k8sfv/scale_test.go
```diff
@@ -160,6 +164,49 @@ var _ = Context("with a k8s clientset", func() {
...
+					time.Sleep(2 * time.Second)
```
fasaxc: Can you poll for the file to be created or something instead of Sleeping?

k8sfv/pod.go
neiljerram: Does that mean I don't need the dance with making my own buffers?

k8sfv/scale_test.go
neiljerram: Bit fiddly, as I'd have to have the same logic as Felix for the file name (template + timestamp).

k8sfv/pod.go
neiljerram: Nvm, I can go and look that up.

PR #1464, opened 2017-05-22
Add hostname tagged metric

obourdon:
This potentially allow prometheus smarter mapping like having:
sum by (host) (felix_active_local_endpoints * on (job, instance) group_left(host) felix_host)
instead of:
sum by (instance) (felix_active_local_endpoints)

because Felix metrics do not include any tags by default and according to
[https://groups.google.com/d/msgid/prometheus-developers/71a37ceb-e4bb-4894-be62-3ffc853c177f%40googlegroups.com](https://groups.google.com/d/msgid/prometheus-developers/71a37ceb-e4bb-4894-be62-3ffc853c177f%40googlegroups.com) adding host tag to all metrics is not the way to go

### DIFF
--- a/felix.go
+++ b/felix.go
@@ -397,6 +397,13 @@ configRetry:
 
 	if configParams.PrometheusMetricsEnabled {
 		log.Info("Prometheus metrics enabled.  Starting server.")
+		gaugeHost := prometheus.NewGauge(prometheus.GaugeOpts{
+			Name:        "felix_host",
+			Help:        "Configured Felix hostname (as a label), typically used in grouping/aggregating stats; the label defaults to the hostname of the host but can be overridden by configuration. The value of the gauge is always set to 1.",
+			ConstLabels: prometheus.Labels{"host": configParams.FelixHostname},
+		})
+		gaugeHost.Set(1)
+		prometheus.MustRegister(gaugeHost)
 		go servePrometheusMetrics(configParams)
 	}
 


marvin-tigera:

felix.go
```diff
@@ -466,6 +466,12 @@ func servePrometheusMetrics(configParams *config.Config) {
 				prometheus.Unregister(prometheus.NewProcessCollector(os.Getpid(), ""))
 			}
 		}
+		gaugeHost := prometheus.NewGaugeVec(prometheus.GaugeOpts{
```
fasaxc: Feels a bit odd to use `GaugeVec` here when there's only one metric, can you do something like this instead: 
```
prometheus.NewGauge(prometheus.GaugeOpts{
    Name: "felix_host",
    Help: "Felix host.",
    ConstLabels: prometheus.Labels{...}, 
})
```

felix.go
```diff
@@ -466,6 +466,12 @@ func servePrometheusMetrics(configParams *config.Config) {
...
+			Name: "felix_host",
+			Help: "Felix host.",
```
fasaxc: For the help, how about "Configured Felix hostname (as a label), typically used in grouping/aggregating stats; the label defaults to the hostname of the host but can be overridden by configuration.  The value of the gauge is always set to 1."

obourdon: Totally agree, will do

felix.go
obourdon: Makes sense, due to my current lack of in-depth knowledge in Prometheus GoLang client lib for now, just took the 1st function which seemed to work. Currently changing this

felix.go
```diff
@@ -466,6 +466,13 @@ func servePrometheusMetrics(configParams *config.Config) {
 				prometheus.Unregister(prometheus.NewProcessCollector(os.Getpid(), ""))
 			}
 		}
+		gaugeHost := prometheus.NewGauge(prometheus.GaugeOpts{
+			Name:        "felix_host",
+			Help:        "Configured Felix hostname (as a label), typically used in grouping/aggregating stats; the label defaults to the hostname of the host but can be overridden by configuration. The value of the gauge is always set to 1.",
+			ConstLabels: prometheus.Labels{"host": configParams.FelixHostname},
+		})
+		prometheus.MustRegister(gaugeHost)
```
fasaxc: Just spotted this was inside the loop.  Does it need to be?  Is registering idempotent?

The `configParams` can't change BTW, Felix restarts if they do.

felix.go
```diff
@@ -398,6 +398,13 @@ configRetry:
 	if configParams.PrometheusMetricsEnabled {
 		log.Info("Prometheus metrics enabled.  Starting server.")
 		go servePrometheusMetrics(configParams)
+		gaugeHost := prometheus.NewGauge(prometheus.GaugeOpts{
```
fasaxc: Please move this above the `go servePrometheusMetrics()` so that there's no race condition where we server metrics without this one being defined.

obourdon: Done

obourdon:
@fasaxc could you please review latest changes and if ok, merge. Thanks

fasaxc:
ok to test

PR #1465, opened 2017-05-30
Ensure our Docker image runs with a sane init daemon.

fasaxc:
## Description
Previously Felix ran as PID 1 inside our container.  That is not sensible because PID 1 has responsibilities; it must:

- reap zombie processes
- handle and fan out signals correctly (e.g. to make Ctrl-C work)

This PR adds the `tini` init daemon to wrap Felix.  `tini` is designed for precisely this purpose (to handle the responsibilities of PID 1 while running exactly one child process).

### DIFF
--- a/docker-image/Dockerfile
+++ b/docker-image/Dockerfile
@@ -1,6 +1,10 @@
 FROM alpine:3.4
 MAINTAINER Shaun Crampton <shaun@tigera.io>
 
+# Since our binary isn't designed to run as PID 1, run it via the tini init daemon.
+RUN apk --no-cache add --update tini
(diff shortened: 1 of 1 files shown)


docker-image/Dockerfile
```diff
@@ -1,6 +1,10 @@
 FROM alpine:3.4
 MAINTAINER Shaun Crampton <shaun@tigera.io>
 
+# Since our binary isn't designed to run as PID 1, run it via the tini init daemon.
+RUN apk add --update tini
```
tomdee: I think it's better to use `--no-cache` to keep the image size down

PR #1466, opened 2017-06-01
Add support for Typha, the sync-offload daemon.

fasaxc:
## Description
This PR allows Felix to use a remote Syncer connection via the [Typha](https://github.com/projectcalico/typha) fan-out daemon.  By offloading the work of the Syncer, we reduce the watcher burden on the datastore.

This support is currently experimental and the API is subject to breaking changes!.

## Todos
- [x] Documentation (in plan)

### DIFF
--- a/config/config_params.go
+++ b/config/config_params.go
@@ -103,9 +103,9 @@ type Config struct {
 	EtcdCaFile    string   `config:"file(must-exist);;local"`
 	EtcdEndpoints []string `config:"endpoint-list;;local"`
 
-	TyphaServiceEnvVarPfx string `config:"string;"`
-	TyphaK8sServiceName   string `config:"string;"`
-	TyphaAddr             string `config:"authority;;"`
+	TyphaAddr           string `config:"authority;;"`
+	TyphaK8sServiceName string `config:"string;"`
+	TyphaK8sNamespace   string `config:"string;kube-system;non-zero"`
 
 	Ipv6Support    bool `config:"bool;true"`
 	IgnoreLooseRPF bool `config:"bool;false"`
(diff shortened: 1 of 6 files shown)


neiljerram:
@fasaxc This fix seems to be needed to the coreos/etcd Git ID: https://github.com/neiljerram/felix/commit/27fd85e4fe1a4f6ef305d26a09d623b2db43dff0

felix.go
```diff
@@ -778,3 +809,51 @@ func (fc *DataplaneConnector) Start() {
...
+	if err != nil {
+		log.WithError(err).Error("Unable to create Kubernetes config.")
+		return "", err
+	}
+	clientset, err := kubernetes.NewForConfig(k8sconf)
+	if err != nil {
+		log.WithError(err).Error("Unable to create Kubernetes client set.")
+		return "", err
+	}
+	svcClient := clientset.CoreV1().Services(configParams.TyphaK8sNamespace)
```
neiljerram: Does namespace need to be non-"" ?  If so, we should check that above with TyphaK8sServiceName.

config/config_params.go
```diff
@@ -103,6 +103,11 @@ type Config struct {
 	EtcdCaFile    string   `config:"file(must-exist);;local"`
 	EtcdEndpoints []string `config:"endpoint-list;;local"`
 
+	TyphaServiceEnvVarPfx string `config:"string;"`
```
neiljerram: Now unused?

felix.go
fasaxc: I've tackled that in the config; marking that config param as non-zero.

PR #1467, opened 2017-06-02
update k8s FVs to use apimachinery watch and metav1 imports

gunjan5:
current imports conflict with libcalico-go latest dependency versions


caseydavenport:
Included in https://github.com/projectcalico/felix/pull/1469

PR #1474, opened 2017-06-08
Scan iptables-save output when calculating hashes.

fasaxc:
Reduce CPU usage and garbage generated by the iptables read loop.  A bit of profiling last week suggested these improvements:

- Read the input incrementally.
- Avoid copying each line of input to string by using the `[]byte` versions of the regex functions.

## Todos
- [x] Perf test; does it actually help!?

### DIFF
--- a/iptables/table.go
+++ b/iptables/table.go
@@ -589,10 +589,14 @@ func (t *Table) attemptToGetHashesFromDataplane() (hashes map[string][]string, e
 
 	stdout, err := cmd.StdoutPipe()
 	if err != nil {
+		log.WithError(err).Warnf("Failed to get stdout pipe for %s", t.iptablesSaveCmd)
 		return
 	}
 	err = cmd.Start()
 	if err != nil {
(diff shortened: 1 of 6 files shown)


iptables/table.go
```diff
@@ -575,69 +576,129 @@ func (t *Table) getHashesFromDataplane() map[string][]string {
...
 // returns a zero string.  Hence, the lengths of the returned values are the lengths of the chains
 // whether written by Felix or not.
-func (t *Table) getHashesFromBuffer(buf *bytes.Buffer) map[string][]string {
-	newHashes := map[string][]string{}
-	for {
+func (t *Table) readHashesFrom(r io.ReadCloser) (hashes map[string][]string, err error) {
+	hashes = map[string][]string{}
+	scanner := bufio.NewScanner(r)
+
+	// Figure out is debug logging is enabled so we can disable some WithFields() calls in the
```
neiljerram: s/is/if/

neiljerram: s/disable/skip/

PR #1475, opened 2017-06-08
Plumb through configuration for the post-write check interval.

fasaxc:
## Description
Make the post-write iptables check interval configurable.  I suspect that the previous default of 50ms is too aggressive, resulting in high CPU usage.

## Todos
- [ ] Documentation

### DIFF
--- a/config/config_params.go
+++ b/config/config_params.go
@@ -109,7 +109,7 @@ type Config struct {
 	Ipv6Support    bool `config:"bool;true"`
 	IgnoreLooseRPF bool `config:"bool;false"`
 
-	IptablesRefreshInterval            int           `config:"int;10"`
+	IptablesRefreshInterval            time.Duration `config:"seconds;10"`
 	IptablesPostWriteCheckIntervalSecs time.Duration `config:"seconds;1"`
 
 	MetadataAddr string `config:"hostname;127.0.0.1;die-on-fail"`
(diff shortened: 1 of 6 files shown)


iptables/table.go
```diff
@@ -29,7 +29,8 @@ import (
 )
 
 const (
-	MaxChainNameLength = 28
+	MaxChainNameLength       = 28
+	defaultPostWriteInterval = 50 * time.Millisecond
```
neiljerram: This is really a minimum, isn't it, so `minPostWriteInterval` perhaps?

IIUC it never gets used as a default, because the default IptablesPostWriteCheckInterval is 1s, which is larger than this.

config/config_params.go
```diff
@@ -109,7 +109,8 @@ type Config struct {
 	Ipv6Support    bool `config:"bool;true"`
 	IgnoreLooseRPF bool `config:"bool;false"`
 
-	IptablesRefreshInterval int `config:"int;10"`
+	IptablesRefreshInterval            int           `config:"int;10"`
+	IptablesPostWriteCheckIntervalSecs time.Duration `config:"seconds;1"`
```
neiljerram: Will it be clear to the Felix admin/configurer that IptablesPostWriteCheckIntervalSecs can be set to a float, whereas other interval-ish config settings need to be ints?  If not, is there something we can do to make that clearer?

PR #1476, opened 2017-06-08
Prep for a subset of k8sfv to run in semaphore CI

neiljerram:
Adds a way of identifying tests that take longer than a minute, and a "JUST_A_MINUTE" setting that means to skip those tests.  So, once this has merged, the semaphore CI can do `JUST_A_MINUTE=true make k8sfv-test`

Also:
- fixes up k8sfv imports following recent client-go/apimachinery moves
- fixes up how clients are authorized by the API server, following jump from API server 1.5.3 to 1.6.4.

### DIFF
--- a/iptables/table.go
+++ b/iptables/table.go
@@ -17,6 +17,7 @@ package iptables
 import (
 	"bytes"
 	"fmt"
+	"os/exec"
 	"reflect"
 	"regexp"
 	"strings"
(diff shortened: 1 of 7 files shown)


neiljerram:
@fasaxc I added quite a bit more here, after seeing that the test ends prematurely on semaphore, so think it makes sense for you to review again.

PR #1482, opened 2017-06-15
Skip recalculation of selector matches if selector hasn't changed.

fasaxc:
## Description
Optimisation: skip recalculation of selector matches if selector hasn't changed.

I noticed that the recalculation was being done often in our churn testing and it's fairly expensive so should be worth skipping.

## Todos
- [x] Unit tests (full coverage) - was already covered.

### DIFF
--- a/labelindex/label_inheritance_index.go
+++ b/labelindex/label_inheritance_index.go
@@ -190,10 +190,17 @@ func (l *InheritIndex) OnUpdate(update api.Update) (_ bool) {
 }
 
 func (idx *InheritIndex) UpdateSelector(id interface{}, sel selector.Selector) {
-	log.Infof("Updating selector %v", id)
 	if sel == nil {
 		log.WithField("id", id).Panic("Selector should not be nil")
 	}
+	oldSel := idx.selectorsById[id]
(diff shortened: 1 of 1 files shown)


labelindex/label_inheritance_index.go
```diff
@@ -190,10 +190,17 @@ func (l *InheritIndex) OnUpdate(update api.Update) (_ bool) {
...
+	oldSel := idx.selectorsById[id]
+	// Since the selectorRoot struct has cache fields, the easiest way to compare two
+	// selectors is to compare their IDs.
+	if oldSel != nil && oldSel.UniqueID() == sel.UniqueID() {
+		log.WithField("selID", id).Info("Skipping unchanged selector")
+		return
+	}
+	log.WithField("selID", id).Info("Updating selector")
 	idx.scanAllLabels(id, sel)
 	idx.selectorsById[id] = sel
```
neiljerram: Might we still want to do this line - i.e. not return before it - even when the old and new selectors have the same UniqueID?

fasaxc: As discussed, I don't think there's going to be any differing internal state here and I can't convince myself it's better to throw away the older copy of an object.

PR #1483, opened 2017-06-15
Add support for multiple CIDRs in a match rule.

fasaxc:
## Description
Allow a single rule to contain more than one CIDR.  Multiple CIDRs are or-ed together.

## Todos
- [x] Finalise datamodel change
- [x] Finalise mixed IP version handling.
- [x] Documentation
- [x] Switch to main libcalico-go repo once change is merged
- [x] Markups
- [ ] Integration tests in calico repo?

## Release Note

```release-note
Felix now supports policy match rules that match on multiple source/destination CIDRs.
```


neiljerram:
Documentation at https://github.com/projectcalico/calico/pull/895

neiljerram:
Looks like you can check the "Switch to main libcalico-go repo once change is merged" item now (in top comment above).

rules/policy.go
```diff
@@ -59,9 +59,202 @@ func (r *DefaultRuleRenderer) ProtoRulesToIptablesRules(protoRules []*proto.Rule
...
+	//     positive matches on dest address
+	//     negated matches on source address
+	//     negated matches on dest address
+	//     rule containing rest of match critera
+	//
+	// We use one match bit to record whether all the blocks accept the packet and one as a
+	// scratch bit for each block to use.  As an invariant, at the end of each block, the
+	// "all blocks pass" bit should only be set if all previous blocks match the packet.
+	//
+	// We do a some optimisations to keep the number of rules down:
```
neiljerram: delete " a "

rules/policy.go
```diff
@@ -59,9 +59,202 @@ func (r *DefaultRuleRenderer) ProtoRulesToIptablesRules(protoRules []*proto.Rule
...
+		}
+
+		if len(ruleCopy.SrcNet) > 1 {
+			// More than one positive match on source IP, need to render a block.  We
+			// prioritise the positive match over the negative match and avoid
+			// rendering a block if it'd only contain one IP.
+			appendCIDRMatchBlock(ruleCopy.SrcNet, src)
+			// Since we're using a block for this, nil out the match.
+			ruleCopy.SrcNet = nil
+		}
```
neiljerram: assert here that SrcNet is either nil or length 1 ? - it would be nice (for the code reader) to know that it can't be non-nil length 0.

rules/policy.go
```diff
@@ -59,9 +59,202 @@ func (r *DefaultRuleRenderer) ProtoRulesToIptablesRules(protoRules []*proto.Rule
...
+
+		if len(ruleCopy.DstNet) > 1 {
+			// More than one positive match on dest IP, need to render a block.  We
+			// prioritise the positive match over the negative match and avoid
+			// rendering a block if it'd only contain one IP.
+			appendCIDRMatchBlock(ruleCopy.DstNet, dst)
+			// Since we're using a block for this, nil out the match.
+			ruleCopy.DstNet = nil
+		}
```
neiljerram: Sim here, if you agree with above.

rules/policy.go
```diff
@@ -83,28 +276,48 @@ func (r *DefaultRuleRenderer) ProtoRuleToIptablesRules(pRule *proto.Rule, ipVers
...
 				return nil
 			}
-
-			markBit, actions := r.CalculateActions(match, &ruleCopy, ipVersion)
+			if usingCIDRBlocks {
+				// The CIDR matches in the rule overflowed and we rendered them
+				// as additional rules, which set the markAllBlocksPass bit on
+				// success.  Add a match on that bit to the calculated rule.
+				match = match.MarkSet(markAllBlocksPass)
+			}
```
neiljerram: Not sure this loop works if you have multiple port ranges with `usingCIDRBlocks`.  At the start of the first iteration, the relevant mark bit is markAllBlocksPass, but at the end of the first iteration - and therefore the start of the second, if there is one - the relevant mark bit may be markBit.

rules/policy_test.go
```diff
@@ -184,13 +184,15 @@ var ruleTestData = []TableEntry{
...
-		IptablesMarkPass:   0x10,
-		IptablesLogPrefix:  "calico-packet",
+		IPIPEnabled:          true,
+		IPIPTunnelAddress:    nil,
+		IPSetConfigV4:        ipsets.NewIPVersionConfig(ipsets.IPFamilyV4, "cali", nil, nil),
+		IPSetConfigV6:        ipsets.NewIPVersionConfig(ipsets.IPFamilyV6, "cali", nil, nil),
+		IptablesMarkAccept:   0x8,
+		IptablesMarkPass:     0x10,
+		IptablesMarkScratch0: 0x20,
+		IptablesMarkScratch1: 0x40,
```
neiljerram: How about setting these marks to some different values for this block?  You never know...

rules/policy_test.go
```diff
@@ -285,6 +287,129 @@ var _ = Describe("Protobuf rule to iptables rule conversion", func() {
...
+		Entry("1 src, 1 !src, 1 dst, 1 !dst", 1, 1, 1, 1, []string{
+			preSetAllBlocksMarkRule,
+			"-A test --source 11.0.0.0/24 --jump MARK --set-mark 0/0x20",
+			"-A test --destination 13.0.0.0/24 --jump MARK --set-mark 0/0x20",
+			"-A test --source 10.0.0.0/24 --destination 12.0.0.0/24 -m mark --mark 0x20/0x20 --jump MARK --set-mark 0x8/0x8",
+			returnRule,
+		}),
+
+		// Two of everything; everything overflows.
+		Entry("1 src, 1 !src, 1 dst, 1 !dst", 2, 2, 2, 2, []string{
```
neiljerram: test name needs correcting: "2 src, 2 !src, 2 dst, 2 !dst"

rules/rule_defs.go
```diff
@@ -201,8 +205,38 @@ type Config struct {
 	DisableConntrackInvalid bool
 }
 
+func (c *Config) validate() {
+	// Scan for unset iptables mark bits.  We use reflection so taht we have a hope of catching
```
neiljerram: "taht"

rules/rule_defs.go
```diff
@@ -201,8 +205,38 @@ type Config struct {
...
+			if usedBits&bits > 0 {
+				log.WithField("field", fieldName).Panic(
+					"IptablesMarkXXX field overlapped with another's bits.")
+			}
+			usedBits |= bits
+			found++
+		}
+	}
+	if found == 0 {
+		// Check the reflection found everything we were expecting.
```
neiljerram: more correctly: "found something that we were expecting"

rules/policy.go
fasaxc: Might be missing something but I _think_ it's correct; the expansion of ports works a bit differently than the new feature.  We only ever expect one of the port match rules to fire, all the rules in the port match block are checking the same thing using the same bit, i.e. whether the preceding CIDR blocks matched.

The logic is something like this:

Simplifying slightly by assuming we're only allowed 1 port in a rule and we need to match `source port  in {1234, 12345} and dest port in {80, 8080}`.  We do the cross-product of the ports:

- `(1234, 80)`
- `(1234, 8080)`
- `(12345, 80)`
- `(12345, 8080)`

and render something like this:

```
<cidr blocks>
# at this point all-blocks bit is set iff the CIDR matches were successful
if all-blocks-bit && src-port == 1234 && dest port == 80 then <rule action>
if all-blocks-bit && src-port == 1234 && dest port == 8080 then <rule action>
if all-blocks-bit && src-port == 12345 && dest port == 80 then <rule action>
if all-blocks-bit && src-port == 12345 && dest port == 8080 then <rule action>
```

(In reality, the kernel can support 15 ports in a rule so we do the cross-product of chunks of 15 ports but the logic is similar.)

rules/policy.go
fasaxc: As discussed, it's not very golang-ish to distinguish nil and 0-length for slices since `for .. range` and `len()` don't distinguish them either.

PR #1487, opened 2017-06-22
[WiP] Add a grace period before deleting routes.

fasaxc:
## Description
The CNI plugin pre-adds the routes for containers that it networks.  This can lead to flapping the route when Felix spots the interface before it hears about the endpoint.  Add a grace period before cleaning up the route to avoid this.

## Todos
- [ ] Check IP re-use cases work; do we need to bypass the grace period if the IP is added to another endpoint?
- [ ] Documentation
- [ ] Backport


matthewdupre:
Continued as #1518

PR #1489, opened 2017-06-23
Implement health endpoints for Felix

neiljerram:
## Todos
- [x] FV tests
- [ ] k8s integration testing
- [ ] ?Moving health package elsewhere so that Typha can share it
- [x] ?Commit squashing

### DIFF
--- a/calc/async_calc_graph.go
+++ b/calc/async_calc_graph.go
@@ -25,6 +25,7 @@ import (
 
 	"github.com/projectcalico/felix/config"
 	"github.com/projectcalico/felix/dispatcher"
+	"github.com/projectcalico/felix/health"
 	"github.com/projectcalico/felix/proto"
 	"github.com/projectcalico/libcalico-go/lib/backend/api"
 )
@@ -77,22 +78,33 @@ type AsyncCalcGraph struct {
 	eventBuffer      *EventSequencer
 	beenInSync       bool
 	needToSendInSync bool
+	syncStatusNow    api.SyncStatus
+	healthAggregator *health.HealthAggregator
 
 	flushTicks       <-chan time.Time
 	flushLeakyBucket int
 	dirty            bool
 }
 
-func NewAsyncCalcGraph(conf *config.Config, outputEvents chan<- interface{}) *AsyncCalcGraph {
+const (
+	healthName     = "async_calc_graph"
+	healthInterval = 10 * time.Second
+)
+
+func NewAsyncCalcGraph(conf *config.Config, outputEvents chan<- interface{}, healthAggregator *health.HealthAggregator) *AsyncCalcGraph {
 	eventBuffer := NewEventBuffer(conf)
 	disp := NewCalculationGraph(eventBuffer, conf.FelixHostname)
 	g := &AsyncCalcGraph{
-		inputEvents:  make(chan interface{}, 10),
-		outputEvents: outputEvents,
-		Dispatcher:   disp,
-		eventBuffer:  eventBuffer,
+		inputEvents:      make(chan interface{}, 10),
+		outputEvents:     outputEvents,
+		Dispatcher:       disp,
+		eventBuffer:      eventBuffer,
+		healthAggregator: healthAggregator,
 	}
 	eventBuffer.Callback = g.onEvent
+	if healthAggregator != nil {
+		healthAggregator.RegisterReporter(healthName, &health.HealthReport{Live: true, Ready: true}, healthInterval*2)
+	}
 	return g
 }
 
@@ -112,6 +124,8 @@ func (acg *AsyncCalcGraph) OnStatusUpdated(status api.SyncStatus) {
 
 func (acg *AsyncCalcGraph) loop() {
 	log.Info("AsyncCalcGraph running")
+	healthTicks := time.NewTicker(healthInterval).C
+	acg.reportHealth()
 	for {
 		select {
 		case update := <-acg.inputEvents:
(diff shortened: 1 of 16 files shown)


health/health.go
```diff
@@ -0,0 +1,119 @@
...
+				"ready": ready,
+				"live":  live,
+			}).Debug("Health now")
+			mutex.Unlock()
+		}
+	}()
+
+	for {
+		log.WithField("port", port).Info("Starting health endpoints")
+		http.HandleFunc("/readiness", func(rsp http.ResponseWriter, req *http.Request) {
```
fasaxc: Do these need to be reregistered each time around the loop?  Are there any side-effects of doing that?

neiljerram: No, I don't think so - have changed to as only to register once.

neiljerram:
@fasaxc - modulo the Todos listed above, I think this is ready for full code review now.

health/health.go
```diff
@@ -0,0 +1,137 @@
...
+
+// For a component that provides health indications, return the sources that it provides to indicate
+// readiness, and those that it provides to indicate liveness.
+type HealthProvider interface {
+	ReadySources() []HealthSource
+	LiveSources() []HealthSource
+}
+
+func MonitorHealth(
+	ready *bool,
```
fasaxc: I think this would be cleaner as a method on a struct rather than a function with lots of parameters.  Then you have a place to document the mutex and what it protects, and you avoid needing to pass in pointers to bools, which is an odd pattern.

health/health_test.go
```diff
@@ -0,0 +1,171 @@
...
+
+	AfterEach(func() {
+		close(healthChannel)
+		Eventually(getReady).Should(BeFalse())
+		Eventually(getLive).Should(BeFalse())
+	})
+
+	It("initially reports false", func() {
+		Expect(getReady()).To(BeFalse())
+		Expect(getLive()).To(BeFalse())
```
fasaxc: Liveness should probably initially report true

set/set.go
```diff
@@ -131,3 +132,15 @@ func (set mapSet) Equals(other Set) bool {
 	}
 	return true
 }
+
+func (set mapSet) ContainsAll(other Set) bool {
```
fasaxc: I think set has thorough UTs so it'd be good to add one for this new method.

neiljerram:
@fasaxc - Markups done; could you take a look and check that they are as you were expecting?

health/health.go
```diff
@@ -0,0 +1,171 @@
...
+	return &HealthState{ready: false, live: true, mutex: &sync.Mutex{}}
+}
+
+func MonitorHealth(
+	state *HealthState,
+	neededForReady set.Set,
+	neededForLive set.Set,
+	c <-chan HealthIndicator,
+) {
+	currentHealth := set.New()
```
fasaxc: `s/currentHealth/healthySources/` ?

felix.go
```diff
@@ -449,6 +468,16 @@ configRetry:
 		go servePrometheusMetrics(configParams)
 	}
 
+	if configParams.HealthEnabled {
+		log.Info("Health enabled.  Starting server.")
```
fasaxc: Useful to log the port?

health/health.go
```diff
@@ -0,0 +1,171 @@
...
+	Source HealthSource
+
+	// How long the indicator is valid for.  In other words, if it continues operating normally,
+	// the source expects to refresh this indicator before this timeout.
+	Timeout time.Duration
+}
+
+// For a component that provides health indications, return the sources that it provides to indicate
+// readiness, and those that it provides to indicate liveness.
+type HealthProvider interface {
```
fasaxc: Suggest using this with a `Register()` method on new `HeathMonitor` struct.

calc/async_calc_graph.go
```diff
@@ -77,22 +78,33 @@ type AsyncCalcGraph struct {
...
+	healthState      *health.HealthState
 
 	flushTicks       <-chan time.Time
 	flushLeakyBucket int
 	dirty            bool
 }
 
-func NewAsyncCalcGraph(conf *config.Config, outputEvents chan<- interface{}) *AsyncCalcGraph {
+const (
+	HEALTH_NAME     = "async_calc_graph"
```
fasaxc: I think go's naming convention for consts is `camelCase`

health/health.go
```diff
@@ -0,0 +1,166 @@
...
+	timeout time.Duration
+
+	// The most recent report.
+	latest HealthReport
+
+	// Time of that most recent report.
+	timestamp time.Time
+}
+
+type HealthState struct {
```
fasaxc: `HealthState` feels like the name of a value struct holding an particular health value to me.  How about `HealthAggregator` or `HealthServer` or some such?

health/health.go
```diff
@@ -0,0 +1,166 @@
...
+	// Mutex to protect concurrent access to this health state.
+	mutex *sync.Mutex
+
+	// Map from reporter name to corresponding state.
+	reporters map[string]*reporterState
+}
+
+func (state *HealthState) RegisterReporter(
+	// Unique name for this reporter.
```
fasaxc: This comments-interlaced-with-params approach is certainly readable but does it work with godoc etc?

intdataplane/int_dataplane.go
```diff
@@ -326,6 +334,15 @@ func NewIntDataplaneDriver(config Config) *InternalDataplane {
 		dp.allIptablesTables = append(dp.allIptablesTables, t)
 	}
 
+	// Register that we will report a liveness indicator.
+	if config.HealthState != nil {
+		config.HealthState.RegisterReporter(
+			HEALTH_NAME,
+			&health.HealthReport{Live: true, Ready: true},
+			HEALTH_INTERVAL*2,
```
fasaxc: Might be best to increase the grace period on the dataplane?  We've seen conntrack block for up to 12s (although we mean to fix that).

neiljerram: As discussed, we already have 60s here, with the k8s default failureThreshold - so leaving this as is.

Issue #1470, opened 2017-06-06
Tell iptables-restore to use its lock if available


fasaxc:
Recently, iptables-restore has been updated to support the xtables lock, which prevents concurrent access to iptables os that different users can avoid clobberring each others updates.  We should add support for the lock and turn it on when available.

- [x] Figure out what settings kube-proxy uses (I think there are multiple options for the lock now)
- [x] Add support for setting the flag
- [ ] ~~Detect support (version check or trial call?)~~
- [ ] Update container image to make it easy to mount in the relevant files (lock is file-based so we need to share the file).
- [ ] Make iptables retry/resync less aggressive since it won't be needed to avoid kube-proxy conflicts.

fasaxc:
Hmm, the iptables-restore locking function hasn't landed yet.  `kube-proxy` re-implements the iptables-restore lock itself: https://github.com/kubernetes/kubernetes/blob/master/pkg/util/iptables/iptables.go#L360

caseydavenport:
@fasaxc I think this is done now?

fasaxc:
@caseydavenport Kind-of.  We hit some issues so it's off-by-default. #1504

PR #1491, opened 2017-06-27
Grab the iptables lock before running iptables-restore.

fasaxc:
## Description
To avoid conflicts with kube-proxy (and others), take the iptables lock before running iptables-restore.

See #1470

## Todos
- [x] Integration tests (delete as appropriate) Done
- [ ] Double check license requirements

## Release Note

```release-note
Felix now acquires the iptables lock while manipulating iptables.  This prevents conflicts 
with other applications, such as kube-proxy (as long as they also honor the lock).  
Upgrade note: to be effective if Felix is running in a container, this feature requires the 
directory containing the iptables lock file, `/run`, to be mounted into the container.  
```


fasaxc:
retest this please

iptables/lock.go
```diff
@@ -0,0 +1,214 @@
...
+)
+
+var (
+	summaryLockAcquisitionTime = prometheus.NewSummary(prometheus.SummaryOpts{
+		Name: "felix_iptables_lock_acquire_secs",
+		Help: "Time in seconds that it took to acquire the iptables lock(s).",
+	})
+	countLockTimeouts = prometheus.NewCounter(prometheus.CounterOpts{
+		Name: "felix_iptables_lock_timeouts",
+		Help: "Number of lock timeouts while waiting for hte iptables lock(s).",
```
neiljerram: "hte"

iptables/lock.go
```diff
@@ -0,0 +1,214 @@
...
+		l.Lock14, err = net.ListenUnix("unix", &net.UnixAddr{Name: socketName, Net: "unix"})
+		if err == nil {
+			break
+		}
+		if time.Since(startTime14) > timeout {
+			countLockTimeouts.Inc()
+			return nil, Err14LockTimeout
+		}
+		time.Sleep(probeInterval)
+		countLockRetries.Inc()
```
neiljerram: Feels like it would be useful for the 14 and 16 metrics to be distinct - perhaps with labels.  In a failing system, it's probably going to be one of them in particular that is problematic.

config/param_types.go
```diff
@@ -140,6 +140,20 @@ func (p *SecondsParam) Parse(raw string) (result interface{}, err error) {
...
+	Metadata
+}
+
+func (p *MillisParam) Parse(raw string) (result interface{}, err error) {
+	millis, err := strconv.ParseFloat(raw, 64)
+	if err != nil {
+		err = p.parseFailed(raw, "invalid float")
+		return
+	}
+	result = time.Duration(millis * float64(time.Millisecond))
```
neiljerram: Is this cast needed?  I'm curious to understand, if so.

intdataplane/int_dataplane.go
```diff
@@ -203,38 +215,45 @@ func NewIntDataplaneDriver(config Config) *InternalDataplane {
...
+		RefreshInterval:       config.IptablesRefreshInterval,
+		PostWriteInterval:     config.IptablesPostWriteCheckInterval,
+	}
+
+	// However, the NAT tables need an extra cleanup regex.
+	iptablesNATOptions := iptablesOptions
+	iptablesNATOptions.ExtraCleanupRegexPattern = rules.HistoricInsertedNATRuleRegex
+
+	// Create the shared iptables lock.  This allows us to block other processes from
+	// manipulating iptables but it still allows us to do our work in parallel.
```
neiljerram: I didn't understand the last clause on first reading this.  Perhaps "but it still allows our own goroutines to modify iptables in parallel - which is OK because they are operating on different tables"?

config/param_types.go
fasaxc: Yes, it is.  I want to allow the input to be a float, so you could say `1.2` (milliseconds).  Go's casting rules are very strict so you can't do `float * duration` (which is really an int64 number of nanoseconds) without a cast.  Then I need to cast back to `time.Duration` at the end.

Casting millis to `time.Duration` first then multiplying doesn't work because that would lose precision.

neiljerram: That makes sense - thanks.

Makefile
```diff
@@ -324,6 +330,16 @@ ut combined.coverprofile: vendor/.up-to-date $(FELIX_GO_FILES)
 	@echo Running Go UTs.
 	$(DOCKER_GO_BUILD) ./utils/run-coverage
 
+bin/fv.test: vendor/.up-to-date $(FELIX_GO_FILES)
+	$(DOCKER_GO_BUILD) go test ./fv -c --tags fvtests -o bin/fv.test
```
neiljerram: When I added the `k8sfv` directory, I needed to exclude it from some of the infrastructure for running UTs:

```
./utils/run-coverage:11:	      grep -vE '/vendor/|\./proto/|.glide|/k8sfv/' | \
./utils/run-coverage:24:ginkgo -cover -covermode=count -coverpkg=${go_dirs} -r -skipPackage k8sfv
./Makefile:352:	$(DOCKER_GO_BUILD) ginkgo -r -skipPackage k8sfv $(GINKGO_OPTIONS)
./Makefile:357:	$(DOCKER_GO_BUILD) ginkgo watch -r -skipPackage k8sfv $(GINKGO_OPTIONS)
```

Do you need to do that for `fv` as well?

fv/iptables_lock_test.go
```diff
@@ -0,0 +1,112 @@
...
+	cmdInContainer := func(cmd ...string) *exec.Cmd {
+		arg := []string{"exec", containerName}
+		arg = append(arg, cmd...)
+		return exec.Command("docker", arg...)
+	}
+
+	BeforeEach(func() {
+		containerName = fmt.Sprintf("felix-fv-%d-%d", os.Getpid(), containerIdx)
+		containerIdx++
+		felixDir, err := os.Getwd()
```
neiljerram: Do these tests need felixDir to be correct?  I would expect Getwd() here to give `<reporoot>/fv`, not `<reporoot>` - and I'm not sure that is what you intend with the name `felixDir`.

Makefile
fasaxc: I saw a neat trick in that Cloudflare article: using a build tag to identify which tests to build/run and I went for that approach.  WDYT?

fv/iptables_lock_test.go
fasaxc: Hmm, it works from the makefile due to the way we run it.  Maybe I should be tolerant to that though, so that `ginkgo watch` does the right thing.

caseydavenport:
@fasaxc I think we've got duplicated release note information coming from this PR and https://github.com/projectcalico/calico/pull/935.  Can we remove the label from this one?

fasaxc:
Sure.

PR #1493, opened 2017-06-28
Allow SNAT to be performed on all fixed IP traffic

logan2211:
## Description
Previously SNAT was only being perfomed on traffic from the fixed IP to itself. This doesn't perform full aliasing on the fixed IP address which is expected by most floating IP implementations as this allows them to reach the external networks without any further instance configuration.

## Todos
- [ ] Documentation

### DIFF
--- a/config/config_params.go
+++ b/config/config_params.go
@@ -106,8 +106,9 @@ type Config struct {
 	TyphaK8sServiceName string `config:"string;"`
 	TyphaK8sNamespace   string `config:"string;kube-system;non-zero"`
 
-	Ipv6Support    bool `config:"bool;true"`
-	IgnoreLooseRPF bool `config:"bool;false"`
+	Ipv6Support        bool   `config:"bool;true"`
+	IgnoreLooseRPF     bool   `config:"bool;false"`
+	NatSnatDestination string `config:"oneof(private,any);private;non-zero,die-on-fail"`
 
 	IptablesRefreshInterval            time.Duration `config:"seconds;10"`
 	IptablesPostWriteCheckIntervalSecs time.Duration `config:"seconds;1"`
--- a/felix.go
+++ b/felix.go
@@ -280,6 +280,8 @@ configRetry:
 				FailsafeOutboundHostPorts: configParams.FailsafeOutboundHostPorts,
 
 				DisableConntrackInvalid: configParams.DisableConntrackInvalidCheck,
+
+				NatSnatDestination: configParams.NatSnatDestination,
 			},
 			IPIPMTU:                        configParams.IpInIpMtu,
 			IptablesRefreshInterval:        configParams.IptablesRefreshInterval,
--- a/felix.go
+++ b/felix.go
@@ -239,10 +239,12 @@ configRetry:
 		markAccept := configParams.NextIptablesMark()
 		markPass := configParams.NextIptablesMark()
 		markWorkload := configParams.NextIptablesMark()
+		markSnatSkip := configParams.NextIptablesMark()
 		log.WithFields(log.Fields{
 			"acceptMark":   markAccept,
 			"passMark":     markPass,
 			"workloadMark": markWorkload,
+			"snatSkipMark": markSnatSkip,
 		}).Info("Calculated iptables mark bits")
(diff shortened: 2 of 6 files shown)


marvin-tigera:

fasaxc:
@Logan2211, I see a couple of problems with this:

- If there's more than one workload behind the VIP then return traffic could be routed back to the wrong instance.
- Calico policy is based on source IP so if you had policy that said "allow from X" but then you SNAT all of X's outgoing traffic then the policy won't match.

Can you give some examples of other systems where they SNAT the outgoing VIP?

Logan2211:
I'm approaching it from the Openstack perspective because that's where I'm using this.

> If there's more than one workload behind the VIP then return traffic could be routed back to the wrong instance.

In an Openstack Nova/Neutron environment, the FIP is routed to one instance.

> Calico policy is based on source IP so if you had policy that said "allow from X" but then you SNAT all of X's outgoing traffic then the policy won't match.

Openstack security groups only control inbound traffic

> Can you give some examples of other systems where they SNAT the outgoing VIP?

Most clouds floating ip implementations I've seen do not require a loopback configuration on the instance to function properly. Prime example is elastic IPs on ec2. The instance is unaware of the floating IP, only the private IP is bound and outbound traffic from the instance is snatted.

neiljerram:
When @nbartos originally implemented our floating IP support, we discussed what the behaviour should be as regards outbound traffic - and my recollection from that time is that it was really difficult to find any definition of what expected behaviour was - even restricting just to the OpenStack context.

On the other hand, now that you've raised this, @Logan2211 , it's trivial to look at the Nova and Neutron code and see that both nova-network and Neutron _do_ do the SNAT that you are proposing [1][2]

[1] https://github.com/openstack/nova/blob/master/nova/network/linux_net.py#L742:L757
[2] https://github.com/openstack/neutron/blob/master/neutron/agent/l3/router_info.py#L163:L169

If we compare what Neutron does with Calico (including your proposed change), though, there is a key difference, which is that Neutron's DNAT/SNAT for floating IPs happens on a network node that is distinct from the relevant compute host - which has the happy consequences that
- it doesn't interfere with any policy enforcement, using fixed IPs, that might be being performed on the compute host
- it doesn't interfere with the path from an instance to the metadata API
- it probably (?) doesn't interfere with the data path between instances on the same Neutron network.

Not sure yet what that all means for what we should do in Calico.... but food for further thought.

nbartos:
I recall that in our original discussion, we thought that the best solution would be some configurability about when the SNAT was applied, as the Calico network model didn't map cleanly into what other OpenStack SDNs were doing (using virtual routers to decide boundaries).  The single SNAT rule that I added was only so that a VM could connect to it's own floating IP.  I like the idea of finishing the implementation, I'm just not sure what the minimum configurability should be.  I suppose the more options, the better.

fasaxc:
I guess the ideal would be to do something similar to our "NAT outgoing" feature that we use for containers.  That preserves the source IP within the cluster but it SNATs traffic that is leaving the cluster.  That has a couple of problems in OpenStack:

- In container-land, we know what the container IP range is because we're doing the IPAM there; our rule matches against the IPAM blocks that we know about and SNATs if the packet is leaving those.  In OpenStack, we don't do the IPAM so we haven't plumbed the OpenStack network information into our datamodel (although you can configure IPAM blocks manually to get this behaviour).
- In container-land, we SNAT to the host's IP, which is different to this proposed behaviour, so we end up with yet another flag :-/

In any case, I'm wary of a global flag that changes this behaviour (but maybe it'd be OK as a stop-gap).  We have hybrid systems deployed (OpenStack + k8s + bare-metal, for example) and we could paint ourselves into a corner if OpenStack needs the flag set one way and k8s needs it set another way.  What about a field that is set per-NAT entry, then OpenStack can set it to `"nat_type": "bidirectional"`, say.

CC @lxpollitt

nbartos:
I think a field per-NAT entry would probably be a good approach, provided that it's possible to set a default per type of environment (OpenStack, k8s).

tomdee:
Closing this stale PR. It can be reopened if anyone plans to do any work on it?

PR #1496, opened 2017-06-30
k8sfv: use flags instead of positional args

neiljerram:


k8sfv/run-test
```diff
@@ -183,21 +183,21 @@ felix_ip=$(get_container_ip ${prefix}-felix)
...
+	   "${run} -ginkgo.focus \"${GINKGO_FOCUS}\""
 elif ${JUST_A_MINUTE}; then
     # Run all the tests that are _not_ marked as "[slow]", and fail
     # the test if any of those takes longer than 2 minutes.  A k8sfv
     # test should be marked as "[slow]" if it normally takes longer
     # than 1 minute - so this allows a buffer of 1 extra minute to
     # allow for executor slowness.
     docker exec ${prefix}-felix /bin/sh -c \
-	   "${run} -ginkgo.skip \"\\[slow\\]\" -ginkgo.slowSpecThreshold 120 ${k8s_api_endpoint} ${felix_ip} ${felix_hostname} \"${PROMPG_URL}\" \"${CODE_LEVEL}\"" | perl -pe 'END { exit $status } $status=1 if /SLOW TEST|FAIL!/;'
+	   "${run} -ginkgo.skip \"\\[slow\\]\" -ginkgo.slowSpecThreshold 120" | perl -pe 'END { exit $status } $status=1 if /SLOW TEST|FAIL!/;'
```
fasaxc: Why doesn't this need the args?

neiljerram: They are now in the expansion of `${run}` - set above in a more common place.

PR #1498, opened 2017-07-07
Do conntrack deletions in the background.

fasaxc:
Block on pending deletions if IP is reused.  Improves the slow performance observed in https://github.com/projectcalico/felix/issues/1501

## Todos
- [x] Redo manual tests to verify improvement. (Completed by @frnkdny and @doublek)

### DIFF
--- a/routetable/route_table.go
+++ b/routetable/route_table.go
@@ -357,6 +357,10 @@ func (r *RouteTable) syncRoutesForLink(ifaceName string) error {
 
 // startConntrackDeletion starts the deletion of conntrack entries for the given CIDR in the background.  Pending
 // deletions are tracked in the pendingConntrackCleanups map so we can block waiting for them later.
+//
+// It's important to do the conntrack deletions in the background because scanning the conntrack
+// table is very slow if there are a lot of entries.  Previously, we did the deletion synchronously
+// but that led to lengthy Apply() calls on the critical path.
 func (r *RouteTable) startConntrackDeletion(ipAddr ip.Addr) {
 	log.WithField("ip", ipAddr).Debug("Starting goroutine to delete conntrack entries")
 	done := make(chan struct{})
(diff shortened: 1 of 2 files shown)


fasaxc:
@neiljerram, @doublek tested this and found it to be a strict improvement over what we have now in his high-connection-load testing.  Please can you review?

routetable/route_table.go
```diff
@@ -347,6 +355,45 @@ func (r *RouteTable) syncRoutesForLink(ifaceName string) error {
 	return nil
 }
 
+// startConntrackDeletion starts the deletion of conntrack entries for the given CIDR in the background.  Pending
+// deletions are tracked in the pendingConntrackCleanups map so we can block waiting for them later.
```
neiljerram: I think we need a bit more context somewhere to record _why_ we're making this change to move conntrack deletions into the background.  I'd be happy with any of:
- adding to the comment here
- adding to the commit message
- adding more to the PR comments - either inline explanation or a link to an issue where it's explained.

At the moment - unless I missed it - I don't think we have any of those.

routetable/route_table_test.go
```diff
@@ -107,6 +107,43 @@ var _ = Describe("RouteTable", func() {
...
+				// Initial apply starts a background thread to delete
+				// 10.0.0.1 and 10.0.0.3.
+				rt.Apply()
+				// We try to add 10.0.0.10, which hasn't been seen before.
+				rt.SetRoutes("cali1", []Target{
+					{CIDR: ip.MustParseCIDR("10.0.0.10/32"), DestMAC: mac1},
+				})
+				start := time.Now()
+				rt.Apply()
+				Expect(time.Since(start)).To(BeNumerically("<", 10*time.Millisecond))
```
neiljerram: This looks fragile if we're running on a really slow test VM.  One option would be a factor of 10, as 0.3s would still be nearly unnoticeable.  Is there anything better we can do?

fasaxc: I couldn't think of a better way to interlock the test without quite a bit of plumbing.  Let's just up the timer.

routetable/route_table.go
fasaxc: I have added an issue and updated the comment.

fasaxc:
retest this please

PR #1500, opened 2017-07-11
Only delete conntrack entries for orig-src and reply-src.

fasaxc:
This reduces the number of (expensive) scans of the conntrack table we need to do.

## Description
This was one of the ideas that @doublek / @matthewdupre  / I came up with to reduce the time spent removing conntrack flows.   After closer inspection of the conntrack tuple, I'm convinced that we don't need to look in the other two fields when looking for flows to delete.

## Todos

Improves https://github.com/projectcalico/felix/issues/1501

### DIFF
--- a/conntrack/conntrack.go
+++ b/conntrack/conntrack.go
@@ -22,11 +22,20 @@ import (
 	log "github.com/Sirupsen/logrus"
 )
 
-var conntrackDirections = []string{
+// For TCP/UDP, each conntrack entry holds two copies of the tuple
+// (src addr, dst addr, src port, dst port).  One copy for the original direction and one copy for
+// the reply direction.  This is how the kernel handles NAT: by looking up the tuple for a packet
+// by its original tuple and mapping onto the corresponding reply direction tuple (or vice versa).
+//
+// When we delete conntrack entries by IP address, we need to specify which element of the tuple
+// to look in.  This slice holds the flags corresponding to the fields we care about.  Since we're
+// deleting entries for local workload endpoints, either the endpoint originated the traffic, or it
+// received the traffic and replied to it.  In the originating case, the "original source" will be
+// set to the endpoint's IP; in the other case, the "reply source". Hence, it's sufficient to only
+// look in those two fields.
+var deleteDirections = []string{
 	"--orig-src",
-	"--orig-dst",
 	"--reply-src",
-	"--reply-dst",
 }
 
 const numRetries = 3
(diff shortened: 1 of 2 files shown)


conntrack/conntrack.go
```diff
@@ -22,11 +22,20 @@ import (
...
+// (src addr, dst addr, src port, dst port).  One copy for the original direction and one copy for
+// the reply direction.  This is how the kernel handles NAT: by looking up the tuple for a packet
+// by its original tuple and mapping onto the corresponding reply direction tuple (or vice versa).
+//
+// When we delete conntrack entries by IP address, we need to specify which element of the tuple
+// to look in.  This slice holds the flags corresponding to the fields we care about.  Since we're
+// deleting entries for local workload endpoints, either the endpoint originated the traffic, or it
+// received the traffic and replied to it.  In the originating case, the "original source" will be
+// set to the endpoint's IP; in the other case, the "reply source". Hence, it's sufficient to only
+// look in those two fields.
```
neiljerram: This comment is very helpful, but I think it may need even a bit more in order to explain why only looking at orig-src and reply-src is watertight for our purposes.

The concerns are, I think that:
- There could be a (partial) flow with `orig-dst==<endpoint IP>` but no `reply-src`, if the endpoint chooses not to respond at all.  At least with UDP that would be quite easy to arrange; probably more tricky with TCP.
- Is there any way to get a flow with `reply-dst==<endpoint IP>!=orig-src` ?

I think the possible answers are that:
- It isn't possible to get conntrack state at all for such cases, because they fall outside the definition of what conntrack considers to be a flow; and/or
- There can be conntrack state for such cases, but it naturally expires so quickly that we don't care about Felix deleting it here.  (Hypothesis:) Conntrack state is only long-lived for complete flows with `reply-src==orig-dst` and `reply-dst==orig-src`, and it's only that long-lived state that we are concerned about deleting here.

Feel free to cut and paste if any of this is right or useful.

fasaxc: I think the conntrack tuples are produced as a result of the NAT calculation; it doesn't matter that there hasn't yet been a reply; the reply direction would still exist and have the _expected_ reply source IP in it.

neiljerram: Ah, I see now, I had completely misunderstood.  (So this exchange was very useful!)

neiljerram: To double-check my new understanding, then:

- If `orig-dst` is different from `reply-src`, that would only be because of a DNAT affecting connection _to_ the local workload, with `orig-dst` being the service or floating IP and `reply-src` being the endpoint IP.   Since we are deleting by endpoint IP, `reply-src` is the only useful lookup.

- If `reply-dst` is different from `orig-src`, that would only be because of an SNAT affecting outgoing connections _from_ the local workload, with `orig-src` being the endpoint IP - and again, because we are deleting by endpoint IP, `orig-src` is the only useful lookup.

That all seems watertight to me, so I'll approve now.

fasaxc: That's my understanding.  I can't think of any other cases.

PR #1503, opened 2017-07-13
Move logutils functionality to libcalico-go

mgleung:
## Description
Moved some of the compoenents of logutils to libcalico-go so that the functionality can be shared across repos.  Depends upon https://github.com/projectcalico/libcalico-go/pull/474 in order to pass tests and function correctly.

## Todos
- [x] Test log message format

```release-note
None required
```


bcreane:
Looks good. This would have been helpful for https://github.com/projectcalico/calico/issues/876 since I copy/pasted (and modified) felix logging code for calico_node.

glide.yaml
```diff
@@ -50,6 +50,6 @@ import:
 - package: github.com/emicklei/go-restful
   version: v1.2
 - package: github.com/containernetworking/cni
-  version: ^0.5.2
+  version: v0.5.2
```
fasaxc: Why this change?

fasaxc:
retest this please

glide.yaml
mgleung: Since v0.6.0-rc1 was released a couple weeks ago, any time that glide updates, it grabs the newest version (and does not skip rc) so this was grabbing a version that no longer has `pkg/ns` package that some of the logic depends on.  We eventually should update to reflect those changes, but for this change, it is unnecessary, so we should pin the version until we are ready to handle the newer one.

PR #1509, opened 2017-07-21
Update glide pin for libcalico-go

caseydavenport:
## Description

Pin to libcalico-go 1.5 series.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```


fasaxc:
Not 100% sure that's the right call but we can adjust easily enough later on if we change our minds.

PR #1511, opened 2017-07-24
Add static check for Typha pin.

fasaxc:
Prevent Felix and Typha from being pinned to different libcalico-go versions, which would be a problem if the datamodel changed.

## Description

If the datamodel changes to add a new field and we only update Felix, but not Typha, then Typha will send Felix updates without the new field, which would be very confusing.  Add a static check to make sure that the version of Typha that Felix is pinned against uses the same libcalico-go pin.

## Release Note

```release-note
None required
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -358,9 +358,24 @@ go-meta-linter: vendor/.up-to-date $(GENERATED_GO_FILES)
 	                                --enable=goimports \
 	                                --vendor ./...
 
+.PHONY: check-typha-pins
+check-typha-pins: vendor/.up-to-date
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -358,9 +358,24 @@ go-meta-linter: vendor/.up-to-date $(GENERATED_GO_FILES)
 	                                --enable=goimports \
 	                                --vendor ./...
 
+.PHONY: check-typha-pins
+check-typha-pins: vendor/.up-to-date
+	@echo "Checking Typha's libcalico-go pin matches ours (so that any datamodel"
+	@echo "changes are reflected in the Typha-Felix API."
```
neiljerram: close bracket

PR #1516, opened 2017-07-25
Adding support for ppc64le.

djlwilder:
The build architecture is select by setting the ARCH environment variable.
For example: When building on ppc64le you could use ARCH=ppc64le make <..>.
When ARCH is undefined the build defaults to amd64.

Based on work by: Hitomi Takahashi

## Description
Hi -
I have made progress on a ppc64le port and have changes ready for most of the Calico repos.  Let's start with felix so we can review my strategy for non-x86 builds.  This is my approach:

1) The build architecture is select by setting the ARCH environment variable.
For example: When building on ppc64le you could use ARCH=ppc64le make <....>.
When ARCH is undefined builds defaults to amd64.

2) Dockefile names are appended with -$ARCH, for x86 file-names are unchanged.  For example on ppc64le I use Dockerfile-ppc64le.

3) Generated container image names are appended with -$ARCH, for x86 image names are unchanged.  Example calico/felix-ppc64le.

Testing:  
On ppc64le I have verified make targets: clean, all and ut.
fv: is currently failing on ppc64le, I will address that shortly in another pull request.

### DIFF
--- a/docker-build-images/centos7-build.Dockerfile-ppc64le
+++ b/docker-build-images/centos7-build.Dockerfile-ppc64le
@@ -1,0 +1,22 @@
+FROM ppc64le/centos:7
+MAINTAINER Shaun Crampton <shaun@tigera.io>
+ENV STREAM el7
+
+ARG UID
+ARG GID
+
+# Some commands that would typically be run at container build time must be run in a privileged container.
+# Therefor, we do a two step image build process, in step 1 (now) scripts are placed in the image, in step 2
+# a privileged container is started and the scripts are executed.
+
+ADD install-centos-build-deps install-centos-build-deps
+
+# rpmbuild requires the current user to exist inside the container, copy in
+# some user/group entries calculated by the makefile.
+# use `--force` and `-o` since tests can run under root and command will fail with duplicate error
+
+RUN echo "#!/usr/bin/env bash" > /setup-user; \
+    echo "groupadd --force --gid=$GID user && useradd -o --home=/ --gid=$GID --uid=$UID user" >> /setup-user; \
+    chmod +x /setup-user
+
+WORKDIR /code
--- a/docker-build-images/centos7-build.Dockerfile-ppc64le
+++ b/docker-build-images/centos7-build.Dockerfile-ppc64le
@@ -6,7 +6,7 @@ ARG UID
 ARG GID
 
 # Some commands that would typically be run at container build time must be run in a privileged container.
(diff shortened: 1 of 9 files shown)


marvin-tigera:

fasaxc:
ok to test

fasaxc:
@djlwilder Looks like there's a build failure on Semaphore, please can you take a look?  The error doesn't immediately make sense to me.

docker-build-images/centos7-build.Dockerfile-ppc64le
```diff
@@ -0,0 +1,22 @@
+FROM ppc64le/centos:7
+MAINTAINER Shaun Crampton <shaun@tigera.io>
+ENV STREAM el7
+
+ARG UID
+ARG GID
+
+# Some commands that would typically be run at container build time must be run in a privileged container.
```
fasaxc: Can you explain this a bit more?  I've never had privilege issues at build time before, don't builds run as root (I thought that was the point!)?

I wonder if now would be a good time to switch to the `entrypoint` approach used in https://github.com/projectcalico/go-build since that doesn't require a rebuild of the container to add the user.

djlwilder: I was surprised by this error as well.  I found others complaining about the same issue.  Yes "docker build" runs as root,  however the root user's capabilities may be limited by docker when it starts a container, or in this case the build process.  If my Dockerfile has:
  RUN groupadd --force --gid=1000 user && useradd -o --home=/ --gid=1000 --uid=1000 user

The docker build reports this error, as the process is missing some required capability.
>groupadd: /etc/group.7: Operation not permitted

Unfortunately there is no way to increase the privileged level of a docker build.I suspect this is a bug in docker,  I don't know why the problem is not seen on x86. My workaround should continue to work even if the problem is solved in the future. I will try using the entrypoint approach as you suggested.

djlwilder: Unfortunately the entrypoint solution wont solve my problem. I hit the same permissions issue running yum in a non-privileged container (or from a docker build). To make the entrypoint solution work I need to install sudo during the container build.
$ docker run ppc64le/centos:7 yum
error: db5 error(1) from dbenv->open: Operation not permitted
---------
this works ok.
$ docker run --privileged ppc64le/centos:7 yum

I have not resolved the root of this problem, it only occurs in a ubuntu ppc64le VM and only with the centos container.  For now I need to stick with my workaround.

PR #1517, opened 2017-07-25
Do not depend on the returned text string of iptables -w.

djlwilder:
The iptables_lock_test is looking for a specific text string to be
returned by iptables. Unfortunately the output string varies depending
on version. Checking the returned status is sufficient for this test.
This problem was seen on a ppc64le. I verified the change with "make
fv" on both x86 and ppc64le.

### DIFF
--- a/fv/iptables_lock_test.go
+++ b/fv/iptables_lock_test.go
@@ -115,7 +115,7 @@ var _ = Describe("with running container", func() {
 		It("iptables should succeed in getting the lock after 3s", func() {
 			iptCmd := cmdInContainer("iptables", "-w", "3", "-A", "FORWARD")
 			out, err := iptCmd.CombinedOutput()
-			Expect(string(out)).To(ContainSubstring("Another app is currently holding the xtables lock"))
+			log.Printf("iptables output='%s'", out)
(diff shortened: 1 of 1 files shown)


marvin-tigera:

fasaxc:
ok to test

fv/iptables_lock_test.go
```diff
@@ -115,7 +115,7 @@ var _ = Describe("with running container", func() {
 		It("iptables should succeed in getting the lock after 3s", func() {
 			iptCmd := cmdInContainer("iptables", "-w", "3", "-A", "FORWARD")
 			out, err := iptCmd.CombinedOutput()
-			Expect(string(out)).To(ContainSubstring("Another app is currently holding the xtables lock"))
+			log.Printf("iptables output='%s'", out)
```
fasaxc: Since we're using Logrus, probably best to use `Infof` to avoid confusion (Logrus' Printf behaves differently to the built in one)

djlwilder:
I ran the go tests for conntrack, ipsets and iptables on Ubuntu 16.04.2 LTS ppc64le, all three passed.
I reviewed the code for all three. They all have some dependency on returned text strings,  but I cant see how that can be avoided.

PR #1521, opened 2017-08-03
Add watching ippool

fkakuma:
In the case of using gobgp as network backend, confd is only
used to execute allocate_ipip_addr command to sync ippool data.
So we can simplify calico-node if felix has process to watch ippool.
Then we will not need to run confd in the case of using gobgp.

This patch add the following functions.
- Add NetworkingBackend in config parameter.
- Watching ippool directory by thread, if any events occurs,
  it executes allocate_ipip_addr.

Signed-off-by: Fumihiko Kakuma <kakuma@valinux.co.jp>

## Description
- This patch aims to simpify calico-node by adding a some feature of confd to felix.
- After this, I am going to make a patch preparing an environment to use this feature in calico/calico_node.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/config/config_params.go
+++ b/config/config_params.go
@@ -95,6 +95,8 @@ type Config struct {
 
 	FelixHostname string `config:"hostname;;local,non-zero"`
 
+	NetworkingBackend string `config:"oneof(bird,gobgp,none);bird"`
+
 	EtcdAddr      string   `config:"authority;127.0.0.1:2379;local"`
 	EtcdScheme    string   `config:"oneof(http,https);http;local"`
 	EtcdKeyFile   string   `config:"file(must-exist);;local"`
--- a/felix.go
+++ b/felix.go
@@ -46,6 +46,7 @@ import (
 	_ "github.com/projectcalico/felix/config"
 	"github.com/projectcalico/felix/extdataplane"
 	"github.com/projectcalico/felix/intdataplane"
+	"github.com/projectcalico/felix/ippool"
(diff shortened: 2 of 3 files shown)


marvin-tigera:

fasaxc:
@fkakuma Thanks for your contribution.  Unfortunately, I don't think I can take it under the current design for a few reasons:

- It bypasses our libcalico-go abstraction layer and duplicates a bunch of function in that layer
- It calls out to an external binary to do the IP allocation
- In general, for security reasons, we try to avoid writing to the datastore from Felix so that Felix can be run a a read-only API user.  This last point would be acceptable as long as the feature is disabled by default.

Maybe we could discuss on slack to find a way forward?  We have a `#calico-dev` channel on [our slack](https://slack.projectcalico.org/)

fkakuma:
@fasaxc thank you for your review and sorry for my late reply.(I was on vacation.)
I thought the following along your comments and also put them in the slack.

- add watch method to backend datastore interface in libcalico-go.
- add an environment variable to allow felix to watch the ip pool. (for example FELIX_WATCH_IPPOOL)
- add watching ippool and an equivalent feature based on libcalico-go to allocate_ipip_addr command to felix.
  If FELIX_WATCH_IPPOOL is specified, they will be executed.

fkakuma:
I close this pull-req from discussion on #calico-dev once.

PR #1524, opened 2017-08-04
Deb/RPM packaging for Calico 2.4.0

neiljerram:


debian/changelog
```diff
@@ -1,3 +1,112 @@
...
+    - Improve testing and test coverage (#1486, #1494, #1496, #1497).
+    - Make test suites produce junit reports (#1488).
+    - Allow selection of policy allow action (#1492).
+    - Implement liveness and readiness endpoints for Felix (#1489).
+    - Improve Calico version reporting (#1499).
+    - Streamline conntrack state deletions (#1500, #1498).
+    - Add release note to PR template (#1502).
+    - Add support for multiple CIDRs in a match rule (#1483, #1505).
+    - Support using a lock to coordinate iptables programming with other
+    - software (#1491, #1504).
```
lwr20: spurious bullet?

debian/changelog
```diff
@@ -1,3 +1,112 @@
...
+    - Move logutils functionality to libcalico-go (#1503).
+    - Add pre-DNAT policy support (#1506).
+    - Update glide pin for logrus (#1509).
+    - Allow for time fuzziness in route table UT (#1510).
+    - Update to Typha v0.3.0 (#1512).
+    - Only report ready after first apply() completes.
```
lwr20: These have no PR numbers and are duplicated in the rc2 section below?  Duplication might be expected I guess.

debian/changelog
```diff
@@ -1,3 +1,112 @@
...
+    The headline feature in this release is a significant increase in scale when using the
+    Kubernetes datastore driver by introducing support for a new daemon, Typha. Typha
+    connects to the Kubernetes API server and fans out updates to a number of Felixes.
+    By having only a handful of Typha instances connected to the API server instead of
+    many Felixes, we place a lot less load on the API server.  In addition, Typha is able
+    to squash unimportant updates form the API server, significantly reducing the
+    number of mesages each Felix has to handle.
+
+    Typha is disabled by default. The documentation for how to configure Typha and Felix
+    together will follow as part of the wider Calico 2.3.0 meta-release.
```
lwr20: Did it? Might want to revise that time estimate - maybe the docs will appear in 2.4?

debian/changelog
```diff
@@ -1,3 +1,112 @@
...
+    - mechanism for only running quick tests
+    - fix client authorization to API server 1.6
+    - Ensure that ip6_tables module is loaded
+    - put Typha in the loop.
+
+    The DockerHub and Quay.io `calico/felix` images have been updated.  The `calico/node` image
+    based on this release will follow shortly.
+
+    Since we're targetting Kubernetes for this release, we haven't yet updated the OpenStack
+    debs and RPMs, those should follow in a few days.
```
lwr20: Do we still want this comment?

debian/changelog
```diff
@@ -1,3 +1,112 @@
...
+
+    - Performance: Scan iptables-save output incrementally when calculating hashes.
+    - Performance: Disable WithFields logging in the ipsets resync parse loop.
+    - Performance: Disable some WithField calls in the inner iptables resync loop.
+    - Allow the iptables post-write check interval to be set.
+    - Improve diagnostics around ipset restore failures
+    - Log ip6tables-save stderr if it fails
+    - Convert interval config parameters to time.Duration for increased precision.
+    - Rev libcalico-go to v1.4.4 and Typha to v0.2.2.
+    - Make JUST_A_MINUTE the default, as it's more useful for devs
```
lwr20: This is a k8sfv feature belonging in the section below.  

This implies that this whole section is based on the flawed 2.3 tag text - probably easiest to replace it with the edited github version.

rpm/felix.spec
```diff
@@ -2,7 +2,7 @@
 
 Name:           felix
 Summary:        Project Calico virtual networking for cloud data centers
-Version:        2.2.1
+Version:        2.4.0
```
lwr20: Erk.  The fact this jumped from 2.2.1 to 2.4.0 - did we miss a change for 2.3.0?

rpm/felix.spec
```diff
@@ -152,6 +152,109 @@ rm -rf $RPM_BUILD_ROOT
...
+    - Add release note to PR template (#1502).
+    - Add support for multiple CIDRs in a match rule (#1483, #1505).
+    - Support using a lock to coordinate iptables programming with other
+    - software (#1491, #1504).
+    - Move logutils functionality to libcalico-go (#1503).
+    - Add pre-DNAT policy support (#1506).
+    - Update glide pin for logrus (#1509).
+    - Allow for time fuzziness in route table UT (#1510).
+    - Update to Typha v0.3.0 (#1512).
+    - Only report ready after first apply() completes.
```
lwr20: Duplicated.  Actually - its clear this is cut-and-paste from the other text file above - presumably same comments apply to both.

debian/changelog
neiljerram: I'll try to find and add the PR numbers.
Think it also makes sense - given that I'm manually editing this - to remove the duplication, i.e. the 2.4.0-rc1 and 2.4.0-rc2 sections - agree?

debian/changelog
neiljerram: This is just a repeat of the 2.3 changes that we've already made.  I don't think we should change those now.

debian/changelog
neiljerram: 2.3 again, as you say.  I think it would be confusing if this was any different from the 2.3 notes that we've already merged into 2.3.x-series and master.

debian/changelog
neiljerram: 2.3 again

rpm/felix.spec
neiljerram: No, I did that yesterday.  This is just because you're looking at a combined change.

PR #1526, opened 2017-08-09
Add a configurable timeout to netlink operations.

### LINKED ISSUE #1522: Felix sometimes stops updating iptables.
Sometimes felix stops updating iptables, breaking networking for new deployed containers.
This happens every month or so on our beta cluster. 

## Expected Behavior

Felix shouldn't stop updating the iptables.

## Current Behavior

When the bug happens, iptables stops getting updated with new policies or newly started containers.

When starting a new container, the route to it gets created:

    root@nodes-kr1m:~$ ip route | grep 100.92.223.207
    100.92.223.207 dev cali59ef8132c6c  scope link 

but if you search the interface name in iptablres, it's not there:

    root@nodes-kr1m:~# iptables-save | grep cali59ef8132c6c
    # shows no output.

Here are the calico-node container logs. Note that around `14:03` all logs about iptables and ipset stop happening, which is the time calico stops working.
https://gist.github.com/Dirbaio/7d529d7a4e60953187d1ca5d7d3484bc

There doesn't seem to be any bad error message in the logs.

## Possible Solution

As a workaround, restarting calico-node container fixes it.

1. Deploy calico
2. Wait
3. The bug may happen after months of having everyting up.

## Your Environment

* And/Or, `calico/node` container version (if running Felix in a container): quay.io/calico/node:v1.3.0
* Which orchestrator are you using (e.g. Kubernetes, OpenStack, Docker, Mesos): Kubernetes
* Orchestrator version (if used): 1.5.2
* Etcd version (if using etcd datastore driver): gcr.io/google_containers/etcd:2.2.1
* Operating System and version: Ubuntu 16.04

fasaxc:
## Description
Fixes that Felix could wait forever if a netlink response is dropped by the kernel (#1522).  (Since the netlink protocol doesn't have guaranteed delivery, this is possible at high load.)

Switch to using the Handle API, which exposes a timeout setting on the socket.  Default the timeout to 10s but allow it to be configured.

Since the Handle is stateful and caches a socket, which is not something we were doing before, defensively reconnect after any errors.

Fixes #1522 

## Todos
- [x] Refactor to mock out Handle instead
  - [x] Test connection failure
  - [x] Test timeout set failure
  - [x] Test persistent failure causes panic
- [x] Integration tests (already exist)
- [x] Documentation: https://github.com/projectcalico/calico/pull/1088
- [ ] Backport
- [x] Release note

## Release Note

```release-note
Felix now imposes a (configurable) timeout on netlink operations. This prevents Felix from hanging if the kernel fails to respond to a netlink request.  It may also improve performance of netlink operations by re-using the netlink socket.
```


fasaxc:
retest this please

fasaxc:
retest this please

neiljerram:
Hmm, I just realized that the Todos section of the PR template is ambiguous: is it (1):
- unchecked -> I don't think this item is needed for this PR
- checked -> this item is needed for this PR, and may or may not have been done yet

or (2):
- line deleted -> that item wasn't needed for this PR
- unchecked -> needed for this PR, but I haven't done it yet
- checked -> needed for this PR, and I've done it.

I have been using interpretation (2), but wonder (here) if you are using interpretation (1) ?

fasaxc:
I'm using checked = done, unchecked = no yet done/not yet considered, deleted = n/a.

neiljerram:
Ah, OK, that's the same as me then.

routetable/testing_shims.go
```diff
@@ -17,60 +17,130 @@ package routetable
...
 	return cmd.Run()
 }
 
-func (r realDataplane) RemoveConntrackFlows(ipVersion uint8, ipAddr net.IP) {
+func (r *realDataplane) RemoveConntrackFlows(ipVersion uint8, ipAddr net.IP) {
 	r.conntrack.RemoveConntrackFlows(ipVersion, ipAddr)
 }
 
-var _ dataplaneIface = realDataplane{}
+var _ dataplaneIface = &realDataplane{}
```
neiljerram: What is this for?  (Although I appreciate that you haven't really changed it.)

routetable/testing_shims.go
```diff
@@ -17,60 +17,130 @@ package routetable
 import (
```
neiljerram: I think it's a bit confusing for this file to be named so that it looks like it's part of our UT, when in fact - IIUC - it is very much product code.

Is this also related to your query about how you can test UT code?  If it isn't actually UT code, presumably that problem is no longer there...?

routetable/testing_shims.go
fasaxc: It's a compile-time assertion that a `*realDataplane` should implement that interface.  In this case, we'd likely get a compile-time error anyway due to the ay we're using the struct but this makes it explicit/local.

routetable/testing_shims.go
fasaxc: Agreed `dataplane_shims.go`?  I'm now thinking we should shim the Handle and move the reconnect logic into the wrapper code.

routetable/testing_shims.go
```diff
@@ -17,60 +17,130 @@ package routetable
...
+	h, err := netlink.NewHandle(syscall.NETLINK_ROUTE)
+	if err != nil {
+		logrus.WithError(err).Error("Failed to (re)connect to netlink")
+		return nil, err
+	}
+	if r.nlTimeout > 0 {
+		err := h.SetSocketTimeout(r.nlTimeout)
+		if err != nil {
+			logrus.WithError(err).Error("Failed to set netlink socket timeout")
+			return nil, err
```
fasaxc: Looks like I missed that, good spot.

routetable/testing_shims.go
neiljerram: Ah, neat.

fasaxc:
Reworked to mock out the handle.  I think that's a lot cleaner now.

fasaxc:
I think the concern what motivated us to switch to the pkg handle was that it is stateless.  Using a dedicated handle caches the socket(s) so you get better performance, but could potentially hit an issue where the socket fails persistently.  I've mitigated that in the current code by closing the handle and recreating it after any sort of failure.

routetable/route_table.go
```diff
@@ -161,13 +191,59 @@ func (r *RouteTable) QueueResync() {
...
+			r.numConsistentNetlinkFailures++
+			log.WithError(err).WithField("numFailures", r.numConsistentNetlinkFailures).Error(
+				"Failed to connect to netlink")
+			return nil, err
+		}
+		err = nlHandle.SetSocketTimeout(r.netlinkTimeout)
+		if err != nil {
+			r.numConsistentNetlinkFailures++
+			log.WithError(err).WithField("numFailures", r.numConsistentNetlinkFailures).Error(
+				"Failed to connect to netlink")
```
neiljerram: "Failed to set timeout on netlink handle" ?

routetable/route_table.go
```diff
@@ -390,14 +473,14 @@ func (r *RouteTable) syncRoutesForLink(ifaceName string) error {
...
-			if err := r.dataplane.RouteAdd(&route); err != nil {
+			if err := nl.RouteAdd(&route); err != nil {
 				logCxt.WithError(err).Warn("Failed to add route")
 				updatesFailed = true
 			}
 		}
 		if r.ipVersion == 4 && target.DestMAC != nil {
 			// TODO(smc) clean up/sync old ARP entries
-			err := r.dataplane.AddStaticArpEntry(cidr, target.DestMAC, ifaceName)
+			err := r.addStaticArpEntry(cidr, target.DestMAC, ifaceName)
```
neiljerram: If this member function was still capitalized, it would be a bit harder for someone to accidentally (and wrongly) change this to ```err := addStaticArpEntry(cidr, target.DestMAC, ifaceName)```  Just a thought.

(And I believe that that would still compile, because the non-test impl `addStaticArpEntry` is also in the routetable package.)

fasaxc: I'm not keen on making the field public again.  I tried a few spelling variants but didn't find any I liked.  If you do sub in the wrong one then the UT should fail with a permissions error so I think we should live with.

PR #1528, opened 2017-08-10
Allow Policy to explicitly govern ingress and/or egress

### LINKED ISSUE #1180: Can't create a policy that only affects egress, or only ingress
Not sure if this is a Felix issue or a libcalico-go issue yet, but I can't seem to figure out how to create a policy which only affects egress.

I've created the following:

```
- apiVersion: v1
  kind: policy
  metadata:
    name: advanced-policy-demo.deny-egress
  spec:
    egress:
    - action: deny
      destination:
        selector: calico/k8s_ns != 'advanced-policy-demo'
      source:
        selector: calico/k8s_ns == 'advanced-policy-demo'
    order: 500
    selector: calico/k8s_ns == 'advanced-policy-demo'
```

It shows up in etcd like so:

```
{"order":500,"outbound_rules":[{"action":"deny","src_selector":"calico/k8s_ns == 'advanced-policy-demo'","dst_selector":"calico/k8s_ns != 'advanced-policy-demo'"}],"selector":"calico/k8s_ns == 'advanced-policy-demo'"}
```

The above policy is dropping ingress traffic to the selected workloadendpoints, even though there are other policies (at a lower order) which allow that traffic.

I would have thought that since no ingress rules are specified, then ingress would not be affected.

CC @robbrockbank @fasaxc

neiljerram:
Fixes #1180

## Description
This is the Felix part of an enhancement to allow a Calico Policy to specify explicitly whether it applies on ingress or on egress, or both.

## Todos
- [x] Tests
- [x] Documentation - in https://github.com/projectcalico/calico/pull/1016
- [x] Release note - in https://github.com/projectcalico/calico/pull/1016


calc/policy_sorter.go
```diff
@@ -91,6 +96,35 @@ func (p PolKV) String() string {
...
+			return true
+		}
+	}
+	return false
+}
+
+func (p PolKV) GovernsIngress() bool {
+	if p.Ingress == nil {
+		governsIngress := p.governsType("ingress")
+		p.Ingress = &governsIngress
```
fasaxc: This method has value type so this write will be lost

neiljerram: Excellent point.

As I was going to sleep last night, I was thinking I'd change it to (p *PolKV) for efficiency, but I didn't spot that it was actually bugged as is.

neiljerram:
Note: test failure was only in a flaky OpenStack VM-rebuild test, so not a concern.

neiljerram:
@fasaxc - Markups and UT all now done on this PR, so back with you for re-review.

neiljerram:
Note: shouldn't actually merge yet.  Need https://github.com/projectcalico/libcalico-go/pull/494 to merge first, and then I (or someone) will have to update the glide files again to point at libcalico-go master.

Would we also need a Typha update at that point?

(CC @bcreane for awareness of the overall position with this work)

calc/policy_sorter.go
```diff
@@ -48,6 +50,21 @@ func (poc *PolicySorter) OnUpdate(update api.Update) (dirty bool) {
 				oldPolicy.DoNotTrack != newPolicy.DoNotTrack ||
 				oldPolicy.PreDNAT != newPolicy.PreDNAT {
 				dirty = true
+			} else {
+				var oldTypes, newTypes set.Set
+				if oldPolicy.Types == nil {
+					oldTypes = set.New()
```
neiljerram: In fact for the nil cases we can avoid using set.Set at all.

calc/policy_sorter.go
```diff
@@ -48,6 +50,21 @@ func (poc *PolicySorter) OnUpdate(update api.Update) (dirty bool) {
...
+				} else {
+					oldTypes = set.FromArray(oldPolicy.Types)
+				}
+				if newPolicy.Types == nil {
+					newTypes = set.New()
```
neiljerram: Well it doesn't matter here because we're just comparing for equality.

calc/policy_sorter.go
```diff
@@ -91,6 +114,35 @@ func (p PolKV) String() string {
...
+		if strings.EqualFold(t, wanted) {
+			return true
+		}
+	}
+	return false
+}
+
+func (p *PolKV) GovernsIngress() bool {
+	if p.ingress == nil {
+		governsIngress := p.governsType("ingress")
```
neiljerram: Those are only for api.Policy.

PR #1529, opened 2017-08-16
Update to libcalico v1.6.0 and typha v0.4.0.

bcreane:
## Description
Update to libcalico v1.6.0 and typha v0.4.0.

## Release Note

```release-note
None required
```

### DIFF
--- a/k8sfv/run-test
+++ b/k8sfv/run-test
@@ -10,7 +10,7 @@
 : ${FELIX_VERSION:=latest}
 #
 # What version of the k8s API server to use.
-: ${K8S_VERSION:=1.6.4}
+: ${K8S_VERSION:=1.7.3}
 #
 # A string to insert into the container names that we use; a calling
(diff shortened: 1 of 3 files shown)


k8sfv/run-test
```diff
@@ -58,6 +58,9 @@
 #
 # Typha logging level.
 : ${TYPHA_LOG_LEVEL:=info}
+#
+# CRD Manifest to use
+: ${CRD_URL:=https://raw.githubusercontent.com/projectcalico/libcalico-go/master/test/crds.yaml}
```
fasaxc: This doesn't look right to me.  I don't think we should introduce a dependency on libcalico-go master for the FV tests when we have libcalico-go pinned in our vendor directory.  Why not get this from the file system at `./vendor/github.com/projectcalico/libcalico-go/test/crds.yaml`?

bcreane: good idea, will take a look soon.

PR #1531, opened 2017-08-16
Reintegrate release branches

fasaxc:
## Description
I noticed that we'd never merged some of our release branches back to master to record changelogs etc.  This PR does that merge.

## Release Note

```release-note
None required
```


fasaxc:
1.4.x-series has diverged heavily so I left that out.

Issue #1532, opened 2017-08-16
Update default MTU for Ip-in-Ip as 1430


gyliu513:
Based on https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux_openstack_platform/7/html/networking_guide/sec-mtu, in Openstack with vxlan environment, the default MTU is 1450 which is even less than GCE 1460, so the default MTU for the IP-in-IP should be update to 1430.

The proposal is adding a new line for `OpenStack VXLAN` and set the default MTU for Ip-in-Ip as 1430.

tmjd:
@gyliu513 In general, I don't believe it is necessary to use IP-in-IP when running on OpenStack.  Have you attempted to disable source/destination IP checks in OpenStack and then turn off IP-in-IP?  If you are able to do that then you don't need to use the tunnel which this setting is for.

gyliu513:
@tmjd can you please show more detail for how to disable the source/destination IP checks in OpenStack?

tmjd:
@kprabhak Could you help with this? I think the source/destination IP check needs disabling, but not 100% sure about that.

kprabhak:
In OpenStack, this would be by disabling Port Security on the port to the instance, as described here:
https://wiki.openstack.org/wiki/Neutron/ML2PortSecurityExtensionDriver

Your OpenStack deployment would need to have the Port Security ML2 extension enabled. Also Port security is currently mutually exclusive with security groups, so you wouldn't be using a security group for the VM (and therefore you would want to ensure that you have tight Calico policy protection for your K8s pods (and potentially also for the host instance within your VM).

An alternate approach of course is to just use IP-in-IP mode as you've discussed above.

gyliu513:
Thanks, but the problem is that we cannot guarantee all of the OpenStack cluster have such configurations, so updating the MTU to 1430 would be a good option. @kprabhak

caseydavenport:
Closing - I don't think we can reasonably support a default value for all MTUs, and we have another issue open for auto-detection support which I believe is the correct path forward. For now, we have documentation on how to configure MTU as needed based on environment.

PR #1533, opened 2017-08-16
Updated default IpInIpMtu to 1430.

gyliu513:
## Description
A few sentences describing the overall goals of the pull request's commits.
Please include
- the type of fix - (e.g. bug fix, new feature, documentation)
documentation & bug fix

- some details on _why_ this PR should be merged
Based on https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux_openstack_platform/7/html/networking_guide/sec-mtu, in Openstack with vxlan environment, the default MTU is 1450 which is even less than GCE 1460, so the default MTU for the IP-in-IP should be update to 1430.

The proposal is adding a new line for OpenStack VXLAN and set the default MTU for Ip-in-Ip as 1430.

- the details of the testing you've done on it (both manual and automated)
- which components are affected by this PR

## Todos
- [ ] Documentation
- [ ] Backport
- [X] Release note

## Release Note

```release-note
Updated default IpInIpMtu to 1430.
```

Fixed https://github.com/projectcalico/felix/issues/1532

### DIFF
--- a/config/config_params.go
+++ b/config/config_params.go
@@ -136,7 +136,7 @@ type Config struct {
 	LogSeveritySys    string `config:"oneof(DEBUG,INFO,WARNING,ERROR,CRITICAL);INFO"`
 
 	IpInIpEnabled    bool   `config:"bool;false"`
-	IpInIpMtu        int    `config:"int;1440;non-zero"`
+	IpInIpMtu        int    `config:"int;1430;non-zero"`
 	IpInIpTunnelAddr net.IP `config:"ipv4;"`
 
 	ReportingIntervalSecs time.Duration `config:"seconds;30"`


fasaxc:
ok to test

fasaxc:
@gyliu513 Please can you fill out the PR comment to explain the change?

gyliu513:
@fasaxc done

fasaxc:
Hmm, not sure we can change the default now without doing work in the CNI plugin.  What if the current value works in your cluster but you have CNI workloads already created with the large MTU and then we reduce the MTU on the tunnel device; now the CNI workload's MTU will be too big to fit in the tunnel.

gyliu513:
@fasaxc I think that we cannot handle the case if the end user update the MTU manually, here I want to make sure we can handle default value well, based on investigation, seems OpenStack VXLAN should be the smallest one. Comments?

fasaxc:
@gyliu513 I'm worried about the upgrade case for existing users.  If they haven't explicitly set the value then, when they upgrade, we'll change it to the new default.  That could break their systems by suddenly making the MTU too small to handle packets that it was handling before.

gyliu513:
Yes, but default 1440 will cause Kubernetes on OpenStack does not work https://developer.ibm.com/recipes/tutorials/troubleshooting-network-issues-for-ibm-cloud-private-icp-on-openstack/ , so I still prefer update to 1430 to make make Kubernetes on OpenStack works.

fasaxc:
@gyliu513 I agree that 1430 would be a better default if we were starting over but I don't want to break existing users.  To avoid that, I think we'd need a more advanced patch that uses the new default for new systems but keeps the old default for upgraded systems.  Do you agree?

gyliu513:
+1 for this, any comments for how we can proceed? BTW: I was changing master, and I think this will  not impact old systems? @fasaxc

fasaxc:
@gyliu513 The problem is if someone upgrades from old version to new.  The old version defaulted to the larger value, the new one defaults to the smaller one.  That could cause fragmentation or packet loss if the old value was correct for their deployment.

tomdee:
@fasaxc Any thoughts on whether we can merge this or whether we should just close it?

tomdee:
Unfortunately, I don't think we can merge this without more work on the upgrade path, so closing for now.

PR #1536, opened 2017-08-22
Record 2.5.0 deb/RPM changelogs.

fasaxc:
## Description
Re-integrate deb/RPM changelogs for 2.5.0 release.

## Release Note

```release-note
None required
```


fasaxc:
@bcreane Did you mean jenkins?  Looks like the test passed but then the script failed to tear down the GCE VMs it was using.  Good enough for me!

PR #1538, opened 2017-08-29
Add read/write timeout options for Typha connection.

### LINKED ISSUE #1537: Felix stops updating the dataplane is Typha connection is severed without a FIN/RST
If the Typha connection is severed, for example due to a routing blackhole the Felix doesn't spot the problem and waits for an indefinite period.

More details in https://github.com/projectcalico/typha/issues/38

fasaxc:
## Description
Fixes #1537

## Todos
- [x] Move pin back to projectcalico/typha repo once fix is merged to client.
- [x] Release note
- [ ] Docs

## Release Note

```release-note
Fix that, in some cases, Felix wouldn't spot if it got disconnected from Typha (#1537)
```

### DIFF
--- a/config/config_params.go
+++ b/config/config_params.go
@@ -102,9 +102,11 @@ type Config struct {
 	EtcdCaFile    string   `config:"file(must-exist);;local"`
 	EtcdEndpoints []string `config:"endpoint-list;;local"`
 
-	TyphaAddr           string `config:"authority;;"`
-	TyphaK8sServiceName string `config:"string;"`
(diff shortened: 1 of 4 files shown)


fasaxc:
The read timeout is set to 3x the default Typha ping interval; I figure missing 3 pings (with no other traffic) is a pretty bad sign.  

I don't have much to back up the choice of write timeout, except that it matches the existing connection timeout.  I really don't expect writes to take very long at all since we only write a packet every 10s to respond to pings.

PR #1543, opened 2017-08-31
Replace log.Fatal with log.Panic in most cases.

fasaxc:
Fatal() results in a silent process exit when running under Ginkgo, which
is very annoying to diagnose.

Ginkgo is able to catch panics and logs them sensibly with context.

I didn't remove _all_ uses of Fatal.  The ones in felix.go are appropriate, I think.  I was worried about swapping out the ones in logutils in case there was some unfortunate cycle created.

## Release Note

```release-note
None required
```

### DIFF
--- a/calc/active_rules_calculator.go
+++ b/calc/active_rules_calculator.go
@@ -169,7 +169,7 @@ func (arc *ActiveRulesCalculator) OnUpdate(update api.Update) (_ bool) {
 			// longer matches.
 			sel, err := selector.Parse(policy.Selector)
 			if err != nil {
-				log.Fatal(err)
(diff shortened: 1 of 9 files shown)


fasaxc:
@bcreane I think the logrus version calls os.Exit(), which can't be caught :-(

fasaxc:
I doubt we'd try to recover from any old panic.  I think if we introduced recover() into the codebase, we'd want to be passing a specific error type to panic so that we can tell the difference between an expected and unexpected panic.  These changes would naturally fall under the unexpected category, which we wouldn't try to recover.

PR #1546, opened 2017-09-12
FV test for ingress/egress policy

neiljerram:
## Description
Felix FV test for https://github.com/projectcalico/calico/pull/1016

### DIFF
--- a/fv/containers/containers.go
+++ b/fv/containers/containers.go
@@ -24,6 +24,7 @@ import (
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	"github.com/projectcalico/felix/fv/utils"
+	"github.com/projectcalico/libcalico-go/lib/set"
 	log "github.com/sirupsen/logrus"
 )
 
@@ -33,6 +34,7 @@ type Container struct {
 	Hostname string
 	Stop     func()
 	stopped  bool
+	binaries set.Set
 }
 
 var containerIdx = 0
@@ -65,11 +67,12 @@ func Run(namePrefix string, args ...string) (c *Container) {
 			c.stopped = true
 			runCmd.Process.Signal(os.Interrupt)
 			c.WaitNotRunning(10 * time.Second)
+
+			// And now to be really sure that the container is cleaned up.
+			utils.RunMayFail("docker", "rm", "-f", c.Name)
 		}
-		// And now to be really sure that the container is cleaned up - and regardless of
-		// whether we already tried before...
-		utils.RunMayFail("docker", "rm", "-f", c.Name)
 	}
+	c.binaries = set.New()
 	log.WithField("container", c).Info("Container now running")
 	return
 }
@@ -132,3 +135,10 @@ var _ = AfterEach(func() {
 	}
 	runningContainers = []*Container{}
 })
+
+func (c *Container) EnsureBinary(name string) {
+	if !c.binaries.Contains(name) {
+		exec.Command("docker", "cp", "../bin/"+name, c.Name+":/"+name).Run()
+		c.binaries.Add(name)
+	}
+}
--- a/fv/containers/containers.go
+++ b/fv/containers/containers.go
@@ -53,7 +53,10 @@ func Run(namePrefix string, args ...string) (c *Container) {
 	runCmd := exec.Command("docker", runArgs...)
 	err := runCmd.Start()
 	Expect(err).NotTo(HaveOccurred())
-	c.WaitRunning(10 * time.Second)
+
+	// It might take a very long time for the container to show as running, if the image needs
+	// to be downloaded - e.g. when running on semaphore.
+	c.WaitRunning(20 * 60 * time.Second)
 
 	// Remember that this container is now running.
 	runningContainers = append(runningContainers, c)
--- a/fv/containers/containers.go
+++ b/fv/containers/containers.go
@@ -32,8 +32,7 @@ type Container struct {
 	Name     string
 	IP       string
 	Hostname string
-	Stop     func()
-	stopped  bool
+	runCmd   *exec.Cmd
 	binaries set.Set
 }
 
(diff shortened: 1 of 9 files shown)


fv/ingress_egress_test.go
```diff
@@ -0,0 +1,177 @@
...
+		felix.Stop()
+
+		if CurrentGinkgoTestDescription().Failed {
+			utils.Run("docker", "exec", etcd.Name, "etcdctl", "ls", "--recursive", "/")
+		}
+		etcd.Stop()
+	})
+
+	It("full connectivity to and from workload 0", func() {
+		Expect(w[1].CanConnectTo(w[0].IP, w[0].Port)).To(BeTrue())
```
fasaxc: WIBNI: you could make a custom ginkgo predicate here:

```Expect(w[1]).To(HaveConnectivityTo(w[0]))```

that'd let you control the diagnostic output if the test fails so that it'd say something more useful than true != false.

File: fv/ingress_egress_test.go [hunk 2]
```diff
@@ -0,0 +1,177 @@
...
+			policy := api.NewPolicy()
+			policy.Metadata.Name = "policy-1"
+			allowFromW1 := api.Rule{
+				Action: "allow",
+				Source: api.EntityRule{
+					Selector: "name=='" + w[1].Name + "'",
+				},
+			}
+			policy.Spec.IngressRules = []api.Rule{allowFromW1}
+			policy.Spec.Selector = "name=='" + w[0].Name + "'"
```
fasaxc: Adding a  `w[0].NameSelector()` method might be useful in a few places.

fv/workload/workload.go
```diff
@@ -85,8 +110,38 @@ func (w *Workload) Configure(client *client.Client) {
...
+	errPipe, err := connectionCmd.StderrPipe()
+	Expect(err).NotTo(HaveOccurred())
+	err = connectionCmd.Start()
+	Expect(err).NotTo(HaveOccurred())
+
+	wOut, err := ioutil.ReadAll(outPipe)
+	Expect(err).NotTo(HaveOccurred())
+	wErr, err := ioutil.ReadAll(errPipe)
+	Expect(err).NotTo(HaveOccurred())
+	err = connectionCmd.Wait()
```
fasaxc: Would it be worth using a specific return code for "connection failed" and checking for it here?  I'm thinking that would make it easier to spot if the test binary was failing for an unexpected reason (such as passing it bad arguments).

neiljerram: I think I'd rather leave this one for future refinement, given current task pressure.

fasaxc: Sure, no problem.  All of these were enhancement ideas rather than must haves

fv/containers/containers.go
```diff
@@ -24,21 +24,38 @@ import (
...
+	binaries set.Set
 }
 
 var containerIdx = 0
 
 var runningContainers = []*Container{}
 
+func Stop(c *Container) {
+	if c == nil {
+		log.Info("Stop no-op because nil container")
```
fasaxc: I think the go idiom is to do an early return here and below (rather than an if/elseif/else.

fv/containers/containers.go
```diff
@@ -24,21 +24,38 @@ import (
...
-	stopped  bool
+	runCmd   *exec.Cmd
+	binaries set.Set
 }
 
 var containerIdx = 0
 
 var runningContainers = []*Container{}
 
+func Stop(c *Container) {
```
fasaxc: Why not do `func (c *Container) Stop() {`?

fasaxc:
Looks like one of the tests failed on semaphore

fv/ingress_egress_test.go
```diff
@@ -0,0 +1,180 @@
...
+			}
+			policy.Spec.IngressRules = []api.Rule{allowFromW1}
+			policy.Spec.Selector = w[0].NameSelector()
+			_, err := client.Policies().Create(policy)
+			Expect(err).NotTo(HaveOccurred())
+		})
+
+		It("only w1 can connect into w0, but egress from w0 is unrestricted", func() {
+			Expect(w[1]).To(HaveConnectivityTo(w[0]))
+			Expect(w[2]).NotTo(HaveConnectivityTo(w[0]))
```
fasaxc: As discussed, I think this might race so probably needs to be an `Eventually`

fv/ingress_egress_test.go
```diff
@@ -0,0 +1,180 @@
...
+			policy.Spec.Selector = w[0].NameSelector()
+			_, err := client.Policies().Create(policy)
+			Expect(err).NotTo(HaveOccurred())
+		})
+
+		It("ingress to w0 is unrestricted, but w0 can only connect out to w1", func() {
+			Expect(w[1]).To(HaveConnectivityTo(w[0]))
+			Expect(w[2]).To(HaveConnectivityTo(w[0]))
+			Expect(w[0]).To(HaveConnectivityTo(w[1]))
+			Expect(w[0]).NotTo(HaveConnectivityTo(w[2]))
```
fasaxc: Ditto.

fv/ingress_egress_test.go
```diff
@@ -0,0 +1,180 @@
...
+			policy.Spec.IngressRules = []api.Rule{allowFromW1}
+			policy.Spec.Selector = w[0].NameSelector()
+			policy.Spec.Types = []api.PolicyType{api.PolicyTypeIngress, api.PolicyTypeEgress}
+			_, err := client.Policies().Create(policy)
+			Expect(err).NotTo(HaveOccurred())
+		})
+
+		It("only w1 can connect into w0, and all egress from w0 is denied", func() {
+			Expect(w[1]).To(HaveConnectivityTo(w[0]))
+			Expect(w[2]).NotTo(HaveConnectivityTo(w[0]))
```
fasaxc: and here

PR #1549, opened 2017-09-14
Minor container FV improvements.  Add multiple port support, parallel conn checker.

fasaxc:
## Description

Broke this out of my named ports work in the hope of avoiding more conflicts...

- Add support for multiple ports to the container FV framework.
- Add support for explicit names to make diags more meaningful.

## Release Note

```release-note
None required
```

### DIFF
--- a/fv/host_port_test.go
+++ b/fv/host_port_test.go
@@ -19,6 +19,8 @@ package fv_test
 import (
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+	log "github.com/sirupsen/logrus"
+
 	"github.com/projectcalico/felix/fv/containers"
 	"github.com/projectcalico/felix/fv/metrics"
 	"github.com/projectcalico/felix/fv/utils"
@@ -26,7 +28,6 @@ import (
 	"github.com/projectcalico/libcalico-go/lib/api"
 	"github.com/projectcalico/libcalico-go/lib/client"
 	"github.com/projectcalico/libcalico-go/lib/numorstring"
-	log "github.com/sirupsen/logrus"
 )
 
 func RunEtcd() *containers.Container {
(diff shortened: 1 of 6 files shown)


fv/host_port_test.go
```diff
@@ -41,6 +42,7 @@ func RunFelix(etcdIP string) *containers.Container {
 	return containers.Run("felix-fv",
 		"--privileged",
 		"-e", "CALICO_DATASTORE_TYPE=etcdv2",
+		"-e", "FELIX_LOGSEVERITYSCREEN=debug",
```
neiljerram: I guess this is OK for now, with all existing tests being low-scale and functional.  If and when we migrate the k8sfv scale tests here, we will probably need to parameterize this.

fv/test-workload/test-workload.go
```diff
@@ -118,20 +128,25 @@ func main() {
...
 		}
 
-		l, err := net.Listen("tcp", ipAddress+":"+port)
-		if err != nil {
-			return err
+		// Listen on each port.
+		for _, port := range ports {
+			log.WithField("port", port).Info("About to listen for connections")
+			l, err := net.Listen("tcp", ipAddress+":"+port)
+			panicIfError(err)
```
neiljerram: Within the namespace.Do func, I was returning the err, rather than panicking directly, because I wasn't sure that it would be OK to panic while in the workload namespace.  Is that a valid concern at all?

fv/workload/workload.go
```diff
@@ -102,16 +104,35 @@ func Run(c *containers.Container, interfaceName, ip, port string) (w *Workload)
...
+	go func() {
+		for {
+			line, err := stdoutReader.ReadString('\n')
+			if err != nil {
+				log.WithError(err).Info("End of workload stdout")
+				return
+			}
+			log.Infof("Workload %s stdout: %s", name, strings.TrimSpace(string(line)))
+		}
+	}()
```
neiljerram: This is really nice - but unfortunately it means that my much more hacky workload-output-logging code, in the Stop() method, needs to be deleted.  Do you mind doing that?

fv/test-workload/test-workload.go
fasaxc: Looks like all their cleanup code is nicely `defer`-ed so it just works.

neiljerram: OK, cool.

PR #1550, opened 2017-09-15
Use docker stop instead of sigkill in fv tests

ozdanborne:
Not sure this is exactly the right fix, but wanted to submit before I go on holiday.

I'm finding that the k8s-apiserver container doesn't die in response to an os.Interrupt. My test runs end up panicking during teardown as a result.

This PR fixes my panics by issuing a stop signal using docker directly, which actually stops the k8s-apiserver pod.

related PR: https://github.com/projectcalico/k8s-policy/pull/139

### DIFF
--- a/fv/containers/containers.go
+++ b/fv/containers/containers.go
@@ -21,6 +21,7 @@ import (
 	"os/exec"
 	"strings"
 	"sync"
+	"syscall"
 	"time"
 
 	. "github.com/onsi/ginkgo"
@@ -49,10 +50,10 @@ func (c *Container) Stop() {
 	if c == nil {
 		log.Info("Stop no-op because nil container")
 	} else if c.runCmd == nil {
-		log.WithField("container", c).Info("Stop no-op because container is not running")
+		log.WithField("container", c.Name).Info("Stop no-op because container is not running")
 	} else {
 		log.WithField("container", c).Info("Stop")
-		c.runCmd.Process.Signal(os.Interrupt)
+		c.runCmd.Process.Signal(syscall.SIGTERM)
 		c.WaitNotRunning(10 * time.Second)
 		c.runCmd = nil
 
@@ -68,7 +69,7 @@ func Run(namePrefix string, args ...string) (c *Container) {
 	c = &Container{Name: fmt.Sprintf("%v-%d-%d-", namePrefix, os.Getpid(), containerIdx)}
 
 	// Start the container.
-	log.WithField("container", c).Info("About to run container")
+	log.WithField("container", c.Name).Info("About to run container")
 	runArgs := append([]string{"run", "--name", c.Name}, args...)
 	c.runCmd = exec.Command("docker", runArgs...)
 	err := c.runCmd.Start()
@@ -120,7 +121,7 @@ func (c *Container) WaitRunning(timeout time.Duration) {
 			break
 		}
 		if time.Since(start) > timeout {
-			log.Panic("Timed out waiting for container to be listed.")
+			log.WithField("container", c.Name).Panic("Timed out waiting for container to be listed.")
 		}
 	}
 }
@@ -136,7 +137,7 @@ func (c *Container) WaitNotRunning(timeout time.Duration) {
 			break
 		}
 		if time.Since(start) > timeout {
-			log.Panic("Timed out waiting for container not to be listed.")
+			log.WithField("container", c.Name).Panic("Timed out waiting for container not to be listed.")
 		}
 	}
 }


neiljerram:
@ozdanborne I'm inclined to be conservative and not to make this change unless we can be sure that it's needed.  In particular, the felix PR at #1551 adds a k8s-apiserver container, _and_ uses the existing Stop() implementation to stop it - and in the felix FV that seems to work reliably every time.

neiljerram:
@ozdanborne By complete coincidence I just came across https://github.com/Yelp/dumb-init, which includes:
```
Normally, when you launch a Docker container, the process you're executing becomes PID 1, giving it the quirks and responsibilities that come with being the init system for the container.

There are two common issues this presents:

In most cases, signals won't be handled properly.

The Linux kernel applies special signal handling to processes which run as PID 1.

When processes are sent a signal on a normal Linux system, the kernel will first check for any custom handlers the process has registered for that signal, and otherwise fall back to default behavior (for example, killing the process on SIGTERM).

However, if the process receiving the signal is PID 1, it gets special treatment by the kernel; if it hasn't registered a handler for the signal, the kernel won't fall back to default behavior, and nothing happens. In other words, if your process doesn't explicitly handle these signals, sending it SIGTERM will have no effect at all.
```

I wonder if this is the difference between how your FV is running k8s-apiserver, and how mine is - and hence yours perhaps is ignoring SIGINT?

FYI, my k8s-apiserver is run with:
```
	apiServerContainer = containers.Run("apiserver",
		"gcr.io/google_containers/hyperkube-amd64:v"+config.K8sVersion,
		"/hyperkube", "apiserver",
		fmt.Sprintf("--etcd-servers=http://%s:2379", etcdContainer.IP),
		"--service-cluster-ip-range=10.101.0.0/16",
		"-v=10",
		"--authorization-mode=RBAC",
	)
```

fasaxc:
@neiljerram We use `tini` in felix/typha for that reason.  I believe newer versions of docker even have some support for tini built in, so you can run `docker run --tini` and it'll put in a `tini` shim around the process that you run.

neiljerram:
@fasaxc Right, thanks.

@ozdanborne So if the container that you're using has tini, or a similar init shim, SIGINT will work; and interrupting a process that the test code has in hand feels more elegant to me than `docker stop`.

fv/containers/containers.go
```diff
@@ -85,7 +85,7 @@ func Run(namePrefix string, args ...string) (c *Container) {
 	c.IP = c.GetIP()
 	c.Hostname = c.GetHostname()
 	c.binaries = set.New()
-	log.WithField("container", c).Info("Container now running")
+	log.WithField("container", c.Name).Info("Container now running")
```
neiljerram: I'd like to leave this one logging the whole Container struct.

ozdanborne:
Thanks for the info guys. I looked into it some more and noticed that it was etcd that was failing - not k8s-apiserver. I fixed it by pulling a more recent `etcd:latest` image. 

I've updated as per your comments. Hoping we can merge these logging improvements. 

For reference, this is what they look like now: `time="2017-09-19T13:45:08-07:00" level=info msg="About to run container" container=&{calico-policy-controller-3750-3-   <nil> {0 0} <nil>} `

I can see how logging the entire struct is more useful once its actually running: `time="2017-09-19T13:45:08-07:00" level=info msg="Container now running" container=&{calico-policy-control
ler-3750-3- 172.17.0.7 d001318c60f9 0xc420432420 {0 0} map[]} `

ozdanborne:
Scratch that, I'm still seeing intermittent issues killing apiserver and etcd when using the sigint approach. 

I strongly recommend we switch to the `docker stop` approach. It has builtin support where it waits `-t` seconds for the container to respond to the signal before issuing a hard kill.

ozdanborne:
How about this compromise - switch to `SIGTERM` instead, which is the sig that `docker stop` sends? (https://docs.docker.com/engine/reference/commandline/stop/)

fasaxc:
I think the key thing to note here is that we're sending the interrupt to the `docker run` CLI command.  That's like sending a Ctrl-C to that command, which should tell it to stop the container that it's running.

If that's not working in all cases, maybe we should switch to `docker stop` or do belt-and-braces.  

In any case, the idea behind using `docker run` is that it helps to make sure the container gets cleaned up if the test is killed.  I guess we could name our containers according to the pid of the test process, allowing us to reap it if the process no longer exists when the next test starts, or something like that.  

As discussed, the key things we were trying to maintain in this framework are:

- parallel execution by unique naming
- avoiding test pollution by tearing down containers between tests
- automatic cleanup if test process dies

ozdanborne:
BTW, its possible I was hitting errors that you guys weren't because I was executing tests by simply running `ginkgo`. I haven't had time to circle back and confirm that, just a thought that occurred to me later on. Once I switched to building then running `e2e.test` binary, I stopped hitting this issue. Not sure if coincidence and something else was going on?

neiljerram:
@ozdanborne As we seem to be struggling to agree about the docker stop vs SIGINT, how about putting the logging changes into a separate PR, so we can at least merge those?

djlwilder:
Before this issue is closed I wanted to add a comment.  I am seeing the same issue on ppc64le ( etcd and felix containers not stopping in teardown).  Here are a few things I found:

-The containers do eventual stop,  (they are no longer running after the test completes). Therefore I don't think the method of stopping them is the problem.

- Running top during the test shows dockerd at times consuming up to 200% of a cpu.

- My working theory is that this is a performance issue,  the Ginkgo runs parallel test (-p) 15 in my case.  I don't see the issue on a smaller x86 box.  Reducing the number of parallel tests helps a little.
  
- WaitNotRunning() runs  "docker ps" in a tight loop,  I added a 1 second sleep in the loop (simmer to in WaitUntilRunning() ),  dockerd peek cpu usage dropped to roughly 50%.  However, the problem still occurred.

- I increased the time to wait for a container to stop from 10 seconds to 120 seconds and increased  Ginkgo's slowSpecThreshold from 40 seconds to 120 seconds.  The tests runs ok in this configuration.  

Note:  I am only running a subset of the full fv tests at this point.

ozdanborne:
closing since this is outdated

PR #1515, opened 2017-07-25
[WiP] FV tests for Felix/Typha readiness

fasaxc:
## Description
This PR adds FV tests for Felix and Typha's readiness endpoints.

## Todos
- [ ] Unit tests (full coverage)
- [ ] Integration tests (delete as appropriate) In plan/Not needed/Done
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

PR #1551, opened 2017-09-15
Readiness/Liveness FV testing

neiljerram:
## Description
FV for Felix and Typha health endpoints.  This PR takes the logic of #1515 and merges it into the new FV infrastructure.


fv/containers/containers.go
```diff
@@ -109,27 +128,44 @@ func (c *Container) GetHostname() string {
...
+		cmd := utils.Command("docker", "ps")
 		out, err := cmd.CombinedOutput()
 		Expect(err).NotTo(HaveOccurred())
 		if strings.Contains(string(out), c.Name) {
 			break
 		}
-		if time.Since(start) > timeout {
-			log.Panic("Timed out waiting for container to be listed.")
-		}
+		time.Sleep(1 * time.Second)
```
fasaxc: Can we make this sleep smaller?  If it takes about a second to start a container we may add a second to every test.

fv/containers/containers.go
```diff
@@ -201,3 +235,24 @@ func (c *Container) CanConnectTo(ip, port string) bool {
 
 	return err == nil
 }
+
+func RunEtcd() *Container {
+	return Run("etcd",
+		"quay.io/coreos/etcd",
+		"etcd",
+		"--advertise-client-urls", "http://127.0.0.1:2379",
```
fasaxc: Hmm, this should be the real IP of the etcd container or a smart client may end up trying to reconnect to 127.0.0.1 instead of the main etcd IP.  Is that easy to fix?

fv/health_test.go
```diff
@@ -0,0 +1,469 @@
...
+		)
+		if err != nil {
+			log.Info("Waiting for API server to accept cluster role binding")
+			time.Sleep(2 * time.Second)
+			continue
+		}
+		break
+	}
+
+	// Allow anonymous connections to the API server.  We also use this command to wait
```
fasaxc: Copy/paste error?

fv/health_test.go
```diff
@@ -0,0 +1,469 @@
...
+			time.Sleep(2 * time.Second)
+			continue
+		}
+		break
+	}
+
+	k8sAPIEndpoint = fmt.Sprintf("https://%s:6443", apiServerContainer.IP)
+	badK8sAPIEndpoint = fmt.Sprintf("https://%s:1234", apiServerContainer.IP)
+	for {
+		resp, err := insecureHTTPClient.Get(k8sAPIEndpoint + "/apis/extensions/v1beta1/thirdpartyresources")
```
fasaxc: Should we be testing CRDs now?

fv/health_test.go
```diff
@@ -0,0 +1,469 @@
...
+		}
+		err = calicoClient.EnsureInitialized()
+		if err != nil {
+			log.WithError(err).Warn("Failed to init datastore")
+			continue
+		}
+		break
+	}
+
+	for {
```
fasaxc: Should we time out this loop? (and above?)

fv/health_test.go
```diff
@@ -0,0 +1,469 @@
...
+	Expect(etcdContainer).NotTo(BeNil())
+
+	// Start the k8s API server.
+	apiServerContainer = containers.Run("apiserver",
+		"gcr.io/google_containers/hyperkube-amd64:v"+config.K8sVersion,
+		"/hyperkube", "apiserver",
+		fmt.Sprintf("--etcd-servers=http://%s:2379", etcdContainer.IP),
+		"--service-cluster-ip-range=10.101.0.0/16",
+		"-v=10",
+		"--authorization-mode=RBAC",
```
ozdanborne: Is there a reason you are turning RBAC on only to add a clusterrolebinding that gives anonymous full access? Why not `--authorization-mode=AlwaysAllow` right from the start?

fv/containers/containers.go
neiljerram: Unfortunately I don't see any nice way of doing this.  We could start using `--ip=` and managing all the Docker container IPs ourselves, but obviously that would be far from nice.  All of the examples that I could find about running etcd in a container use `--net=host`, so they already know the host's IP.

fv/health_test.go
fasaxc: 30s should be enough for anyone

fv/health_test.go
neiljerram: Yes: the clients in this test - Felix, Typha and the test code itself - all connect anonymously to the API server, because (a) they aren't running in pods in a proper Kubernetes cluster, and (b) they don't provide client TLS certificates, and (c) they don't use any of the other non-anonymous mechanisms that Kubernetes supports.  But, as of 1.6, the API server doesn't allow anonymous users with the default "AlwaysAllow" authorization mode.  So we specify the "RBAC" authorization mode instead, and create a ClusterRoleBinding that gives the "system:anonymous" user unlimited power (aka the "cluster-admin" role).

(This is noted in the k8sfv test at https://github.com/projectcalico/felix/blob/master/k8sfv/run-test#L108-L118, but I will add it here as well.)

neiljerram:
(Rebased to fix conflicts w.r.t. the merge of pre-DNAT FV)

calc/async_calc_graph.go
```diff
@@ -101,6 +103,11 @@ func NewAsyncCalcGraph(conf *config.Config, outputEvents chan<- interface{}, hea
 		eventBuffer:      eventBuffer,
 		healthAggregator: healthAggregator,
 	}
+	if conf.DebugSimulateCalcGraphHangAfter != 0 {
+		log.WithField("delay", conf.DebugSimulateCalcGraphHangAfter).Warn(
+			"Simulating a calculation graph hang.")
+		g.debugHangC = time.After(conf.DebugSimulateDataplaneHangAfter)
```
fasaxc: Oops, this is the wrong parameter vs the if condition above.

PR #1555, opened 2017-09-19
Fix AllowAction flags behavior, so it replaces all ALLOWs with RETURNs.

Dirbaio:
## Description
This PR fixes the `IptablesFilterAllowAction` and `IptablesMangleAllowAction` flags so they return all packets when they're set to RETURN.

As discussed on Slack, the original intent of these flags is to allow operators to insert their own iptables rules in the top-level iptables chains after the Calico chains, to do further filtering or manipulation of the packets. Therefore, Calico should return all allowed packets to the original chain, instead of outright accepting them.

Unfortunately, setting the flags to RETURN only change some ACCEPTs to RETURNs, but not all, so most of the traffic is still ACCEPTed by Calico, making the flags not very useful.

This PR fixes the flags so all the allowed Calico traffic gets returned to the top-level iptables chains.

The only exception is the failsafe traffic. Calico still ACCEPTs it. It can be changed to RETURN too, but:
- it would require the cali-failsafe chains to return the acceptance status via a mark bit, like the policy chains do, incurring in extra load.
- The failsafe wouldn't protect against user misconfigurations in their own rules. (Maybe that's OK since it's outside Calico's responsibility)

IMO this is not a big deal, if you want all traffic to hit your chains you can just disable the failsafe.

To be more specific, this PR ensure that Calico behavior is the following: 

- If IptablesAllowAction == ACCEPT, the cali-INPUT, cali-OUTPUT, cali-FORWARD chains in the filter table should:
  - ACCEPT packets allowed by calico failsafe
  - ACCEPT packets accepted by calico policy
  - DROP packets denied by calico policy
  - RETURN packets for endpoints not managed by calico

- If IptablesAllowAction == RETURN, the cali-INPUT, cali-OUTPUT, cali-FORWARD chains in the filter table should:
  - ACCEPT packets allowed by calico failsafe
  - RETURN packets accepted by calico policy
  - DROP packets denied by calico policy
  - RETURN packets for endpoints not managed by calico
- Same for the chains in the mangle table.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
Fixed the behavior when setting IptablesFilterAllowAction or IptablesMangleAllowAction to RETURN: now all packets allowed by Calico are returned to the top-level chains for further processing.
```

### DIFF
--- a/rules/endpoints.go
+++ b/rules/endpoints.go
@@ -41,6 +41,7 @@ func (r *DefaultRuleRenderer) WorkloadEndpointToIptablesChains(
 			"", // No fail-safe chains for workloads.
 			chainTypeNormal,
 			adminUp,
+			r.filterAllowAction, // Workload endpoint chains are only used in the filter table
 		),
 		// Chain for traffic _from_ the endpoint.
 		r.endpointIptablesChain(
@@ -53,6 +54,7 @@ func (r *DefaultRuleRenderer) WorkloadEndpointToIptablesChains(
 			"", // No fail-safe chains for workloads.
 			chainTypeNormal,
 			adminUp,
+			r.filterAllowAction, // Workload endpoint chains are only used in the filter table
 		),
 	}
 }
@@ -76,6 +78,7 @@ func (r *DefaultRuleRenderer) HostEndpointToFilterChains(
 			ChainFailsafeOut,
 			chainTypeNormal,
 			true, // Host endpoints are always admin up.
+			r.filterAllowAction,
 		),
 		// Chain for traffic _from_ the endpoint.
 		r.endpointIptablesChain(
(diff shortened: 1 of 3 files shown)


marvin-tigera:

Dirbaio:
I found that the failsafe wasn't working in RETURN mode with my first patch.

I've updated the PR with a new patch that doesn't do RETURN for failsafe traffic (see description for why), and updated the description accordingly.

fasaxc:
ok to test

fasaxc:
retest this please

fasaxc:
retest this please

fasaxc:
Ah build failure is caused by a bug in our build script; it doesn't like the `/` in your branch name.

Dirbaio:
@fasaxc do you want me to resend the PR with a different branch name?

fasaxc:
@Dirbaio that'd work as a quick fix, I've also filed an issue against that script to fix it up.

Dirbaio:
Cool. New PR is #1561

PR #1558, opened 2017-09-22
Explicitly allow IPIP packets from/to Calico hosts.

fasaxc:
## Description
Rather than fixing https://github.com/projectcalico/calico/issues/888 by doccing that a rule is needed, maybe we should make felix auto-generate the rule?  WDYT?

This removes the need for an explicit policy rule when using IPIP and host protection together.

Fixes https://github.com/projectcalico/calico/issues/888
Fixes https://github.com/projectcalico/calico/issues/1357

## Todos
- [x] Documentation - checked the host protection guide, doesn't mention IPIP
- [x] Release note

## Release Note

```release-note
When IPIP is enabled, Calico now installs an explicit allow rule for its IPIP traffic.  This removes the need to manually add such a rule when using host protection policy. 
```

### DIFF
--- a/felix.go
+++ b/felix.go
@@ -80,6 +80,13 @@ const (
 	// heap can be lost to garbage.  We reduce it to this value to trade increased CPU usage for
 	// lower occupancy.
 	defaultGCPercent = 20
+
+	// String sent on the failure report channel to indicate we're shutting down for config
+	// change.
+	reasonConfigChanged = "config changed"
+	// Process return code used to report a config change.  This is the same as the code used
+	// by SIGHUP, which means that the wrapper script also restarts Felix on a SIGHUP.
+	configChangedRC = 129
 )
 
 // main is the entry point to the calico-felix binary.
@@ -571,9 +578,11 @@ func servePrometheusMetrics(configParams *config.Config) {
 }
 
 func monitorAndManageShutdown(failureReportChan <-chan string, driverCmd *exec.Cmd, stopSignalChans []chan<- bool) {
-	// Ask the runtime to tell us if we get a term signal.
-	termSignalChan := make(chan os.Signal, 1)
-	signal.Notify(termSignalChan, syscall.SIGTERM)
+	// Ask the runtime to tell us if we get a term/int signal.
+	signalChan := make(chan os.Signal, 1)
+	signal.Notify(signalChan, syscall.SIGTERM)
+	signal.Notify(signalChan, syscall.SIGINT)
+	signal.Notify(signalChan, syscall.SIGHUP)
 
 	// Start a background thread to tell us when the dataplane driver stops.
 	// If the driver stops unexpectedly, we'll terminate this process.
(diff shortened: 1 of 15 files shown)


Dirbaio:
+1000 to this. Our prod cluster had IPIP + host endpoint policy and I overlooked the "allow IPIP" detail. The resulting issues were very hard to debug, since conntrack would allow incoming IPIP packets if the host had sent an outgoing IPIP packet before. It would only fail when spinning up a new node, only before that node had tried to send out IPIP traffic.

re: the security concern: IIRC IPIP traffic goes through iptables two times, first wrapped, then unwrapped, so you can't circunvent hostendpoint policies sending traffic through the tunnel destined for the node's IP.

fasaxc:
Yes, I think @Dirbaio is correct, the unwrapped packet is re-injected and seems to come from the tunl0 device, where our other policy will get applied as normal.

fv/containers/containers.go
```diff
@@ -308,3 +312,84 @@ func StartSingleNodeEtcdTopology() (felix, etcd *Container, client client.Interf
 	success = true
 	return
 }
+
+// StartSingleNodeEtcdTopology starts an etcd container and a single Felix container; it initialises
```
neiljerram: Single -> Two

neiljerram: (also "a single Felix container")

fv/containers/containers.go
```diff
@@ -308,3 +312,84 @@ func StartSingleNodeEtcdTopology() (felix, etcd *Container, client client.Interf
...
+// the datastore and installs a Node resource for the Felix node.
+func StartTwoNodeEtcdIPIPTopology() (felixes []*Container, etcd *Container, client client.Interface) {
+	log.Info("Starting a single-node etcd topology.")
```
neiljerram: single again

fv/containers/containers.go
```diff
@@ -308,3 +312,84 @@ func StartSingleNodeEtcdTopology() (felix, etcd *Container, client client.Interf
...
+		}
+		ctx, cancel = context.WithTimeout(context.Background(), 10*time.Second)
+		defer cancel()
+		ipPool := api.NewIPPool()
+		ipPool.Name = "test-pool"
+		ipPool.Spec.CIDR = "10.65.0.0/16"
+		ipPool.Spec.IPIPMode = api.IPIPModeAlways
+		_, err = client.IPPools().Create(ctx, ipPool, options.SetOptions{})
+		return err
+	}).ShouldNot(HaveOccurred())
```
neiljerram: I guess all that is shared with the single node setup case - can it be commonized?

neiljerram: Or you could just make the number of felixes a parameter to this whole function, and use it in the single case too.

fv/containers/containers.go
```diff
@@ -308,3 +312,84 @@ func StartSingleNodeEtcdTopology() (felix, etcd *Container, client client.Interf
...
+		felixes = append(felixes, felix)
+	}
+
+	for i, iFelix := range felixes {
+		for j, jFelix := range felixes {
+			if i == j {
+				continue
+			}
+
+			jBlock := fmt.Sprintf("10.65.%d.0/24", j)
```
neiljerram: Possible comment about not using IPAM here.

fv/ipip_test.go
```diff
@@ -0,0 +1,178 @@
...
+				_, err := client.HostEndpoints().Create(ctx, hep, options.SetOptions{})
+				Expect(err).NotTo(HaveOccurred())
+			}
+		})
+
+		It("should have workload connectivity but not host connectivity", func() {
+			// Host endpoints block host-host traffic.
+			cc.ExpectNone(felixes[0], hostW[1])
+			cc.ExpectNone(felixes[1], hostW[0])
+			// But the rules to allow IPIP between our hosts let the workload traffic through.
```
neiljerram: Add "and the lack of applyOnForward"?

felix.go
```diff
@@ -639,18 +653,36 @@ func monitorAndManageShutdown(failureReportChan <-chan string, driverCmd *exec.C
...
 	}
 
-	if !receivedSignal {
+	if !receivedFatalSignal {
 		// We're exiting due to a failure or a config change, wait
 		// a couple of seconds to ensure that we don't go into a tight
-		// restart loop (which would make the init daemon give up trying
-		// to restart us).
-		logCxt.Info("Shutdown wasn't caused by signal, pausing to avoid tight restart loop")
+		// restart loop (which would make the init daemon in calico/node give
```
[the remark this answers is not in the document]
fasaxc: I'm talking about the init daemon used in calico/node, which is `runit`, not `tini`, and it does restart `calico-felix`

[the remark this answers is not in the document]
neiljerram: Ah, right, sorry.

fv/ipip_test.go
fasaxc: Discussed offline and improved surrounding comment.

PR #1565, opened 2017-10-05
Support workload ingress policy for kube-proxy running in ipvs mode.

song-jiang:
## Description
Added workload ingress policy support for kube-proxy running in ipvs mode.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

### DIFF
--- a/config/config_params.go
+++ b/config/config_params.go
@@ -125,7 +125,8 @@ type Config struct {
 	MetadataAddr string `config:"hostname;127.0.0.1;die-on-fail"`
 	MetadataPort int    `config:"int(0,65535);8775;die-on-fail"`
 
-	InterfacePrefix string `config:"iface-list;cali;non-zero,die-on-fail"`
+	InterfacePrefix  string `config:"iface-list;cali;non-zero,die-on-fail"`
+	InterfaceExclude string `config:"iface-list;kube-ipvs0"`
 
 	ChainInsertMode             string `config:"oneof(insert,append);insert;non-zero,die-on-fail"`
 	DefaultEndpointToHostAction string `config:"oneof(DROP,RETURN,ACCEPT);DROP;non-zero,die-on-fail"`
@@ -215,6 +216,10 @@ func (config *Config) UpdateFrom(rawData map[string]string, source Source) (chan
 
 func (c *Config) InterfacePrefixes() []string {
 	return strings.Split(c.InterfacePrefix, ",")
+}
+
+func (c *Config) InterfaceExcludes() []string {
+	return strings.Split(c.InterfaceExclude, ",")
 }
 
 func (config *Config) OpenstackActive() bool {
--- a/ifacemonitor/iface_monitor.go
+++ b/ifacemonitor/iface_monitor.go
@@ -43,7 +43,13 @@ const (
 type InterfaceStateCallback func(ifaceName string, ifaceState State)
 type AddrStateCallback func(ifaceName string, addrs set.Set)
 
+type Config struct {
+	// List of interface names that dataplane receives no callbacks from them.
+	InterfaceExcludes []string
+}
 type InterfaceMonitor struct {
+	Config
+
 	netlinkStub  netlinkStub
 	resyncC      <-chan time.Time
 	upIfaces     set.Set
@@ -53,14 +59,15 @@ type InterfaceMonitor struct {
 	ifaceAddrs   map[int]set.Set
 }
 
-func New() *InterfaceMonitor {
+func New(config Config) *InterfaceMonitor {
 	// Interface monitor using the real netlink, and resyncing every 10 seconds.
 	resyncTicker := time.NewTicker(10 * time.Second)
-	return NewWithStubs(&netlinkReal{}, resyncTicker.C)
+	return NewWithStubs(config, &netlinkReal{}, resyncTicker.C)
 }
 
-func NewWithStubs(netlinkStub netlinkStub, resyncC <-chan time.Time) *InterfaceMonitor {
+func NewWithStubs(config Config, netlinkStub netlinkStub, resyncC <-chan time.Time) *InterfaceMonitor {
 	return &InterfaceMonitor{
+		Config:      config,
 		netlinkStub: netlinkStub,
 		resyncC:     resyncC,
 		upIfaces:    set.New(),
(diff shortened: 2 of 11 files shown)


rules/static.go
```diff
@@ -313,12 +313,18 @@ func (r *DefaultRuleRenderer) filterOutputChain() *Chain {
...
 	for _, prefix := range r.WorkloadIfacePrefixes {
-		// If the packet is going to a worklaod endpoint, RETURN.
+		// If the packet is going to a workload endpoint, apply workload ingress policy if traffic
+		// belongs to an IPVS connection and return at the end.
 		log.WithField("ifacePrefix", prefix).Debug("Adding workload match rules")
 		ifaceMatch := prefix + "+"
 		rules = append(rules,
 			Rule{
+				Match:  Match().OutInterface(ifaceMatch).IPVSConnection(),
+				Action: JumpAction{Target: ChainToWorkloadDispatch},
```
neiljerram: I just noticed that we use Return here, when we have logically allowed a packet, whereas in the forward chain we use AcceptAction.  Do you know why that is?

ifacemonitor/ifacemonitor_test.go
```diff
@@ -347,6 +350,15 @@ var _ = Describe("ifacemonitor", func() {
 		<-nl.userSubscribed
 	})
 
+	It("should skip netlink address updates for ipvs", func() {
+		SetDefaultConsistentlyDuration(1 * time.Second)
```
neiljerram: I think the code should have "1s" on the specific Consistently statements for which this is appropriate, instead of setting this default for the whole test suite.  (Unless there is some way of making this change temporarily for a group of tests, using BeforeEach and AfterEach?)

ifacemonitor/ifacemonitor_test.go
```diff
@@ -347,6 +350,15 @@ var _ = Describe("ifacemonitor", func() {
...
+		nl.addLink("kube-ipvs0")
+		resyncC <- time.Time{}
+		dp.expectAddrStateCb("kube-ipvs0", "", true)
```
neiljerram: Two points here:

1. I'm not sure I understand the logic for *which* callbacks are skipped, for an excluded interface.  Specifically (a) is it correct that there is no LinkStateCb? - what bits of Felix are influenced by that?  (b) why *do* we get the initial AddrStateCb as shown above?  (It may be that this is all correct, but then it would be good to write a comment to explain it.)

2. What does `"", true` mean?  expectAddrStateCb doesn't have a precise expectation for this combination.

config/config_params.go
```diff
@@ -125,7 +125,8 @@ type Config struct {
 	MetadataAddr string `config:"hostname;127.0.0.1;die-on-fail"`
 	MetadataPort int    `config:"int(0,65535);8775;die-on-fail"`
 
-	InterfacePrefix string `config:"iface-list;cali;non-zero,die-on-fail"`
+	InterfacePrefix  string `config:"iface-list;cali;non-zero,die-on-fail"`
+	InterfaceExclude string `config:"iface-list;kube-ipvs0"`
```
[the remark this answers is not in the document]
song-jiang: As discussed, we are not going to advertise this config parameters so leave as it is for the time being.

rules/static.go
song-jiang: Discussed. No change needed.

ifacemonitor/ifacemonitor_test.go
```diff
@@ -351,12 +370,29 @@ var _ = Describe("ifacemonitor", func() {
...
 		nl.addAddr("kube-ipvs0", "10.100.0.1/32")
+		// Should not expect an address notification from interface monitor to dataplane.
 		dp.notExpectAddrStateCb()
+
+		// Should not expect a link up/down/delete if link has been added but excluded for moinitoring.
+		nl.changeLinkState("kube-ipvs0", "up")
+		dp.notExpectLinkStateCb()
+		nl.changeLinkState("kube-ipvs0", "down")
+		nl.delLink("kube-ipvs0")
+		dp.notExpectLinkStateCb()
```
neiljerram: Is there an AddrStateCb here?  It feels asymmetric if there is a callback when the link is created, but not when it is deleted.  Could we either:
- ensure that there is a symmetric callback when the link is deleted, or
- suppress the callback when the link is created?

neiljerram: (I think probably the latter would be preferable.)

ifacemonitor/ifacemonitor_test.go
```diff
@@ -280,7 +292,14 @@ func (dp *mockDataplane) addrStateCallback(ifaceName string, addrs set.Set) {
...
 func (dp *mockDataplane) notExpectAddrStateCb() {
-	Consistently(dp.addrC).ShouldNot(Receive())
+	Consistently(func() bool {
+		select {
+		case <-dp.addrC:
+			return true
+		default:
+			return false
+		}
+	}, "1s", "200ms").ShouldNot(BeTrue())
```
neiljerram: Does `Consistently(dp.addrC, "1s", "200ms").ShouldNot(Receive())` not work?  (I didn't think that you would need a change as big as this.)

ifacemonitor/iface_monitor.go
```diff
@@ -44,6 +44,9 @@ type InterfaceStateCallback func(ifaceName string, ifaceState State)
 type AddrStateCallback func(ifaceName string, addrs set.Set)
 
 type Config struct {
+	// List of interface names that should be excluded from being monitored on events.
+	// Dataplane recieves link add/delete events for an excluded interface as a normal interface.
```
neiljerram: "recieves" should be "receives".

ifacemonitor/iface_monitor.go
```diff
@@ -44,6 +44,9 @@ type InterfaceStateCallback func(ifaceName string, ifaceState State)
...
+	// Dataplane ignores ip address update and link up/down events for an excluded interface.
```
neiljerram: I think this is still difficult to understand.  The reader would have to know that "link add/delete events" means AddrStateCallback with no address information, and that "link up/down events" means InterfaceStateCallback.

The best thing, I think, would be if we can eliminate the one remaining AddrStateCallback.  Then we can just say that there are no callbacks at all for an excluded interface.

PR #1566, opened 2017-10-11
Test when hostEp has expectedIPs matching 2 host interfaces

neiljerram:


neiljerram:
Shaun, an extra test here for something that occurred to me while working on Host * protection.  Please could you review?

PR #1567, opened 2017-10-12
Run FV tests in parallel by copying ginkgo out of the build container.

fasaxc:
## Description
So far we've struggled to get the FV tests running inside the build container due to their docker dependency.  This PR copies the ginkgo tool out of the container so we can use it for executing the tests in parallel.

## Todos
- [x] Run this in CI several times to build confidence in it.

## Release Note

```release-note
None required
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -388,7 +388,6 @@ fv: calico/felix bin/iptables-locker bin/test-workload bin/test-connection $(FV_
 	# if we could give the build container access to the docker API but we've so-far struggled
 	# to get that working.
 	@echo Running Go FVs.
-	mkdir -p bin
 	$(DOCKER_GO_BUILD) cp /go/bin/ginkgo bin/ginkgo
 	# fv.test is not expecting a container name with an ARCHTAG.
 	-docker tag calico/felix$(ARCHTAG) calico/felix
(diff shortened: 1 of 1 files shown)


fasaxc:
@neiljerram I pulled out the questionable interface retry so I think this is good to go.

Makefile
```diff
@@ -383,12 +383,18 @@ $(FV_TESTS): vendor/.up-to-date $(FELIX_GO_FILES)
 
 .PHONY: fv
 fv: calico/felix bin/iptables-locker bin/test-workload bin/test-connection $(FV_TESTS)
+	# Copy the ginkgo binary out of the container since we need to run the fv tests directly
+	# on the host (because they need to be able to manipulate docker).  It'd be even nicer
+	# if we could give the build container access to the docker API but we've so-far struggled
+	# to get that working.
 	@echo Running Go FVs.
+	mkdir -p bin
```
neiljerram: We know that bin already exists, because of the deps in the `fv:` line above.  So this will always be a no-op, and I think it's worth saving a line here.

Alternatively, another approach that would be usable for other future targets, is:
```
bin/ginkgo:
       mkdir -p bin
       $(DOCKER_GO_BUILD) cp /go/bin/ginkgo bin/ginkgo
```
and then add bin/ginkgo as another dep of the `fv:` line.

PR #1571, opened 2017-10-19
Make it optional to auto-remove containers and add support for restarting stopped ones

ozdanborne:
## Description
Improves FV framework by allowing tests to restart containers.

## Release Note

```release-note
None required
```

### DIFF
--- a/fv/containers/containers.go
+++ b/fv/containers/containers.go
@@ -60,7 +60,11 @@ func (c *Container) Stop() {
 	}
 }
 
-func Run(namePrefix string, args ...string) (c *Container) {
+type RunOpts struct {
+	AutoRemove bool
+}
+
+func Run(namePrefix string, opts RunOpts, args ...string) (c *Container) {
 
 	// Build unique container name and struct.
 	containerIdx++
@@ -68,7 +72,15 @@ func Run(namePrefix string, args ...string) (c *Container) {
 
 	// Prep command to run the container.
 	log.WithField("container", c).Info("About to run container")
-	runArgs := append([]string{"run", "--rm", "--name", c.Name, "--hostname", c.Name}, args...)
+	runArgs := []string{"run", "--name", c.Name, "--hostname", c.Name}
+
+	if opts.AutoRemove {
+		runArgs = append(runArgs, "--rm")
+	}
+
+	// Add remaining args
+	runArgs = append(runArgs, args...)
+
 	c.runCmd = utils.Command("docker", runArgs...)
 
 	// Get the command's output pipes, so we can merge those into the test's own logging.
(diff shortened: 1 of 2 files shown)


fv/containers/containers.go
```diff
@@ -66,7 +66,7 @@ func Run(namePrefix string, args ...string) (c *Container) {
 
 	// Prep command to run the container.
 	log.WithField("container", c).Info("About to run container")
-	runArgs := append([]string{"run", "--rm", "--name", c.Name, "--hostname", c.Name}, args...)
+	runArgs := append([]string{"run", "--name", c.Name, "--hostname", c.Name}, args...)
```
fasaxc: The `--rm` was very-much intentional here.  It basically makes cleanup automatic if the test is `Ctrl-C`ed, which is doubly important because Felix's FVs can be run in parallel making automated cleanup tricky.  The cleanup takes advantage of the fact that we leave the docker CLI process running and it automatically gets killed if our process dies.

neiljerram: @ozdanborne and I discussed yesterday.  He has a requirement for being able to stop and then restart a container.  @ozdanborne Is it the case that a container with `--rm` will be destroyed completely when it is stopped?  If so, then clearly the containers that you want to restart can't have `--rm`.  However, as @fasaxc says, we do still want `--rm` for most containers - or else some equally robust cleanup solution - and in any case you'll also need a robust cleanup solution for the containers that you want to stop and restart.

ozdanborne: Why do we need to clean up stopped containers?

fasaxc: As discussed offline, cleanup was one of the problems we were trying to solve here for the felix FVs but happy for this to be extended with a new use-case as long as we don't lose the cleanup that we wanted for felix FVs.

ozdanborne:
@fasaxc @neiljerram can you guys take a look, let me know if this is an OK approach? (low priority though :smiley:  )

ozdanborne:
@fasaxc @neiljerram Any guidance on this guy? Would like to see this moved along within the v3-beta as [our resiliency tests in the kube-controllers repo](https://github.com/projectcalico/kube-controllers/blob/master/tests/fv/fv_resiliency_test.go) are skipped until this is merged:

fv/containers/containers.go
```diff
@@ -243,6 +288,7 @@ func (c *Container) CanConnectTo(ip, port, protocol string) bool {
 
 func RunEtcd() *Container {
 	return Run("etcd",
+		RunOpts{true},
```
fasaxc: Please can you spell out the field name here an below?  If we add another option field then this will break.

ozdanborne:
Addressed the comment then rebased and tests are now passing. I'll let you hit the merge button when ready @fasaxc

PR #1572, opened 2017-10-20
Use the v2 backend API to load configuration.

fasaxc:
## Description
The v2 backend API no longer supports listing the old models.  Get the new models and run them through the update processor, which is already used by the syncer, to convert them to v1 KVs.  Then extract the KVs as we used to.

I needed to rev libcalico-go to pick up some fixes that came in overnight (the etcdv3 key templates were changed).  I've manually kicked the tyres on this by using calicoctl to write some config and checking felix could load it but it could probably do with an FV.

## Todos
- [ ] Documentation
- [x] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/felix.go
+++ b/felix.go
@@ -53,9 +53,12 @@ import (
 	"github.com/projectcalico/felix/rules"
 	"github.com/projectcalico/felix/statusrep"
 	"github.com/projectcalico/felix/usagerep"
+	apiv2 "github.com/projectcalico/libcalico-go/lib/apis/v2"
 	"github.com/projectcalico/libcalico-go/lib/backend"
 	bapi "github.com/projectcalico/libcalico-go/lib/backend/api"
 	"github.com/projectcalico/libcalico-go/lib/backend/model"
+	"github.com/projectcalico/libcalico-go/lib/backend/syncersv1/updateprocessors"
+	errors2 "github.com/projectcalico/libcalico-go/lib/errors"
 	"github.com/projectcalico/libcalico-go/lib/health"
 	"github.com/projectcalico/typha/pkg/syncclient"
 )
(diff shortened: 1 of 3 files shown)


felix.go
```diff
@@ -647,51 +655,98 @@ func monitorAndManageShutdown(failureReportChan <-chan string, driverCmd *exec.C
...
+func convertV2ConfigToMap(configType string, v2Config *model.KVPair) (map[string]string, error) {
+	logCxt := log.WithField("type", configType)
+	if v2Config == nil {
+		logCxt.Info("No config of this type")
+		return nil, nil
+	}
+	// Re-use the update processor logic implemented for hte syncer.  We give it a v2 config
+	// object in a KVPair and it uses the annotations defined on it to split it into v1-style
+	// KV pairs.
+	configConvertor := updateprocessors.NewFelixConfigUpdateProcessor()
```
neiljerram: nit: I think the spelling in that case is normally "converter"

felix.go
```diff
@@ -647,51 +655,98 @@ func monitorAndManageShutdown(failureReportChan <-chan string, driverCmd *exec.C
...
+	v1kvs, err := configConvertor.Process(v2Config)
```
neiljerram: Does this encapsulate the logic for _not_ emitting an update for a particular key/value that is not set in the FelixConfiguration?  (As an aside, I wasn't sure how "not set" is represented for string values.  I got that nil pointers are used for the other cases.)

fasaxc: I believe they use an empty string for "not set".  In that case it emits a KVPair with a nil `Value` field.  (The Value field is an `interface{}` so we're talking about a nil interface rather than a  nil pointer inside the interface.)

neiljerram:
Looks like you'll need a Typha bump as well, to match the move of the libcalico-go pin.

PR #1576, opened 2017-10-21
re-enable k8sfv tests, bump libcalico-go

bcreane:
## Description
re-enable k8sfv tests, bump libcalico-go

## Todos

## Release Note

```release-note
None required
```

### DIFF
--- a/k8sfv/k8sfv.go
+++ b/k8sfv/k8sfv.go
@@ -27,9 +27,6 @@ import (
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/client-go/tools/clientcmd/api"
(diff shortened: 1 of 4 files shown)


k8sfv/k8sfv.go
```diff
@@ -158,7 +158,7 @@ func initialize(k8sServerEndpoint string) (clientset *kubernetes.Clientset) {
 
 func initializeCalicoDeployment(k8sServerEndpoint string) {
```
gunjan5: why cant we just get rid of this entire function? We don't need any initialization anymore, and the client created is not used (as it's `_`'d out)

bcreane: yep, good suggestion, will do.

PR #1578, opened 2017-10-22
[WIP] Query v2 Node resource to get its config

caseydavenport:
## Description

In the v2 data model, there is some information that Felix treats as config that is represented on the Node resource - namely the IpInIpTunnelAddr for this node.

This PR adds a call to the node object and merges it with other node config.

Not sure what tests need to be updated here.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/felix.go
+++ b/felix.go
@@ -683,6 +683,11 @@ func loadConfigFromDatastore(
 		Name:      "node." + hostname,
 		Namespace: "",
 	}, "")
+	n, err := client.Get(ctx, model.ResourceKey{
+		Kind:      apiv2.KindNode,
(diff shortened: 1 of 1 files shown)


felix.go
```diff
@@ -683,6 +683,11 @@ func loadConfigFromDatastore(
 		Name:      "node." + hostname,
 		Namespace: "",
```
robbrockbank: Should be checking for error here

robbrockbank:
Closing in favor of #1579

PR #1582, opened 2017-10-23
Felix v3 config

neiljerram:
## Description
Change Felix's default datastore type to etcdv3.  Also enhance how Felix works out its datastore config so that it works with both CALICO and FELIX environment variables.

### DIFF
--- a/config/config_params.go
+++ b/config/config_params.go
@@ -90,7 +90,7 @@ type Config struct {
 	UseInternalDataplaneDriver bool   `config:"bool;true"`
 	DataplaneDriver            string `config:"file(must-exist,executable);calico-iptables-plugin;non-zero,die-on-fail,skip-default-validation"`
 
-	DatastoreType string `config:"oneof(kubernetes,etcdv2,etcdv3);etcdv2;non-zero,die-on-fail"`
+	DatastoreType string `config:"oneof(kubernetes,etcdv3);etcdv3;non-zero,die-on-fail"`
 
 	FelixHostname string `config:"hostname;;local,non-zero"`
 
@@ -357,9 +357,9 @@ func (config *Config) resolve() (changed bool, err error) {
 }
 
 func (config *Config) DatastoreConfig() apiconfig.CalicoAPIConfig {
-	// Special case for etcdv2 datastore, where we want to honour established Felix-specific
+	// Special case for etcdv3 datastore, where we want to honour established Felix-specific
 	// config mechanisms.
-	if config.DatastoreType == "etcdv2" {
+	if config.DatastoreType == "etcdv3" {
 		// Build a CalicoAPIConfig with the etcd fields filled in from Felix-specific
 		// config.
 		var etcdEndpoints string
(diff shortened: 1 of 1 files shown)


config/config_params.go
```diff
@@ -356,45 +356,50 @@ func (config *Config) resolve() (changed bool, err error) {
...
+	return set
+}
 
-	// Build CalicoAPIConfig from the environment.  This means that any XxxYyy field in
-	// CalicoAPIConfigSpec can be set by a corresponding XXX_YYY or CALICO_XXX_YYY environment
-	// variable, and that the datastore type can be set by a DATASTORE_TYPE or
-	// CALICO_DATASTORE_TYPE variable.  (Except in the etcdv2 case which is handled specially
-	// above.)
+func (config *Config) explicitlySet(name string) bool {
+	return config.setBy(name, ConfigFile) || config.setBy(name, EnvironmentVariable)
```
fasaxc: I feel like the content of this method doesn't really do what I'd expect form the name (how is a datastore setting less explicit than an environment variable setting?), how about searching all the config types?
```
for _, rawConfig := range config.sourceToRawConfig {
    if _, set := rawConfig[name] {
        return true
    }
}
return false
```

config/config_params.go
```diff
@@ -356,45 +356,50 @@ func (config *Config) resolve() (changed bool, err error) {
...
 		log.WithError(err).Panic("Failed to create datastore config")
 	}
-	// If that didn't set the datastore type (in which case the field will have been set to its
-	// default 'etcdv2' value), copy it from the Felix config.
-	if cfg.Spec.DatastoreType == "etcdv2" {
-		cfg.Spec.DatastoreType = apiconfig.DatastoreType(config.DatastoreType)
+
+	// Now allow FELIX_XXXYYY variables to override that, when they are explicitly set, in the
+	// etcd case.
+	if config.explicitlySet("DatastoreType") && config.DatastoreType == "etcdv3" {
```
fasaxc: Any chance we'll reintroduce etcdv2 support as we head to GA?  Maybe we should match on etcdv2 or etcdv3?

config/config_params.go
neiljerram: Well,

1. Wouldn't that include the `Default` source, which I definitely don't want to include?

2. For these specific settings here, it's now not possible for them to have datastore settings.  (i.e. those have been removed from the FelixConfiguration object).

For these particular settings, being set by config file or environment are definitely the two conditions that we are looking for.  Perhaps I just need to improve the naming so as to better reflect that?

config/config_params.go
neiljerram: I'd be happy to do that if we also still had 'etcdv2' as a valid value for DatastoreType.  I think we should leave adding `|| config.DatastoreType == 'etcdv2'` here until the same time as we reinstate 'etcdv2' as a valid value - and I think we would be naturally prompted for that, if we also reinstated any etcdv2 testing, or added etcdv2 with similar config as we have for etcdv3.  WDYT?

config/config_params.go
fasaxc: Yes, could equally be solved with naming.  I don't think we store Default config in that map but I could be wrong.

config/config_params.go
fasaxc: SGTM

PR #1583, opened 2017-10-23
Update libcalico and corresponding label / profile names

caseydavenport:
## Description

This updates to the latest libcalico, which uses slightly modified names for k8s profile's and namespace labels.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/fv/ingress_egress_test.go
+++ b/fv/ingress_egress_test.go
@@ -49,10 +49,10 @@ var _ = Context("with initialized Felix, etcd datastore, 3 workloads", func() {
 		defaultProfile := api.NewProfile()
 		defaultProfile.Name = "default"
 		defaultProfile.Spec.LabelsToApply = map[string]string{"default": ""}
-		defaultProfile.Spec.EgressRules = []api.Rule{{Action: "allow"}}
+		defaultProfile.Spec.EgressRules = []api.Rule{{Action: api.Allow}}
 		defaultProfile.Spec.IngressRules = []api.Rule{{
-			Action: "allow",
-			Source: api.EntityRule{Tag: "default"},
+			Action: api.Allow,
+			Source: api.EntityRule{Selector: "default == ''"},
 		}}
 		_, err := client.Profiles().Create(utils.Ctx, defaultProfile, utils.NoOptions)
 		Expect(err).NotTo(HaveOccurred())
@@ -97,7 +97,7 @@ var _ = Context("with initialized Felix, etcd datastore, 3 workloads", func() {
 			policy.Namespace = "fv"
 			policy.Name = "policy-1"
 			allowFromW1 := api.Rule{
-				Action: "allow",
+				Action: api.Allow,
 				Source: api.EntityRule{
 					Selector: w[1].NameSelector(),
 				},
@@ -123,7 +123,7 @@ var _ = Context("with initialized Felix, etcd datastore, 3 workloads", func() {
 			policy.Namespace = "fv"
 			policy.Name = "policy-1"
 			allowToW1 := api.Rule{
-				Action: "allow",
+				Action: api.Allow,
 				Destination: api.EntityRule{
 					Selector: w[1].NameSelector(),
 				},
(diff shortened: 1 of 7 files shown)


fv/named_ports_test.go
```diff
@@ -719,8 +719,8 @@ var _ = Describe("with a simulated kubernetes nginx and client", func() {
 			"tcp",
 		)
 		nginx.WorkloadEndpoint.Labels = map[string]string{
```
robbrockbank: and orchestrator

robbrockbank: oh hang on.  Is this using the main clientv2 or the backend client.

Assuming the former then we should remove the namespace label.
If the latter then we may need to add the orchestrator label as well.

caseydavenport: It's using the main v2 client, so we should just remove.

fv/pre_dnat_test.go
```diff
@@ -122,16 +122,15 @@ var _ = Context("with initialized Felix, etcd datastore, 2 workloads", func() {
...
 				policy.Name = "deny-ingress"
 				order := float64(20)
 				policy.Spec.Order = &order
 				policy.Spec.PreDNAT = true
 				policy.Spec.ApplyOnForward = true
-				policy.Spec.IngressRules = []api.Rule{{Action: "deny"}}
+				policy.Spec.IngressRules = []api.Rule{{Action: api.Deny}}
 				policy.Spec.Selector = "has(host-endpoint)"
-				_, err := client.NetworkPolicies().Create(utils.Ctx, policy, utils.NoOptions)
+				_, err := client.GlobalNetworkPolicies().Create(utils.Ctx, policy, utils.NoOptions)
```
caseydavenport: Need to switch to GlobalNetworkPolicy for these, because a NP can't select host endpoints since they are not a namespaced resource.

fv/ingress_egress_test.go
```diff
@@ -49,10 +49,10 @@ var _ = Context("with initialized Felix, etcd datastore, 3 workloads", func() {
 		defaultProfile := api.NewProfile()
 		defaultProfile.Name = "default"
 		defaultProfile.Spec.LabelsToApply = map[string]string{"default": ""}
-		defaultProfile.Spec.EgressRules = []api.Rule{{Action: "allow"}}
+		defaultProfile.Spec.EgressRules = []api.Rule{{Action: api.Allow}}
 		defaultProfile.Spec.IngressRules = []api.Rule{{
-			Action: "allow",
-			Source: api.EntityRule{Tag: "default"},
+			Action: api.Allow,
+			Source: api.EntityRule{Selector: "default == ''"},
```
robbrockbank: `has(default)` i think is the preferred way of doing this

caseydavenport: Figured this way you don't match say, `default: whatever` that might be on some other pod for some reason - it strictly matches the profile's label of `default: ""`

fasaxc: Either should be fine; it's up to the user really.  Probably good to have a mix in the tests.

k8sfv/namespace.go
```diff
@@ -105,21 +105,14 @@ func createNetworkPolicy(clientset *kubernetes.Clientset, namespace string) {
...
-			PodSelector: metav1.LabelSelector{
-				MatchLabels: map[string]string{"calico/k8s_ns": namespace},
-			},
+			// An empty PodSelector selects all pods in this Namespace.
+			PodSelector: metav1.LabelSelector{},
 			Ingress: []networkingv1.NetworkPolicyIngressRule{
 				networkingv1.NetworkPolicyIngressRule{
-					Ports: []networkingv1.NetworkPolicyPort{
-						networkingv1.NetworkPolicyPort{},
-					},
```
neiljerram: Does not specifying Ports at all here, have the same meaning as specifying a single empty Port struct?

caseydavenport: Yep, both cases mean "don't restrict based on port", as far as the k8s NP API is concerned.

PR #1564, opened 2017-09-27
Allow override of etcd image in FVs

ozdanborne:
## Description

Docker images used in FVs are currently hardcoded to x86 images. Users building and testing for different architectures have a hard time running them unless they can override the image names at runtime. 

See community contribution for similar functionality at https://github.com/projectcalico/kube-controllers/pull/153 and https://github.com/projectcalico/kube-controllers/issues/150

Currently, policy controller FVs use their own `RunEtcd` function. In order to use the upstream mainline and maintain cross-architecture compatibility, this PR is needed.

## Todos
- [ ] Unit tests (full coverage)
- [ ] Integration tests (delete as appropriate) In plan/Not needed/Done
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

PR #1594, opened 2017-10-30
Allow override of Typha and Hyperkube images in FVs

djlwilder:
To override hard-coded image names set environment variables:
K8_IMAGE and/or TYPHA_IMAGE.

## Description
This PR works together with PR #1564 to select image names for functional verifications.  This is required to make fv arch agnostic.  The default image names are unchanged.
- which components are affected by this PR
-->  fv

## Todos
Set environment variables in the Makefile to override image names of non amd64 arch builds.  
Since this patch is not changing the default image names this patch is not dependent on the Makefile change.  I will submit a  separate PR for the Makefile change.

## Release Note
```release-note
None required
```

### DIFF
--- a/fv/health_test.go
+++ b/fv/health_test.go
@@ -41,9 +41,9 @@ import (
 	"math/rand"
 	"net"
 	"net/http"
+	"os"
 	"time"
 
-	"github.com/kelseyhightower/envconfig"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	"github.com/onsi/gomega/types"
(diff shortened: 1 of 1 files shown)


fasaxc:
Unless I'm missing something, I think the code (using the `envconfig` library) you removed was _already_ loading environment variables for that purpose.  Is there a reason the existing code didn't work for ppc64?

neiljerram:
Agreed, I'd prefer to keep EnvConfig  and meet the need by adding fields to the Config struct.

We need to merge #1596 first, then this PR could be reworked to do that.

djlwilder:
Agreed,  I will test #1596 on ppc,  then do the rework on this PR.

djlwilder:
Closing this PR,  continuing in PR#1599.

PR #1596, opened 2017-10-31
Allow override of etcd image in FVs

neiljerram:
Continuation of #1564


neiljerram:
@djlwilder Please could you check that this version works for your PPC64 use case?

neiljerram:
(The variable to set is ETCD_IMAGE or FV_ETCD_IMAGE)

djlwilder:
@neiljerram I verified your change works for the ppc test case.  LGTM
BTW: I found the environment variable to set is FV_ETCDIMAGE.
After this is merged, I can re-work #1594 to to do the same for the Typha and HyperKube image.

Thanks

PR #1598, opened 2017-10-31
Rev go build to v0.9

fasaxc:
## Description

- Pull in go 1.9.2.
- Close some CVEs.
- Remove monotime dependency since go 1.9's time implementation has monotonic comparisons.

## Release Note

```release-note
Felix is now built against go 1.9.2.
```

### DIFF
--- a/intdataplane/int_dataplane.go
+++ b/intdataplane/int_dataplane.go
@@ -696,7 +696,7 @@ func (d *InternalDataplane) loopUpdatingDataplane() {
 
 				// Record stats.
 				applyTime := time.Since(applyStart)
-				summaryApplyTime.Observe(time.Since(applyStart).Seconds())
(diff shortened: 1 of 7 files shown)


intdataplane/int_dataplane.go
```diff
@@ -690,14 +689,14 @@ func (d *InternalDataplane) loopUpdatingDataplane() {
...
+				applyStart := time.Now()
 
 				// Actually apply the changes to the dataplane.
 				d.apply()
 
 				// Record stats.
-				applyTime := monotime.Since(applyStart)
-				summaryApplyTime.Observe(applyTime.Seconds())
+				applyTime := time.Since(applyStart)
+				summaryApplyTime.Observe(time.Since(applyStart).Seconds())
```
neiljerram: Do we need repeat of `time.Since(applyStart)` ? (as opposed to using `applyTime`)

PR #1599, opened 2017-11-01
Allow override of Typha and Hyperkube images in FVs.

djlwilder:
Continuation of PR #1594 

This PR allows for the typha and hyperkube image names and versions to be overwritten for fv.  Defaults stay the same.    Image names/version can be selected in the Makefile according to the value of ARCH.  See also PR:  #1564.

## Description
This is required for multi-arch support. 
I verified the changes by running fv on both amd64 and ppc64le.

## Release Note
```release-note
None required
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,21 +45,22 @@
 # The build architecture is select by setting the ARCH variable.
 # For example: When building on ppc64le you could use ARCH=ppc64le make <....>.
 # When ARCH is undefined it defaults to amd64.
-ifdef ARCH
-	ARCHTAG:=-$(ARCH)
-endif
 ARCH?=amd64
-ARCHTAG?=
-
 ifeq ($(ARCH),amd64)
-GO_BUILD_VER:=v0.9
+	ARCHTAG?=
+	GO_BUILD_VER?=v0.9
+	FV_TYPHAIMAGE?=calico/typha:v0.5.1-27-g49eaa9b
 endif
 
 ifeq ($(ARCH),ppc64le)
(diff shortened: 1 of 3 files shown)


djlwilder:
@fasaxc  @neiljerram this PR is a continuation of PR #1594

neiljerram:
Sorry for the delay on this, @djlwilder. I'm planning a proper review tomorrow morning.

neiljerram:
@djlwilder There are a few extra concepts bundled into this PR, which make it less of a no-brainer to review than if it was just allowing the images to be overridden:

1. Allowing GO_BUILD_VER to be overriden by the make caller's environment.

2. Moving where defaults are specified - in the amd64 case - from the FV code to the Makefile.

3. Incorporating image *version* config/override into the overall *image* specification.

I'm happy that (3) makes sense (because I believe that tags may be scoped to the relevant hub), and (1) seems harmless enough, but still thinking a bit about (2).

djlwilder:
@neiljerram thanks for the review.
I feel the values for all archs should be specified the same place,  to me the Makefile is the best spot. 
 If not overwritten in the Makefile the FV code is still supplying a default value for the etcd, typha and K8 images and those are x86 defaults.  This is bad as both spots would need to be updated if we want to change the x86 values.   I think I should eliminate the defaults in the FV code.

djlwilder:
@neiljerram what are your thoughts.  Should I keep the x86 defaults in the FV code and remove it from the Makefile?   Then just use the makefile to overwrite the default for other arch builds?

PR #1613, opened 2017-11-09
Wait for stderr after an FV workload failure.

fasaxc:
## Description

Should give us better diags in the event of a failure.

### DIFF
--- a/fv/workload/workload.go
+++ b/fv/workload/workload.go
@@ -107,11 +107,11 @@ func Run(c *containers.Container, name, interfaceName, ip, ports string, protoco
 	// Read the workload's namespace path, which it writes to its standard output.
 	stdoutReader := bufio.NewReader(w.outPipe)
 	stderrReader := bufio.NewReader(w.errPipe)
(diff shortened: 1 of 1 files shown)


fv/workload/workload.go
```diff
@@ -121,6 +121,16 @@ func Run(c *containers.Container, name, interfaceName, ip, ports string, protoco
 			log.Infof("Workload %s stderr: %s", name, strings.TrimSpace(string(line)))
 		}
 	}()
+
+	namespacePath, err := stdoutReader.ReadString('\n')
+	if err != nil {
+		// (Only) if we fail here, wait for the stderr to be output before returning.
+		defer errDone.Wait()
```
neiljerram: Just wondering what difference it makes to use `defer` here, instead of directly calling `errDone.Wait()` ?

fasaxc: Think the idea was that failing the test first makes it easier to diagnose if the Wait() call hangs.  Pretty spurious though!

PR #1612, opened 2017-11-09
bump libcalico-go

bcreane:
## Todos
- [ ] Unit tests (full coverage)
- [ ] Integration tests (delete as appropriate) In plan/Not needed/Done
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

PR #1614, opened 2017-11-09
Bump libcalico-go

neiljerram:
Like #1612 but without the go-sockaddr change in glide.lock.


bcreane:
Hey @neiljerram, @fasaxc , thanks for fixing up/replacing the libcalico-go PR. Agree with Shaun that an explicit pin is a good idea.

neiljerram:
Thanks @bcreane ; I missed @fasaxc 's comment earlier, but will add that explicit pin now.

PR #1620, opened 2017-11-13
k8sfv.test updated to use image definitions from the Makefile.

djlwilder:
Container versions previously defined in k8sfv.test are replace by
image names:tags as defined in the Makefile with the variables:
FV_ETCDIMAGE, FV_TYPHAIMAGE and FV_K8SIMAGE.

## Description
Currently the etcd, typha and hyperkube images used by the k8sfv-test are defined in the k8sfv-test script and in the case of amd64 are different than the version used during fv. This patch moves these definitions into the Makefile by using the already existing environment variables FV_TYPHAIMAGE, FV_ETCDIMAGE, FV_K8SIMAGE.  This provides two advantages:

- The same images are used in both fv and k8sfv-test.
- The images names and tags are defined for the specific architecture under test.

Note: For amd64 the default version of hyperkube changes from v1.7.3 to 1.7.5. The default version of TYPHA changes from "latest" to v0.6.0-alpha1-17-gc6c5726 (in the current Makefile).

## Release Note

```release-note
None required
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -192,7 +192,10 @@ k8sfv-test: calico/felix k8sfv-test-existing-felix
 # container image.  To use some existing Felix version other than
 # 'latest', do 'FELIX_VERSION=<...> make k8sfv-test-existing-felix'.
 k8sfv-test-existing-felix: bin/k8sfv.test
-	TYPHA_VERSION=$(TYPHA_VERSION) k8sfv/run-test
+	FV_ETCDIMAGE=$(FV_ETCDIMAGE) \
+	FV_TYPHAIMAGE=$(FV_TYPHAIMAGE) \
+	FV_K8SIMAGE=$(FV_K8SIMAGE) \
+	k8sfv/run-test
 
 PROMETHEUS_DATA_DIR := $$HOME/prometheus-data
 K8SFV_PROMETHEUS_DATA_DIR := $(PROMETHEUS_DATA_DIR)/k8sfv
--- a/Makefile
+++ b/Makefile
@@ -49,15 +49,13 @@ ARCH?=amd64
 ifeq ($(ARCH),amd64)
 	ARCHTAG?=
 	GO_BUILD_VER?=v0.9
-	TYPHA_VERSION?=v0.6.0-alpha1-17-gc6c5726
-	FV_TYPHAIMAGE?=calico/typha:$(TYPHA_VERSION)
+	FV_TYPHAIMAGE?=calico/typha:v0.6.0-alpha1-17-gc6c5726
 endif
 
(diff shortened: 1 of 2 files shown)


djlwilder:
I am not sure why fv failed.  Runs ok for me.  Can fv be re-run?

tmjd:
I kicked the test off again

djlwilder:
thanks @tmjd ,  it failed again :(   I will look closer and see if I can figure it out.

tmjd:
I kicked it off again because I thought maybe some updates that have gone into felix might fix this (because I didn't see how your changes would have caused a problem) and it passed.

djlwilder:
Cool,  thanks.

Makefile
```diff
@@ -192,7 +192,10 @@ k8sfv-test: calico/felix k8sfv-test-existing-felix
 # container image.  To use some existing Felix version other than
 # 'latest', do 'FELIX_VERSION=<...> make k8sfv-test-existing-felix'.
 k8sfv-test-existing-felix: bin/k8sfv.test
-	TYPHA_VERSION=$(TYPHA_VERSION) k8sfv/run-test
```
neiljerram: Please also now retire the remaining uses of TYPHA_VERSION in the Makefile.

k8sfv/run-test
```diff
@@ -10,7 +10,7 @@
 : ${FELIX_VERSION:=latest}
 #
 # What version of the k8s API server to use.
-: ${K8S_VERSION:=1.7.3}
+# Update: See FV_K8SIMAGE in the Makefile for version information.
```
neiljerram: I think we should either (1) continue allowing `run-test` to have sensible defaults of its own, or (2) make clear that it requires these variables to be set in the environment.

(1) would mean something like `: ${FV_K8SIMAGE:=gcr.io/google_containers/hyperkube-amd64:v1.7.5}` here

(2) would mean something like

```
if ! test -n "${FV_K8SIMAGE}"; then
    echo FV_K8SIMAGE must be set to indicate the hyperkube image to use
    exit -1
fi
```

(and, in either case, similarly for the other variables affected by this change)

WDYT?  (I think I slighly prefer (2), to avoid duplicating the default values.)

djlwilder:
Agreed @neiljerram,  I made both your suggested changes.

neiljerram:
Guessing that the FV failure is a timing flake.  I'll kick off a rebuild now to see if it happens again.

(`[Fail] health tests with Felix and Typha running after removing iptables-restore [It] should become unready, then die `)

neiljerram:
CI rebuild passed, so merging this now.

PR #1631, opened 2017-11-21
2.6.x: Handle config changes for ClusterGUID, ClusterType, CalicoVersion.

fasaxc:
## Description

- Add whitelist set for config changes that we can handle.
- Implement support for updating those fields.  (Little bit of work but I'm very keen to maintain the invariant that restarting Felix should be a graceful no-op.)

## Todos
- [x] Documentation
- [x] Backport
- [x] Release note

## Release Note

```release-note
In order to make upgrade from 2.6.x to 3.0 smoother, Felix now handles certain config changes without restarting.
```

### DIFF
--- a/config/config_params.go
+++ b/config/config_params.go
@@ -163,10 +163,12 @@ type Config struct {
 	FailsafeInboundHostPorts  []ProtoPort `config:"port-list;tcp:22,udp:68;die-on-fail"`
 	FailsafeOutboundHostPorts []ProtoPort `config:"port-list;tcp:2379,tcp:2380,tcp:4001,tcp:7001,udp:53,udp:67;die-on-fail"`
 
-	UsageReportingEnabled bool   `config:"bool;true"`
-	ClusterGUID           string `config:"string;baddecaf"`
-	ClusterType           string `config:"string;"`
-	CalicoVersion         string `config:"string;"`
+	UsageReportingEnabled          bool          `config:"bool;true"`
+	UsageReportingInitialDelaySecs time.Duration `config:"seconds;300"`
+	UsageReportingIntervalSecs     time.Duration `config:"seconds;86400"`
+	ClusterGUID                    string        `config:"string;baddecaf"`
+	ClusterType                    string        `config:"string;"`
+	CalicoVersion                  string        `config:"string;"`
 
 	DebugMemoryProfilePath          string        `config:"file;;"`
 	DebugDisableLogDropping         bool          `config:"bool;false"`
--- a/felix.go
+++ b/felix.go
@@ -405,11 +405,12 @@ configRetry:
 			}
 		}()
 
-		go usagerep.PeriodicallyReportUsage(
+		usageRep := usagerep.New(
 			24*time.Hour,
 			statsChanOut,
 			configUpdChan,
 		)
+		go usageRep.PeriodicallyReportUsage(context.Background())
 	} else {
 		// Usage reporting disabled, but we still want a stats collector for the
 		// felix_cluster_* metrics.  Register a no-op function as the callback.
--- a/felix.go
+++ b/felix.go
@@ -321,11 +321,13 @@ configRetry:
 	// Initialise the glue logic that connects the calculation graph to/from the dataplane driver.
 	log.Info("Connect to the dataplane driver.")
 	failureReportChan := make(chan string)
-	var configUpdChan chan map[string]string
+	var connToUsageRepUpdChan chan map[string]string
 	if configParams.UsageReportingEnabled {
-		configUpdChan = make(chan map[string]string, 1)
+		// Make a channel for the connector to use to send updates to the usage reporter.
+		// (Otherwise, we pass in a nil channel, which disables such updates.)
+		connToUsageRepUpdChan = make(chan map[string]string, 1)
 	}
-	dpConnector := newConnector(configParams, configUpdChan, datastore, dpDriver, failureReportChan)
+	dpConnector := newConnector(configParams, connToUsageRepUpdChan, datastore, dpDriver, failureReportChan)
 
 	// Now create the calculation graph, which receives updates from the
 	// datastore and outputs dataplane updates for the dataplane driver.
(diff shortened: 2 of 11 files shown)


bcreane:
@fasaxc - is this just awaiting @neiljerram's lgtm?

felix.go
```diff
@@ -321,7 +321,11 @@ configRetry:
 	// Initialise the glue logic that connects the calculation graph to/from the dataplane driver.
 	log.Info("Connect to the dataplane driver.")
 	failureReportChan := make(chan string)
-	dpConnector := newConnector(configParams, datastore, dpDriver, failureReportChan)
+	var configUpdChan chan map[string]string
+	if configParams.UsageReportingEnabled {
+		configUpdChan = make(chan map[string]string, 1)
+	}
+	dpConnector := newConnector(configParams, configUpdChan, datastore, dpDriver, failureReportChan)
```
neiljerram: I first read this as "creating a connector object with an initial set of config, and with a channel so that we can also pass config changes to the connector".  Whereas in fact I think the channel is for the connector to pass config changes to *its* downstream(s).

Given that, would it be clearer for the channel creation to move inside newConnector?  It might also need moving the whole of the usagerep setup inside newConnector - but perhaps that would make sense too?

Alternatively, perhaps just a comment here to explain.

usagerep/usagerep.go
```diff
@@ -28,39 +27,104 @@ import (
...
+		if config == nil {
+			return
+		}
+		if initialDelayStarted {
+			return
+		}
+
+		// To avoid thundering herd, inject some startup jitter.
+		initialDelay := u.calculateInitialDelay(stats.NumHosts)
+		func() {
```
neiljerram: Should this be in a goroutine?  Otherwise looks like this will block the main loop below.

felix.go
```diff
@@ -401,13 +405,12 @@ configRetry:
...
 		}()
 
-		go usagerep.PeriodicallyReportUsage(
+		usageRep := usagerep.New(
 			24*time.Hour,
-			configParams.ClusterGUID,
-			configParams.ClusterType,
-			configParams.CalicoVersion,
 			statsChanOut,
+			configUpdChan,
```
neiljerram: Is it possible that Felix reads ClusterType etc. in its initial read from the datastore (`loadConfigFromDatastore`), and that those settings *never* change after that, and so there are never any updates through `configUpdChan`?

Would the usagerep still report correctly, in that case?

usagerep/usagerep.go
fasaxc: Yes!

felix.go
fasaxc: The calculation graph doesn't report deltas from the initial load of config, it reports all the config form the initial snapshot so it should send the initial update.

felix.go
fasaxc: Added a comment and made the naming clearer.

felix.go
neiljerram: Ah right - fine.

fv/containers/containers.go
```diff
@@ -204,6 +222,17 @@ func (c *Container) ExecMayFail(cmd ...string) error {
...
 }
 
+func (c *Container) ExecOutput(args ...string) (string, error) {
+	arg := []string{"exec", c.Name}
+	arg = append(arg, args...)
+	cmd := exec.Command("docker", arg...)
+	out, err := cmd.Output()
+	if err != nil {
+		return "", err
+	}
```
neiljerram: I'd expect it often to be useful to return the output too, when an error occurs.

usagerep/usagerep_test.go
```diff
@@ -21,13 +21,180 @@ import (
...
+		var err error
+		tcpListener, err = net.Listen("tcp", ":0")
+		Expect(err).NotTo(HaveOccurred())
+		httpHandler = &requestRecorder{}
+		go func() {
+			defer GinkgoRecover()
+			http.Serve(tcpListener, httpHandler)
+		}()
+
+		// Channels to sent data to the UsageReporter.
```
neiljerram: nit, sent -> send

usagerep/usagerep_test.go
```diff
@@ -21,13 +21,180 @@ import (
...
+					NumHostEndpoints:     2,
+					NumWorkloadEndpoints: 3,
+				}
+			})
+
+			It("should do first check ins correctly", func() {
+				By("checking in within 2s")
+				startTime := time.Now()
+				Eventually(httpHandler.GetRequestURIs, "2s", "100ms").Should(HaveLen(1))
+				By("waiting until after the intial delay")
```
neiljerram: nit, "initial"

config/config_params.go
```diff
@@ -163,10 +163,12 @@ type Config struct {
 	FailsafeInboundHostPorts  []ProtoPort `config:"port-list;tcp:22,udp:68;die-on-fail"`
 	FailsafeOutboundHostPorts []ProtoPort `config:"port-list;tcp:2379,tcp:2380,tcp:4001,tcp:7001,udp:53,udp:67;die-on-fail"`
 
-	UsageReportingEnabled bool   `config:"bool;true"`
-	ClusterGUID           string `config:"string;baddecaf"`
-	ClusterType           string `config:"string;"`
-	CalicoVersion         string `config:"string;"`
+	UsageReportingEnabled          bool          `config:"bool;true"`
+	UsageReportingInitialDelaySecs time.Duration `config:"seconds;300"`
+	UsageReportingIntervalSecs     time.Duration `config:"seconds;86400"`
```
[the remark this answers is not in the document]
fasaxc: Ugh, probably

fv/config_update_test.go
```diff
@@ -0,0 +1,175 @@
...
+
+		start := time.Now()
+		for {
+			pids := getFelixPIDs()
+			Expect(pids).NotTo(BeEmpty())
+			if len(pids) == 1 {
+				felixInitialPID = pids[0]
+				break
+			}
+			Expect(time.Since(start)).To(BeNumerically("<", time.Second))
```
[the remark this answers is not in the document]
fasaxc: SGTM

fv/config_update_test.go
```diff
@@ -0,0 +1,175 @@
...
+			felix.Exec("ip", "r")
+		}
+
+		for ii := range w {
+			w[ii].Stop()
+		}
+		felix.Stop()
+
+		if CurrentGinkgoTestDescription().Failed {
+			etcd.Exec("etcdctl", "ls", "--recursive", "/")
```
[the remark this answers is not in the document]
fasaxc: This is a 2.6.x PR so that only needs to be fixed when this is merged to master.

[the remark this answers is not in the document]
neiljerram: Oh yes of course - d'oh.

PR #1633, opened 2017-11-27
Fix config batcher log spam

fasaxc:
## Description

- Unset the dirty flag after sending an update.
- Avoid info-level logs for no-op updates.

## Todos
- [ ] Unit tests (full coverage)
- [ ] Integration tests (delete as appropriate) In plan/Not needed/Done
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

PR #1634, opened 2017-11-28
Fix config batcher: unset dirty flag after flush.

fasaxc:
## Description

Forward port of #1633 

- Reset the config batcher's dirty flag after sending updates.
- Tune logs to avoid Info-level spam for no-op updates.
- Add UT for ConfigBatcher.

## Todos
- [x] Documentation
- [x] Backport
- [x] Release note

## Release Note

```release-note
Fix log spam from Felix caused by no-op config changes.
```


fasaxc:
Grr, it repro'd locally after running all the tests but not when I run that one on its own...

PR #1645, opened 2017-12-04
Fix that felix didn't restart when config was deleted.

### LINKED ISSUE #1644: Felix does not handle removal of IP-in-IP tunnel address for its node
Expected: Felix should detect a config change and restart; then the new Felix should remove the previously assigned address from the tunnel device.

Observed: address persists on the tunnel device for at least 15 seconds after the relevant IPPool was disabled.

fasaxc:
## Description

Fixes #1644; needs backport to 2.6.x

## Todos
- [x] Documentation
- [x] Backport
- [x] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/fv/config_update_test.go
+++ b/fv/config_update_test.go
@@ -35,32 +35,35 @@ import (
 var _ = Context("Config update tests, after starting felix", func() {
 
 	var (
-		etcd            *containers.Container
-		felix           *containers.Container
-		felixInitialPID int
-		client          client.Interface
-		w               [3]*workload.Workload
+		etcd     *containers.Container
+		felix    *containers.Container
+		felixPID int
+		client   client.Interface
+		w        [3]*workload.Workload
 	)
 
 	getFelixPIDs := func() []int {
 		return felix.GetPIDs("calico-felix")
 	}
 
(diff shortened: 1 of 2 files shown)


fv/config_update_test.go
```diff
@@ -35,32 +35,35 @@ import (
...
 		for {
 			pids := getFelixPIDs()
-			Expect(pids).NotTo(BeEmpty())
 			if len(pids) == 1 {
-				felixInitialPID = pids[0]
+				felixPID = pids[0]
 				break
 			}
 			Expect(time.Since(start)).To(BeNumerically("<", time.Second))
+			time.Sleep(50 * time.Millisecond)
```
neiljerram: Logic here doesn't feel quite right to me.  Shouldn't it be 'wait until there's 1 Felix PID that is stable for 0.5s' ?  As it stands, I think the code can be fooled by seeing a PID of a Felix on the verge of restarting.

fasaxc: That's OK, I think: we want to get the old PID, then the tests use that to see whether felix restarted or not.

When we call this update method in the new test, it's after waiting for the old PID to disappear.

PR #1648, opened 2017-12-13
Change the version regex to require a "v"

fasaxc:
For consistency with other repos

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/utils/lib.sh
+++ b/utils/lib.sh
@@ -55,7 +55,7 @@ function validate_version {
     version=$1
 
     # We allow.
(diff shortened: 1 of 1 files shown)


caseydavenport:
@fasaxc can we merge this?

fasaxc:
@caseydavenport Is this code still used after your recent changes?

caseydavenport:
@fasaxc No I don't think that it is.

I'm not against adding the check back in though if we now want to check it _does_ have a `v`. 

No other repos have it, but no other repos have the history of _not_ using a `v` so there might be a period of confusion.

PR #1650, opened 2017-12-14
FV infrastructure enhancements

neiljerram:
## Description
This PR contains enhancements to the FV infrastructure that I made while running on upgrade scale testing.  We're not going to merge the upgrade scale test code itself, because there's no continuing value of having that running all the time in our CI (and making each CI take longer), but the enhancements are valuable, and worth merging to v2.6.x-series as well as to master so as to avoid divergence between the FV infrastructure in those two branches.

This PR is for merging to v2.6.x-series.  Once that's agreed I'll follow up with a similar PR for master.

### DIFF
--- a/fv/health_test.go
+++ b/fv/health_test.go
@@ -17,13 +17,7 @@
 package fv_test
 
 import (
-	"errors"
-	"fmt"
 	"net/http"
-	"time"
-
-	"crypto/tls"
-	"net"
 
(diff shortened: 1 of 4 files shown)


neiljerram:
@fasaxc `make test` and `make k8sfv-test` pass for me locally with this PR; semaphore failure above is caused by semaphore being generally broken at the moment.

PR #1654, opened 2017-12-15
Update Typha liveness test to reflect post-migration behaviour.

fasaxc:
## Description

Update Felix repo health tests to reflect new Typha behaviour (after adding the migration step)

Note: the Typha manifest has a 30s grace period before it even starts the liveness check, so Typha will get quite a bit of time before being restarted (and the KDD migration should be trivially fast because it's only for a couple of small config objects).

## Release Note

```release-note
None required
```


fasaxc:
Verbal OK from Neil

PR #1661, opened 2017-12-19
Avoid reconnecting just to check the ready flag.

fasaxc:
## Description

During v2 -> v3 upgrade, Felix can be waiting for the ready flag to be true for a considerable length of time.  Erik found that we were leaking connections in that case. 

This change re-uses the same connection when rechecking the ready flag and adds a restart if we detect that we've leaked too many connections.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/felix.go
+++ b/felix.go
@@ -167,6 +167,7 @@ func main() {
 	var backendClient bapi.Client
 	var configParams *config.Config
 	var typhaAddr string
+	var numClientsCreated int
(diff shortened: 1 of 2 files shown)


fasaxc:
Fixes https://github.com/projectcalico/calico/issues/1531

PR #1680, opened 2017-12-29
Deduplicate doc on package publication

neiljerram:
Remove Tigera-specific package publication details from RELEASING.md.
Also remove the para on building development packages, as this is
unrelated to releasing.

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -76,6 +76,11 @@ To build Debian and RPM packages for a release:
 
       FORCE_VERSION=<new version> make deb rpm
 
-  to rebuild packages with those changes in.
+  to rebuild packages with those changes in.  (Where `<new version>` is exactly
+  the same as when you ran `make release VERSION=<new version>` above.)
 
 - Once the changes are approved and any testing looks good, merge the PR.
+
+Tigera note: the wider Calico process release, including how to publish those
+packages, is now documented in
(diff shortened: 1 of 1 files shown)


RELEASING.md
```diff
@@ -67,47 +67,15 @@ To build Debian and RPM packages for a release:
...
+  and packages built under `dist/`.
 
-To build Debian and RPM packages for any Git HEAD:
+- Create a PR to get those changes, in particular the release notes, reviewed.
+  If you need to make changes, do so, and run
 
-- Run `make deb rpm`.  You should see changes similar to the release case
-  above, except that you probably don't want to save the temporary
-  debian/changelog and rpm/felix.spec updates that are made.
+      FORCE_VERSION=<new version> make deb rpm
```
fasaxc: Is `<new version>` the same as the above "new version" or does it need to be different?

neiljerram: Yes, it should be the same as the `<new version>` in `make release VERSION=<new version>`.

neiljerram: But that's a bit pants, isn't it?  It would perhaps be better and clearer to have something like `KEEP_CHANGELOGS=true make deb rpm`.

neiljerram: That said, I think this PR should be just doc, and that if we want to make a change like that we should do it separately.

neiljerram:
@fasaxc After updating to remove that Tigera-specific link, I think this PR is in principle good now, and could be merged.  WDYT?

PR #1683, opened 2018-01-17
Port network sets to v3.x

fasaxc:
## Description

Add support for GlobalNetworkSets, which allow a set of CIDRs to be associated with labels in policy.

- Plumb through new datatype - it follows a similar path to a HostEndpoint, but it contains CIDRs instead of IPs.
- Update dataplane to use a `hash:net` instead of `hash:ip` so that it can contain CIDRs (where a single IP is represented as a /32 or /128).

## Todos
- [x] Documentation https://github.com/projectcalico/calico/pull/1630
- [x] Backport
- [x] Release note

## Release Note

```release-note
Felix now supports the GlobalNetworkSet resource type, which allows for a (potentially large)set of external IPs to be referenced (by label selector) in policy.  On Linux, the sets are rendered efficiently using IP sets.
```


fv/containers/containers.go
```diff
@@ -337,25 +337,36 @@ func RunEtcd() *Container {
...
+type TopologyOptions struct {
+	FelixLogSeverity string
+}
+
+func (o TopologyOptions) felixLogSeverity() string {
+	if o.FelixLogSeverity == "" {
+		return "debug"
+	}
+	return o.FelixLogSeverity
+}
```
neiljerram: If we imagine TopologyOptions having more fields with default values, I think I'd prefer a single constructor func that establishes all the defaults, rather than a method like this for each field.

fv/containers/containers.go
```diff
@@ -380,8 +391,12 @@ func StartTwoNodeEtcdTopology() (felixes []*Container, etcd *Container, client c
 // - Configures routes between the hosts, giving each host 10.65.x.0/24, where x is the
 //   index in the returned array.  When creating workloads, use IPs from the relevant block.
 // - Configures the Tunnel IP for each host as 10.65.x.1.
-func StartNNodeEtcdTopology(n int) (felixes []*Container, etcd *Container, client client.Interface) {
+func StartNNodeEtcdTopology(n int, topoOptions ...TopologyOptions) (felixes []*Container, etcd *Container, client client.Interface) {
 	log.Infof("Starting a %d-node etcd topology.", n)
+	var opts TopologyOptions
+	if len(topoOptions) > 0 {
+		opts = topoOptions[0]
+	}
```
neiljerram: Think we should take the one time hit of adding topoOptions in all existing calls to these functions, and *not* make it an optional arg as here.  As it is, I misread the ... as meaning that there could be one set of options for the first Felix, a different set for the second Felix, and so on.

fv/netsets_test.go
```diff
@@ -0,0 +1,359 @@
...
+	"github.com/projectcalico/libcalico-go/lib/net"
+)
+
+const (
+	numPolicies    = 20
+	numNetworkSets = 100
+)
+
+var _ = FContext("Network sets connectivity tests with initialized Felix and etcd datastore", func() {
+
```
neiljerram: Seems to be nothing here - remove?

fv/netsets_test.go
```diff
@@ -0,0 +1,359 @@
...
+		for {
+			pids := getFelixPIDs()
+			if len(pids) == 1 {
+				felixPID = pids[0]
+				break
+			}
+			Expect(time.Since(start)).To(BeNumerically("<", time.Second))
+			time.Sleep(50 * time.Millisecond)
+		}
+	}
```
neiljerram: We should move getFelixPIDs and updateFelixPID into a utility package, and use those from both here and config_update_test.go.

fv/netsets_test.go
```diff
@@ -0,0 +1,359 @@
...
+	})
+
+	AfterEach(func() {
+		if CurrentGinkgoTestDescription().Failed {
+			felix.Exec("iptables-save", "-c")
+		}
+		felix.Stop()
+
+		if CurrentGinkgoTestDescription().Failed {
+			etcd.Exec("etcdctl", "ls", "--recursive", "/")
```
neiljerram: That won't give much for etcdv3...  I guess that's also a clue that this recurring logic should be utility-ized.

But perhaps those changes should be in another PR.  In that case best to leave this as is here, for ease of identifying it as one of the places that needs changing.

neiljerram: (BTW, another Felix repo change that's needed in the same area is in calico-diags.  I'm planning to prepare that this week.)

fv/netsets_test.go
```diff
@@ -0,0 +1,359 @@
...
+			cc.ExpectNone(w[0], w[3])
+			cc.ExpectNone(w[1], w[0])
+			cc.ExpectNone(w[1], w[2])
+			cc.ExpectNone(w[1], w[3])
+			cc.ExpectNone(w[2], w[0])
+			cc.ExpectNone(w[2], w[1])
+			cc.ExpectNone(w[2], w[3])
+			cc.ExpectNone(w[3], w[0])
+			cc.ExpectNone(w[3], w[1])
+			cc.ExpectNone(w[3], w[2])
```
neiljerram: Need to actually check the connectivity!
```
			Eventually(cc.ActualConnectivity, "10s", "100ms").Should(Equal(cc.ExpectedConnectivity()))
```

neiljerram: (If I'm right, looks like this is missing throughout this file - so please check and fix throughout.)

fv/netsets_test.go
```diff
@@ -0,0 +1,359 @@
...
+		churnPolicies := func(iterations int) {
+			log.Info("Churning policies...")
+			created := false
+			for i := 0; i < iterations; i++ {
+				if created {
+					for _, pol := range policies {
+						log.WithField("policy", pol.Name).Info("Deleting policy")
+						cxt, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+						_, err := client.GlobalNetworkPolicies().Delete(cxt, pol.Name, options.DeleteOptions{})
+						cancel()
```
neiljerram: What does the cancel do here?  If the Delete call returns immediately, then (a) I don't see the point of the 10s timeout, and (b) we could be cancelling before the deletion has had any time to happen.

ipsets/ipsets_test.go
```diff
@@ -114,24 +98,63 @@ var _ = Describe("IPSetType", func() {
...
 
 var _ = Describe("IPPort types", func() {
 	It("V4 should stringify correctly", func() {
 		Expect(V4IPPort{
-			IP:       ip.FromString("10.0.0.1").(ip.V4Addr),
+			IP:       ip.FromString("10.0.0.0").(ip.V4Addr),
 			Protocol: labelindex.ProtocolTCP,
 			Port:     1234,
-		}.String()).To(Equal("10.0.0.1,tcp:1234"))
+		}.String()).To(Equal("10.0.0.0,tcp:1234"))
```
neiljerram: Why the change here from .0 to .1 ?

labelindex/named_port_index.go
```diff
@@ -226,13 +226,28 @@ func (idx *SelectorAndNamedPortIndex) OnUpdate(update api.Update) (_ bool) {
...
+			log.Debugf("Updating NamedPortIndex for network set %v", key)
+			netSet := update.Value.(*model.NetworkSet)
+			idx.UpdateEndpoint(
+				key,
+				netSet.Labels,
+				extractCIDRsFromNetworkSet(netSet),
+				nil,
+				nil)
+		} else {
+			log.Debugf("Deleting host endpoint %v from NamedPortIndex", key)
```
neiljerram: 'host endpoint'

labelindex/named_port_index.go
```diff
@@ -255,24 +270,43 @@ func (idx *SelectorAndNamedPortIndex) OnUpdate(update api.Update) (_ bool) {
...
 	return combined
 }
 
-func convertNets(a, b []net.IPNet) []ip.Addr {
-	combined := make([]ip.Addr, 0, len(a)+len(b))
-	for _, addr := range a {
-		combined = append(combined, ip.FromNetIP(addr.IP))
+// extractCIDRsFromWorkloadEndpoint converts the IPv[46]Nets fields of the WorkloadEndpoint into
+// /32 and /128 CIDRs.  It ignores any prefix length (but our validation ensures those nets are
+// /32s or /128s in any case).
```
neiljerram: Why not propagate the prefix length here?  ISTM that would be in the right direction for adding e.g. IPv6 prefix delegation.

rpm/felix.spec
```diff
@@ -210,6 +210,28 @@ rm -rf $RPM_BUILD_ROOT
...
+* Mon Oct 16 2017 Neil Jerram <neil@tigera.io> 2.7.0-0.1.pre2
+  - Felix 2.7.0-pre2 (from Git commit 942db10).
+    [Changes recorded in 2.7.0-pre2 tag]
+    - Add support for network sets.
+
+* Mon Oct 16 2017 Neil Jerram <neil@tigera.io> 2.7.0-0.1.pre1
+  - Felix 2.7.0-pre1 (from Git commit d0576c7).
+    [Changes recorded in 2.7.0-pre1 tag]
+    - Implement support for network sets.
+    [Changes recorded in 2.6.0 tag]
```
neiljerram: As with debian/changelog, remove the 2.6.0 material here.

fv/netsets_test.go
fasaxc: The delete is synchronous but you're always supposed to call `cancel()` to correctly clean up the context.  In this case, I expect that calling cancel frees the resources associated with the timeout timer.

rpm/felix.spec
fasaxc: I might just remove all the 2.7-pre release stanzas, they were pulled in by the merge and I'm not sure they're that useful.

labelindex/named_port_index.go
fasaxc: I did consider it but I was worried about making that change without giving it enough thought.  I _think_ the validation prevents non-/32s from popping out here so we could change this part but I'm pretty sure that downstream requires /32s right now.

dataplane/linux/ipsets_mgr.go
```diff
@@ -48,6 +48,8 @@ func (m *ipSetsManager) OnUpdate(msg interface{}) {
 		switch msg.Type {
 		case proto.IPSetUpdate_IP:
```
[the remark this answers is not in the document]
fasaxc: I think all the complexity is in the dataplane layer, which we are still using (for the all-hosts IP set). This look-up table seems fairly safe to leave in to me.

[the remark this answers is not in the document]
neiljerram: Fine.  (I agree that there doesn't appear to be much specifically related code, anyway.)

fv/netsets_test.go
neiljerram: OK, although that strikes me as quite weird.

In my experience, at least with k8s, the delete-that-matters (i.e. the resource actually disappearing) is asynchronous w.r.t. the Delete call (which returns immediately) and relatively slow, as in ~10s.  So, with the k8s API,

- surely the point of the cancelling the context would be to (try to) cancel the asynchronous deletion operation?

- otherwise what is the point of the context at all?

Now here I appreciate that we have the Calico API and not k8s.  But all the same it doesn't seem to make sense.

- If the desired operation has always completed before the Delete call returns, why should we ask the caller to cancel the context, when our API code could just do that itself?

- If the desired operation isn't known to have completed, we shouldn't risk cancelling the context so soon.

I'm probably missing something...

labelindex/named_port_index.go
neiljerram: OK, that makes sense.  If we did implement non-/32s, there'd be no chance of missing that we had to relax this code again.

rpm/felix.spec
neiljerram: Ok, if you prefer.  I thought you were intentionally wanting to record the 2.7.0-pre1 delivery, and that that was quite a nice touch.

fv/netsets_test.go
fasaxc: The purpose of the context is to allow the IO that the `Delete()` call does to be cancelled (in this case, I'm just using that to impose a timeout, but, in general, you could call `cancel()` from another goroutine). 

The delete returns after doing the write to the datastore.  If there are watchers, they will be asynchronously notified of update (the context doesn't have anything to do with that, though).

The API call can't cancel the context, because it doesn't have the `cancel()` function; in general, a context receiver isn't supposed to cancel the context, because it doesn't know how broad the context is.  For example, you might want to do several IO operations sequentially all within a single timeout.  A classic example is a web server creating a context for each HTTP request with a processing timeout:

- If the sum of all the backend operations is too long, the context times out and all the backend operations get stopped.

- If the user kills the connection then the request machinery can call cancel() to pro-actively stop everything.

ipsets/ipsets_test.go
fasaxc: This checks that the CIDR is correctly masked since we really do want the IP sets members to be canonical CIDRs, not "IPs with prefix lengths".

fv/netsets_test.go
neiljerram: Thanks for explaining - this all makes sense now.

fv/workload/workload.go
```diff
@@ -399,6 +413,35 @@ func (c *ConnectivityChecker) ExpectedConnectivity() []string {
...
+func (c *ConnectivityChecker) CheckConnectivity(optionalDescription ...interface{}) {
+	// Make sure that we retry at least once even if the check itself takes >10s.
+	expConnectivity := c.ExpectedConnectivity()
+	if reflect.DeepEqual(c.ActualConnectivity(), expConnectivity) {
+		return
+	}
+	start := time.Now()
+	completedAttempts := 0
+	var actualConn []string
+	for time.Since(start) < 10*time.Second || completedAttempts <= 1 {
```
neiljerram: Think || should be && here.

labelindex/named_port_index.go
```diff
@@ -306,7 +308,29 @@ func extractCIDRsFromNetworkSet(netSet *model.NetworkSet) []ip.CIDR {
...
+				combined = append(combined,
+					ip.MustParseCIDROrIP("::/1"),
+					ip.MustParseCIDROrIP("8000::/1"),
+				)
+			default:
+				log.WithField("cidr", cidr).Panic("Unknown IP version")
+			}
+			continue
+		}
+		combined = append(combined, cidr)
```
neiljerram: For me this would be clearer in an `else` branch.  (I know it's OK as is because of the `continue`, but it took me a while to see that.)

fv/containers/containers.go
```diff
@@ -45,20 +45,96 @@ type Container struct {
...
-		log.WithField("container", c).Info("Stop no-op because container is not running")
-	} else {
-		log.WithField("container", c).Info("Stop")
-		c.runCmd.Process.Signal(os.Interrupt)
-		c.WaitNotRunning(60 * time.Second)
+		return
+	}
+
+	logCxt := log.WithField("container", c.Name)
+	if c.runCmd == nil {
```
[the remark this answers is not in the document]
fasaxc: Yes, I think we should.

fv/fv_suite_test.go
```diff
@@ -23,15 +23,31 @@ import (
...
+
 	"github.com/projectcalico/libcalico-go/lib/testutils"
 )
 
 func init() {
 	testutils.HookLogrusForGinkgo()
 }
 
+var _ = BeforeEach(func() {
+	testutils.HookLogrusForGinkgo()
```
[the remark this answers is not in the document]
fasaxc: Removed; I don't think it helped with the problem I was trying to address (which was lack of logs on certain tests).

fv/k8sapiserver/k8sapiserver.go
```diff
@@ -106,32 +120,48 @@ func SetUp() *Server {
 		//"-v=10",
 		"--authorization-mode=RBAC",
 	)
-	Expect(server.apiServerContainer).NotTo(BeNil())
+	if server.apiServerContainer == nil {
+		return nil, errors.New("failed to create k8s API server container")
```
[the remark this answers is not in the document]
fasaxc: If the container is nil, is there still a possibility that it _was_ created?

fv/workload/workload.go
```diff
@@ -399,6 +413,35 @@ func (c *ConnectivityChecker) ExpectedConnectivity() []string {
 	return result
 }
 
-func (c *ConnectivityChecker) CheckConnectivity() {
-	EventuallyWithOffset(1, c.ActualConnectivity(), "10s", "100ms").Should(Equal(c.ExpectedConnectivity()))
+func (c *ConnectivityChecker) CheckConnectivity(optionalDescription ...interface{}) {
+	// Make sure that we retry at least once even if the check itself takes >10s.
+	expConnectivity := c.ExpectedConnectivity()
+	if reflect.DeepEqual(c.ActualConnectivity(), expConnectivity) {
+		return
```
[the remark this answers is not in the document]
fasaxc: OK, will fine-tune the comment; we don't retry if we succeed on the first attempt.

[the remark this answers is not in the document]
fasaxc: Actually, looks like I ended up with two solutions to the problem in here.  I'll update the code.

fv/workload/workload.go
fasaxc: It's correct now I removed the extra connectivity check at the start of the function.  The intention is to make sure that we run a minimum of two checks if the first one fails.

fv/k8sapiserver/k8sapiserver.go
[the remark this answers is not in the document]
neiljerram: I was thinking of the previously created etcd container.  Does that need to be cleaned up in this case?

fv/workload/workload.go
neiljerram: OK - but the problem is that two checks might completely successfully in less than 10s, and then the loop will continue iterating because 10s have not elapsed yet.  Surely that is wrong?

fv/k8sapiserver/k8sapiserver.go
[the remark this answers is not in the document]
fasaxc: Aha, yes

fv/workload/workload.go
fasaxc: It does an early return if the connectivity is ever correct.

neiljerram: Ah, right, I get it now: "**If the first connectivity check fails**, we want [to do at least one retry]".

PR #1686, opened 2018-01-25
calico-diags: Dump etcdv3 data as well as etcdv2

neiljerram:
## Todos
- [ ] Testing

### DIFF
--- a/etc/calico-diags
+++ b/etc/calico-diags
@@ -64,7 +64,7 @@ echo "  copying contents of etcd..."
 
 ${ETCDCTL} ls --recursive -p | grep -v '/$' | xargs -n1 -I '{}' sh -c "echo; echo {}; ${ETCDCTL} get {} " | tee etcdctl_get
 echo >> etcdctl_get
-${ETCDCTL_V3} get --prefix /calico | tee -a etcdctl_get
+eval ${ETCDCTL_V3} get --prefix /calico | tee -a etcdctl_get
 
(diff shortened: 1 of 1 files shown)


neiljerram:
@fasaxc I've seen this working as intended now on both CentOS and Ubuntu Trusty - so please could you review?

etc/calico-diags
```diff
@@ -62,6 +63,8 @@ rabbitmqctl status rabbitmq_status 2>&1
 echo "  copying contents of etcd..."
 
 ${ETCDCTL} ls --recursive -p | grep -v '/$' | xargs -n1 -I '{}' sh -c "echo; echo {}; ${ETCDCTL} get {} " | tee etcdctl_get
+echo >> etcdctl_get
+eval ${ETCDCTL_V3} get --prefix /calico | tee -a etcdctl_get
```
fasaxc: What does the eval do here?  Does etcdctl in v3 mode print keys as well as values?  The etcdv2 code habndles that above with the xargs business.

neiljerram: > What does the eval do here?

It allows the value of ETCDCTL_V3 to include an initial environment variable setting "ETCDCTL_API=3".  Without the `eval`, the shell tries to interpret "ETCDCTL_API=3" as the name of the command to run.

> Does etcdctl in v3 mode print keys as well as values? 

Yes.  So the xargs business isn't needed for v3.

PR #1692, opened 2018-02-06
Add latency measurement tests

fasaxc:
## Description

I built a bit of infrastructure to do the latency tests for network sets.  It's a bit rough around the edges but it seems sensible to get it checked in and running in semaphore so we can re-use it later.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -59,7 +59,12 @@ ifeq ($(ARCH),ppc64le)
 endif
 
 GO_BUILD_CONTAINER?=calico/go-build$(ARCHTAG):$(GO_BUILD_VER)
-FV_FELIXIMAGE?=calico/felix$(ARCHTAG)
+
+# Name used to tag the Felix container image when building.
+FELIX_IMAGE_NAME?=calico/felix$(ARCHTAG):latest
+
+# Name of the images to run FV tests against.
+FV_FELIXIMAGE?=$(FELIX_IMAGE_NAME)
 FV_ETCDIMAGE?=quay.io/coreos/etcd:v3.2.5$(ARCHTAG)
 FV_K8SIMAGE?=gcr.io/google_containers/hyperkube$(ARCHTAG):v1.7.5
 FV_GINKGO_NODES?=4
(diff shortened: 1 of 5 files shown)


fv/latency_test.go
```diff
@@ -0,0 +1,230 @@
...
+		})
+
+		It("with allow-all should have good latency", func() {
+			meanRtt := w[0].LatencyTo(w[1].IP, w[1].DefaultPort)
+			_, err := fmt.Fprintf(resultsFile, "allow-all: %v\n", meanRtt)
+			Expect(meanRtt).To(BeNumerically("<", 10*time.Millisecond))
+			Expect(err).NotTo(HaveOccurred())
+		})
+
+		Describe("with all() selector", func() {
```
neiljerram: Saying "source selector" would be a touch clearer, given that there's already another use of `all()` above.

fv/latency_test.go
```diff
@@ -0,0 +1,230 @@
...
+				BeforeEach(func() {
+					w[1].WorkloadEndpoint.Spec.IPNetworks = append(w[1].WorkloadEndpoint.Spec.IPNetworks,
+						generateIPs(10000)...)
+					wep := w[1].WorkloadEndpoint
+					wep.Namespace = "fv"
+					_, err := client.WorkloadEndpoints().Update(utils.Ctx, wep, utils.NoOptions)
+					Expect(err).NotTo(HaveOccurred())
+
+					Eventually(func() int {
+						return getNumIPSetMembers(felix.Container, "cali4-s:5y5I3VdRZfDU01O--xXAPx2")
```
neiljerram: Is that IP set name predictable?

fv/latency_test.go
```diff
@@ -0,0 +1,230 @@
...
+					}, "100s", "1000ms").Should(Equal(10002))
+				})
+
+				It("should have good latency", func() {
```
neiljerram: Perhaps worth commenting here that the `all()` source selector will now expand to an IP set with 10000+ entries.

fv/latency_test.go
```diff
@@ -0,0 +1,230 @@
...
+				w[ii].Stop()
+			}
+		})
+	}
+
+	Context("IPv4: Network sets tests with initialized Felix and etcd datastore", func() {
+		describeLatencyTests(latencyConfig{ipVersion: 4})
+	})
+
+	// TODO: IPv6
```
neiljerram: Worth filling this in?  Does it just need an IPv6 version of generateIPs?

Makefile
```diff
@@ -59,6 +59,7 @@ ifeq ($(ARCH),ppc64le)
 endif
 
 GO_BUILD_CONTAINER?=calico/go-build$(ARCHTAG):$(GO_BUILD_VER)
+FV_FELIXIMAGE?=calico/felix$(ARCHTAG)
```
neiljerram: Should this have a `:latest` tag?

fv/latency_test.go
fasaxc: Yes, it's a hash of the selector name with a fixed prefix.  Given this test isn't really about that, though, maybe I should calculate the value here.

PR #1696, opened 2018-02-07
WIP Add etcd restart FV test

caseydavenport:
## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/fv/etcd_restart_test.go
+++ b/fv/etcd_restart_test.go
@@ -1,0 +1,167 @@
+// +build fvtests
+
+// Copyright (c) 2017-2018 Tigera, Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package fv_test
+
+import (
+	. "github.com/onsi/ginkgo"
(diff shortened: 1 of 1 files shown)


caseydavenport:
Test seems to be failing now, need to investigate why.

caseydavenport:
@fasaxc test is passing now, but it also appears to pass on v3.0.0 felix, which isn't expected.

I think it's because etcd is being terminated gracefully. I haven't been able to get docker kill / start working properly yet.

fv/etcd_restart_test.go
```diff
@@ -0,0 +1,147 @@
...
+
+	It("should survive an ungraceful etcd termination", func() {
+		By("having workload to workload connectivity", func() {
+			cc.ExpectSome(w[0], w[1])
+			cc.ExpectSome(w[1], w[0])
+			cc.CheckConnectivity()
+		})
+
+		By("killing etcd", func() {
+			// Kill etcd so that it cannot gracefully terminate connections.
```
fasaxc: I think, even a `kill -9` will gracefully terminate connections because the connection is socket gets cleaned up by the kernel.

One option is to nuke the conntrack entry inside the felix container.  I think you can delete by destination IP, and you have etcd's IP in-hand as etcd.IP (IIRC).  Another is to sneak in a raw iptables rule inside one of the containers, that lets you silently drop packets.

fasaxc: It's a good idea to check the graceful restart case too, of course!

fv/etcd_restart_test.go
```diff
@@ -0,0 +1,147 @@
...
+			err := exec.Command("docker", "restart", "-t", "0", etcd.Name).Run()
```
fasaxc: Does this preserve the data in etcd?

caseydavenport: hehe, nope. But hopefully we can just drop packets without restarting.

fasaxc:
Continued under #1699

PR #1699, opened 2018-02-12
Add FV test that silently cuts the etcd connection.

fasaxc:
## Description

- Use the raw table to silently drop packets from one connection.
- Check that felix detects and reconnects.

## Todos
- [x] Documentation
- [ ] Backport to 3.0.x
- [x] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/fv/etcd_restart_test.go
+++ b/fv/etcd_restart_test.go
@@ -1,6 +1,6 @@
 // +build fvtests
 
-// Copyright (c) 2017-2018 Tigera, Inc. All rights reserved.
+// Copyright (c) 2018 Tigera, Inc. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


fv/etcd_restart_test.go
```diff
@@ -0,0 +1,174 @@
...
+			var portRegexp = regexp.MustCompile(`sport=(\d+).*dport=2379`)
+			for _, felix := range felixes {
+				// Use conntrack to identify the source port that Felix is using.
+				out, err := felix.ExecOutput("conntrack", "-L")
+				Expect(err).NotTo(HaveOccurred())
+				logrus.WithField("output", out).WithError(err).Info("Conntrack entries")
+				found := false
+				for _, line := range strings.Split(out, "\n") {
+					matches := portRegexp.FindStringSubmatch(line)
+					if len(matches) == 0 {
```
neiljerram: Looks like "< 2" would be a more relevant check, given that code below uses `matches[1]`.

fv/etcd_restart_test.go
```diff
@@ -0,0 +1,174 @@
...
+						continue
+					}
+					found = true
+
+					// Use the raw table to drop the single TCP connection that felix is using,
```
neiljerram: You say "single", but this loop will handle and blackhole multiple connections to 2379, I believe.  I'm unclear whether that is (a) OK, or (b) your intention.

fv/etcd_restart_test.go
```diff
@@ -0,0 +1,174 @@
...
+						"-d", etcd.IP,
+						"-m", "multiport", "--source-ports", matches[1],
+						"-j", "DROP")
+				}
+				Expect(found).To(BeTrue(), "Failed to detect any felix->etcd connections")
+				felix.Exec("conntrack", "-D", "--orig-dst", etcd.IP)
+			}
+		})
+
+		By("starting to update policy again", func() {
```
neiljerram: "starting to"?  I think this should just be "updating policy again".

fv/etcd_restart_test.go
fasaxc: Intentional to sever any and all connections.  (Quite possible that gRPC uses more than one, if it so chooses.)

PR #1705, opened 2018-02-15
Enable host endpoint policy with kube-proxy running in IPVS mode

song-jiang:
## Description
Enable host endpoint policy with kube-proxy running in IPVS mode.

## Todos
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Enable host endpoint policy with kube-proxy running in IPVS mode.
```

### DIFF
--- a/dataplane/driver.go
+++ b/dataplane/driver.go
@@ -17,6 +17,7 @@
 package dataplane
 
 import (
+	"math/bits"
 	"net"
 	"os/exec"
 
@@ -59,12 +60,15 @@ func StartDataplaneDriver(configParams *config.Config, healthAggregator *health.
 				"MarkMask": configParams.IptablesMarkMask,
 			}).Panic("Not enough mark bits available for endpoint mark.")
 		}
+		// Take first bit position (position 1) from endpoint mark mask as reserved generic endpoint mark.
+		markEndpointMarkGeneric := uint32(1) << uint(bits.TrailingZeros32(markEndpointMark))
 		log.WithFields(log.Fields{
-			"acceptMark":   markAccept,
-			"passMark":     markPass,
-			"scratch0Mark": markScratch0,
-			"scratch1Mark": markScratch1,
-			"endpointMark": markEndpointMark,
+			"acceptMark":          markAccept,
+			"passMark":            markPass,
+			"scratch0Mark":        markScratch0,
+			"scratch1Mark":        markScratch1,
+			"endpointMark":        markEndpointMark,
+			"endpointMarkGeneric": markEndpointMarkGeneric,
 		}).Info("Calculated iptables mark bits")
 
 		dpConfig := intdataplane.Config{
@@ -94,11 +98,12 @@ func StartDataplaneDriver(configParams *config.Config, healthAggregator *health.
 				OpenStackMetadataIP:          net.ParseIP(configParams.MetadataAddr),
 				OpenStackMetadataPort:        uint16(configParams.MetadataPort),
 
-				IptablesMarkAccept:   markAccept,
-				IptablesMarkPass:     markPass,
-				IptablesMarkScratch0: markScratch0,
-				IptablesMarkScratch1: markScratch1,
-				IptablesMarkEndpoint: markEndpointMark,
+				IptablesMarkAccept:          markAccept,
+				IptablesMarkPass:            markPass,
+				IptablesMarkScratch0:        markScratch0,
+				IptablesMarkScratch1:        markScratch1,
+				IptablesMarkEndpoint:        markEndpointMark,
+				IptablesMarkEndpointGeneric: markEndpointMarkGeneric,
 
 				IPIPEnabled:       configParams.IpInIpEnabled,
 				IPIPTunnelAddress: configParams.IpInIpTunnelAddr,
--- a/dataplane/driver.go
+++ b/dataplane/driver.go
@@ -60,7 +60,7 @@ func StartDataplaneDriver(configParams *config.Config, healthAggregator *health.
 				"MarkMask": configParams.IptablesMarkMask,
 			}).Panic("Not enough mark bits available for endpoint mark.")
 		}
-		// Take first bit position (position 1) from endpoint mark mask as reserved generic endpoint mark.
+		// Take lowest bit position (position 1) from endpoint mark mask reserved for non-calico endpoint.
 		markEndpointNonCaliEndpoint := uint32(1) << uint(bits.TrailingZeros32(markEndpointMark))
 		log.WithFields(log.Fields{
 			"acceptMark":          markAccept,
@@ -68,7 +68,7 @@ func StartDataplaneDriver(configParams *config.Config, healthAggregator *health.
 			"scratch0Mark":        markScratch0,
 			"scratch1Mark":        markScratch1,
 			"endpointMark":        markEndpointMark,
-			"endpointMarkGeneric": markEndpointNonCaliEndpoint,
+			"endpointMarkNonCali": markEndpointNonCaliEndpoint,
 		}).Info("Calculated iptables mark bits")
 
 		dpConfig := intdataplane.Config{
--- a/dataplane/driver.go
+++ b/dataplane/driver.go
@@ -61,14 +61,14 @@ func StartDataplaneDriver(configParams *config.Config, healthAggregator *health.
 			}).Panic("Not enough mark bits available for endpoint mark.")
 		}
 		// Take first bit position (position 1) from endpoint mark mask as reserved generic endpoint mark.
-		markEndpointMarkGeneric := uint32(1) << uint(bits.TrailingZeros32(markEndpointMark))
+		markEndpointNonCaliEndpoint := uint32(1) << uint(bits.TrailingZeros32(markEndpointMark))
 		log.WithFields(log.Fields{
 			"acceptMark":          markAccept,
 			"passMark":            markPass,
 			"scratch0Mark":        markScratch0,
 			"scratch1Mark":        markScratch1,
 			"endpointMark":        markEndpointMark,
-			"endpointMarkGeneric": markEndpointMarkGeneric,
+			"endpointMarkGeneric": markEndpointNonCaliEndpoint,
 		}).Info("Calculated iptables mark bits")
 
 		dpConfig := intdataplane.Config{
@@ -103,7 +103,7 @@ func StartDataplaneDriver(configParams *config.Config, healthAggregator *health.
 				IptablesMarkScratch0:        markScratch0,
 				IptablesMarkScratch1:        markScratch1,
 				IptablesMarkEndpoint:        markEndpointMark,
-				IptablesMarkEndpointGeneric: markEndpointMarkGeneric,
+				IptablesMarkNonCaliEndpoint: markEndpointNonCaliEndpoint,
 
 				IPIPEnabled:       configParams.IpInIpEnabled,
 				IPIPTunnelAddress: configParams.IpInIpTunnelAddr,
--- a/dataplane/linux/endpoint_mgr_test.go
+++ b/dataplane/linux/endpoint_mgr_test.go
@@ -78,7 +78,7 @@ var wlDispatchEmpty = []*iptables.Chain{
 			},
 			{
 				Action:  iptables.SetMaskedMarkAction{Mark: 0x0100, Mask: 0xff00},
-				Comment: "Generic endpoint mark",
+				Comment: "Non-Cali endpoint mark",
 			},
 		},
 	},
@@ -446,7 +446,7 @@ func chainsForIfaces(ifaceMetadata []string,
 			},
 			iptables.Rule{
 				Action:  iptables.SetMaskedMarkAction{Mark: 0x0100, Mask: 0xff00},
-				Comment: "Generic endpoint mark",
+				Comment: "Non-Cali endpoint mark",
 			},
 		)
 		epMarkFrom = append(epMarkFrom,
--- a/dataplane/linux/endpoint_mgr_test.go
+++ b/dataplane/linux/endpoint_mgr_test.go
@@ -64,8 +64,23 @@ var wlDispatchEmpty = []*iptables.Chain{
 		},
 	},
 	{
-		Name:  "cali-set-endpoint-mark",
-		Rules: []iptables.Rule{},
+		Name: "cali-set-endpoint-mark",
+		Rules: []iptables.Rule{
+			iptables.Rule{
+				Match:   iptables.Match().InInterface("cali+"),
+				Action:  iptables.DropAction{},
+				Comment: "Unknown endpoint",
+			},
+			iptables.Rule{
+				Match:   iptables.Match().InInterface("tap+"),
+				Action:  iptables.DropAction{},
+				Comment: "Unknown endpoint",
+			},
+			{
+				Action:  iptables.SetMaskedMarkAction{Mark: 0x0100, Mask: 0xff00},
+				Comment: "Generic endpoint mark",
+			},
+		},
 	},
 }
 
@@ -137,6 +152,7 @@ func chainsForIfaces(ifaceMetadata []string,
 	epMarkSetName := "cali-set-endpoint-mark"
 	epMarkFromName := "cali-from-endpoint-mark"
 	epMarkSetOnePrefix := "cali-sm-"
+	epmarkFromPrefix := outPrefix[:6]
 
 	if host {
 		hostOrWlLetter = "h"
@@ -147,6 +163,7 @@ func chainsForIfaces(ifaceMetadata []string,
 		}
 		outPrefix = "cali-to-"
 		inPrefix = "cali-from-"
+		epmarkFromPrefix = inPrefix[:6]
 	}
 	for _, ifaceMetadata := range ifaceMetadata {
 		var ifaceName, polName string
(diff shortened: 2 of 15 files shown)


rules/dispatch.go
```diff
@@ -38,59 +37,71 @@ func (r *DefaultRuleRenderer) WorkloadDispatchChains(
...
-			)...,
-		)
-	}
 	return result
 }
 
+// In some scenario, e.g. packet goes to an kuberentes ipvs service ip. Traffic goes through input/output filter chain
+// instead of forward filter chain. It is not feasible to match on an incoming workload/host interface with service ips.
+// Assemble a set-endpoint-mark chain to set the endpoint mark matching on the incoming workload/host interface and
+// a from-endpoint-mark chain to jump to a corresponding endpoint chain matching on its' endpoint mark.
```
fasaxc: typo: `s/its'/its/` (this part of English makes no sense whatsoever)

rules/dispatch.go
```diff
@@ -102,112 +113,69 @@ func (r *DefaultRuleRenderer) hostDispatchChains(
...
 			ChainDispatchToHostEndpointForward,
 			false,
 			false,
-			false,
 		)...,
 	)
 }
 
-func (r *DefaultRuleRenderer) dispatchChains(
+func (r *DefaultRuleRenderer) dispatchEndpointChains(
```
fasaxc: How bout calling this `interfaceNameDispatchChains`, to emphasise that they match on the name?

rules/dispatch.go
```diff
@@ -102,112 +113,69 @@ func (r *DefaultRuleRenderer) hostDispatchChains(
...
-		}
-		if _, present := prefixToNames[prefix]; !present {
-			// Record the prefixes in sorted order (if we iterate over the map, we get a
-			// random order, which we don't want).
-			prefixes = append(prefixes, prefix)
-		}
-		prefixToNames[prefix] = append(prefixToNames[prefix], name)
-		lastName = name
-	}
+	commonPrefix, prefixes, prefixToNames := r.divideEndpointNamesToPrefixTree(names)
```
fasaxc: This refactoring is much cleaner, thanks!

rules/dispatch.go
```diff
@@ -102,112 +113,69 @@ func (r *DefaultRuleRenderer) hostDispatchChains(
...
 
 	rootFromEndpointRules := make([]Rule, 0)
 	rootToEndpointRules := make([]Rule, 0)
```
fasaxc: Comment on Line 185 now out of date since this function no longer deals with the mark chains.

rules/dispatch.go
```diff
@@ -387,3 +333,193 @@ func (r *DefaultRuleRenderer) dispatchChains(
 
 	return chains
 }
+
+func (r *DefaultRuleRenderer) dispatchEndPointMarkChains(
```
fasaxc: If you make the above change, then you should change this to `endpointMarkDispatchChains` for consistency

rules/dispatch.go
```diff
@@ -102,112 +113,69 @@ func (r *DefaultRuleRenderer) hostDispatchChains(
...
-	epMarkMapper EndpointMarkMapper,
 	fromEndpointPfx,
 	toEndpointPfx,
 	dispatchFromEndpointChainName,
 	dispatchToEndpointChainName string,
 	dropAtEndOfFromChain bool,
 	dropAtEndOfToChain bool,
-	useEndPointMark bool,
 ) []*Chain {
 	// Sort interface names so that rules in the dispatch chain are ordered deterministically.
```
fasaxc: Why not move this sort into the `divideEndpointNamesToPrefixTree` function so it can be shared?

rules/dispatch.go
```diff
@@ -387,3 +333,193 @@ func (r *DefaultRuleRenderer) dispatchChains(
...
+				Action: GotoAction{
+					Target: EndpointChainName(setMarkPfx, ifaceName),
+				},
+			})
+		}
+	}
+
+	// At the end of set mark chain, set generic endpoint mark. A generic endpoint mark is used when a forward packet
+	// whose incoming interface is neither a workload nor a host endpoint.
+	rootSetMarkRules = append(rootSetMarkRules, Rule{
```
fasaxc: I think we need to make sure that we *don't* set this for `caliXXX|tapXXX` interfaces.  `caliXXX` interfaces that we haven't programmed yet shouldn't be allowed to send traffic.

rules/dispatch.go
```diff
@@ -387,3 +333,193 @@ func (r *DefaultRuleRenderer) dispatchChains(
...
+	log.WithField("ifaceNames", names).Debug("Rendering endpoint mark dispatch chains")
+
+	rootSetMarkRules := make([]Rule, 0)
+	rootFromMarkRules := make([]Rule, 0)
+
+	// start rendering set mark rules.
+	// Since there can be >100 endpoints, putting them in a single list adds some latency to
+	// endpoints that are later in the chain.  To reduce that impact, we build a shallow tree of
+	// chains based on the prefixes of the chains.
+	commonPrefix, prefixes, prefixToNames := r.divideEndpointNamesToPrefixTree(names)
```
fasaxc: D'oh, just spotted that this won't work well for a mixed workload and host endpoint chain.  The common prefix of `"caliABC"` and `"eth0"` is `""`, you might need to do host and workload endpoints separately?

rules/epmark.go
```diff
@@ -19,9 +19,15 @@ import (
 	"hash/fnv"
 	"io"
 
+	log "github.com/sirupsen/logrus"
+
 	"github.com/projectcalico/felix/markbits"
 )
 
+const (
+	pseudoGenericEndpointName = "caliPseudoGenericEndpoint"
```
fasaxc: Please make sure this is an invalid interface name, I think it's enough to make it >32 chars, but you could add a 0-byte in there (which is definitely not valid in an interface name)

rules/static.go
```diff
@@ -411,23 +449,29 @@ func (r *DefaultRuleRenderer) filterOutputChain(ipVersion uint8) *Chain {
...
+					Comment: "Allow IPIP packets from tunnel device to other Calico hosts",
+				},
+			)
+		}
+
+		// Jump to forward-endpoint-mark chain if endpoint mark is not zero, which means
 		// packet has been through filter INPUT chain. There could be policies apply to its ingress interface.
 		rules = append(rules,
 			Rule{
 				Match:  Match().MarkNotClear(r.IptablesMarkEndpoint),
```
fasaxc: I think we might need to clear the mark in case the user has their own tunnels set up.  For example, what if they're using a GRE tunnel and the GRE packets go through the OUTPUT chain again, just like our IPIP packets?

rules/dispatch.go
```diff
@@ -387,3 +333,193 @@ func (r *DefaultRuleRenderer) dispatchChains(
...
+	rootSetMarkRules = append(rootSetMarkRules, Rule{
+		Action: SetMaskedMarkAction{
+			Mark: r.IptablesMarkEndpointGeneric,
+			Mask: epMarkMapper.GetMask()},
+		Comment: "Generic endpoint mark",
+	})
+
+	// start rendering from mark rules.
+	// Rendering rules for workload endpoints.
+	for _, name := range wlNames {
```
fasaxc: `wlNames` and `hepNames` need to be sorted too, I think (unless we know they're already sorted?)

iptables/match_builder.go
```diff
@@ -76,6 +76,20 @@ func (m MatchCriteria) MarkMatchesWithMask(mark, mask uint32) MatchCriteria {
 	return append(m, fmt.Sprintf("-m mark --mark %#x/%#x", mark, mask))
 }
 
+func (m MatchCriteria) MarkNotMatchesWithMask(mark, mask uint32) MatchCriteria {
```
fasaxc: How about `NotMarkMatchesWithMask()`  (for consistency with the other `NotXXX` matches below)

rules/dispatch.go
```diff
@@ -387,3 +333,193 @@ func (r *DefaultRuleRenderer) dispatchChains(
...
+	// Since there can be >100 endpoints, putting them in a single list adds some latency to
+	// endpoints that are later in the chain.  To reduce that impact, we build a shallow tree of
+	// chains based on the prefixes of the chains.
+	commonPrefix, prefixes, prefixToNames := r.divideEndpointNamesToPrefixTree(names)
+
+	// Now, iterate over the prefixes.  If there are multiple names in a prefix, we render a
+	// child chain for that prefix.  Otherwise, we render the rule directly to avoid the cost
+	// of an extra goto.
+	var chains []*Chain
+	for _, prefix := range prefixes {
```
fasaxc: I find this much clearer than the old code but it might be worth a try to put his into a utility function to build a single chain: `buildDispatchChain(rootPrefix, childPrefix string, getActionForEndpoint func(name string) iptables.Action, prefixes <...etc>)`

Then above, you'd call it twice, to build the to/from chains, whereas here, you'd only call it once.

dataplane/driver.go
```diff
@@ -94,11 +98,12 @@ func StartDataplaneDriver(configParams *config.Config, healthAggregator *health.
...
-				IptablesMarkPass:     markPass,
-				IptablesMarkScratch0: markScratch0,
-				IptablesMarkScratch1: markScratch1,
-				IptablesMarkEndpoint: markEndpointMark,
+				IptablesMarkAccept:          markAccept,
+				IptablesMarkPass:            markPass,
+				IptablesMarkScratch0:        markScratch0,
+				IptablesMarkScratch1:        markScratch1,
+				IptablesMarkEndpoint:        markEndpointMark,
+				IptablesMarkEndpointGeneric: markEndpointMarkGeneric,
```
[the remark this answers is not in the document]
song-jiang: Good point.

rules/dispatch.go
song-jiang: Good point.

rules/dispatch.go
song-jiang: Yes.

rules/dispatch.go
song-jiang: Good point. Sort can be shared.

rules/dispatch.go
song-jiang: Done.

rules/dispatch.go
song-jiang: Done.

rules/epmark.go
song-jiang: Done.

iptables/match_builder.go
song-jiang: Done.

rules/dispatch.go
song-jiang: We need to call it twice too for workload and host endpoint. It is a bit tricky to combine those but it is doable. I created a `buildSingleDispatchChains` and return `childChains, rootChain rootRule`. The caller can make tweeks (merge, add rules to rootRules) as needed.

rules/dispatch.go
song-jiang: Fixed.Done.

rules/dispatch.go
song-jiang: Yes. Changed to do host and workload endpoints separately. Done.

rules/static.go
song-jiang: Done. Replace `ReturnAction`  to `GotoAction{Target: cali-clear-endpoint-mark}` .

rules/dispatch.go
song-jiang: Done. Added drop rules to `cali-set-endpoint-mark` chain.

dataplane/linux/endpoint_mgr_test.go
```diff
@@ -64,8 +64,23 @@ var wlDispatchEmpty = []*iptables.Chain{
...
 	},
 	{
-		Name:  "cali-set-endpoint-mark",
-		Rules: []iptables.Rule{},
+		Name: "cali-set-endpoint-mark",
+		Rules: []iptables.Rule{
+			iptables.Rule{
+				Match:   iptables.Match().InInterface("cali+"),
+				Action:  iptables.DropAction{},
+				Comment: "Unknown endpoint",
```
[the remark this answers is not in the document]
song-jiang: It means the interface prefix is `cali` or `tab` but this endpoint is unknown to felix. For instance, cni created a pod but felix has not got the workload endpoint update yet.

dataplane/driver.go
```diff
@@ -59,12 +60,15 @@ func StartDataplaneDriver(configParams *config.Config, healthAggregator *health.
...
-			"passMark":     markPass,
-			"scratch0Mark": markScratch0,
-			"scratch1Mark": markScratch1,
-			"endpointMark": markEndpointMark,
+			"acceptMark":          markAccept,
+			"passMark":            markPass,
+			"scratch0Mark":        markScratch0,
+			"scratch1Mark":        markScratch1,
+			"endpointMark":        markEndpointMark,
+			"endpointMarkGeneric": markEndpointNonCaliEndpoint,
```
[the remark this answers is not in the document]
song-jiang: Oh. Yes.

dataplane/driver.go
```diff
@@ -59,12 +60,15 @@ func StartDataplaneDriver(configParams *config.Config, healthAggregator *health.
 				"MarkMask": configParams.IptablesMarkMask,
 			}).Panic("Not enough mark bits available for endpoint mark.")
 		}
+		// Take first bit position (position 1) from endpoint mark mask as reserved generic endpoint mark.
```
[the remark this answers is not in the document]
song-jiang: Good Point.

dataplane/linux/endpoint_mgr_test.go
```diff
@@ -64,8 +64,23 @@ var wlDispatchEmpty = []*iptables.Chain{
...
+			},
+			iptables.Rule{
+				Match:   iptables.Match().InInterface("tap+"),
+				Action:  iptables.DropAction{},
+				Comment: "Unknown endpoint",
+			},
+			{
+				Action:  iptables.SetMaskedMarkAction{Mark: 0x0100, Mask: 0xff00},
+				Comment: "Generic endpoint mark",
```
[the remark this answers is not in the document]
song-jiang: This endpoint mark is used to mark any incoming interface which is not part of workload or host endpoint.

rules/dispatch.go
```diff
@@ -38,59 +37,71 @@ func (r *DefaultRuleRenderer) WorkloadDispatchChains(
...
-				false, // Non forwarded packet will pass through set endpoint mark chain.
-				true,
-				true,
-			)...,
-		)
-	}
 	return result
 }
 
+// In some scenario, e.g. packet goes to an kuberentes ipvs service ip. Traffic goes through input/output filter chain
```
neiljerram: s/kuberentes/kubernetes/

dataplane/linux/endpoint_mgr_test.go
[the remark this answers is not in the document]
song-jiang: Added comment. Done.

rules/static.go
```diff
@@ -126,6 +126,37 @@ func (r *DefaultRuleRenderer) StaticFilterInputForwardCheckChain(ipVersion uint8
 	}
 }
 
+// With kube-proxy running in ipvs mode, forward traffic gets an endpoint mark in INPUT chain.
```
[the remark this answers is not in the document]
song-jiang: Yes. I have updated initial comment that an endpoint mark is representing the source endpoint.

rules/static.go
```diff
@@ -126,6 +126,37 @@ func (r *DefaultRuleRenderer) StaticFilterInputForwardCheckChain(ipVersion uint8
...
+// Forward endpoint mark chain implements rules to process forward packet with an endpoint mark in OUTPUT chain.
+func (r *DefaultRuleRenderer) StaticFilterOutputForwardEndpointMarkChain() *Chain {
+	var fwRules []Rule
+
+	fwRules = append(fwRules,
+		// Jump to from-endpoint-mark dispatch chain if endpoint mark is not a generic mark, which means
```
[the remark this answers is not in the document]
song-jiang: `generic` means not a calico workload or host endpoint. Agreed this name is not a good choice. Replaced all appearance of `generic` to `non-cali` and added a comment alongside definition.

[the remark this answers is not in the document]
song-jiang: It means `packet having a non-generic endpoint mark => it has been through filter INPUT chain with a real ingress endpoint`. I have updated the comment. Done.

rules/dispatch.go
song-jiang: Oh. Thanks. Done.

rules/static.go
```diff
@@ -126,6 +126,37 @@ func (r *DefaultRuleRenderer) StaticFilterInputForwardCheckChain(ipVersion uint8
...
+		// packet has been through filter INPUT chain with a real ingress endpoint. There could be policies
+		// apply to its ingress interface, e.g. workload egress or host endpoint ingress policies.
+		Rule{
+			Match:  Match().NotMarkMatchesWithMask(r.IptablesMarkNonCaliEndpoint, r.IptablesMarkEndpoint),
+			Action: JumpAction{Target: ChainDispatchFromEndPointMark},
+		},
+		// For any forward packet with an endpoint mark, apply host endpoint egress forward policies.
```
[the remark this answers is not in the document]
song-jiang: Agreed. Updated comments. Done.

PR #1707, opened 2018-02-16
Replace FlexVol Management API with binder

spikecurtis:
## Description
Replaces Flex Volume driver management API, which handles creating per-pod domain sockets for the
Policy Sync API, with a `Binder` object that watches the file system and binds the Policy Sync API server
to per-pod domain sockets.

Depends on https://github.com/istio/istio/tree/master/security/cmd/flexvolume as the Flex Volume Driver.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/binder/watcher.go
+++ b/binder/watcher.go
@@ -17,6 +17,7 @@ package binder
 import (
 	"io/ioutil"
 	"log"
+	"os"
 	"path/filepath"
 	"strings"
 	"time"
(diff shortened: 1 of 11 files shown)


binder/watcher.go
```diff
@@ -0,0 +1,108 @@
...
+func (p *pollWatcher) watch() <-chan workloadEvent {
+	c := make(chan workloadEvent)
+	go p.poll(c)
+	return c
+}
+
+func (p *pollWatcher) poll(events chan<- workloadEvent) {
+	// The workloads we know about.
+	known := make(map[string]bool)
+	credPath := filepath.Join(p.path + CredentialsSubdir)
```
saumoh: Can u merge the changes from https://github.com/colabsaumoh/proto-udsuspver/commit/63f881f8634f581f14b25430914b6ed0ac4ce204
at least the change in `watcher.go`.
When testing in my rig i had to make changes here to make it work.
other than that all of this looks good to me.

spikecurtis: Fixed!

PR #1708, opened 2018-02-16
Update iptables_lock_test and health_test to use FV_FELIXIMAGE.

djlwilder:
## Description
Update iptables_lock_test and health_test to use FV_FELIXIMAGE.

Remove the hard coded felix image name in these tests, the default
image name will not change if FV_FELIXIMAGE is unset.

## Release Note

```release-note
None required
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -200,6 +200,7 @@ k8sfv-test: calico/felix k8sfv-test-existing-felix
 # container image.  To use some existing Felix version other than
 # 'latest', do 'FELIX_VERSION=<...> make k8sfv-test-existing-felix'.
 k8sfv-test-existing-felix: bin/k8sfv.test
+	FV_FELIXIMAGE=$(FV_FELIXIMAGE) \
 	FV_ETCDIMAGE=$(FV_ETCDIMAGE) \
 	FV_TYPHAIMAGE=$(FV_TYPHAIMAGE) \
 	FV_K8SIMAGE=$(FV_K8SIMAGE) \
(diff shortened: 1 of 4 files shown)


djlwilder:
I added a second commit (and squashed) to also include this change in k8sfv/run-test.
This change is needed for native build and test on ppc64le.   I ran verified the change on both amd64 and ppc64le with:
```
make fv
JUST_A_MINUTE=true CLEANUP=true USE_TYPHA=false make k8sfv-test
JUST_A_MINUTE=true CLEANUP=true USE_TYPHA=true make k8sfv-test
```

djlwilder:
FYI @fasaxc

fasaxc:
That's odd, I thought the k8sfv tests did honour some flag.  @neiljerram ?

neiljerram:
No, they have that for Typha, but not for Felix.  There is a ${FELIX_VERSION} knob, but not Felix image.

I think the explanation is:

- Originally our expectation was that k8sfv would always run with <the Felix code that I have in front of me>, which would have just been built to calico/felix:latest as part of the process, and hence there was no need to customize either Felix image or Felix version.

- Then we realized that - effectively because of how we're mixing product and test in one repo - it could be useful to run changed test code on another version of Felix, and hence ${FELIX_VERSION}.

Then I presume that customizing the image name is the logical next step, given multiarch.

neiljerram:
@fasaxc So these changes look good to me.  Do you have remaining concerns before we merge?

PR #1715, opened 2018-03-01
Break binder event loop on stop

spikecurtis:
Fix bug in code that had binder break out of its event loop.


spikecurtis:
@saumoh you probably want to pull this change into the code you're writing for Istio

spikecurtis:
`watcher.go` isn't bugged in this way, if that's what you're asking.

However, we don't have any mechanism for stopping the `poll()` like we the Binder event loop, so the go routine will never terminate gracefully. I'll whip up a PR that propagates the stop event to the `poll()`

PR #1718, opened 2018-03-02
Whitelist failsafe port response traffic in the raw table only

fasaxc:
## Description

This fixes a nasty interaction between do-not-track policy and failsafe ports.  Previously, adding do-not-track policy to a host endpoint resulted in failsafe port traffic being blocked because one leg of the connection would get conntracked, but the return leg would hit the do-not-track policy, which would either drop it (thus breaking the failsafe port) or mark it as NOTRACK, which results in future outbound packets failing the `--ctstate INVALID` test (because conntrack has only seen an outbound SYN, it isn't expecting the outbound to move past the SYN stage of the handshake).

This change whitelists the response traffic in the raw table only so that failsafe ports are now forced to be conntracked, bypassing any do-not-track policy in both directions.  Since the change is made only in the raw table (and we don't mark failsafe port traffic with the accepted mark), the response traffic still has to pass the `--ctstate ESTABLISHED` test in the filter table.

## Todos
- [x] Documentation
- [ ] Backport
- [x] Release note

## Release Note

```release-note
Fixes an interaction between failsafe inbound/outbound ports and do-not-track policy that resulted in failsafe ports being blocked if do-not-track policy was added.
```

### DIFF
--- a/fv/donottrack_test.go
+++ b/fv/donottrack_test.go
@@ -1,6 +1,6 @@
 // +build fvtests
 
-// Copyright (c) 2017-2018 Tigera, Inc. All rights reserved.
+// Copyright (c) 2018 Tigera, Inc. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


fv/donottrack_test.go
```diff
@@ -0,0 +1,275 @@
...
+					Action: api.Deny,
+					Destination: api.EntityRule{
+						Selector: host0Selector,
+					},
+				},
+			}
+			host0Pol, err = client.GlobalNetworkPolicies().Update(ctx, host0Pol, options.SetOptions{})
+			Expect(err).NotTo(HaveOccurred())
+
+			host1Pol, err = client.GlobalNetworkPolicies().Update(ctx, host1Pol, options.SetOptions{})
```
neiljerram: Does this update do anything?

fv/workload/workload.go
```diff
@@ -492,5 +500,5 @@ func (c *ConnectivityChecker) CheckConnectivityWithTimeout(timeout time.Duration
 		strings.Join(actualConn, "\n    "),
 		strings.Join(expConnectivity, "\n    "),
 	)
-	Fail(message, 1)
+	Fail(message, callerSkip)
```
neiljerram: Note: I think default `callerSkip` here has changed from 1 to 2 - but I presume that that is what you want (for nicest debugging/reporting).

fasaxc: Yes, exactly, I want it to report the failure on the test case line, not in `workload.go`.

PR #1720, opened 2018-03-02
Sync service accounts

spikecurtis:
## Description
Felix finds profiles that are K8s Service Accounts and passes them to the dataplane.
PolicySync API fans these out to all connected workloads.

## Todos

## Release Note

```release-note
None required
```

### DIFF
--- a/calc/profile_decoder.go
+++ b/calc/profile_decoder.go
@@ -33,21 +33,21 @@ const (
 	KindServiceAccount
 )
 
-// profileDecoder takes updates from a dispatcher, determines if the profile is a Kubernetes Service Account, and
+// ProfileDecoder takes updates from a dispatcher, determines if the profile is a Kubernetes Service Account, and
 // if it is, generates a dataplane update or remove for it.
-type profileDecoder struct {
+type ProfileDecoder struct {
 	callbacks passthruCallbacks
 }
 
-func NewProfileDecoder(callbacks passthruCallbacks) *profileDecoder {
-	return &profileDecoder{callbacks}
+func NewProfileDecoder(callbacks passthruCallbacks) *ProfileDecoder {
+	return &ProfileDecoder{callbacks}
 }
 
-func (p *profileDecoder) RegisterWith(d *dispatcher.Dispatcher) {
+func (p *ProfileDecoder) RegisterWith(d *dispatcher.Dispatcher) {
 	d.Register(model.ProfileKey{}, p.OnUpdate)
 }
 
-func (p *profileDecoder) OnUpdate(update api.Update) (filterOut bool) {
+func (p *ProfileDecoder) OnUpdate(update api.Update) (filterOut bool) {
 	// This type assertion is safe because we only registered for Profile updates.
 	key := update.Key.(model.ProfileKey)
 	id, kind := classifyProfile(key)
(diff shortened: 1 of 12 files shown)


calc/profile_decoder.go
```diff
@@ -0,0 +1,90 @@
...
+	} else {
+		labels := update.Value.(map[string]string)
+		msg := proto.ServiceAccountUpdate{Id: &id, Labels: decodeServiceAccountLabels(labels)}
+		p.callbacks.OnServiceAccountUpdate(&msg)
+	}
+	return false
+}
+
+func classifyProfile(key model.ProfileLabelsKey) (proto.ServiceAccountID, kind) {
+	if strings.HasPrefix(key.Name, conversion.ServiceAccountProfileNamePrefix) {
```
fasaxc: Nit: I think it's a Go idiom to use early returns for "guards" rather than nesting
```
if ! strings.HasPrefix(...) {
    return
}
parts := strings.Split(...)
if len(parts) != 3 {
    return
}
...
```

calc/profile_decoder.go
```diff
@@ -0,0 +1,90 @@
...
+		} else {
+			log.WithField("name", key.Name).Warn("Profile with SA prefix could not be parsed.")
+			return proto.ServiceAccountID{}, KindUnknown
+		}
+	} else {
+		return proto.ServiceAccountID{}, KindUnknown
+	}
+}
+
+func decodeServiceAccountLabels(in map[string]string) map[string]string {
```
fasaxc: Worth a comment to explain why this is needed?

dataplane/mock/mock_dataplane.go
```diff
@@ -330,6 +342,10 @@ func (d *MockDataplane) OnEvent(event interface{}) {
 		delete(d.endpointToPolicyOrder, id.String())
 		delete(d.endpointToUntrackedPolicyOrder, id.String())
 		delete(d.endpointToPreDNATPolicyOrder, id.String())
+	case *proto.ServiceAccountUpdate:
+		d.serviceAccounts[*event.Id] = event
+	case *proto.ServiceAccountRemove:
+		delete(d.serviceAccounts, *event.Id)
```
fasaxc: Worth an assert that the key did exist?

policysync/processor.go
```diff
@@ -396,6 +428,17 @@ func (p *Processor) syncRemovedProfiles(ei *EndpointInfo) {
 	}
 }
 
+// sendServiceAccounts sends all known ServiceAccounts to the endpoint
+func (p *Processor) sendServiceAccounts(ei *EndpointInfo) {
+	for _, update := range p.serviceAccountByID {
+		log.WithFields(log.Fields{
+			"serviceAccount": update.Id.String(),
```
fasaxc: Won't the logging framework lazily stringify these?

calc/profile_decoder.go
spikecurtis: Good to know!

I rewrote this function to also handle Namespaces (PR coming soon), so I don't think it's worth fixing.

fasaxc:
LGTM, would be good to have a green Semaphore run (excluding the Typha pin check, which we can fix up before we merge to master)

PR #1722, opened 2018-03-06
Update calico/felix alpine base image to 3.7

matthewdupre:
## Description
calico/felix is still on alpine 3.4.  Rev it to 3.7.

## Release Note
```release-note
None required
```


docker-image/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM alpine:3.4
+FROM alpine:3.7
```
fasaxc: Please can you copy the change to the ppc version?

matthewdupre: DOne

PR #1723, opened 2018-03-06
Sync namespace labels

spikecurtis:
## Description
Decode Calico Profiles which represent K8s Namespaces and pass them over the Policy Sync API.

## Todos

## Release Note

```release-note
None required
```

### DIFF
--- a/calc/profile_decoder.go
+++ b/calc/profile_decoder.go
@@ -26,14 +26,6 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
-type kind int
-
-const (
-	KindUnknown kind = iota
-	KindServiceAccount
-	KindNamespace
-)
-
 // ProfileDecoder takes updates from a dispatcher, determines if the profile is a Kubernetes Service Account or
 // Kubernetes Namespace, and if it is, generates a dataplane update or remove for it.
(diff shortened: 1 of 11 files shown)


calc/profile_decoder.go
```diff
@@ -51,41 +53,56 @@ func (p *ProfileDecoder) OnUpdate(update api.Update) (filterOut bool) {
...
-		return false
-	}
-	if update.Value == nil {
-		p.callbacks.OnServiceAccountRemove(id)
-	} else {
-		labels := update.Value.(map[string]string)
-		msg := proto.ServiceAccountUpdate{Id: &id, Labels: decodeServiceAccountLabels(labels)}
-		p.callbacks.OnServiceAccountUpdate(&msg)
+	id, knd := p.classifyProfile(key)
+	switch knd {
```
fasaxc: This feels like a good place to use Go's type switch:
```
switch id := id.(type) {
    case proto.ServiceAccountID: // "id := " above removes need for a type assertion inside this block
        if update.Value == nil {
            p.callbacks....
        } else {
            ....
        }
    case proto.NamespaceID:
        ...
    case nil: // hit for an interface with a nil type pointer
        ...
}
```

PR #1724, opened 2018-03-07
Fix that workload ingress policy wasn't applied to IPVS traffic

fasaxc:
## Description

- Add the to-workload dispatch jumps to the forward-endpoint-mark chain.
- Fix that a return from the forward-endpoint-mark chain would only return to
  the cali-OUTPUT chain, not to the OUTPUT chain.
- Remove some IPVS ingress support rules taht can no longer be hit (since any
  traffic that hit those will be redirected to the forward-endpoint-mark
  chain instead.

## Todos
- [x] Integration tests (delete as appropriate) Done (k8s E2Es)
- [x] Documentation
- [x] Release note

## Release Note

```release-note
None required
```


fasaxc:
Thanks @song-jiang

PR #1726, opened 2018-03-07
Windows dataplane driver should apply profiles when policies are not …

nwoodmsft:
…applicable for a given direction

## Description
Type of fix: Bug fix

Tested on: Windows RS3 / RS4 pre-release

Components affected: Windows Dataplane driver

Details: When workload tier Policies are not applicable for a direction (e.g. egress) the driver was applying a default block rule for that direction instead of applying any applicable Profiles. In some configurations, this would result in traffic being dropped by default rather than allowed.

## Todos
- [X ] Unit tests (full coverage)
- [X ] Documentation

## Release Note

```release-note
None required
```

### DIFF
--- a/dataplane/windows/policysets/policysets.go
+++ b/dataplane/windows/policysets/policysets.go
@@ -91,18 +91,23 @@ func (s *PolicySets) RemovePolicySet(setId string) {
 }
 
 // GetPolicySetRules receives a list of Policy set ids and it computes the complete
-// set of resultant hns rules which are needed to enforce all of the Policy sets. As
-// the Policy sets are processed, we increment a priority number and assign it to each rule
-// from the current set. By incremening the rule priority for each set, we ensure that all of
-// the sets will be enforced and considered by the dataplane in the order intended by felix.
-// Once all rules are gathered, we add a final pair of rules to default deny any traffic which
-// has not matched any rules from any Policy sets.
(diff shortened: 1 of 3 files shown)


dataplane/windows/policysets/policysets.go
```diff
@@ -559,3 +564,11 @@ func filterNets(mixedCIDRs []string, ipVersion uint8) (filtered []string, filter
 	}
 	return
 }
+
+// AppendPrefixToPolicyNames appends a prefix to each policy name
+func AppendPrefixToPolicyNames(prefix string, policyNames []string) (appendedNames []string) {
```
fasaxc: Nit: I think you're "prepending", not "appending"

Since this function just takes some strings and returns updated strings, I'd be tempted to move it to the package where it's used and make it more generic (and a bit shorter in name) `prependAll(prefix string, in []strings) []string`

nwoodmsft:
Correct, the prefixing logic should have been present and was missing. Originally, the driver did not add any prefixes, but as part of PR feedback we made a late change to add prefixing (to avoid the case where a policy and profile had the same name). Unfortunately, that change was incomplete as it only touched Policy Manager. If I recall, this meant that policies/profiles were still being applied but not correctly updated when a ipset change changed.

dataplane/windows/policysets/policysets.go
nwoodmsft: Sounds good, i'll make this change.

PR #1734, opened 2018-03-12
Sync IPSets over Policy Sync API

spikecurtis:
## Description
Sync IPSets over the Policy Sync API

Handles splitting up large sets into multiple messages.

## Todos
- [x] Documentation in plan

## Release Note

```release-note
None required
```

### DIFF
--- a/policysync/ipset.go
+++ b/policysync/ipset.go
@@ -1,0 +1,105 @@
+// Copyright (c) 2018 Tigera, Inc. All rights reserved.
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package policysync
+
+import (
+	"fmt"
+
+	"github.com/projectcalico/felix/ipsets"
+	"github.com/projectcalico/felix/proto"
+	"github.com/projectcalico/libcalico-go/lib/set"
+)
+
+type ipSetInfo struct {
+	ipsets.IPSetMetadata
+	members set.Set
+}
+
+func newIPSet(update *proto.IPSetUpdate) *ipSetInfo {
+	s := &ipSetInfo{}
+
+	switch update.GetType() {
+	case proto.IPSetUpdate_IP:
+		s.Type = ipsets.IPSetTypeHashIP
+	case proto.IPSetUpdate_IP_AND_PORT:
+		s.Type = ipsets.IPSetTypeHashIPPort
+	case proto.IPSetUpdate_NET:
+		s.Type = ipsets.IPSetTypeHashNet
+	}
+
+	s.SetID = update.GetId()
+
+	// Note: We ignore MaxSize.
+
+	s.replaceMembers(update)
+	return s
+}
+
+func (s *ipSetInfo) replaceMembers(update *proto.IPSetUpdate) {
+	s.members = set.New()
+	for _, ms := range update.GetMembers() {
+		s.members.Add(s.Type.CanonicaliseMember(ms))
+	}
+}
+
+func (s *ipSetInfo) deltaUpdate(update *proto.IPSetDeltaUpdate) {
+	for _, ms := range update.GetAddedMembers() {
+		s.members.Add(s.Type.CanonicaliseMember(ms))
+	}
+	for _, ms := range update.GetRemovedMembers() {
+		s.members.Discard(s.Type.CanonicaliseMember(ms))
+	}
+}
+
+func (s *ipSetInfo) getIPSetUpdate() *proto.IPSetUpdate {
+	u := &proto.IPSetUpdate{Id: s.SetID}
+	s.members.Iter(func(item interface{}) error {
+		m := item.(fmt.Stringer)
+		u.Members = append(u.Members, m.String())
+		return nil
+	})
+	return u
+}
+
+type ruleList interface {
+	GetInboundRules() []*proto.Rule
+	GetOutboundRules() []*proto.Rule
+}
+
+func addIPSetsRuleList(rl ruleList, s map[string]bool) {
+	for _, rule := range rl.GetInboundRules() {
+		addIPSetsRule(rule, s)
+	}
+	for _, rule := range rl.GetOutboundRules() {
+		addIPSetsRule(rule, s)
+	}
+}
+
+func addIPSetsRule(r *proto.Rule, s map[string]bool) {
+	addAll(r.SrcIpSetIds, s)
+	addAll(r.DstIpSetIds, s)
+	addAll(r.SrcNamedPortIpSetIds, s)
+	addAll(r.DstNamedPortIpSetIds, s)
+	addAll(r.NotSrcIpSetIds, s)
+	addAll(r.NotDstIpSetIds, s)
+	addAll(r.NotSrcNamedPortIpSetIds, s)
+	addAll(r.NotDstNamedPortIpSetIds, s)
+}
+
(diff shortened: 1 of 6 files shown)


spikecurtis:
I don't think this requires specific FVs, since it's entirely in the Processor and UT'd, but shout if you disagree @fasaxc

policysync/ipset.go
```diff
@@ -0,0 +1,105 @@
...
+func addIPSetsRuleList(rl ruleList, s map[string]bool) {
+	for _, rule := range rl.GetInboundRules() {
+		addIPSetsRule(rule, s)
+	}
+	for _, rule := range rl.GetOutboundRules() {
+		addIPSetsRule(rule, s)
+	}
+}
+
+func addIPSetsRule(r *proto.Rule, s map[string]bool) {
```
fasaxc: Any way you can write a UT that will fail if a new IpSetIds field gets added without this code being updated?  Always a risk that a function like this will get forgotten in maintenance so elsewhere in Felix I've got UTs that use reflection to scan structs for new fields and compare, for example proto.Rule vs felix's ParsedRule.

For example, you could scan the proto struct for fields that end in `IpSetIds` and populate them with a slice that contains their name, then pass them through this function and check that all the names show up.  If the fields change, we'll then be forced to fix this code or the test.

policysync/ipset.go
```diff
@@ -0,0 +1,105 @@
...
+	s := &ipSetInfo{}
+
+	switch update.GetType() {
+	case proto.IPSetUpdate_IP:
+		s.Type = ipsets.IPSetTypeHashIP
+	case proto.IPSetUpdate_IP_AND_PORT:
+		s.Type = ipsets.IPSetTypeHashIPPort
+	case proto.IPSetUpdate_NET:
+		s.Type = ipsets.IPSetTypeHashNet
+	}
```
fasaxc: Worth a `default: log.Warn("Unknown IP set type")`?

policysync/processor_test.go
```diff
@@ -293,6 +298,314 @@ var _ = Describe("Processor", func() {
...
+				})
+
+				It("should send IPSetUpdate to only to ref'd endpoint", func(done Done) {
+					msg := updateIpSet(IPSetName, 2)
+					updates <- msg
+					g := <-refdOutput
+					Expect(g).To(Equal(proto.ToDataplane{Payload: &proto.ToDataplane_IpsetUpdate{msg}}))
+
+					assertInactiveNoUpdate()
+					close(done)
```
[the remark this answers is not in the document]
spikecurtis: closing the `Done` channel just indicates to Ginkgo that the testcase is over. It's not a test cleanup operation.

policysync/ipset.go
spikecurtis: Yes, but should be a panic.

policysync/processor_test.go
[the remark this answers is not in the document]
fasaxc: Ah, I didn't spot where that came from; I'm so used to seeing `It` lambdas with no arguments.  Why do you need the done argument here?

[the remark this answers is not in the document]
spikecurtis: I find the `done` construction easier to read than using `Eventually` everywhere.  It also gives more flexibility for matching specific fields of the channel message; `Eventually` only matches on the top level message coming out of the channel.

policysync/ipset.go
spikecurtis: I ended up making this method public so I could test it, but don't feel great about it. What do you think?  Should the UT suite be a different package? Should I spin up a new Ginkgo suite just for this test?

policysync/processor.go
```diff
@@ -372,10 +399,68 @@ func (p *Processor) handleNamespaceRemove(update *proto.NamespaceRemove) {
...
+	id := update.Id
+	log.WithField("ID", id).Debug("Processing IPSetDeltaUpdate")
+
+	// We trust the calc graph to never send us Delta updates for non-existent sets.
+	s := p.ipSetsByID[id]
+	s.deltaUpdate(update)
+
+	// gRPC has limits on message size, so break up large update if necessary.
+	updates := splitIPSetDeltaUpdate(update)
+	for _, ei := range p.updateableEndpoints() {
```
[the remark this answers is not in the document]
spikecurtis: There are enough subtle differences that I don't want to do that.

In the `IPSetUpdate` version, we also add the IPSet ID to a set.

I don't want to use `sendIPSetUpdate(ei *EndpointInfo, id string)` because I want the splitting to happen just once, outside the endpoint loop.

spikecurtis:
I tried writing an FV with a large number of updates, but it took ages to run since we have to push WEPs to the datastore and thru the calc graph.

Instead, I've refactored the UTs that test the splitting to happen using a *real* gRPC connection, so they should be sensitive to changes in that library ruining our day.

policysync/ipset.go
fasaxc: There are examples in felix where we have tests in `package name_test` for "FV" and tests in `package name` for private UTs.  You end up needing two ginkgo suite files but it does work OK.

I don't find it particularly worrying to export the odd method for test convenience, either.  It's not like we're exporting an important public API from this package.

policysync/processor_test.go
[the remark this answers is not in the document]
fasaxc: OK, so the only purpose of adding the Done argument is to trigger Ginkgo's timeout logic.  Can't say I'm keen on their API choice; there's nothing about adding a Done argument to a test that says "now this will run with a timeout" to me.  It's unfortunate that they don't give you a way to shut down a test after it times out.  Presumably it'll be leaked.

policysync/processor_test.go
```diff
@@ -640,19 +836,74 @@ func deltaUpdateIpSet(id string, add, del int) *proto.IPSetDeltaUpdate {
...
+		grpc.WithDialer(getDialer("unix"))}
+}
+
+func getDialer(proto string) func(string, time.Duration) (net.Conn, error) {
+	return func(target string, timeout time.Duration) (net.Conn, error) {
+		return net.DialTimeout(proto, target, timeout)
+	}
+}
+
+const ListenerDir = "/tmp/felixut"
```
fasaxc: It'd be best to use a per-test temp dir to avoid any cross-contamination if tests are run in parallel (or if any garbage is left from a previous failure).

policysync/ipset.go
```diff
@@ -37,6 +37,8 @@ func newIPSet(update *proto.IPSetUpdate) *ipSetInfo {
 		s.Type = ipsets.IPSetTypeHashIPPort
 	case proto.IPSetUpdate_NET:
 		s.Type = ipsets.IPSetTypeHashNet
+	default:
+		panic("unknown IPSetType")
```
fasaxc: You should use `log.WithField("IPSetType", ...).Panic("...")`.  `log.Panic()` makes sure the log gets properly flushed.

policysync/processor_test.go
```diff
@@ -865,16 +870,23 @@ func getDialer(proto string) func(string, time.Duration) (net.Conn, error) {
...
-	}
-	lis, err := net.Listen("unix", ListenerPath)
+const ListenerSocket = "policysync.sock"
+
+// Single source of random numbers insures repeated calls to makeTmpListenDir get unique paths.  But! source is not
+// threadsafe so we protect it with a Mutex. Not the most idiomatic, but simple and fast.
+var rng = rand.New(rand.NewSource(time.Now().UnixNano()))
+var rngLock = sync.Mutex{}
+
+func makeTmpListenerDir() string {
```
fasaxc: I think there's a library function that does exactly what you're looking for here: `ioutil.TempDir()`

policysync/processor_test.go
```diff
@@ -649,14 +652,16 @@ var _ = Describe("Processor", func() {
...
+				var socketPath string
 
 				BeforeEach(func() {
 					uidAllocator := policysync.NewUIDAllocator()
 					syncServer = policysync.NewServer(uut.JoinUpdates, uidAllocator.NextUID)
 
 					gRPCServer = grpc.NewServer(grpc.Creds(testCreds{}))
 					proto.RegisterPolicySyncServer(gRPCServer, syncServer)
-					listener = openListener()
+					socketPath = makeTmpListenerDir()
```
fasaxc: Should delete this dir in an AfterEach()

PR #1736, opened 2018-03-14
Run HostEndpoint tests against KDD also

tmjd:
## Release Note

```release-note
None required
```


fv/datastore/datastore_describe.go
```diff
@@ -0,0 +1,55 @@
...
+	"fmt"
+
+	. "github.com/onsi/ginkgo"
+)
+
+type DatastoreType int
+
+const (
+	DatastoreEtcdV3 DatastoreType = 1 << iota
+	DatastoreK8s
```
neiljerram: How about using the DatastoreType type and values from libcalico-go/lib/apiconfig ?

fv/datastore/datastore_describe.go
```diff
@@ -0,0 +1,55 @@
...
+
+	DatastoreAll = DatastoreEtcdV3 | DatastoreK8s
+)
```
neiljerram: Those are strings, so you'd need a different way of representing this (`DatastoreAll`), but I guess that wouldn't be hard.

fv/datastore/datastore_describe.go
```diff
@@ -0,0 +1,55 @@
...
+// easy construction of end-to-end tests covering multiple different datastore
+// drivers.
+//
+// The *datastores* parameter is a bit-wise OR of the required datastore drivers
+// that will be tested.
+func DatastoreDescribe(description string, datastores DatastoreType, body func(getInfra func() (DatastoreInfra, error))) bool {
+	if datastores&DatastoreEtcdV3 != 0 {
+		Describe(fmt.Sprintf("%s (etcdv3 backend)", description),
+			func() {
+				body(CreateEtcdDatastoreInfra)
```
neiljerram: Guess these infra funcs aren't used outside this package, so should begin with lowercase.

fv/datastore/infra_etcd.go
```diff
@@ -0,0 +1,100 @@
...
+		"-e", "FELIX_DATASTORETYPE=etcdv3",
+		"-e", "CALICO_ETCD_ENDPOINTS=http://" + eds.etcdContainer.IP + ":1234",
+	}
+}
+
+func (eds *EtcdDatastoreInfra) GetCalicoClient() client.Interface {
+	return utils.GetEtcdClient(eds.etcdContainer.IP, "")
+}
+
+func (eds *EtcdDatastoreInfra) AddNode(felix *containers.Felix, idx int, count int) {
```
neiljerram: `count int` -> `numNodes int` or `needBGP bool`.

tmjd:
Thanks for the review @neiljerram. To your point 1, I realized I was duplicating a lot of code and my thinking was that I'm not sure if what I am building will work every place that makes use of the code I would have been modifying.  I think I'll continue on with the way I have it and get through the additional HostEndpoint test changes and then, try to replace all instances with my new versions and see if that is possible.  And then assuming it is possible I can rework the moves or just make the changes to the original code if that makes sense.

neiljerram:
@tmjd That sounds good; thanks for thinking about my points.

fv/infrastructure/datastore_describe.go
```diff
@@ -0,0 +1,49 @@
...
+				func() {
+					body(createEtcdDatastoreInfra)
+				})
+		case apiconfig.Kubernetes:
+			Describe(fmt.Sprintf("%s (kubernetes backend)", description),
+				func() {
+					body(createK8sDatastoreInfra)
+				})
+		default:
+			return false
```
fasaxc: I think this should panic.  Otherwise we'll silently fail to run tests, right?

tmjd: I wasn't sure so I checked it out and you are right it just silently fails. So I've switched to panic.

tmjd:
I've pushed updates.
Before this is merged I should squash.
Should I squash all the changes together or should I leave the initial move commit on it's own?

fasaxc:
Feel free to squash down your markups commits but I prefer to keep separate commits for the different chunks of work.

fasaxc:
Hmm, there was a semaphore failure but no error in the log that I could see.  Kicked off another run...

PR #1739, opened 2018-03-19
Automatic update of logrus to v1.0.4

fasaxc:


fasaxc:
Looks like known flakes on the CI run

PR #1743, opened 2018-03-20
policysync UT

spikecurtis:
## Description
This PR adds full UT to the PolicySync package in Felix

## Todos

## Release Note

```release-note
None required
```


spikecurtis:
These constructions now look like:

`Expect(&g).To(HavePayload(expectedMessage))`

And it will print out both the actual payload and the expected.  This avoids the awkward job of having to construct a `To_Dataplane` object for comparison in the test code.

PR #1746, opened 2018-03-22
standard makefile targets and multi-arch push

deitch:
## Description

* adds calico-project-standard `build` and `image` targets to wrap `bin/calico-felix` and `calico/felix`, respectively
* adds calico-project-standard `build-all` and `image-all` targets to build and image for all supported architectures 
* adds calico-project-standard `push` and `push-all` targets to push images to docker hub and quay.io, pushing arch-less tags as `linux-amd64` until such time as multi-arch manifests are supported
* updates `README.md` to describe the targets

This should make it much easier for our process to simply "build and deploy for all architectures"

As discussed on slack with @fasaxc and @caseydavenport 

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

Multi-arch image support

```release-note
None required
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -50,3 +50,4 @@ junit.xml
 /k8sfv/prometheus/data
 /k8sfv/prometheus/prometheus.yml
 /k8sfv/output/
+
--- a/Makefile
+++ b/Makefile
@@ -46,13 +46,18 @@
 # The target architecture is select by setting the ARCH variable.
 # When ARCH is undefined it is set to the detected host architecture.
 # When ARCH differs from the host architecture a crossbuild will be performed.
-ARCHES=$(patsubst docker-image/Dockerfile.%,%,$(wildcard docker-image/Dockerfile.*))
+
+# restore this when all arches work. For now, there are qemu issues related to cross-compile on s390x
+#   see https://github.com/projectcalico/go-build/pull/32
+#ARCHES=$(patsubst docker-image/Dockerfile.%,%,$(wildcard docker-image/Dockerfile.*))
+ARCHES=amd64 arm64 ppc64le
 
 
 # BUILDARCH is the host architecture
 # ARCH is the target architecture
 # we need to keep track of them separately
 BUILDARCH ?= $(shell uname -m)
+BUILDOS ?= $(shell uname -s | tr A-Z a-z)
 
 # canonicalized names for host architecture
 ifeq ($(BUILDARCH),aarch64)
@@ -73,10 +78,12 @@ ifeq ($(ARCH),x86_64)
     override ARCH=amd64
 endif
 
-GO_BUILD_VER ?= v0.14
+CONTAINER_NAME=calico/felix
+GO_BUILD_VER ?= v0.15
 # For building, we use the go-build image for the *host* architecture, even if the target is different
 # the one for the host should contain all the necessary cross-compilation tools
-GO_BUILD_CONTAINER = calico/go-build:$(GO_BUILD_VER)-$(BUILDARCH)
+# we do not need to use the arch since go-build:v0.15 now is multi-arch manifest
+GO_BUILD_CONTAINER = calico/go-build:$(GO_BUILD_VER)
 PROTOC_VER ?= v0.1
 PROTOC_CONTAINER ?= calico/protoc:$(PROTOC_VER)-$(BUILDARCH)
 FV_ETCDIMAGE ?= quay.io/coreos/etcd:v3.2.5-$(BUILDARCH)
@@ -110,7 +117,11 @@ help:
 	@echo "For any target, set ARCH=<target> to build for a given target."
 	@echo "For example, to build for arm64:"
 	@echo
-	@echo "  make calico/felix ARCH=arm64"
+	@echo "  make build ARCH=arm64"
+	@echo
+	@echo "To generate a docker image for arm64:"
+	@echo
+	@echo "  make image ARCH=arm64"
 	@echo
 	@echo "By default, builds for the architecture on which it is running. Cross-building is supported"
 	@echo "only on amd64, i.e. building for other architectures when running on amd64."
@@ -122,10 +133,15 @@ help:
 	@echo
 	@echo "Builds:"
 	@echo
-	@echo "  make all           Build all the binary packages."
-	@echo "  make deb           Build debs in ./dist."
-	@echo "  make rpm           Build rpms in ./dist."
-	@echo "  make calico/felix  Build calico/felix docker image."
+	@echo "  make all                    Build all the binary packages."
+	@echo "  make deb                    Build debs in ./dist."
+	@echo "  make rpm                    Build rpms in ./dist."
+	@echo "  make build                  Build binary."
+	@echo "  make image                  Build docker image."
+	@echo "  make build-all              Build binary for all supported architectures."
+	@echo "  make image-all              Build docker images for all supported architectures."
+	@echo "  make push IMAGETAG=tag      Deploy docker image with the tag IMAGETAG for the given ARCH, e.g. calico/felix:<IMAGETAG>-<ARCH>."
+	@echo "  make push-all IMAGETAG=tag  Deploy docker images with the tag IMAGETAG all supported architectures"
 	@echo
 	@echo "Tests:"
 	@echo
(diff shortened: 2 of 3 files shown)


deitch:
@fasaxc and @caseydavenport , several open issues here:

1. I left the targets named `release`, `continue-release` and `release-once-tagged`. @caseydavenport seemed to indicate that we wanted those standardized on `release` (unchanged), `release-verify` and `release-publish`. Should we rename or leave them?
2. `release-once-tagged` does not actually push anything, just builds the artifacts, some `docker tag`ging, some verification, and then lots of "you should now push this out". Do we want to make the actual `docker push` part of the target?
3. `manifest-tool`. There is a target for it, but it isn't called by anything. Once the image push (manual or moved into `release-once-tagged` / `release-publish`) is done, it needs to be called. Move that into the `-all` target for `release-once-tagged`?

djlwilder:
The build-all and image-all targets have a bug,  The "all" portion of the target name is seen as a part of the list of ARCHES.  ie:  it eventually runs "$(MAKE) build ARCH=all" and fails.  Flipping the name around to all-build works ok.  I don't know why.

djlwilder:
Since s390 is not cross compiling due to qemu issue we need to remove it from the list of ARCHES.

deitch:
> The build-all and image-all targets have a bug, The "all" portion of the target name is seen as a part of the list of ARCHES. ie: it eventually runs "$(MAKE) build ARCH=all" and fails. Flipping the name around to all-build works ok. I don't know why

Nice catch. I will figure it out. Some `Makefile` magic. :-)

deitch:
> Since s390 is not cross compiling due to qemu issue we need to remove it from the list of ARCHES

Well, it _is_ cross-compiling, just not running correctly. :-) But you are right. I will remove it.

Nice to get some love from the s390x folk who can figure it out.

deitch:
Removed `s390x` from the list. I am going to flatten the commits so it is a single commit. I kept the `patsubst` magic in a comment to make it easy to restore when we get `s390x` working correctly.

deitch:
Fixed both the issues. Thanks for finding @djlwilder . 

@fasaxc and @caseydavenport take a look please?

djlwilder:
The new build-all and image-all targets LGTM.     I'm not sure about the push-manifest target,  the $(call join_platforms,$(ARCHES)) bit is not expanding as I expect and nothing defines VERSION.

$ make push-manifest
./manifest-tool push from-args --platforms  --template calico/felix:-ARCH --target calico/felix:
FATA[0000] You must specify all three arguments --platforms, --template and --target 
Makefile:183: recipe for target 'push-manifest' failed

deitch:
Rebased after the merge of #1752 . 

@fasaxc and @caseydavenport can we address the three open questions from above so this can move ahead?

deitch:
> the $(call join_platforms,$(ARCHES)) bit is not expanding as I expect

No it isn't. I must have messed up on commits, because the code isn't there now and was before. Putting back.

> nothing defines VERSION

Eh. I should add enforcement for defining `VERSION`. Will add that too.

Thanks for catching these @djlwilder

deitch:
Nice catch @djlwilder . Fixed both of the issues.

Makefile
```diff
@@ -158,6 +163,30 @@ TOPDIR:=$(shell pwd)
 all: deb rpm calico/felix-$(ARCH)
 test: ut fv
 
+all-multi: $(addprefix sub-all-,$(ARCHES))
+sub-all-%:
+	$(MAKE) all ARCH=$*
+
+manifest-tool:
```
fasaxc: Bit of a hack to download this here.  I think we should add this to the go-build container so that it's available to all the builds at a pinned / blessed version.

mkumatag: Nice idea,! instead of downloading the flow will be copying the manifest-tool from `go-build` container to host machine for building manifest image(as it requires docker daemon access, credentials to push to docker hub etc..)

Is my understanding correct?

fasaxc: Yes, perhaps that's the best we can do.  We also do that for ginkgo, in order to run the FV tests is a way that lets them use docker.  Suggest copying the binary out into ./bin, which I think is what we do for ginkgo when needed.

mkumatag: cool.

mkumatag: I'll add a PR to `go-build` to add manifest-tool binary. -Thanks

mkumatag: @fasaxc : submitted the PR to include manifest-tool in `go-build` - https://github.com/projectcalico/go-build/pull/39

deitch: Why copy it to the host machine? Why not just run it in a container and bind-mount the docker config in?

mkumatag: well, theoretically yes. But not really sure whether container will also get the docker credentials also inside the container to push images to the registry.

deitch: Either way, once @mkumatag's PR adding it to go-build is in, I will change this PR to leverage it.

deitch: Updated to use the `manifest-tool` in `calico/go-build:v0.14`

.gitignore
```diff
@@ -50,3 +50,6 @@ junit.xml
 /k8sfv/prometheus/data
 /k8sfv/prometheus/prometheus.yml
 /k8sfv/output/
+
+# tools
+manifest-tool
```
mkumatag: We don't need this anymore as we are using this tool inside the container.

deitch: Good point. Updating and squashing.

Makefile
```diff
@@ -254,6 +297,22 @@ ifeq ($(ARCH),amd64)
 	docker tag calico/felix:latest-$(ARCH) calico/felix:latest
 endif
 
+push-all: image-all imagetag $(addprefix sub-push-,$(ARCHES))
+sub-push-%:
+	$(MAKE) push ARCH=$* IMAGETAG=$(IMAGETAG)
+
+push: image imagetag
+	docker tag calico/felix:latest-$(ARCH) calico/felix:$(IMAGETAG)-$(ARCH)
```
fasaxc: Again, think I'd prefer push-only with failure if the image isn't there.  Better to fail fast than silently do the wrong thing.

Makefile
```diff
@@ -329,6 +388,7 @@ DOCKER_GO_BUILD := mkdir -p .go-pkg-cache && \
                               -e LOCAL_USER_ID=$(MY_UID) \
                               -v $${PWD}:/go/src/github.com/projectcalico/felix:rw \
                               -v $${PWD}/.go-pkg-cache:/go/pkg:rw \
+                              -v $${HOME}/.docker:/home/user/.docker \
```
fasaxc: What's this change for?

Makefile
deitch: Yeah @fasaxc and @caseydavenport convinced me of it a few minutes ago. I tried to address it in the latest commit to the [cni-plugin PR](https://github.com/projectcalico/cni-plugin/pull/516) by having the `push` target *only* push and a `tagimages` target tag correctly. It still simplifies greatly, but ensures that it only pushes the images if they exist.

deitch: I will apply that here too (I accidentally hit "Comment" before I was done).

Makefile
deitch: Ah, this is a relic from when this PR included multi-arch manifest. We decided to punt on it for now, doing everything short of it, so this should be removed. Will fix.

.gitignore
```diff
@@ -50,3 +50,5 @@ junit.xml
 /k8sfv/prometheus/data
 /k8sfv/prometheus/prometheus.yml
 /k8sfv/output/
+
+# tools
```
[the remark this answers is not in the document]
deitch: Also a relic from multi-arch. Will remove.

deitch:
Updated. Back at you @fasaxc

Makefile
```diff
@@ -254,6 +298,40 @@ ifeq ($(ARCH),amd64)
...
+push: imagetag
+	docker push $(CONTAINER_NAME):$(IMAGETAG)-$(ARCH)
+	docker push quay.io/$(CONTAINER_NAME):$(IMAGETAG)-$(ARCH)
+ifeq ($(ARCH),amd64)
+	docker push $(CONTAINER_NAME):$(IMAGETAG)
+	docker push quay.io/$(CONTAINER_NAME):$(IMAGETAG)
+endif
+
+## tag images of one arch
+tagimages: imagetag
```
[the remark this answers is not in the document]
deitch: Updated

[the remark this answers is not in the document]
fasaxc: You sure?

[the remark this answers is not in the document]
deitch: Ha! I pushed without committing. Doing again.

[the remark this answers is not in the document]
deitch: Now I am sure. :-)

PR #1751, opened 2018-03-27
Pin to using v0.1 of calico/protoc container

neiljerram:


djlwilder:
Humm..   PROTOC_VER is defined twice in the Makefile. For amd64 builds (and for cross builds) it is set to v0.1,  for all other arch native builds PROTOC_VER set to latest.    Only amd64 images have been tagged to v0.1 and pushed to the hub.

The bug here is that the second definition is using "=?"  (PROTOC_VER?=v0.1)  since PROTOC_VER was defined earlier this definition was never used.

I believe the correct fix should be to change ?= to = in the second define of PROTOC_VER.

neiljerram:
Sorry @djlwilder - just looking again at this now.

djlwilder:
@neiljerram See PR 1752

neiljerram:
@djlwilder Cool, thanks; I'll go with yours rather than mine, subject to CI passing, as it's simpler.

It leaves a somewhat inconsistent picture as regards whether it's possible to override all those vars - equally for all arches - when running make.  But that's really an independent matter, and the picture was already inconsistent in that way before my change earlier today.

PR #1752, opened 2018-03-27
Non-amd native build to use the latest build of protoc.

djlwilder:

## Description
Non-amd native build will use the latest protoc version.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -86,8 +86,6 @@ FV_FELIXIMAGE ?= calico/felix:latest-$(BUILDARCH)
 
 # If building on amd64 omit the arch in the container name.  Fixme!
 ifeq ($(BUILDARCH),amd64)
-        PROTOC_VER?=v0.1
-        PROTOC_CONTAINER=calico/protoc:$(PROTOC_VER)
         FV_ETCDIMAGE=quay.io/coreos/etcd:v3.2.5
         FV_K8SIMAGE=gcr.io/google_containers/hyperkube:v1.7.5
         FV_TYPHAIMAGE=calico/typha:v0.6.0-beta1-16-g512a0f2


djlwilder:
This PR is in response to PR1751  @fasaxc @neiljerram @deitch 

Another option is to force all arch native builds to use the v0.10 of protoc.  It would be up to the builder to insure calico/protoc:v0.10-$(BUILDARCH) is available.  If we use that approach then we only need a single definition of PROTOC_VER so it is a little cleaner.

Ultimately,  we should rename the amd64  protoc container  to calico/protoc:v0.10-amd64,  to make the name consistent for all archs.  That would be much cleaner.

deitch:
Oh sure. Force me to rebase! :-)

LGTM

neiljerram:
@djlwilder I'm sorry, but semaphore hasn't picked this up for some reason.  Please could you give it a kick by adding another commit like https://github.com/projectcalico/felix/pull/1751/commits/406a4a526d24dd66b38374e1c34c9d7e41509769

fasaxc:
I repushed v0.1 as `calico/protoc:v0.1-amd64`, does that help?

Makefile
```diff
@@ -86,7 +86,7 @@ FV_FELIXIMAGE ?= calico/felix:latest-$(BUILDARCH)
 
 # If building on amd64 omit the arch in the container name.  Fixme!
 ifeq ($(BUILDARCH),amd64)
-        PROTOC_VER?=v0.1
+        PROTOC_VER=v0.1
```
fasaxc: This prevents us from using an invocation like `make <something> PROTOC_VER=v0.2` for trying out a new protoc.  I'd prefer if we could keep the `?=`.  Maybe the non-amd64 case could be handled in an `else`?

djlwilder: >>  I repushed v0.1 as calico/protoc:v0.1-amd64, does that help?
yes that helps.  I can eliminate "PROTOC_CONTAINER=calico/protoc:$(PROTOC_VER)" from inside the amd ifeq.  The entire ifeq ($(BUILDARCH),amd64) block is a hack to handle the transition to the new image names.  The more I can eliminate from inside the ifeq the better.   

We could define PROTOC_VER ?= v0.1 (and GO_BUILD_VER ?= v0.10) just once outside the amd "if".  The amd build should work,  other arch native builds can override the values to meet the builders needs.

For example:
make <something> ARCH=ppc64le GO_BUILD_VER=<latest> PROTOC_VER=<latest>  

I think this is cleaner and easer to keep updated.  In the future if we push multi-arch manifest for the protoc and go-build images then all arch builds should just work.

djlwilder:
I updated and squashed.  The new commit message is:

Remove extra definitions of PROTOC_CONTAINER and PROTOC_VER.

Both values can be overridden to meet the builder's needs.
For example, you could do something like:

make calico/felix ARCH=ppc64le GO_BUILD_VER=latest PROTOC_VER=latest

In this example these images would be used:
calico/go-build:latest-ppc64le
calico/protoc:latest-ppc64le

djlwilder:
semaphoreci failure looks unrelated.

tmjd:
That does looks like a test flake I'm investigating. I've kicked off a re-run of the Semaphore tests.

deitch:
They like you better than me, @djlwilder . Look how quickly yours gets merged in! :-)

But now I need to rebase....

PR #1753, opened 2018-03-27
Fix protoc pinning change to work for multiarch

neiljerram:
Follow on from #1751


neiljerram:
Abandoning in favour of #1752

PR #1754, opened 2018-03-28
Fix package building following recent multiarch changes

neiljerram:


neiljerram:
@fasaxc Thanks, good point.

PR #1756, opened 2018-03-28
[WIP] Tweaks to get info on flake

tmjd:
## Description
Several commits here that were done while trying to fix/get info on some flakes:
- Changed check that was causing a flake in host_port_test.go checking the metrics port for reachablity
- Added to clean up HostEndpoints with KDD after tests.
- Added dumping the HostEndpoints when dumping out KDD when the DumpErrorData is used.
- Added specific GINKGO_FOCUS for fv tests because I was having trouble creating a proper focus with just GINKGO_ARGS
- Removed the parallelism of the batch tests. There was a flake that seemed to just stop mid test and I thought I'd remove any complexity around there. (See https://semaphoreci.com/calico/felix-2/branches/master/builds/950)
  - Looking at the current felix fv batches I did not see any use of the previous parallelism so thought this would be ok.

## Release Note

```release-note
None required
```

### DIFF
--- a/fv/host_port_test.go
+++ b/fv/host_port_test.go
@@ -17,8 +17,6 @@
 package fv_test
 
 import (
-	"time"
-
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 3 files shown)


tmjd:
Closing as everything we want to keep here has gone in under other PRs.

PR #1757, opened 2018-03-29
Ensure RPM directory exists

neiljerram:
This is another fix following recent multiarch changes - but really the
error was rpm/build-rpms assuming that a directory would exist, so we
fix that here.


neiljerram:
(grrr... `make deb rpm` worked, but just `make rpm` didn't, on a clean checkout)

neiljerram:
Tested locally, and semaphore is loaded and flaky today, so bypassing...

PR #1761, opened 2018-04-02
Make slow test failures obvious

tmjd:
## Description
Tests that caused a "SLOW TEST" failure with the k8sfv tests was not very obvious, the run returned a failure but then no failing tests were reported.  This PR causes a summary of the failing tests to be printed hopefully making that obvious.

## Release Note

```release-note
None required
```

### DIFF
--- a/k8sfv/run-test
+++ b/k8sfv/run-test
@@ -166,15 +166,15 @@ done
 if ${USE_TYPHA}; then
     # Run Typha.
     docker run --detach --privileged --name ${prefix}-typha \
-	   -e CALICO_DATASTORE_TYPE=kubernetes \
-	   -e TYPHA_LOGSEVERITYSCREEN=${TYPHA_LOG_LEVEL} \
-	   -e TYPHA_DATASTORETYPE=kubernetes \
-	   -e K8S_API_ENDPOINT=${k8s_api_endpoint} \
-	   -e K8S_CA_FILE=/testcode/k8sfv/output/apiserver.crt \
-	   -v ${PWD}:/testcode \
-	   -w /testcode/k8sfv/output \
-	   $FV_TYPHAIMAGE \
-	   /bin/sh -c "for n in 1 2; do calico-typha; done"
+       -e CALICO_DATASTORE_TYPE=kubernetes \
+       -e TYPHA_LOGSEVERITYSCREEN=${TYPHA_LOG_LEVEL} \
+       -e TYPHA_DATASTORETYPE=kubernetes \
+       -e K8S_API_ENDPOINT=${k8s_api_endpoint} \
+       -e K8S_CA_FILE=/testcode/k8sfv/output/apiserver.crt \
+       -v ${PWD}:/testcode \
+       -w /testcode/k8sfv/output \
+       $FV_TYPHAIMAGE \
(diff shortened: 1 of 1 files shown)


k8sfv/run-test
```diff
@@ -211,13 +211,27 @@ if test -n "${GINKGO_FOCUS}"; then
...
     # Run all the tests that are _not_ marked as "[slow]", and fail
     # the test if any of those takes longer than 2 minutes.  A k8sfv
     # test should be marked as "[slow]" if it normally takes longer
     # than 1 minute - so this allows a buffer of 1 extra minute to
     # allow for executor slowness.
     docker exec ${prefix}-felix /bin/sh -c \
-	   "${run} -ginkgo.skip \"\\[slow\\]\" -ginkgo.slowSpecThreshold 120" | perl -pe 'END { exit $status } $status=1 if /SLOW TEST|FAIL!/;'
+       "${run} -ginkgo.skip \"\\[slow\\]\" -ginkgo.slowSpecThreshold 120" | tee $tmp
+	# Fail if the docker command returned failure
+	test ${PIPESTATUS[0]} -eq 0
```
fasaxc: Does this definitely fail the whole script?  ISTR that a bit of care is needed inside an `if` block but I could be wrong.

k8sfv/run-test
```diff
@@ -211,13 +211,27 @@ if test -n "${GINKGO_FOCUS}"; then
...
+	# Grab the lines that have SLOW TEST until a line with ---------
```
fasaxc: You've got mixed tabs/spaces in this file, please can you fix up?  I usually go for 3-space indents in shell.

k8sfv/run-test
```diff
@@ -211,13 +211,27 @@ if test -n "${GINKGO_FOCUS}"; then
...
+    slow="$(awk '/SLOW TEST/,/--------/' $tmp)"
+	# If there were any slow tests then fail the test but print out
+	# the slow test lines.
+    if [ -n "$slow" ]; then
+        echo "====Slow tests detected===="
+        echo $slow
+        echo "====End of slow test summary===="
+        exit 1
+    fi
+    rm -f $tmp
```
fasaxc: You should delete this on the failure path too.

k8sfv/run-test
tmjd: I just tried this test script
```
#!/bin/bash -ex

x=1
if [ $x -eq 0 ]; then
    echo "x is 0"
else
    echo "Here"
    test 1 -eq 0
    echo "There"
fi
```
and got this output
```
> ./test.sh 
+ x=1
+ '[' 1 -eq 0 ']'
+ echo Here
Here
+ test 1 -eq 0
```
I am going to go ahead and change it though because I think a simple if statement with a message when it exits will be better.

k8sfv/run-test
tmjd: It looked like the 'tabs' were 4 spaces in there so I standardized on that. If you want me to update it all to 3, I can do that if you think I should.

tmjd:
Thanks for the review, I think I've addressed things here. PTAL

k8sfv/run-test
fasaxc: Also, I typoed above, meant to say 2-space.  2 or 4 is fine.  3 would be crazy!

PR #1763, opened 2018-04-02
Unexpected fix for HEP Policy fix

tmjd:
- Ensured that HostEndpoint is blocking traffic before adding policy to
  allow traffic.
- Changed Eventually test to a loop that checks longer than Eventually
  did to try to get a little more information if the flake happens
  again.  Unexplainably the change seemed to fix the flake.

## Release Note

```release-note
None required
```

### DIFF
--- a/fv/host_port_test.go
+++ b/fv/host_port_test.go
@@ -17,6 +17,8 @@
 package fv_test
 
 import (
+	"time"
+
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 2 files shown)


fv/host_port_test.go
```diff
@@ -145,7 +149,21 @@ var _ = infrastructure.DatastoreDescribe("with initialized Felix", []apiconfig.D
 			})
 
 			It("port should be reachable", func() {
-				Eventually(metricsPortReachable, "10s", "1s").Should(BeTrue())
+				//Eventually(metricsPortReachable, "10s", "1s").Should(BeTrue())
```
fasaxc: The fact that this seems to work around suggests that `metricsPortReachable` has a bug and takes >10s to time out, if that was the case then an Eventually(... 10, 1) would only run metricsPortReachable once so it would be flaky. 

Looking at the code that `metricsPortReachable` calls, i don't see any timeouts specified so I bet it defaults to something large like 60s.

tmjd: I just made that change (incorrectly) and noticed https://github.com/projectcalico/felix/pull/1763/files#diff-c663090db8354a82f2b72fd73f1dc935R28. Looks like there is already a 1 second timeout on the http.Get.
That was a great thought though.

tmjd:
I've ran this current changes 9 times and there hasn't been a failure.  So maybe we should go with this for now.
WDYT @fasaxc?

PR #1765, opened 2018-04-02
Change batch script to try to understand flakes

tmjd:
## Description
Drop the parallelism in the fv batches since I did not see it being used and to simplify the test run output.

## Release Note

```release-note
None required
```


fasaxc:
Not sure about this one, I only recently turned off the parallelism on semaphore to help diagnose the flakes.  It'd be nice to turn it back on (2-3 minute time reduction) if we can get the flakes under control...

PR #1774, opened 2018-04-05
FV: Trying to get info on HEP flake

tmjd:
The test being changed has been a flake and this is trying to get a
bit more information by checking for longer than than passing allows.

#1763 was the first change trying to fix this flake.

### DIFF
--- a/fv/host_port_test.go
+++ b/fv/host_port_test.go
@@ -17,6 +17,8 @@
 package fv_test
 
 import (
+	"time"
+
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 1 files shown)


tmjd:
Here is an instance of the test failing with the previous changes. https://semaphoreci.com/calico/felix-2/branches/pull-request-1773/builds/1

tmjd:
The failure in semaphore is a different flake :disappointed:

fv/host_port_test.go
```diff
@@ -146,7 +148,21 @@ var _ = infrastructure.DatastoreDescribe("with initialized Felix", []apiconfig.D
...
-				Eventually(metricsPortReachable, "10s", "1s").Should(BeTrue())
+				//Eventually(metricsPortReachable, "10s", "1s").Should(BeTrue())
+				// Switching to the loop below, *seems* to have fixed a flake
+				// that was happening here with the Eventually check above.
+				// The loop is purposefully high vs what is being checked so
+				// if another failure occurs we can see if the timeout is just
+				// a little too low or it is something more.
+				start := time.Now()
+				for time.Since(start).Seconds() < 50 {
+					reachable := metricsPortReachable()
```
fasaxc: Suggest you wrap this with ```
startTime=time.Now()
...
log.WithField("time", time.Since(startTime)).Info("Checked metrics port")
```
so that we find out exactly how long it took  (unless you're sure we can already work that out from the logs).

PR #1782, opened 2018-04-09
Stop healthchecks from listening on all interfaces.

ozdanborne:
## Description

Modify healthchecks so that they listen on `localhost` by default instead of `0.0.0.0`.

Exposed the adress as a Felix config option because in DC/OS, healthchecks come from the scheduler (marathon), not the node process (mesos-slave). So that is one known case where the healthcheck endpoint needs to be accessible beyond localhost.

## Todos
- [ ] Documentation
- [ ] Backport
- [x] Release note

## Release Note

```release-note
HealthCheck default behavior changed from listening on all interfaces to just listening on localhost. This behavior can be controlled by the new HealthHost parameter.
```

### DIFF
--- a/felix.go
+++ b/felix.go
@@ -211,7 +211,7 @@ configRetry:
 
 		// Each time round this loop, check that we're serving health reports if we should
 		// be, or cancel any existing server if we should not be serving any more.
-		healthAggregator.ServeHTTP(configParams.HealthEnabled, "", configParams.HealthPort)
+		healthAggregator.ServeHTTP(configParams.HealthEnabled, configParams.HealthHost, configParams.HealthPort)
 
 		// We should now have enough config to connect to the datastore
 		// so we can load the remainder of the config.
@@ -282,7 +282,7 @@ configRetry:
 	healthAggregator.Report(healthName, &health.HealthReport{Live: true, Ready: true})
 
 	// Enable or disable the health HTTP server according to coalesced config.
-	healthAggregator.ServeHTTP(configParams.HealthEnabled, "", configParams.HealthPort)
+	healthAggregator.ServeHTTP(configParams.HealthEnabled, configParams.HealthHost, configParams.HealthPort)
 
 	// If we get here, we've loaded the configuration successfully.
 	// Update log levels before we do anything else.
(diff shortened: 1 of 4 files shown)


felix.go
```diff
@@ -211,7 +211,7 @@ configRetry:
 
 		// Each time round this loop, check that we're serving health reports if we should
 		// be, or cancel any existing server if we should not be serving any more.
-		healthAggregator.ServeHTTP(configParams.HealthEnabled, "", configParams.HealthPort)
+		healthAggregator.ServeHTTP(configParams.HealthEnabled, "127.0.0.1", configParams.HealthPort)
```
ozdanborne: s/127.0.0.1/localhost/g

ozdanborne:
Raised libcalico-go PR: https://github.com/projectcalico/libcalico-go/pull/852

@fasaxc adding you as a reviewer to get a maintainer LGTM

fasaxc:
@ozdanborne looks good but it'd be good to have an FV test.  Since you had to fix up health_test.go to set the new flag, presumably you could start a felix without the new flag and simply assert that you can't access the flag from the other mock host?

ozdanborne:
Ok, I've added some FVs checking default behavior with localhost.

I would have liked to check that the healthchecks are working but since Felix isn't running with `--net=host` our FVs aren't executing in the same namespace as felix and thus cannot.

I looked into ways to run the check from inside felix's networking namespace but the infrastructure isn't there yet and I'm worried about making larger changes to it.

So I followed your advice and just made sure the checks fail when done outside the container with `localhost`.

fv/health_test.go
```diff
@@ -227,9 +228,33 @@ var _ = Describe("health tests", func() {
...
+		It("should run support running healthchecks on '127.0.0.1'", func() {
+			startFelix("", k8sInfra.GetDockerArgs, "", "", "127.0.0.1")
+			Consistently(felixReady, "10s", "100ms").Should(Equal(statusErr))
+		})
+
+		It("should support running healthchecks on 'localhost'", func() {
+			startFelix("", k8sInfra.GetDockerArgs, "", "", "localhost")
+			Consistently(felixReady, "10s", "100ms").Should(Equal(statusErr))
+		})
+
```
fasaxc: We don't have any dedicated infra, but checking from inside the container should be as simple as `felix.Exec("curl", "http://localhost:...")`.  I think that's worth doing assuming it's as easy as that.  WDYT?

PR #1785, opened 2018-04-10
Automatic update of GO_BUILD to v0.12

tomdee:


fasaxc:
Looks like you need to fix up some new lint?

fasaxc:
Ah, it's more subtle than that.  The pre-commit hook (not containerised) and `make static-checks` (containerised) now disagree.

Fixing up under #1792

PR #1791, opened 2018-04-13
Speculative work around for link not found errors in k8sfv.

fasaxc:
## Description

Use a netlink Handle rather than individual netlink sockets when setting up fake pods.  This seems to make it more reliable but I'm not sure of the reason.  It may be:

- reduction in load/churn of sockets
- kernel enforcing stricter concurrency guarantees if we re-use the socket
- socket gets locks to the netns it was created in so we avoid some cross-talk bug?

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```


fasaxc:
CI hit a different flake, which we think is already fixed on master so I've rebased...

PR #1793, opened 2018-04-13
For host endpoint tests, install an allow-to-datastore policy

fasaxc:
## Description

Looks like we were cutting off or own connectivity to the datastore in some tests.  KDD seems more susceptible to that issue; maybe it's more likely to start a new TCP connection than etcd?

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```


tmjd:
LGTM, my only thought is should we just add that allow by default with no selector (that's the same as all, right?)?  I'm thinking just as part of infrastructure setup for all tests.

fasaxc:
> should we just add that allow by default with no selector

I only want to allow from host endpoints (not workloads), and there's no way to say "all host endpoints" in our selector syntax.

PR #1798, opened 2018-04-16
Changes to allow libcalico-go upgrade for Felix.

saumoh:
## Description
Upgrade the `libcalico-go` pin's in Felix. Made changes to ensure that the ft's work.

NOTE: I had a hard time figuring out what the IPSet's sha244 mac should be now that ServiceAccount Labels are part of the selector. So I ended up copying the `SrcIpSetIds`! But if there is a alternate then please do suggest.


fasaxc:
Continued in #1799

Issue #1789, opened 2018-04-12
High felix CPU usage while disconnected from datastore


fasaxc:
Stopped an unrelated Felix FV test to do a bit of diagnosis, noted that top was reporting calico-felix using 180% CPU. 

At this point in the test, etcd had been stopped, only errors/warning in the log were
```
2018-04-12 13:44:57.763 [ERROR][8] etcdv3.go 445: Failed to grant a lease error=context deadline exceeded
2018-04-12 13:44:57.765 [WARNING][8] felix.go 866: Failed to write status to datastore: context deadline exceeded
2018-04-12 13:44:59.765 [WARNING][8] etcdv3.go 228: Apply failed error=context deadline exceeded etcdKey=StatusReport(hostname=felix-95082-2-felixfv) rev="" ttl=0s value=&model.StatusReport{Timestamp:"2018-04-12T13:44:55Z", UptimeSeconds:460.077638951, FirstUpdate:false}
2018-04-12 13:44:59.767 [WARNING][8] felix.go 878: Failed to write status to datastore: context deadline exceeded
```

## Expected Behavior

Felix had nothing to do so should be using almost no CPU.

## Current Behavior

High CPU with no obvious logs to help understand what might be tight looping.

## Steps to Reproduce (for bugs)

1.
2.
3.
4.

## Your Environment

* Felix version (if using Felix binary directly `calico-felix --version`):
* And/Or, `calico/node` container version (if running Felix in a container):
* Which orchestrator are you using (e.g. Kubernetes, OpenStack, Docker, Mesos):
* Orchestrator version (if used):
* Etcd version (if using etcd datastore driver):
* Operating System and version:
* Link to your project (optional):

fasaxc:
Looks like an issue in the etcd client library: https://github.com/coreos/etcd/issues/9578

PR #1802, opened 2018-04-18
Add support for dumping a CPU profile on SIGUSR2.

fasaxc:
## Description

Diags improvements made while investigating #1789 

Add (pended) FV test that verifies CPU usage while etcd is down.  (Currently fails due to #1789, which is caused by an upstream issue.)

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/felix.go
+++ b/felix.go
@@ -500,8 +500,8 @@ configRetry:
 		go servePrometheusMetrics(configParams)
 	}
 
-	// On receipt of SIGUSR1, write out heap profile.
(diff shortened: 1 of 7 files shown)


felix.go
```diff
@@ -501,7 +501,7 @@ configRetry:
 	}
 
 	// On receipt of SIGUSR1, write out heap profile.
```
neiljerram: Probably should update this comment.

PR #1811, opened 2018-04-20
Apply TLS to Felix-Typha connections

neiljerram:


neiljerram:
Yes, of course.  Sorry for forgetting that!

Makefile
```diff
@@ -88,7 +88,7 @@ FV_FELIXIMAGE ?= calico/felix:latest-$(BUILDARCH)
 ifeq ($(BUILDARCH),amd64)
         FV_ETCDIMAGE=quay.io/coreos/etcd:v3.2.5
         FV_K8SIMAGE=gcr.io/google_containers/hyperkube:v1.7.5
-        FV_TYPHAIMAGE=calico/typha:v0.6.0-beta1-16-g512a0f2
+        FV_TYPHAIMAGE=calico/typha:latest
```
fasaxc: As discussed, should be pinned to same commit as glide.yaml

fv/infrastructure/typha.go
```diff
@@ -0,0 +1,116 @@
...
+
+func RunTypha(infra DatastoreInfra, options TopologyOptions) *Typha {
+	log.Info("Starting typha")
+
+	args := infra.GetDockerArgs()
+	args = append(args,
+		"--privileged",
+		"-e", "TYPHA_LOGSEVERITYSCREEN="+options.TyphaLogSeverity,
+		"-e", "TYPHA_PROMETHEUSMETRICSENABLED=true",
+		"-v", "/lib/modules:/lib/modules",
```
fasaxc: Typha shouldn't need any modules mount.

fv/ingress_egress_test.go
```diff
@@ -220,3 +220,65 @@ var _ = Context("with initialized Felix, etcd datastore, 3 workloads", func() {
...
+
+		// Install a default profile that allows workloads with this profile to talk to each
+		// other, in the absence of any Policy.
+		defaultProfile := api.NewProfile()
+		defaultProfile.Name = "default"
+		defaultProfile.Spec.LabelsToApply = map[string]string{"default": ""}
+		defaultProfile.Spec.Egress = []api.Rule{{Action: api.Allow}}
+		defaultProfile.Spec.Ingress = []api.Rule{{
+			Action: api.Allow,
+			Source: api.EntityRule{Selector: "default == ''"},
```
fasaxc: I'd be tempted to make sure there's at least one path that gets denied here, just to be 100% sure that policy got applied rather than some fail-open bug.  Finding it pretty hard to think of a mechanism for _only_ TLS mode to fail open though so I'll leave it up to you.

PR #1812, opened 2018-04-26
Improve calico-diags redaction.

fasaxc:
## Description

Replace sensitive information with REDACTED instead of
removing whole line.

Avoids confusion where config file looks to be missing essential information!
## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/etc/calico-diags
+++ b/etc/calico-diags
@@ -76,12 +76,19 @@ fi
 
 cp -a "$CALICO_CFG" etc_calico
 
-# Use grep -v to attempt to strip out password information
-grep -v -i "password" /etc/nova/nova.conf > nova.conf 2>&1
-grep -v -i "password" /etc/neutron/neutron.conf > neutron.conf 2>&1
-grep -v -i "password" /etc/neutron/plugins/ml2/ml2_conf.ini > ml2_conf.ini 2>&1
-grep -v -i "password" /etc/neutron/dhcp_agent.ini > dhcp_agent.ini 2>&1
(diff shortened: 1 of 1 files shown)


etc/calico-diags
```diff
@@ -77,11 +77,18 @@ fi
 cp -a "$CALICO_CFG" etc_calico
 
 # Use grep -v to attempt to strip out password information
```
neiljerram: Comment needs deleting or updating now.

etc/calico-diags
```diff
@@ -77,11 +77,18 @@ fi
...
-grep -v -i "password" /etc/nova/nova.conf > nova.conf 2>&1
-grep -v -i "password" /etc/neutron/neutron.conf > neutron.conf 2>&1
-grep -v -i "password" /etc/neutron/plugins/ml2/ml2_conf.ini > ml2_conf.ini 2>&1
-grep -v -i "password" /etc/neutron/dhcp_agent.ini > dhcp_agent.ini 2>&1
-grep -v -i "password" /etc/neutron/metadata_agent.ini > metadata_agent.ini 2>&1
+function copy_and_redact() {
+  in=$1
+  out=$2
+
+  sed -e 's/^\([ \t]*\(password\|token\|transport_url\|connection\)[ \t]*=[ \t]*\).*/\1REDACTED/' "$in" > $out 2>&1
```
neiljerram: Wondering about the initial `^` here - is it needed?

Without it, this expression would also catch - for example - `auth_token = xxx`.  But as it stands, I believe it would miss that.

PR #1815, opened 2018-04-26
[WIP] HTTP Path rule convert over to protobuf.

saumoh:
## Description
As the PR for `libcalico-go` is under review (https://github.com/projectcalico/libcalico-go/pull/858) I though it best to get the changes in Felix under review sooner than later.
The key change is the protobuf definition and the rest is glue to covert from model to protobuf.

Assuming that `libcalico-go` model isn't changed during review the Felix side should be stable enough for a preliminary review.

### DIFF
--- a/calc/rule_convert.go
+++ b/calc/rule_convert.go
@@ -21,6 +21,7 @@ import (
 	log "github.com/sirupsen/logrus"
 
 	"github.com/projectcalico/felix/proto"
+	api "github.com/projectcalico/libcalico-go/lib/apis/v3"
 	"github.com/projectcalico/libcalico-go/lib/net"
 	"github.com/projectcalico/libcalico-go/lib/numorstring"
 )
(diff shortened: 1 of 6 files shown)


calc/rule_convert.go
```diff
@@ -129,7 +130,31 @@ func parsedRuleToProtoRule(in *ParsedRule) *proto.Rule {
...
+					"value": value,
+				}).Debug("pathMatch.")
+
+				switch key {
+				case string(api.HTTPPathMatchExact):
+					paths = append(paths, &proto.HTTPMatchPathMatchTypes{PathMatch: &proto.HTTPMatchPathMatchTypes_Exact{Exact: value}})
+				case string(api.HTTPPatchMatchPrefix):
+					paths = append(paths, &proto.HTTPMatchPathMatchTypes{PathMatch: &proto.HTTPMatchPathMatchTypes_Prefix{Prefix: value}})
+				default:
+					log.Debug("Invalid pathMatchkey")
```
fasaxc: I think this should be an Error.

calc/rule_convert.go
```diff
@@ -129,7 +130,31 @@ func parsedRuleToProtoRule(in *ParsedRule) *proto.Rule {
 	}
 
 	if in.HTTPMatch != nil {
-		out.HttpMatch = &proto.HTTPMatch{Methods: in.HTTPMatch.Methods}
+		out.HttpMatch = &proto.HTTPMatch{}
+		paths := []*proto.HTTPMatchPathMatchTypes{}
```
fasaxc: Prefer `var paths []*proto.HTTPMatchPathTypes` over this.  The difference is that it avoids allocating a 0-length slice if it's not needed (some of the linters flag this up).

```
len(nil) == 0
```
so it's safe to do that.

saumoh: good to know. I initially had it as a var.
Will add a few fv's as well. Thanks for the review.

PR #1816, opened 2018-05-01
Allow others to import and run Felix

caseydavenport:
## Description

Minor refactor - lets others import "github.com/projectcalico/felix/felix" and call `felix.Run("")`.

Not perfect, since Felix calls exit a bunch so with just this PR you couldn't, say, run felix as one goroutine amongst many, but might be a neat place to end up at.

With this PR though it can be compiled in with other packages that share the same code in order to reduce binary size by sharing dependencies. 

WDYT?

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -492,7 +492,7 @@ bin/calico-felix-$(ARCH): $(FELIX_GO_FILES) vendor/.up-to-date
 	@echo Building felix for $(ARCH) on $(BUILDARCH)
 	mkdir -p bin
 	$(DOCKER_GO_BUILD) \
-	   sh -c 'go build -v -i -o $@ -v $(LDFLAGS) "github.com/projectcalico/felix" && \
+	   sh -c 'go build -v -i -o $@ -v $(LDFLAGS) "github.com/projectcalico/felix/cmd/calico-felix" && \
 		( ldd $@ 2>&1 | grep -q -e "Not a valid dynamic program" \
 		-e "not a dynamic executable" || \
 		( echo "Error: $@ was not statically linked"; false ) )'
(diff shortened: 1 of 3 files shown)


fasaxc:
Step in the right direction, I think, but I structured this a bit differently when building Typha and I think that worked fairly well (I was just trying to stick much more closely to the recommended go package structure):

- `cmd/calico-typha/calico-typha.go` https://github.com/projectcalico/typha/blob/master/pkg/daemon/daemon.go is minimal and just calls through to a daemon struct:
- `pkg/daemon/daemon.go` https://github.com/projectcalico/typha/blob/master/pkg/daemon/daemon.go

The daemon struct is actually testable (yay!) since it breaks up the startup sequence into chunks.

caseydavenport:
@fasaxc I've updated the package structures to match. Not sure if you want me to also try to split out the Felix startup code into pieces like the Typha code. I haven't done it for now since I think it's not required, but could give it a go if you think it's worthwhile.

I think this separation stands on its own as a useful increment.

fasaxc:
> Not sure if you want me to also try to split out the Felix startup code into pieces like the Typha code

No, I'm not proposing that; it'd be a risky step unless we were going to do the UTs for that at the same time.

I _think_ I'd prefer not to introduce a pkg dir just for this one package.  Maybe we should put this at `./daemon/daemon.go` and then do the move into `pkg` as an orthogonal step.

caseydavenport:
@fasaxc SGTM. Should be good now I think.

PR #1821, opened 2018-05-10
Fix go-meta-linter errors

neiljerram:
/go/src/github.com/projectcalico/felix/calc/calc_graph_test.go:231:6: lastStats declared but not used
couldn't load packages due to errors: github.com/projectcalico/felix/calc_test
fv/infrastructure/infra_k8s.go:273:2: this value of info is never used (SA4006)
fv/infrastructure/topology.go:78:11: this value of client is never used (SA4006)

### DIFF
--- a/fv/infrastructure/topology.go
+++ b/fv/infrastructure/topology.go
@@ -25,7 +25,6 @@ import (
 	"regexp"
 
 	"github.com/projectcalico/felix/fv/containers"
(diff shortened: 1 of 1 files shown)


neiljerram:
@fasaxc The first commit of this PR got merged with the Append test, but there's still a further delta here to  avoid creating an etcd-based client twice.

PR #1831, opened 2018-05-30
Check in protobufs

caseydavenport:
## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```


Makefile
```diff
@@ -433,8 +433,11 @@ ifneq ("$(ARCH)","ppc64le") # no ppc64le support in centos6
 endif
 	utils/make-packages.sh rpm
 
-.PHONY: protobuf
-protobuf: proto/felixbackend.pb.go
+protobuf: clean-protobuf proto/felixbackend.pb.go
```
fasaxc: How about this.  If we add the protobuf target to the same rule as the pb.go one then you should be able to force a rebuild without the extra clean step?
```
.PHONY: protobuf
protobuf proto/felixbackend.pb.go: proto/felixbackend.proto
	$(DOCKER_RUN_RM) -v $${PWD}/proto:/src:rw \
	              $(PROTOC_CONTAINER) \
	              --gogofaster_out=plugins=grpc:. \
	              felixbackend.proto
```

PR #1834, opened 2018-06-10
Replicate all semaphore activities in ci and cd targets

deitch:
## Description

Replicate the workflow from [kube-controllers PR](https://github.com/projectcalico/kube-controllers/pull/258) here:

* Create a `make ci` target that does everything in the various semaphore jobs _except_ tag images and push to registries
* Create a `make cd` target that _only_ tags images and pushes to registries

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

As discussed with @caseydavenport @fasaxc @tomdee

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -466,6 +466,19 @@ fv/fv.test: vendor/.up-to-date $(SRC_FILES)
 	$(DOCKER_GO_BUILD) go test ./$(shell dirname $@) -c --tags fvtests -o $@
 
 .PHONY: fv
+# runs all of the fv tests
+# to run it in parallel, decide how many parallel engines you will run, and in each one call:
+#         $(MAKE) fv FV_BATCHES_TO_RUN="<num>" FV_NUM_BATCHES=<num>
+# where
+#         FV_NUM_BATCHES = total parallel batches
+#         FV_BATCHES_TO_RUN = which number this is
+# e.g. to run it in 10 parallel runs:
+#         $(MAKE) fv FV_BATCHES_TO_RUN="1" FV_NUM_BATCHES=10     # the first 1/10
+#         $(MAKE) fv FV_BATCHES_TO_RUN="2" FV_NUM_BATCHES=10     # the second 1/10
+#         $(MAKE) fv FV_BATCHES_TO_RUN="3" FV_NUM_BATCHES=10     # the third 1/10
+#         ...
+#         $(MAKE) fv FV_BATCHES_TO_RUN="10" FV_NUM_BATCHES=10    # the tenth 1/10
+#         etc.
 fv fv/latency.log: calico/felix bin/iptables-locker bin/test-workload bin/test-connection fv/fv.test
 	cd fv && \
 	  FV_FELIXIMAGE=$(FV_FELIXIMAGE) \
@@ -505,6 +518,10 @@ K8SFV_GO_FILES:=$(shell find ./$(K8SFV_DIR) -name prometheus -prune -o -type f -
 
 .PHONY: k8sfv-test k8sfv-test-existing-felix
 # Run k8sfv test with Felix built from current code.
+# control whether or not we use typha with USE_TYPHA=true or USE_TYPHA=false
+# e.g.
+#       $(MAKE) k8sfv-test JUST_A_MINUTE=true USE_TYPHA=true
+#       $(MAKE) k8sfv-test JUST_A_MINUTE=true USE_TYPHA=false
 k8sfv-test: calico/felix k8sfv-test-existing-felix
 # Run k8sfv test with whatever is the existing 'calico/felix:latest'
 # container image.  To use some existing Felix version other than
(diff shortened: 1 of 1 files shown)


deitch:
Note: once merged, should change Semaphore, as we did with `kube-controllers`

deitch:
This one has been held up because of the parallelism. I have been looking for a way to stick with the `make ci` pattern, but still get potential parallelism. 

The current change creates the `ci` target as follows:

```Makefile
ci: image bin/calico-felix.exe ut upload-to-coveralls static-checks
	$(MAKE) fv FV_BATCHES_TO_RUN="1" FV_NUM_BATCHES=6
	$(MAKE) fv FV_BATCHES_TO_RUN="2" FV_NUM_BATCHES=6
	$(MAKE) fv FV_BATCHES_TO_RUN="3" FV_NUM_BATCHES=6
	$(MAKE) fv FV_BATCHES_TO_RUN="4" FV_NUM_BATCHES=6
	$(MAKE) fv FV_BATCHES_TO_RUN="5" FV_NUM_BATCHES=6
	$(MAKE) fv FV_BATCHES_TO_RUN="6" FV_NUM_BATCHES=6
	$(MAKE) k8sfv-test JUST_A_MINUTE=true USE_TYPHA=true
	$(MAKE) k8sfv-test JUST_A_MINUTE=true USE_TYPHA=false
```

This would do all of the following in serial:

* `image`
* `ut`
* `upload-to-coveralls`
* `static-checks`
* 6 runs each of 1/6 of the `fv` tests (leading to complete coverage)
* 2 runs each of the `k8sfv-test`, one each with or without `typha`

The way that Semaphore does it is to do the following in parallel, each in a separate VM with one core (all of them build the image first):

* each of the 6 `fv` runs
* each of the 2 `k8sfv-test` runs
* `ut` and `upload-to-coveralls`
* `static-checks`

I am going to suggest we do the following to move us at least a good step forward:

1. `make ci` builds binary and image, and runs basic tests - `ut upload-to-coveralls static-checks`
2. `make cd` tages and pushes (like in all the other targets)
3. `make fv` and `make k8sfv-test` do the individual parts, and are kept parallel, not wrapped into `make ci`

The result would look something like this:

```Makefile
ci: image bin/calico-felix.exe ut upload-to-coveralls static-checks

cd:
ifndef CONFIRM
	$(error CONFIRM is undefined - run using make <target> CONFIRM=true)
endif
ifndef BRANCH_NAME
	$(error BRANCH_NAME is undefined - run using make <target> BRANCH_NAME=var or set an environment variable)
endif
	$(MAKE) tag-images push IMAGETAG=${BRANCH_NAME}
	$(MAKE) tag-images push IMAGETAG=$(shell git describe --tags --dirty --always --long)
```

We leave `$(MAKE) fv` and `$(MAKE) k8sfv-test` lines out of `make ci` so they continue to be run in parallel. The build and the deploy at least are kept clean.

I believe this moves us forward, and we can tackle `fv` and `k8sfv-test` and its parallelism (if we choose to) a different time.

Thoughts?

deitch:
Updated as follows:

1. Rebase to remove conflict
2. Placed `ci` and `cd` targets in the correct section under its header
3. Updated header to `CI/CD`
4. Moved all parallel tests (`fv` and `k8sfv-test`) out of `ci`, so can be run in parallel

For the last bit, I was thinking that we might not want `make ci` *not* to run all of those tests, since `make ci` should "just work".

How about the following?

* `make ci` - runs all tests, in serial, including those that might be parallelized
* `make ci SERIALONLY=true` - runs all tests _except_ for those that should be parallelized, i.e. `fv` and `k8sfv-test`

Makefile
```diff
@@ -466,6 +466,16 @@ fv/fv.test: vendor/.up-to-date $(SRC_FILES)
...
 .PHONY: fv
+# runs all of the fv tests
+# to run it in parallel, decide how many parallel engines you will run, and in each one call:
+#         $(MAKE) fv FV_BATCHES_TO_RUN="<num>" FV_NUM_BATCHES=<num>
+# where
+#         FV_NUM_BATCHES = total parallel batches
+#         FV_BATCHES_TO_RUN = which number this is
+# e.g. to run it in 10 parallel runs:
+#         $(MAKE) fv FV_BATCHES_TO_RUN="1" FV_NUM_BATCHES=10     # the first 1/10
+#         $(MAKE) fv FV_BATCHES_TO_RUN="1" FV_NUM_BATCHES=10     # the next 1/10
```
fasaxc: Typo: `FV_BATCHES_TO_RUN="1"` (should be 2 on this line)

Makefile
```diff
@@ -575,8 +589,30 @@ bin/test-connection: $(SRC_FILES) vendor/.up-to-date
...
+
+## Deploy images to registry
+cd:
+ifndef CONFIRM
+	$(error CONFIRM is undefined - run using make <target> CONFIRM=true)
+endif
+ifndef BRANCH_NAME
+	$(error BRANCH_NAME is undefined - run using make <target> BRANCH_NAME=var or set an environment variable)
+endif
+	$(MAKE) tag-images push IMAGETAG=${BRANCH_NAME}
```
fasaxc: I think you need `$(BRANCH_NAME)` here instead of `${BRANCH_NAME}`

Makefile
deitch: Good point, will fix.

Makefile
deitch: Fixing.

deitch:
Fixed both @fasaxc . One of them had carried over to `typha`, so fixed there too.

PR #1836, opened 2018-06-13
Strip the `v` from the version number when building debs

caseydavenport:
## Description

Allows us to use a `v` like other calico components in the `calico/felix` release, but when building the debs they seem to need to start with a number. 

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/utils/lib.sh
+++ b/utils/lib.sh
@@ -40,6 +40,10 @@ function git_commit_id {
     git rev-parse HEAD | cut -c-7
 }
 
+function strip_v {
+	echo $1 | cut -d "v" -f 2
+}
+
 # Convert PEP 440 version to Debian.
 function git_version_to_deb {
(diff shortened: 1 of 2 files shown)


caseydavenport:
@fasaxc does this seem like the right thing to do more or less for allowing Felix to have version numbers starting with a `v` like everyone else?

fasaxc:
Huh, I thought I saw another big Makefile PR go by overnight?  Was this included in that?

caseydavenport:
@fasaxc just needed a rebase

caseydavenport:
> What about when we release debs

hmmm. is this not the script that makes those debs? I had thought it was, but perhaps I'm missing something...

fasaxc:
Re-reading the RELEASING doc, looks like it's changed and it _does_ rely on `make deb rpm`.

PR #1837, opened 2018-06-13
Clean up Makefile

tomdee:
Reordered Makefile to fit the common format
Renamed variables to match other Makefiles
Did a bit of general cleanup

Made vendor behave the same as every other repo
- Created using Docker
- Not PHONY
- No update
- No up-to-date business

Not doing
- Removing .go-pkg.cache from clean (no other build removes this in
  clean)
- Adding CI target. Felix is just too complicated at the moment, doing
  too many things in parallel (and @deitch is already working on this)
- Checking if all the SSH_AUTH_SOCK is really needed (I don't think
    other builds do it)
- Removing or modifying the diagram at the top. I'm not sure if it's
    accurate and/or useful

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,16 +42,21 @@
 #
 #
 ###############################################################################
+# Shortcut targets
+default: build
+
+## Build binary for current platform
+all: build
+
+## Run the tests for the current platform/architecture
+test: ut fv
+
+###############################################################################
 # Both native and cross architecture builds are supported.
 # The target architecture is select by setting the ARCH variable.
 # When ARCH is undefined it is set to the detected host architecture.
 # When ARCH differs from the host architecture a crossbuild will be performed.
-
-# restore this when all arches work. For now, there are qemu issues related to cross-compile on s390x
-#   see https://github.com/projectcalico/go-build/pull/32
-#ARCHES=$(patsubst docker-image/Dockerfile.%,%,$(wildcard docker-image/Dockerfile.*))
-ARCHES=amd64 arm64 ppc64le
-
+ARCHES=$(patsubst docker-image/Dockerfile.%,%,$(wildcard docker-image/Dockerfile.*))
 
 # BUILDARCH is the host architecture
 # ARCH is the target architecture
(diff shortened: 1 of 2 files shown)


tomdee:
@deitch Ooops, I couldn't see anything that used them (and didn't spot them in any of the other repos). I thought all cross builds happen inside the go-build container now?

Makefile
```diff
@@ -332,112 +300,189 @@ sub-tag-images-%:
...
+	$(DOCKER_GO_BUILD) sh -c 'glide nv | xargs -n 3 staticcheck'
+	$(DOCKER_GO_BUILD) gometalinter --deadline=300s \
+	                                --disable-all \
+	                                --enable=goimports \
+	                                --vendor ./...
+
+# Run go fmt on all our go files.
+.PHONY: go-fmt goimports fix
+fix go-fmt goimports:
+	$(DOCKER_GO_BUILD) sh -c 'glide nv -x | \
```
caseydavenport: Indentation looks pretty funky here.

deitch:
> I couldn't see anything that used them

@tomdee , isn't `register` used [here](https://github.com/projectcalico/felix/blob/master/Makefile#L282)? 

> I thought all cross builds happen inside the go-build container now?

(typing this out "stream of consciousness" to think it through)

The compile itself, yes, since we just run, e.g. `go` on the `amd64` arch and build for `arm64`, `ppc64le`, etc. 

The image builder needs to be able to run `apk` (the target arch version, e.g. `arm64`) while running on `amd64`. To do that, we need `qemu`. Looking at the [Dockerfile.arm64](https://github.com/projectcalico/felix/blob/master/docker-image/Dockerfile.arm64) as an example, 3 stages:

1. Get the `qemu` binaries, happens [here](https://github.com/projectcalico/felix/blob/master/docker-image/Dockerfile.arm64#L1-L10)
2. Run the `RUN apk` to install stuff happens [here](https://github.com/projectcalico/felix/blob/master/docker-image/Dockerfile.arm64#L12-L31)
3. Copy all the relevant files from the previous step into a clean container (to avoid all of the `qemu` and everything else polluting it) [here](https://github.com/projectcalico/felix/blob/master/docker-image/Dockerfile.arm64#L35-L46)

The thing is, for step 2 to work `RUN apk` (where `apk` is the `arm64` version), we need `qemu`, which we got in [line 18](https://github.com/projectcalico/felix/blob/master/docker-image/Dockerfile.arm64#L18):

```Dockerfile
COPY --from=qemu /usr/bin/qemu-*-static /usr/bin/
```

But we also need to register that `qemu` to handle the `arm64` binaries the kernel encounters (i.e. `binfmt`). Without it, it doesn't know. That is what `register` is supposed to do. Without it, I don't see how it would work. 

You wrote that you tried it and it built. Are you running on docker4mac? That has `arm64` `binfmt` support shipped natively, so it could be that.

As far as other packages, most are simpler and don't need it (and some may be missing it). E.g. [confd](https://github.com/projectcalico/confd/blob/master/Dockerfile.arm64) just does a straight `ADD`, no `RUN`, so no need for `binfmt`.

```Dockerfile
FROM scratch
LABEL maintainer "David Wilder <wilder@us.ibm.com>"

# Copy in the binary.
ADD bin/confd-arm64 /bin/confd

CMD /bin/confd
```

I think we still need `register`. We might be able to make it part of the `Dockerfile`, but we would have to add it to multiple. Would that buy us anything? Well, maybe, since then it is all encapsulated.

tomdee:
@deitch Thanks for the thorough explanation! I agree with all of that and have added it back to the Makefile :+1:

caseydavenport:
> However, the other repos are wrong, put back the vendor machinery and port it to the other repos if you're worried about the inconsistency.

I disagree. I've never seen problems with it and I think we just need to be comfortable with actually using the glide tool itself. I don't see a reason to encode it all into the makefile - just run `glide up -v` and it works.

It's a huge PITA when the Makefile overwrites the vendor directory I just checked out because it uses some special file flag that glide doesn't know to write.

>  Don't break my repo

Other people build this code too! :laughing:

fasaxc:
I've seen lots of problems with glide; some that affect me personally and some that have bitten (particularly new) developers on the project:

- partial checkouts not being picked up on a rebuild (I've lost hours to this one when go starts picking up the wrong version of a lib from your gopath because it's missing from vendor)
- shooting myself in the foot by accidentally modifying the directory (dir mtimes are too easy to update)
- running with a stale glide cache so reverting someone else's rev of a dependency (less of an issue for libcalico-go now we enforce the rule about using a commit hash but still an issue for other deps that float)
- running the wrong version of glide (not an issue now they've stopped developing it)
- running without `-v` (not an issue now, it's the default)
- glide's cache logic being broken when you switch between branches that use a different repo for their dependencies.

Using an "up-to-date" file and running glide in the container works around all those subtle time-wasting issues. 

The current incarnation is slow, that's the only big downside I'm aware of.  The reason is that I (reluctantly) disabled the cache entirely because I couldn't get it to work consistently and reliably when changing branches with the various bugs in glide :-(

> Other people build this code too! 

Same to you (and I dare say that I build, and `glide up`, this repo a lot more often than you do)!

> It's a huge PITA when the Makefile overwrites the vendor directory

This is hard to argue with; you didn't use the process that I set up and then complain that your own ad-hoc process doesn't work with the process I set up!  We can argue about whether the process is right and whether there'd be advantages to changing it but saying "the process doesn't work if I don't follow it" isn't useful.

tomdee:
I've reverted the vendor changes - running through CI now

caseydavenport:
> This is hard to argue with; you didn't use the process that I set up and then complain that your own ad-hoc 

Not much to say on this. We're using `glide` for dependency management, so I don't see why `glide up` is not an expected workflow.

It works everywhere else. 

> Same to you (and I dare say that I build, and glide up, this repo a lot more often than you do)!

That's not my point. It was more to address the "this is mine" mentality which I think is undesirable. 

Anyway, Tom added the vendor stuff back so let's just get this merged.

PR #1838, opened 2018-06-14
Add option to specify port-range for outgoing NAT.

databus23:
This commit allow to optionally specify a port range for the outgoing NAT rules that are created by felix.

## Description
This option allows to specify the port range used by iptables for outgoing NAT (e.g. `MASQUERADE`). We are running calico on a VMware based OpenStack with security groups implemented by a stateless firewall (VMware distributed virtual switch, DVS). This imposes some limitations how `egress` rules are implemented. In fact we need to ensure that outgoing traffic is coming from the default ephemeral port range (32768-65535). This works fine for traffic initiated by processes from the host network but breaks for the default masquerading case. The reason being that iptables picks arbitrary ports to resolve port collisions by default.

I realise that this is a rather exotic issue but the proposed configuration change is rather generic and fully optional.
I'm hoping to get this merged so that we can use upstream calico in our stack.  Would this change be generally acceptable to you?

## Todos
- [x] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/config/config_params.go
+++ b/config/config_params.go
@@ -180,7 +180,7 @@ type Config struct {
 	FailsafeOutboundHostPorts []ProtoPort `config:"port-list;udp:53,udp:67,tcp:179,tcp:2379,tcp:2380,tcp:6666,tcp:6667;die-on-fail"`
 
 	KubeNodePortRanges []numorstring.Port `config:"portrange-list;30000:32767"`
-	NatPortRange       string             `config:"portrange;"`
+	NatPortRange       string             `config:"portrange;;local"`
 
 	UsageReportingEnabled          bool          `config:"bool;true"`
 	UsageReportingInitialDelaySecs time.Duration `config:"seconds;300"`
--- a/config/config_params.go
+++ b/config/config_params.go
@@ -180,6 +180,7 @@ type Config struct {
 	FailsafeOutboundHostPorts []ProtoPort `config:"port-list;udp:53,udp:67,tcp:179,tcp:2379,tcp:2380,tcp:6666,tcp:6667;die-on-fail"`
 
 	KubeNodePortRanges []numorstring.Port `config:"portrange-list;30000:32767"`
+	NatPortRange       string             `config:"portrange;"`
 
 	UsageReportingEnabled          bool          `config:"bool;true"`
 	UsageReportingInitialDelaySecs time.Duration `config:"seconds;300"`
@@ -516,6 +517,8 @@ func loadParams() {
 			param = &EndpointListParam{}
 		case "port-list":
 			param = &PortListParam{}
+		case "portrange":
+			param = &PortRangeParam{}
 		case "portrange-list":
 			param = &PortRangeListParam{}
 		case "hostname":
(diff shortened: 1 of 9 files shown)


fasaxc:
Sounds like a reasonable tweak to me, to accept it, I think we'd need these other changes:

- Add UTs in the rules package to check that the new rules render as expected.
- So this can be configured with calicoctl, it needs to be exposed in libcalico-go here: https://github.com/projectcalico/libcalico-go/blob/master/lib/apis/v3/felixconfig.go
- So other people can find it, it needs to be added to the docs here: https://github.com/projectcalico/calico/blob/master/master/reference/felix/configuration.md

databus23:
@fasaxc Great! I added a unit test, verified it manually and added to more PR as requested.

rules/nat_test.go
```diff
@@ -55,6 +56,37 @@ var _ = Describe("NAT", func() {
 			},
 		}))
 	})
+	It("should render rules when active with explicit port range", func() {
+
+		//copy struct
+		localConfig := rrConfigNormal
+		localConfig.NATPortRange = numorstring.Port{99, 100, ""}
```
fasaxc: I think there's a utility function for making each type of port.  `numorstring.PortRange(a, b)`, maybe?

rules/nat.go
```diff
@@ -26,13 +27,39 @@ func (r *DefaultRuleRenderer) NATOutgoingChain(natOutgoingActive bool, ipVersion
...
+						Protocol("tcp"),
+				},
+				{
+					Action: iptables.MasqAction{ToPorts: toPorts},
+					Match: iptables.Match().
+						SourceIPSet(masqIPsSetName).
+						NotDestIPSet(allIPsSetName).
+						Protocol("udp"),
+				},
+				{
```
fasaxc: Just occurred to me, is MASQUERADE a "final" action; once you've hit the tcp rule above, will the packet also hit this rule and get double NATted, or something?

fasaxc:
libcalico-go PR: https://github.com/projectcalico/libcalico-go/pull/893

rules/nat.go
databus23: Hmm, good question. I obviously thought it was. Let me verify that...

databus23: Of course its not. Good catch! I updated the code to exit early for tcp and udp.

databus23:
@fasaxc I addressed all your feedback in all three PRs.

config/config_params.go
```diff
@@ -180,6 +180,7 @@ type Config struct {
 	FailsafeOutboundHostPorts []ProtoPort `config:"port-list;udp:53,udp:67,tcp:179,tcp:2379,tcp:2380,tcp:6666,tcp:6667;die-on-fail"`
 
 	KubeNodePortRanges []numorstring.Port `config:"portrange-list;30000:32767"`
+	NatPortRange       numorstring.Port   `config:"portrange;;local"`
```
fasaxc: Just spotted the `local` on here; that shouldn't be needed - no reason to limit this config to env vars only

fasaxc: Maybe you added this to get the tests to pass without the libcalico-go change?

databus23: Yes, I added that to make the tests pass. I removed it as requests.

databus23:
@fasaxc all done

PR #1841, opened 2018-06-18
Allow hep-forwarded traffic by default

neiljerram:


caseydavenport:
@neiljerram this feels like it probably warrants a release note since it's a default behavior change. Could you add one to the PR?

neiljerram:
@caseydavenport Done.

PR #1851, opened 2018-07-02
[WiP] Cache vendor dir on semaphore.

fasaxc:
## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```


tomdee:
@fasaxc Do we/you still want this change?

fasaxc:
Needs tidying up (moving to a script probably), but I think it'd be a good idea.

Rather than the semaphore-specific version here, how about we have a CACHE_DIR variable that defaults to something useful on our machines but gets overridden by the semaphore job?

OTOH, if we're about to switch to vgo, maybe we don't need anything like this?

PR #1852, opened 2018-07-05
ci and cd push all arches; ability to exclude specific

deitch:
#### Build and push by default for all supported arches
Before this PR, `ci` and `cd` targets build the binary and the image and push to registry _only_ for `amd64` unless explicitly choosing to add, e.g `make cd ARCH=arm64`. 

This PR makes `ci` and `cd` call the `*-all` targets, thus building binaries and images and pushing to registries for _all_ known arches.

#### Escape hatch to exclude arches
For example:

* `make cd` - push all known arches
* `make cd EXCLUDEARCH=arm64` - push all known arches except `arm64`
* `make cd EXCLUDEARCH="arm64 s390x"` - push all known arches except `arm64` and `s390x`

Same for `make ci`

#### Pull qemu from go-build instead of downloading every time
We have had the `qemu` binaries in `calico/go-build` since v0.15. Why redownload them? Just install them using multistage

#### Install only relevant qemu
Install only the relevant qemu binaries for the given arch, rather than all

#### Exclude s390x
Until such time as we can get `calico/go-build` to work with qemu (or any other interpreter), there is no point in trying to build here.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
Build binaries and images, and push images, for all known arches.
```

cc @fasaxc @tomdee @caseydavenport

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -96,6 +96,13 @@ register:
 ifneq ($(BUILDARCH),$(ARCH))
 	docker run --rm --privileged multiarch/qemu-user-static:register || true
 endif
+
+
+# list of arches *not* to build when doing *-all
+#    until s390x works correctly
+EXCLUDEARCH ?= s390x
+VALIDARCHES = $(filter-out $(EXCLUDEARCH),$(ARCHES))
+
 ###############################################################################
 CONTAINER_NAME=calico/felix
 PACKAGE_NAME?=github.com/projectcalico/felix
@@ -207,7 +214,7 @@ clean:
 # Building the binary
 ###############################################################################
 build: bin/calico-felix
-build-all: $(addprefix sub-build-,$(ARCHES))
+build-all: $(addprefix sub-build-,$(VALIDARCHES))
 sub-build-%:
 	$(MAKE) build ARCH=$*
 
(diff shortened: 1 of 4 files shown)


deitch:
> Looks like something's broken

Nah, it all is an illusion. It *must* work! :-)

I can replicate it...

deitch:
I think `multiarch/qemu-user-static:register` assumes you already have `qemu-s390x-static` installed somewhere on your system. Which we do not yet. We have it in the `calico/go-build` [here](https://github.com/projectcalico/go-build/blob/master/Dockerfile.amd64#L65), but our `register` process doesn't take advantage of it. I need to recall how we do that. `register` includes the _registration_ , not the binary.

The only prior build in that semaphore run is `amd64`, which doesn't use binfmt_misc, so it doesn't fail.

deitch:
Confirmed. Once I place `qemu-s390x-static` in the right place (well, some register finagling), and run it, I get:

```
$ docker run -it --rm s390x/alpine:3.7 sh
standard_init_linux.go:190: exec user process caused "exec format error"
```

`exec format error` is still a qemu issue, but not the same. The original `no such file or directory` was just because it could not find the `qemu-s390x-static` binary to execute the file.

deitch:
Updating the description above. Please reread.

deitch:
All builds pass now! (prob because I excluded `s390x`...)

vielmetti:
@deitch who should review this? We're blocked here until it has a review.

deitch:
Most recent discussions were with @fasaxc , also @tomdee and @caseydavenport

deitch:
Excellent, thanks @tomdee . If we can get this merged in, I will do the same thing for all of the other dozen or so projects that make up calico.

PR #1854, opened 2018-07-12
Update libcalico-go pin and clean up glide.yaml

tomdee:


tomdee:
Failing some kind of license check @fasaxc any ideas how to fix?

PR #1861, opened 2018-07-17
Glide: Restore CNI pin and remove libcalico-go pin

tomdee:
CNI is actually needed - the pkg/ns package is removed from master
Libcalico-go pin isn't needed - the version from Typha should be used

This PR doesn't change the revision in the glide.lock for the CNI or libcalico packages

### DIFF
--- a/glide.yaml
+++ b/glide.yaml
@@ -3,9 +3,12 @@ homepage: https://projectcalico.org
 license: Apache-2.0
 import:
 - package: github.com/containernetworking/cni
+  version: v0.5.2
   subpackages:
   - pkg/ns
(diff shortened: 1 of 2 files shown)


glide.yaml
```diff
@@ -3,9 +3,12 @@ homepage: https://projectcalico.org
 license: Apache-2.0
 import:
 - package: github.com/containernetworking/cni
+  version: v0.5.2
   subpackages:
   - pkg/ns
+# Need to pin to an older version as our code uses coed that is deprecated in newer versions.
```
fasaxc: Typo: coed

PR #1871, opened 2018-08-06
Update dataplane driver for Windows 1803 builds

nwoodmsft:
## Description
Bug / Compatibility fix - This change updates the Windows Dataplane driver to be compatible with Windows 1803+ builds. We leverage some new features of hcsshim to help us determine which HNS/ACL features the current Windows build supports. This change is backwards compatible, so calico-felix.exe will continue to work on builds prior to 1803.

### DIFF
--- a/dataplane/windows/policysets/policysets.go
+++ b/dataplane/windows/policysets/policysets.go
@@ -33,13 +33,18 @@ type PolicySets struct {
 
 	IpSets []*ipsets.IPSets
 
+	supportedFeatures hns.HNSSupportedFeatures
+
 	resyncRequired bool
 }
 
 func NewPolicySets(ipsets []*ipsets.IPSets) *PolicySets {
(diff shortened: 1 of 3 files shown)


dataplane/windows/policysets/policysets.go
```diff
@@ -530,21 +542,27 @@ func (s *PolicySets) NewRule(isInbound bool, priority uint16) *hns.ACLPolicy {
...
-// NewHostRule returns a new hns rule object scoped to the host. This is only
-// temporarily required for compatibility with RS3.
+// NewHostRule returns a new hns rule object scoped to the host.
 func (s *PolicySets) NewHostRule(isInbound bool) *hns.ACLPolicy {
 	direction := hns.Out
 	if isInbound {
 		direction = hns.In
 	}
 
+	priority := 0
```
fasaxc: If you define this as `var priority uint16` then you wouldn't need the cast below

PR #1901, opened 2018-10-09
Add SNAT random-fully flag.

fasaxc:
## Description

Detect iptables version and pass feature flags around.  

- SNAT --random-fully was aded in iptables v1.6.1
- MASQUERADE --random-fully was added in iptables v1.6.2

Use feature flag to control whether the --random-fully flag is included.

Fixes https://github.com/projectcalico/calico/issues/2073

## Todos
- [x] Check kernel compat?
- [x] Unit tests (full coverage)
- [x] Integration tests (delete as appropriate) In plan/Not needed/Done
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
Where supported, Felix now uses the iptables --random-fully option when programming SNAT and MASQUERADE rules.
```

PR #1903, opened 2018-10-10
Use iptables-restore xtables lock

fasaxc:
## Description

Based on feature-detection work added in #1901 

## Todos
- [x] Need to update manifests to map in the xtables lock file. cf https://github.com/kubernetes-incubator/kubespray/pull/3191/files: https://github.com/projectcalico/calico/pull/2224
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
If deployed with iptables v1.6.2 or above (as included in the calico/felix and calico/node images), all iptables write operations are protected by the xtables global lock.  Previously, the lock was disabled by default but iptables v1.6.2 made the lock mandatory.
```

### DIFF
--- a/iptables/table.go
+++ b/iptables/table.go
@@ -989,8 +989,9 @@ func (t *Table) applyUpdates() error {
 		var outputBuf, errBuf bytes.Buffer
 		args := []string{"--noflush", "--verbose"}
 		if features.RestoreSupportsLock {
-			// Versions of iptables-restore that support the xtables lock also make it mandatory.  Make sure
-			// that we configure it to retry and configure for a short retry interval (the default is not to retry).
+			// Versions of iptables-restore that support the xtables lock also make it impossible to disable.  Make
+			// sure that we configure it to retry and configure for a short retry interval (the default is to try to
+			// acquire the lock only once).
 			lockTimeout := t.lockTimeout.Seconds()
 			if lockTimeout <= 0 {
 				// Before iptables-restore added lock support, we were able to disable the lock completely, which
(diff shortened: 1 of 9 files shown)


iptables/table.go
```diff
@@ -1014,14 +987,37 @@ func (t *Table) applyUpdates() error {
...
+				"probeIntervalMicros": intervalStr,
+			}).Debug("Using native iptables-restore xtables lock.")
+		}
+		cmd := t.newCmd(t.iptablesRestoreCmd, args...)
 		cmd.SetStdin(&inputBuf)
 		cmd.SetStdout(&outputBuf)
 		cmd.SetStderr(&errBuf)
 		countNumRestoreCalls.Inc()
-		t.writeLock.Lock()
+		t.calicoXtablesLock.Lock() // Will be a dummy lock if our xtables lock is disabled.
```
neiljerram: I'd prefer "Will be a dummy lock if iptables-restore is acquiring the xtables lock itself."

iptables/table.go
```diff
@@ -1014,14 +987,37 @@ func (t *Table) applyUpdates() error {
 		t.logCxt.WithField("iptablesInput", input).Debug("Writing to iptables")
 
 		var outputBuf, errBuf bytes.Buffer
-		cmd := t.newCmd(t.iptablesRestoreCmd, "--noflush", "--verbose")
+		args := []string{"--noflush", "--verbose"}
+		if features.RestoreSupportsLock {
+			// Versions of iptables-restore that support the xtables lock also make it mandatory.  Make sure
```
neiljerram: By "mandatory", do you mean that iptables-restore will give up if the lock is not available immediately?  Because it obviously isn't mandatory in the sense that you still need args to tell it to retry...

Mostly just checking my understanding here; I can't think of an obvious wording improvement to the comment.

iptables/table.go
fasaxc: There's more than one reason for it to be disabled but probably worth expanding the comment to explain.

iptables/table.go
fasaxc: I mean, it won't proceed to do any iptables programming without the lock.  Its default behaviour with no args is to quit immediately if it can't acquire the lock.

Will see if I can tighten up the wording.

PR #1924, opened 2018-11-08
Update typha to 8db57062a4bc923a9f1aacbd2c38d8e5d28e219b

doublek:
## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```


neiljerram:
@doublek In future please don't put the commit ID in the title, unless there's a specific reason for that commit ID (in which case explain that).  If you just say "Update Typha" or "Bump Typha" etc, it's implied that you're updating to the latest Typha for the branch - which is actually the case here.  If you mention a specific ID it's begging the question of "why that ID?"

caseydavenport:
@neiljerram our automated scripts do that, not sure if this is a follow on for that.

I don't really have a problem with the commit ID in the title, though. We probably always want to quesiton "why are we bumping libcalico?"

PR #1938, opened 2018-12-20
set proxy_delay before set proxy_arp to avoid kernel panic

### LINKED ISSUE #1937: proxy_delay configure may cause kernel "divide zero" panic
Felix set the `proxy_arp` and `proxy_delay` when monitored the endpoint interface creation. In some concurrent case those two settings order will cause kernel panic.

Kernel panic log:
[kernel panic log](https://gist.github.com/BSWANG/b262578c74323343fd2857e95ac5ee8b)

Kernel code of `proxy_arp` and `proxy_delay` judge and process code:

[line judge the proxy_delay is not zero](https://elixir.bootlin.com/linux/latest/source/net/ipv4/arp.c#L852)

[when process arp package the proxy_delay be set to zero by felix, then divide zero panic](https://elixir.bootlin.com/linux/latest/source/net/core/neighbour.c#L1449)

Felix code of setting `proxy_arp` and `proxy_delay`:
[setting `proxy_arp` then setting `proxy_delay`](https://github.com/projectcalico/felix/blob/bc9660c99b0dcea82dcf2c89c2e52d71605f7c6c/dataplane/linux/endpoint_mgr.go#L892-L901)

## Expected Behavior

Kernel doesn't panic when start containers.

## Current Behavior

Kernel maybe panic when start containers.

## Possible Solution

Setting the `proxy_delay` first, then setting the `proxy_arp`.

1. Prepare a k8s cluster， a big `instance_type`  will easier to reproduce this. I use a "96 Core, 382GB" machine, reproduce after 9 hours stress test.
2. Parellel start and stop containers with `ping 8.8.8.8` `entrypoint`. To make an arp query when container start.
3. Kernel panic after lot of loops.

## Context

Setting `proxy_delay` to `0` before setting `proxy_arp` can solve this problem. 

## Your Environment

* Felix version (if using Felix binary directly `calico-felix --version`): 2.6.8
* Which orchestrator are you using (e.g. Kubernetes, OpenStack, Docker, Mesos): kubernetes
* Orchestrator version (if used): 1.11.5
* Operating System and version: Centos 7.4

BSWANG:
## Description

fix #1937

## Release Note

```release-note
None
```

### DIFF
--- a/dataplane/linux/endpoint_mgr.go
+++ b/dataplane/linux/endpoint_mgr.go
@@ -873,6 +873,12 @@ func (m *endpointManager) configureInterface(name string) error {
 		if err != nil {
 			return err
 		}
+		// Normally, the kernel has a delay before responding to proxy ARP but we know
+		// that's not needed in a Calico network so we disable it.
+		err = m.writeProcSys(fmt.Sprintf("/proc/sys/net/ipv4/neigh/%s/proxy_delay", name), "0")
+		if err != nil {
+			return err
+		}
 		// Enable proxy ARP, this makes the host respond to all ARP requests with its own
 		// MAC.  This has a couple of advantages:
 		//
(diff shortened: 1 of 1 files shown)


neiljerram:
@BSWANG Also do you plan to report this as a kernel bug as well?  If I understand correctly, this is a kernel panic that can happen whenever proxy_delay is set to 0 for an interface, if that interface is processing an ARP request at the same time, and if proxy_arp was already enabled for the interface.

BSWANG:
@neiljerram Thanks for review, The commit message fixed.

PR #1939, opened 2018-12-20
Add region into status reporting

neiljerram:
## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/config/config_params.go
+++ b/config/config_params.go
@@ -139,6 +139,8 @@ type Config struct {
 	MetadataAddr string `config:"hostname;127.0.0.1;die-on-fail"`
 	MetadataPort int    `config:"int(0,65535);8775;die-on-fail"`
 
+	OpenstackRegion string `config:"string;;local"`
+
 	InterfacePrefix  string `config:"iface-list;cali;non-zero,die-on-fail"`
 	InterfaceExclude string `config:"iface-list;kube-ipvs0"`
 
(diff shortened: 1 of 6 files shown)


config/config_params.go
```diff
@@ -139,6 +139,8 @@ type Config struct {
 	MetadataAddr string `config:"hostname;127.0.0.1;die-on-fail"`
 	MetadataPort int    `config:"int(0,65535);8775;die-on-fail"`
 
+	OpenstackRegion string `config:"string;;local"`
```
fasaxc: Bit worried about the prefix since this seem to be being used for non-Openstack data too (host endpoints).  Should we just leave host endpoint status at the old path (or remove it since AIFAIK, it's not used anywhere)?

daemon/daemon.go
```diff
@@ -853,7 +854,7 @@ func (fc *DataplaneConnector) handleProcessStatusUpdate(ctx context.Context, msg
 		FirstUpdate:   !fc.firstStatusReportSent,
 	}
 	kv := model.KVPair{
-		Key:   model.ActiveStatusReportKey{Hostname: fc.config.FelixHostname},
+		Key:   model.ActiveStatusReportKey{Hostname: fc.config.FelixHostname, RegionString: model.RegionString(fc.config.OpenstackRegion)},
```
fasaxc: Do we have a test for this anywhere?  Would it be easy to add?

neiljerram: We have ST for it, in calico-test.

I guess we could also add a Felix FV, but it feels like it's better to get onto the ST, given time pressure.

neiljerram:
> Looking good, holding off on approval until we've got libcalico-go merged.

Just wondering what this means in detail, if - suppose - that libcalico-go is merged end today, and then I want to push through the Typha PR, update this one's pins, and then workflow the networking-calico changes.  Are you happy for me to approve this PR so that that can all happen?

PR #1979, opened 2019-03-11
Capability to build Bionic packages for release-v2.6

neiljerram:
@fasaxc @caseydavenport I found this lying around; I needed it for some reason last year.  Do you think it's worth keeping?


neiljerram:
Thanks @fasaxc . However, it fails CI, and we haven't been running CI on release-v2.6 since last November, so I think that means I have to discard this.

PR #1980, opened 2019-03-11
#1945: Adding support for regular expression values for the Interface Exclude parameter

### LINKED ISSUE #1945: Support exclude interface by pattern?
could you add support that exclude interface by pattern?
felix only support exclude interface by string match now

## Expected Behavior

support exclude interface by a pattern like `veth*` or `veth.*`
so interface change in docker bridge network not affect felix

## Current Behavior

`FELIX_INTERFACEEXCLUDE=kube-ipvs0,veth27afb87`
for each iface in `FELIX_INTERFACEEXCLUDE`, check by string compare (`InterfaceMonitor.isExcludedInterface`)

## Possible Solution

`InterfaceMonitor.isExcludedInterface` add regex/glob support
add pattern to `FELIX_INTERFACEEXCLUDE` or new configuration

use k8s+calico and docker-bridge+in-house PaaS :)

## Context

we try to use k8s to replace our in-house PasS, and now its running on our servers at same time. 

## Your Environment

* `calico/node` container version (if running Felix in a container): v3.3.2
* Which orchestrator are you using (e.g. Kubernetes, OpenStack, Docker, Mesos): k8s
* Orchestrator version (if used): v1.13.1
* Etcd version (if using etcd datastore driver): 3.3.5
* Operating System and version: 4.4.6-gentoo

stevegaossou:
Closes #1945. 

## Description
This PR contains: 
- modifying logic for interface exclude mechanism to allow for regular expressions instead of just exact string matches (#1945) 
- modified existing and added new unit tests (check the commits)
- this component affects `felix` and also documentation for `calico` and `libcalico-go`: 
  - https://github.com/projectcalico/libcalico-go/pull/1061
  - https://github.com/projectcalico/calico/pull/2494

## Todos
    - Modified existing test
- [ ] ~~Integration tests (delete as appropriate) In plan/Not needed/Done~~
- [x] Documentation
- [ ] ~~Backport~~
- [x] Release note

## Release Note

```release-note
Allow the entries for the exclude interface parameter to be regular expression patterns
```

### DIFF
--- a/config/config_params.go
+++ b/config/config_params.go
@@ -150,7 +150,7 @@ type Config struct {
 	OpenstackRegion string `config:"region;;die-on-fail"`
 
 	InterfacePrefix  string `config:"iface-list;cali;non-zero,die-on-fail"`
-	InterfaceExclude string `config:"iface-list-regexp;kube-ipvs0;die-on-fail"`
+	InterfaceExclude string `config:"iface-list-regexp;kube-ipvs0"`
 
 	ChainInsertMode             string `config:"oneof(insert,append);insert;non-zero,die-on-fail"`
 	DefaultEndpointToHostAction string `config:"oneof(DROP,RETURN,ACCEPT);DROP;non-zero,die-on-fail"`
@@ -259,6 +259,19 @@ func (config *Config) InterfaceExcludes() []*regexp.Regexp {
 	strValues := strings.Split(config.InterfaceExclude, ",")
 	regexpValues := []*regexp.Regexp{}
 
+	// Helper function to compile the given string value into regexp format, if invalid
+	// log error instead
+	compileAndAdd := func(s string) {
+		r, err := regexp.Compile(s)
+		if err != nil {
+			log.Errorf(
+				"Failed to parse element in list for %v: %v from source %v. %v",
+				"InterfaceExclude", s, config.InterfaceExclude, err)
+		} else {
+			regexpValues = append(regexpValues, r)
+		}
+	}
+
 	for _, strValue := range strValues {
 		// Ignore empty values
 		if len(strValue) == 0 {
@@ -268,10 +281,10 @@ func (config *Config) InterfaceExcludes() []*regexp.Regexp {
 		// that tests for exact match
 		if !RegexpIfaceElemRegexp.Match([]byte(strValue)) {
 			convertedValue := fmt.Sprintf("^%s$", regexp.QuoteMeta(strValue))
-			regexpValues = append(regexpValues, regexp.MustCompile(convertedValue))
+			compileAndAdd(convertedValue)
 		} else {
 			parsedValue := strValue[1 : len(strValue)-1]
-			regexpValues = append(regexpValues, regexp.MustCompile(parsedValue))
+			compileAndAdd(parsedValue)
 		}
 	}
 	return regexpValues
--- a/config/config_params.go
+++ b/config/config_params.go
@@ -149,8 +149,8 @@ type Config struct {
 
 	OpenstackRegion string `config:"region;;die-on-fail"`
 
-	InterfacePrefix  string `config:"iface-list;cali;non-zero,die-on-fail"`
-	InterfaceExclude string `config:"iface-list-regexp;kube-ipvs0"`
+	InterfacePrefix  string           `config:"iface-list;cali;non-zero,die-on-fail"`
+	InterfaceExclude []*regexp.Regexp `config:"iface-list-regexp;kube-ipvs0"`
 
 	ChainInsertMode             string `config:"oneof(insert,append);insert;non-zero,die-on-fail"`
 	DefaultEndpointToHostAction string `config:"oneof(DROP,RETURN,ACCEPT);DROP;non-zero,die-on-fail"`
@@ -249,45 +249,6 @@ func (config *Config) UpdateFrom(rawData map[string]string, source Source) (chan
 
 func (c *Config) InterfacePrefixes() []string {
 	return strings.Split(c.InterfacePrefix, ",")
-}
-
-// InterfaceExcludes returns a list of Regexp objects, each representing either an
-// exact interface value match or a class of interface matches. This function will
-// panic if any individual value cannot be parsed / compiled into a Regexp.
-func (config *Config) InterfaceExcludes() []*regexp.Regexp {
-	log.Debugf("Compiling regexp values for InterfaceExclude %s \n", config.InterfaceExclude)
-	strValues := strings.Split(config.InterfaceExclude, ",")
-	regexpValues := []*regexp.Regexp{}
-
-	// Helper function to compile the given string value into regexp format, if invalid
-	// log error instead
-	compileAndAdd := func(s string) {
-		r, err := regexp.Compile(s)
-		if err != nil {
-			log.Errorf(
-				"Failed to parse element in list for %v: %v from source %v. %v",
-				"InterfaceExclude", s, config.InterfaceExclude, err)
-		} else {
-			regexpValues = append(regexpValues, r)
-		}
-	}
-
-	for _, strValue := range strValues {
-		// Ignore empty values
-		if len(strValue) == 0 {
-			continue
-		}
-		// Any value not in regexp format will be converted into equivalent regexp
-		// that tests for exact match
-		if !RegexpIfaceElemRegexp.Match([]byte(strValue)) {
-			convertedValue := fmt.Sprintf("^%s$", regexp.QuoteMeta(strValue))
-			compileAndAdd(convertedValue)
-		} else {
-			parsedValue := strValue[1 : len(strValue)-1]
-			compileAndAdd(parsedValue)
-		}
-	}
-	return regexpValues
 }
 
 func (config *Config) OpenstackActive() bool {
--- a/config/config_params.go
+++ b/config/config_params.go
@@ -244,8 +244,17 @@ func (c *Config) InterfacePrefixes() []string {
 	return strings.Split(c.InterfacePrefix, ",")
 }
 
-func (c *Config) InterfaceExcludes() []string {
-	return strings.Split(c.InterfaceExclude, ",")
+func (c *Config) InterfaceExcludes() []*regexp.Regexp {
+	log.Debugf("Compiling regexp values for InterfaceExclude %s \n", c.InterfaceExclude)
+
+	strValues := strings.Split(c.InterfaceExclude, ",")
+	regexpValues := make([]*regexp.Regexp, len(strValues))
+
+	for _, strValue := range strValues {
+		// Will panic if any individual regexp string is invalid (not parsable)
+		regexpValues = append(regexpValues, regexp.MustCompile(strValue))
+	}
+	return regexpValues
 }
 
 func (config *Config) OpenstackActive() bool {
@@ -508,6 +517,7 @@ func loadParams() {
 		case "millis":
 			param = &MillisParam{}
 		case "iface-list":
+			// TODO: What should we allow here given that it's now a regexp?
 			param = &RegexpParam{Regexp: IfaceListRegexp,
 				Msg: "invalid Linux interface name"}
 		case "iface-param":
--- a/config/config_params.go
+++ b/config/config_params.go
@@ -33,19 +33,18 @@ import (
 )
 
 var (
-	// RegexpIfaceListRegexp differs from IfaceListRegexp in that it allows both
-	// normal interface values but also regex pattern values (which are marked
-	// by '/' at the start and end) in the comma delimited list
-	RegexpIfaceListRegexp = regexp.MustCompile(`^([a-zA-Z0-9_-]{1,15}|(\/[^,\s]+\/))(,([a-zA-Z0-9_-]{1,15}|(\/[^,\s]+\/)))*$`)
-	// RegexpIfaceElemRegexp matches an individual element in the overall list;
-	// assumes the value is marked by '/' at the start and end and cannot have commas
-	// or spaces
-	RegexpIfaceElemRegexp = regexp.MustCompile(`^\/[^,\s]+\/$`)
-	IfaceListRegexp       = regexp.MustCompile(`^[a-zA-Z0-9_-]{1,15}(,[a-zA-Z0-9_-]{1,15})*$`)
-	AuthorityRegexp       = regexp.MustCompile(`^[^:/]+:\d+$`)
-	HostnameRegexp        = regexp.MustCompile(`^[a-zA-Z0-9_.-]+$`)
-	StringRegexp          = regexp.MustCompile(`^.*$`)
-	IfaceParamRegexp      = regexp.MustCompile(`^[a-zA-Z0-9:._+-]{1,15}$`)
+	// RegexpIfaceElemRegexp matches an individual element in the overall interface list;
+	// assumes the value represents a regular expression and is marked by '/' at the start
+	// and end and cannot have spaces
+	RegexpIfaceElemRegexp = regexp.MustCompile(`^\/[^\s]+\/$`)
+	// NonRegexpIfaceElemRegexp matches an individual element in the overall interface list;
+	// assumes the value is between 1-15 chars long and only be alphanumeric or - or _
+	NonRegexpIfaceElemRegexp = regexp.MustCompile(`^[a-zA-Z0-9_-]{1,15}$`)
+	IfaceListRegexp          = regexp.MustCompile(`^[a-zA-Z0-9_-]{1,15}(,[a-zA-Z0-9_-]{1,15})*$`)
+	AuthorityRegexp          = regexp.MustCompile(`^[^:/]+:\d+$`)
+	HostnameRegexp           = regexp.MustCompile(`^[a-zA-Z0-9_.-]+$`)
+	StringRegexp             = regexp.MustCompile(`^.*$`)
+	IfaceParamRegexp         = regexp.MustCompile(`^[a-zA-Z0-9:._+-]{1,15}$`)
 )
 
 const (
@@ -248,8 +247,8 @@ func (config *Config) UpdateFrom(rawData map[string]string, source Source) (chan
 	return
 }
 
-func (config *Config) InterfacePrefixes() []string {
-	return strings.Split(config.InterfacePrefix, ",")
+func (c *Config) InterfacePrefixes() []string {
+	return strings.Split(c.InterfacePrefix, ",")
 }
 
 // InterfaceExcludes returns a list of Regexp objects, each representing either an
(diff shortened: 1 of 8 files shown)


stevegaossou:
One outstanding question I have is what we want to allow for the validation off the field not that it allows regexp. 

Namely here in [`./config/config_params.go`](https://github.com/projectcalico/felix/pull/1980/files#diff-5fbf5a733ea1e9a90ce9341bf558beddR36) for the `IfaceListRegexp` parameter. 

For example, should we keep the current validation pattern but only allow the introduction of the wildcard symbol `*` anywhere in the value?
 
cc @caseydavenport

config/config_params.go
```diff
@@ -508,6 +517,7 @@ func loadParams() {
 		case "millis":
 			param = &MillisParam{}
 		case "iface-list":
+			// TODO: What should we allow here given that it's now a regexp?
```
neiljerram: I guess anything for which calling `InterfaceExcludes` doesn't panic?

ifacemonitor/iface_monitor.go
```diff
@@ -136,8 +137,8 @@ readLoop:
 }
 
 func (m *InterfaceMonitor) isExcludedInterface(ifName string) bool {
-	for _, name := range m.InterfaceExcludes {
-		if ifName == name {
+	for _, nameExp := range m.InterfaceExcludes {
+		if nameExp.Match([]byte(ifName)) {
```
neiljerram: I think there's a change of behaviour here for existing config.  With config `xyz,abc`, previously interface names had to be exactly `xyz` or `abc`.  But now, if I understand the regexes correctly, it will match any interface names that _include_ `xyz` or `abc`.

Am I right?  If so, is that a problem?

config/config_params.go
stevegaossou: Fair point. This is now dictated by whatever we're allowing in terms of regexp patterns as values in the list. 

I'm going to separate this out into a new switch case, since it's reused right now between two separate parameters: 
- `InterfacePrefix`
- `InterfaceExclude`

And it's only the second one `InterfaceExclude` that we want to allow regexp for. 

Will also need to update the error message from: 
- `invalid Linux interface name` 

To something that calls out regexp patterns: 
- e.g. `invalid Linux interface name or regular expression pattern`

cc @caseydavenport

ifacemonitor/iface_monitor.go
stevegaossou: > Am I right? If so, is that a problem?

Yes, you are correct @neiljerram! Very good point. 

This would be a problem because it would break backwards compatibility, i.e. with any Calico install where someone has `FELIX_INTERFACEEXCLUDE` set to some values where they expect only exact matches. For example, in an existing Calico install with something like `FELIX_INTERFACEEXCLUDE=veth1` I would want it only exclude an exact match of an interface with name `veth1` and not something with a name `veth11`. 

One solution to this obstacle is have a clear separation between values in the _old format_ (non-regexp exact match values) and values in the _new format_ (regexp values). 

Had a quick chat with @caseydavenport and a simple format could be: 
- Old values remain in the format they are now
- New values must be delimited by `/` on both ends to symbolize it is to be treated as a regex 

Example: 
- The below will exclude an exact interface name of `veth0`
```
FELIX_INTERFACEEXCLUDE=veth0
```
- The below will exclude an exact interface name of `veth0` 
```
FELIX_INTERFACEEXCLUDE=/^veth0$/
```
- The below will exclude (1) any interface name that begins with `veth` (e.g. `veth0`, `veth1`, ... etc) and any interface name with `kube-ipvs` as a substring (e.g. `prefix-kube-ipvs`, `kube-ipvs-suffix`, ``prefix-kube-ipvs-suffix`)
```
FELIX_INTERFACEEXCLUDE=/^veth/,/kube-ipvs/
```


Thoughts?

neiljerram: Yes, that works and sounds OK to me.  Another way would be 
```
		regexpValues = append(regexpValues, regexp.MustCompile("^" + strValue + "$"))
```
in other words require the regex always to match the whole name.

But probably the explicit `/.../` syntax is better because it also avoids another possible back-compatibility problem, if it's possible for any existing settings to include regex metacharacters.

stevegaossou: > But probably the explicit /.../ syntax is better because it also avoids another possible back-compatibility problem, if it's possible for any existing settings to include regex metacharacters.

The existing validation for the interface exclude parameter only allows `[a-zA-Z0-9_-]`. Looking at the regexp constructs it does not look like `-` or `_` symbol would get us in trouble because they would just be interpreted as literals. 

Still, using the `/` on both ends felt like a nice clean way to communicate to users what to expect when looking at the values.

config/config_params.go
```diff
@@ -143,7 +151,7 @@ type Config struct {
 	OpenstackRegion string `config:"region;;die-on-fail"`
 
 	InterfacePrefix  string `config:"iface-list;cali;non-zero,die-on-fail"`
-	InterfaceExclude string `config:"iface-list;kube-ipvs0"`
+	InterfaceExclude string `config:"iface-list-regexp;kube-ipvs0;die-on-fail"`
```
stevegaossou: Calling this out, since it changes the default behaviour. 

Added the `die-on-fail` option, because it seemed sensible for it to halt on failed validation (it's going to panic anyways during the `Config.InterfaceExcludes()` if any of the individual values do not compile).

config/config_params.go
```diff
@@ -510,6 +537,13 @@ func loadParams() {
 		case "iface-list":
 			param = &RegexpParam{Regexp: IfaceListRegexp,
 				Msg: "invalid Linux interface name"}
+		case "iface-list-regexp":
+			param = &RegexpPatternListParam{
+				ListRegexp: RegexpIfaceListRegexp,
+				ElemRegexp: RegexpIfaceElemRegexp,
+				Delimiter:  ",",
+				Msg:        "list contains invalid Linux interface name or regex pattern",
+			}
```
stevegaossou: Turned the parameter into a new category, since it doesn't rely on just a regexp value for validation.

config/param_types.go
```diff
@@ -174,6 +174,45 @@ func (p *RegexpParam) Parse(raw string) (result interface{}, err error) {
...
+			_, compileErr := regexp.Compile(regexpValue)
+			if compileErr != nil {
+				err = p.parseFailed(raw, p.Msg)
+				return
+			}
+		}
+	}
+	result = raw
+	return
+}
```
stevegaossou: New parameter type to handle the special case with validating the interface exclude field that can have values in the old format (non-regexp exact matches) and new format (regexp).

ifacemonitor/ifacemonitor_test.go
```diff
@@ -373,36 +380,50 @@ var _ = Describe("ifacemonitor", func() {
...
+
+		// Repeat test for second interface exclude entry
+		netlinkUpdates("veth1")
 
-		// Clean it.
-		nl.delLink("kube-ipvs0")
-		dp.notExpectAddrStateCb()
-		dp.notExpectLinkStateCb()
+		// Repeat test for third interface exclude entry
+		netlinkUpdates("0dummy1")
```
stevegaossou: Adding more test coverage here for both `InterfaceMonitor.isExcludedInterface(string)` and `Config.InterfaceExcludes()` functions.

stevegaossou:
Woohoo! We're finally 🍏on all tests!

PR for documentation update is here: 
https://github.com/projectcalico/calico/pull/2490

config/config_params.go
neiljerram: I'm not sure I'm fully understanding here, but it feels wrong for this detail of behaviour to change.  I think it's straightforward to avoid "it's going to panic anyways ... if any of the individual values do not compile" by using `regexp.Compile` in that code instead of `regexp.MustCompile`.

neiljerram: (die-on-fail will cause Felix to go into a restart loop, so we generally specify die-on-fail only for essential config parameters)

config/config_params.go
```diff
@@ -33,11 +33,19 @@ import (
...
-	StringRegexp     = regexp.MustCompile(`^.*$`)
-	IfaceParamRegexp = regexp.MustCompile(`^[a-zA-Z0-9:._+-]{1,15}$`)
+	// RegexpIfaceListRegexp differs from IfaceListRegexp in that it allows both
+	// normal interface values but also regex pattern values (which are marked
+	// by '/' at the start and end) in the comma delimited list
+	RegexpIfaceListRegexp = regexp.MustCompile(`^([a-zA-Z0-9_-]{1,15}|(\/[^,\s]+\/))(,([a-zA-Z0-9_-]{1,15}|(\/[^,\s]+\/)))*$`)
+	// RegexpIfaceElemRegexp matches an individual element in the overall list;
+	// assumes the value is marked by '/' at the start and end and cannot have commas
+	// or spaces
+	RegexpIfaceElemRegexp = regexp.MustCompile(`^\/[^,\s]+\/$`)
```
neiljerram: I think this can be just `^\/[^\s]+\/$`.  We don't need to prohibit commas, because the raw value has already been split at commas before applying this regexp.

neiljerram:
@stevegaossou By the way, another thought: when working with Felix config parameters, there are generally a few related places to think about:

1. Equivalent validation in libcalico-go as well, because that allows calicoctl to catch and report errors with a much nicer UX than Felix either restarting or warning.

2. Matching doc update for the [libcalico-go field](https://github.com/projectcalico/libcalico-go/blob/master/lib/apis/v3/felixconfig.go#L109).

3. Matching doc update in the [docs](https://docs.projectcalico.org/v3.6/reference/calicoctl/resources/felixconfig).

In this case, the [libcalico-go field](https://github.com/projectcalico/libcalico-go/blob/master/lib/apis/v3/felixconfig.go#L109) has no `validate` tag:
```
	// InterfaceExclude is a list of interfaces that Felix should exclude when monitoring for host
	// endpoints.  The default value ensures that Felix ignores Kubernetes' IPVS dummy interface,
	// which is used internally by kube-proxy.  [Default: kube-ipvs0]
	InterfaceExclude string `json:"interfaceExclude,omitempty"`
```
which means that there is no detailed libcalico-go validation here.  And it's probably too big a job to start adding that now.

(Yes, it's kinda poor that in principle we have config validation in both Felix and libcalico-go...  It would be good to work towards improving that, but I think it's a big task.)

But the updates (2) and (3) will be needed too.

config/config_params.go
stevegaossou: > I think it's straightforward to avoid "it's going to panic anyways ... if any of the individual values do not compile" by using `regexp.Compile` in that code instead of `regexp.MustCompile`.

Right. 

When I originally considered this I came to a conclusion that using `regexp.Compile` to avoid a panic would require knowing what we want to do when the one of the user provided values fails to compile to regexp. What should we do in that case? We can log the error message as a bare minimum. But if we gracefully continue onwards, it means we probably don't setup that particular interface exclude value. I just wasn't sure from a user experience standpoint what our acceptable behaviour was. E.g. if I wanted to exclude all interfaces with a name starting with `kube` but I somehow mess up the syntax such that it doesn't compile. Ideally I would want `felix` to catch my mistake and halt during startup to tell me so that I can correct myself. 

So that was my original intention. 

> (die-on-fail will cause Felix to go into a restart loop, so we generally specify die-on-fail only for essential config parameters)

 Fair point. I will remove the `die-on-fail` and switch the compiling of values from `regexp.MustCompile` to `regexp.Compile`. 👍

config/config_params.go
```diff
@@ -240,12 +248,30 @@ func (config *Config) UpdateFrom(rawData map[string]string, source Source) (chan
 	return
 }
 
-func (c *Config) InterfacePrefixes() []string {
-	return strings.Split(c.InterfacePrefix, ",")
+func (config *Config) InterfacePrefixes() []string {
+	return strings.Split(config.InterfacePrefix, ",")
```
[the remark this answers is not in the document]
stevegaossou: Sorry about that. It popped up as a recommended change (by `gofmt`?) through my IDE, since all the other functions related to `Config` had `config` instead of `c`. 

I actually had a bunch of other changes like this one, but quickly realized this is not the right place for it.

stevegaossou:
> 3. Matching doc update in the docs.

For the `calico` repo level documentation, this change will be going in the next release `3.7`, right? 

So I only need to update the `./master/` version of the docs. 

👍 or 👎?

config/config_params.go
```diff
@@ -240,12 +248,30 @@ func (config *Config) UpdateFrom(rawData map[string]string, source Source) (chan
...
+	log.Debugf("Compiling regexp values for InterfaceExclude %s \n", config.InterfaceExclude)
+	strValues := strings.Split(config.InterfaceExclude, ",")
+	regexpValues := make([]*regexp.Regexp, len(strValues))
+
+	for i, strValue := range strValues {
+		// All values in interface exclude list will be treated as regular
+		// expressions; any value not in regexp format will be converted into
+		// equivalent regexp that tests for exact match
+		if !RegexpIfaceElemRegexp.Match([]byte(strValue)) {
+			convertedValue := fmt.Sprintf("^%s$", strValue)
```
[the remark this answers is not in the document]
stevegaossou: As you [called out in an earlier thread](https://github.com/projectcalico/felix/pull/1980#discussion_r265684587), the previous validation for this field restricted characters to be from `[a-zA-Z0-9_-]` so it's probably safe without `QuoteMeta`. 

However, I kind of like the idea of using it, since it just completely rules out any possibility of wrong behaviour. I'll add `QuoteMeta`.

config/config_params.go
```diff
@@ -240,12 +248,30 @@ func (config *Config) UpdateFrom(rawData map[string]string, source Source) (chan
...
+			regexpValues[i] = regexp.MustCompile(convertedValue)
+		} else {
+			regexpValues[i] = regexp.MustCompile(strValue)
```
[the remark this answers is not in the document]
stevegaossou: Good catch @neiljerram! I thought this was covered, but it's not.

config/config_params_test.go
```diff
@@ -158,8 +157,15 @@ var _ = DescribeTable("Config parsing",
...
-	Entry("InterfaceExclude list", "InterfaceExclude", "kube-ipvs0,dummy", "kube-ipvs0,dummy"),
+
+	Entry("InterfaceExclude one value no regexp", "InterfaceExclude", "kube-ipvs0", "kube-ipvs0"),
+	Entry("InterfaceExclude list no regexp", "InterfaceExclude", "kube-ipvs0,dummy", "kube-ipvs0,dummy"),
+	Entry("InterfaceExclude one value regexp", "InterfaceExclude", "/kube-ipvs/", "/kube-ipvs/"),
+	Entry("InterfaceExclude list regexp", "InterfaceExclude", "kube-ipvs0,dummy,/^veth*$/", "kube-ipvs0,dummy,/^veth*$/"),
+	Entry("InterfaceExclude no regexp", "InterfaceExclude", "/^kube.*/,/veth/", "/^kube.*/,/veth/"),
+	Entry("InterfaceExclude list regexp invalid", "InterfaceExclude", "kube,//", "kube-ipvs0", true),             // empty regexp value
+	Entry("InterfaceExclude list regexp invalid comma", "InterfaceExclude", "/kube,/,dummy", "kube-ipvs0", true), // bad comma use
+	Entry("InterfaceExclude list regexp invalid comma", "InterfaceExclude", `/^kube\K/`, "kube-ipvs0", true),     // Invalid regexp
```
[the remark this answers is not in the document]
stevegaossou: That would be nice. But I'm hesitant to do it because I'm not sure how the change affects the FelixConfig resource. 

I think I'd rather not make that change just to make this smaller.

config/config_params.go
stevegaossou: Actually looking at the logic for handling `die-on-fail` (`DieOnParseFailure` flag) after parsing a parameter it looks like it wipes the error and then sets the value to default, which for `InterfaceExclude` is `kube-ipvs0`. That seems undesirable (especially now that we could have regex values), no? 

https://github.com/projectcalico/felix/blob/master/config/config_params.go#L335

stevegaossou:
> But the updates (2) and (3) will be needed too.

Done. 

(2) https://github.com/projectcalico/libcalico-go/pull/1061

(3) https://github.com/projectcalico/calico/pull/2494

config/config_params.go
stevegaossou: Fixed.

config/param_types.go
```diff
@@ -174,6 +174,45 @@ func (p *RegexpParam) Parse(raw string) (result interface{}, err error) {
...
+// expressions.
+// Uses two levels of validation:
+// 	(1) Attempt to match raw against a regular expression (for custom validation)
+//  (2) Attempt to compile each list element in raw into a valid regular expression
+func (p *RegexpPatternListParam) Parse(raw string) (result interface{}, err error) {
+	// Ensure list in its entirety matches expected structure
+	if !p.ListRegexp.MatchString(raw) {
+		err = p.parseFailed(raw, p.Msg)
+		return
+	}
```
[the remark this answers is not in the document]
stevegaossou: Fixed.

config/config_params_test.go
```diff
@@ -116,8 +116,7 @@ func fieldsByName(example interface{}) map[string]reflect.StructField {
 var _ = DescribeTable("Config parsing",
 	func(key, value string, expected interface{}, errorExpected ...bool) {
 		config := New()
-		config.UpdateFrom(map[string]string{key: value},
-			EnvironmentVariable)
+		config.UpdateFrom(map[string]string{key: value}, EnvironmentVariable)
```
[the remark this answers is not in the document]
stevegaossou: Fixed.

config/config_params.go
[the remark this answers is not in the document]
stevegaossou: Fixed.

config/config_params.go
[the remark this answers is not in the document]
stevegaossou: Fixed.

config/config_params.go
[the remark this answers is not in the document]
stevegaossou: Fixed.

config/config_params.go
neiljerram: Yes, the behaviour for a parse error for a parameter without `die-on-fail` is to log a warning and then use the default value.  I think that's what we should do here too.

stevegaossou: Okay, sounds good. I thought it over again and you're right, there's no reason to change this behaviour from what it's always done. That's what users would expect since it's established. 

Changing it back.

stevegaossou:
Just leaving a status update on this PR. 

All code review changes have been addressed. 

Doc update PRs can be found here: 
- https://github.com/projectcalico/libcalico-go/pull/1061
- https://github.com/projectcalico/calico/pull/2494

config/config_params.go
```diff
@@ -244,8 +251,43 @@ func (c *Config) InterfacePrefixes() []string {
...
+		// that tests for exact match
+		if !RegexpIfaceElemRegexp.Match([]byte(strValue)) {
+			convertedValue := fmt.Sprintf("^%s$", regexp.QuoteMeta(strValue))
+			compileAndAdd(convertedValue)
+		} else {
+			parsedValue := strValue[1 : len(strValue)-1]
+			compileAndAdd(parsedValue)
+		}
+	}
+	return regexpValues
```
neiljerram: Steve, I'm afraid this still isn't how I was expecting - so I guess I've done a poor job of communicating that.  I've posted a commit - which is a delta on top of your changes - at https://github.com/neiljerram/felix/commit/77a59b0baa9616febed27884ebd08161252ee9bb, which I hope will show more clearly what I've been trying to express.

Could you take a look at that and let me know what you think?

stevegaossou: Hey @neiljerram I just took a look at your changes. 

I see two main changes: 
- Making the parse do more than just validation (so it returns the slice of Regexp objects)
- This makes InterfaceExcludes() redundant, so it's been removed 

I don't think you did a poor job of communicating this as you had a comment about having the Parse return an array of the values instead of the raw (granted this version now returns a slice of Regexp, not just slice of strings). I remember making a comment that I wasn't 100% sure on how that change would affect upstream locations that use `felix` (e.g. FelixConfig), [specifically this change](https://github.com/neiljerram/felix/commit/77a59b0baa9616febed27884ebd08161252ee9bb#diff-5fbf5a733ea1e9a90ce9341bf558beddR154). I recall you said it doesn't affect them and I see what you mean now. The scope of changing that field from string to []*Regexp is still only internal to `felix`.  

If my understanding is correct, your changes look fine to me, shall I just apply your delta on my branch?

neiljerram: Yes, I think so, and then I believe we can call this done!

stevegaossou: Applied and pushed! Was just waiting for CI to pass before I added a comment here confirming. 

Thanks very much for all the feedback @neiljerram! Much appreciated.

PR #1988, opened 2019-03-27
Fix Typha excessive logging when Felix cannot communicate

roshanirathi:
## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/daemon/daemon.go
+++ b/daemon/daemon.go
@@ -442,7 +442,17 @@ configRetry:
 		log.Infof("Starting the Typha connection")
 		err := typhaConnection.Start(context.Background())
 		if err != nil {
-			log.WithError(err).Fatal("Failed to connect to Typha")
+			retry := 0
+			for (err != nil && retry < 10) {
+				// Set Ready and Live to false
+			        healthAggregator.Report(healthName, &health.HealthReport{Live: false, Ready: false})
+				err = typhaConnection.Start(context.Background())
+				log.WithError(err).Warn("Retrying to start Typha")
+				retry++
+			}
+			if (err != nil && retry > 10) {
+				log.WithError(err).Fatal("Failed to connect to Typha")
+			}
 		}
 		go func() {
 			typhaConnection.Finished.Wait()
--- a/daemon/daemon.go
+++ b/daemon/daemon.go
@@ -443,14 +443,14 @@ configRetry:
 		err := typhaConnection.Start(context.Background())
 		if err != nil {
 			retry := 0
-			for (err != nil && retry < 10) {
+			for err != nil && retry < 10 {
 				// Set Ready and Live to false
-			        healthAggregator.Report(healthName, &health.HealthReport{Live: false, Ready: false})
+				healthAggregator.Report(healthName, &health.HealthReport{Live: false, Ready: false})
 				err = typhaConnection.Start(context.Background())
 				log.WithError(err).Warn("Retrying to start Typha")
 				retry++
 			}
-			if (err != nil && retry > 10) {
+			if err != nil && retry > 10 {
 				log.WithError(err).Fatal("Failed to connect to Typha")
 			}
 		}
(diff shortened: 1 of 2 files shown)


daemon/daemon.go
```diff
@@ -442,7 +442,17 @@ configRetry:
 		log.Infof("Starting the Typha connection")
 		err := typhaConnection.Start(context.Background())
 		if err != nil {
-			log.WithError(err).Fatal("Failed to connect to Typha")
+			retry := 0
```
fasaxc: Please log once here at Error level "Failed to connect to Typha, will retry..."

daemon/daemon.go
```diff
@@ -442,7 +442,17 @@ configRetry:
...
+			for err != nil && retry < 10 {
+				// Set Ready and Live to false
+				healthAggregator.Report(healthName, &health.HealthReport{Live: false, Ready: false})
```
fasaxc: I think this should be `Live:true, Ready:false`  If we report non-live then Kubernetes will kill Felix but in this case we're managing the retry and restart ourselves.

daemon/daemon.go
```diff
@@ -442,7 +442,17 @@ configRetry:
 		log.Infof("Starting the Typha connection")
 		err := typhaConnection.Start(context.Background())
 		if err != nil {
-			log.WithError(err).Fatal("Failed to connect to Typha")
+			retry := 0
+			for err != nil && retry < 10 {
```
fasaxc: You should put a `time.Sleep(1 * time.Second)` in this loop to throttle it if the connection fails immediately.

daemon/daemon.go
```diff
@@ -442,7 +442,17 @@ configRetry:
...
+				// Set Ready and Live to false
+				healthAggregator.Report(healthName, &health.HealthReport{Live: false, Ready: false})
+				err = typhaConnection.Start(context.Background())
+				log.WithError(err).Warn("Retrying to start Typha")
```
fasaxc: Please drop this log to debug or we'll emit just as many high-level logs as we used to.

daemon/daemon.go
```diff
@@ -442,7 +442,17 @@ configRetry:
...
+				retry++
+			}
+			if err != nil && retry > 10 {
```
fasaxc: This condition will never fire; retry will be exactly 10 after the loop finishes.

daemon/daemon.go
fasaxc: Made me think about the possiblity that each retry could take 1ms or 30s depending on the type of problem.  Maybe swtich to a time-based loop:
```
startTime := time.Now()
for err!= nil && time.Since(startTime) < 30 * time.Second {
}
```

daemon/daemon.go
roshanirathi: Ack

daemon/daemon.go
roshanirathi: Ack

daemon/daemon.go
roshanirathi: Ack

daemon/daemon.go
roshanirathi: Ack

daemon/daemon.go
roshanirathi: Ack

daemon/daemon.go
```diff
@@ -442,7 +442,18 @@ configRetry:
...
 		err := typhaConnection.Start(context.Background())
 		if err != nil {
-			log.WithError(err).Fatal("Failed to connect to Typha")
+			log.WithError(err).Error("Failed to connect to Typha. Retrying...")
+			startTime := time.Now()
+			for err != nil && time.Since(startTime) < 30*time.Second {
+				// Set Ready to false and Live to true when unable to connect to typha
+				healthAggregator.Report(healthName, &health.HealthReport{Live: true, Ready: false})
+				err = typhaConnection.Start(context.Background())
+				log.WithError(err).Debug("Retrying to start Typha")
```
fasaxc: Need an `if err == nil {break}` above this line so that we don't log/sleep if the retry succeeds.

roshanirathi: Ack

PR #1989, opened 2019-03-30
VXLAN calculation and data plane

caseydavenport:
## Description

Adds VXLAN calculator to the calc graph, as well as the Linux dataplane.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
Felix programs VXLAN routes
```


caseydavenport:
The dataplane is still pretty crude - a few things still need doing, including cleaning up old L2 entries (right now they stay around forever) and also just some general code cleanup.

I think the calc graph piece is getting there though. One thing it doesn't yet handle is nodes with duplicate IP / VXLAN tunnel addresses. 

I think we can probably live with the latter (they are assigned through our IPAM, so should be unique).

For the former, if two nodes both have an IP address, we'll potentially send two VTEP updates to the dataplane, meaning we'll try to set up two static ARP entries for the same IP. Need to think about whether this is acceptable or not - it's clearly a misconfiguration, the question is how should we handle it and will the current code fix itself once the misconfiguration is resolved.

calc/calc_graph.go
```diff
@@ -283,6 +295,21 @@ func NewCalculationGraph(callbacks PipelineCallbacks, hostname string) *CalcGrap
 	hostIPPassthru := NewDataplanePassthru(callbacks)
 	hostIPPassthru.RegisterWith(allUpdDispatcher)
 
+	// Calculate VXLAN routes.
```
caseydavenport: Need to update this diagram.

calc/event_sequencer.go
```diff
@@ -700,6 +716,86 @@ func (buf *EventSequencer) flushNamespaces() {
 	log.Debug("Done flushing Namespaces")
 }
 
+func (buf *EventSequencer) OnVTEPUpdate(update *proto.VXLANTunnelEndpointUpdate) {
+	// We trust the caller not to send us an update with nil ID, so safe to dereference.
```
caseydavenport: Copy/paste comment should be removed.

calc/event_sequencer.go
```diff
@@ -700,6 +716,86 @@ func (buf *EventSequencer) flushNamespaces() {
...
+	log.WithFields(log.Fields{
+		"dst": dst,
+	}).Debug("VTEP removed")
+	delete(buf.pendingVTEPUpdates, dst)
+	if buf.sentVTEPs.Contains(dst) {
+		buf.pendingVTEPDeletes.Add(dst)
+	}
+}
+
+func (buf *EventSequencer) flushVTEPs() {
```
caseydavenport: I think we'll want to reorg this bit so that we can guarantee ordering here:
- Send VTEP adds
- Send route deletes
- Send route adds
- Send VTEP deletes

calc/vxlan_resolver.go
```diff
@@ -0,0 +1,453 @@
...
+			c.sendVTEPRemove(nodeName)
+		}
+	}
+	return
+}
+
+// OnPoolUpdate gets called whenever an IP pool changes. If a new VXLAN pool is added, kick
+// pending routes to see if any should now be programmed. If a VXLAN pool is removed, then
+// find any routes which need to be withdrawn and remove them. Pool CIDRs cannot be modified,
+// so we do not need to handle that here.
```
fasaxc: One man's delete/create can be another man's update.  Imagine this scenario:

- IP pool "my-pool" has CIDR 10.0.0.0/16
- Felix disconnected from datastore
- "my-pool" gets deleted
- New "my-pool" gets created with different CIDR
- Felix resyncs with datastore.  Syncer sends an update, not a delete/add.

Granted, that's pretty unlikely but if we let that type of problem stack up in the codebase, our robustness overall will drop.  Suggest you just sanity check that the CIDR is the same and if not, handle as a delete and re-add.

calc/vxlan_resolver.go
```diff
@@ -0,0 +1,453 @@
...
+)
+
+func NewVXLANResolver(hostname string, callbacks PipelineCallbacks) *VXLANResolver {
+	return &VXLANResolver{
+		hostname:                  hostname,
+		callbacks:                 callbacks,
+		nodeNameToVXLANTunnelAddr: map[string]string{},
+		nodeNameToIPAddr:          map[string]string{},
+		nodeNameToSentRoutes:      map[string]set.Set{},
+		blockToRoutes:             map[string]map[string]*vxlanRoute{},
```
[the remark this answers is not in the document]
fasaxc: Oh, maybe move the constructor below the struct so the lovely design comment comes first?

calc/vxlan_resolver.go
```diff
@@ -0,0 +1,453 @@
...
+// If a VTEP changes (e.g., due to a vxlan tunnel address changing), this component will treat
+// it as a delete followed by an add.
+type VXLANResolver struct {
+	hostname  string
+	callbacks PipelineCallbacks
+
+	// pendingRouteUpdates stores the set of route updates which
+	// we may want to program in the future, but cannot yet because
+	// they are not yet valid (e.g., due to a missing VTEP).
+	pendingRouteUpdates set.Set
```
fasaxc: Wondering if we can do away with the pending and sent sets here by calculating whether the route would already have been sent.

dataplane/driver.go
```diff
@@ -26,8 +26,8 @@ import (
 	"runtime/debug"
 
 	"github.com/projectcalico/felix/config"
-	"github.com/projectcalico/felix/dataplane/external"
-	"github.com/projectcalico/felix/dataplane/linux"
+	extdataplane "github.com/projectcalico/felix/dataplane/external"
```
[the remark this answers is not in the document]
caseydavenport: Yeah, goimports seems to be doing this recently - I definitely didn't do this deliberately!

dataplane/linux/vxlan_mgr.go
```diff
@@ -0,0 +1,307 @@
...
+func (m *vxlanManager) OnUpdate(protoBufMsg interface{}) {
+	logrus.WithField("msg", protoBufMsg).Debug("Received message")
+	switch msg := protoBufMsg.(type) {
+	case *proto.RouteUpdate:
+		if msg.Type == "vxlan" {
+			logrus.Debugf("VXLAN data plane received route update")
+			m.pendingRouteUpdates[msg.Dst] = msg
+		}
+	case *proto.RouteRemove:
+		logrus.WithField("msg", msg).Warnf("VXLAN data plane received route remove")
```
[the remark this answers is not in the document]
caseydavenport: yep, I still need to implement this. I just had at WARNING so it was easier to grep the logs.

dataplane/linux/vxlan_mgr.go
```diff
@@ -0,0 +1,307 @@
...
+	}
+}
+
+// configureVXLANDevice ensures the VXLAN tunnel device is up and configures correctly.
+func (m *vxlanManager) configureVXLANDevice() error {
+	logCxt := logrus.WithFields(logrus.Fields{})
+	logCxt.Debug("Configuring VXLAN tunnel")
+	link, err := netlink.LinkByName("vxlan.calico")
+	if err != nil {
+		parent, err := netlink.LinkByName("ens4")
```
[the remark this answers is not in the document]
caseydavenport: yeah, I think we'll need something. Was wondering if we should just find the interface that has the node's IP address on it.

We could look at `myVTEP` and see what the parent IP is, then find that interface and grab it here. That minimizes the configuration required since it's basically derived from the IP autodetection logic in calico/node which is already fairly flexible (or whatever logic is already setting the node IP in environments where calico/node isn't used).

routetable/route_table.go
```diff
@@ -303,7 +331,26 @@ func (r *RouteTable) Apply() error {
...
+			// Sync L3 routes.
+			err = r.syncRoutesForLink(ifaceName)
+			if err == IfaceNotPresent {
+				logCxt.Info("Interface missing, will retry if it appears.")
+				break
+			} else if err == IfaceDown {
+				logCxt.Info("Interface down, will retry if it goes up.")
+				break
+			} else if err == IfaceGrace {
+				logCxt.Info("Interface in cleanup grace period, will retry after.")
```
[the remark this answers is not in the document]
caseydavenport: Yep. This check is used for all route types though, so I think the right fix is a level lower where we actually check if the route is within the grace period - we should only return this error if the route is not a VXLAN route.

routetable/route_table.go
```diff
@@ -329,6 +376,35 @@ func (r *RouteTable) Apply() error {
...
+	// For each L2 entry we need to program, program it.
+	// Add a static ARP entry.
+	a := &netlink.Neigh{
+		LinkIndex:    linkAttrs.Index,
+		State:        netlink.NUD_PERMANENT,
+		Type:         syscall.RTN_UNICAST,
+		IP:           target.GW.AsNetIP(),
+		HardwareAddr: target.VTEPMAC,
+	}
+	if err := netlink.NeighSet(a); err != nil {
```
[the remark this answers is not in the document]
caseydavenport: It seems to be from the description, but not sure if a "replace" to the same thing that's currently programmed is disruptive to the data plane or not.

I'd assume not, but a quick search doesn't show any obvious results. This is the same operation that flannel does, but flannel won't do this periodically IIUC so it might not be relevant there.

calc/vxlan_resolver.go
```diff
@@ -0,0 +1,453 @@
...
+	return
+}
+
+// OnHostConfigUpdate gets called whenever a node's host config changes. We only care about
+// VXLAN tunnel address updates. On an add/update, we need to check if there are VTEPs or routes which
+// are now valid, and trigger programming of them to the data plane. On a delete, we need to withdraw any
+// routes and VTEPs associated with the node.
+func (c *VXLANResolver) OnHostConfigUpdate(update api.Update) (_ bool) {
+	switch update.Key.(model.HostConfigKey).Name {
+	case "IPv4VXLANTunnelAddr":
```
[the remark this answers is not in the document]
caseydavenport: I can try to rework to use the Node object here - it shouldn't be too hard. And, I agree I like that better. 

One potential side effect of that is that it will no longer be a felixconfig, so won't be configurable via environment variable, etc. This is probably desirable for this though - I think we want it to come through the node.

calc/vxlan_resolver.go
caseydavenport: Sounds good to me so long as we're good with iterating all the routes more frequently. We already have the potential to iterate lots of routes with the current implementation anyway, so I'll see what it will take to adjust to something like this.

routetable/route_table.go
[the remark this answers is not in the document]
caseydavenport: Not sure how to do this generically since this is shared code - I don't think it's really a _problem_ per-se, just might take a bit longer to cleanup VXLAN routes added by someone else. Any ideas on how to clean this up? Or, is it worth it?

routetable/route_table.go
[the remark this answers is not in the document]
caseydavenport: Running an `ip monitor`, I can see that the current code does write new neighbor entries periodically, but I don't see any deletes, so I suspect that we're all good here in terms of data plane.

calc/vxlan_resolver.go
```diff
@@ -0,0 +1,476 @@
...
+		node: b.Host(),
+	}
+	return map[string]vxlanRoute{r.Key(): r}
+}
+
+// determineGatewayForRoute determines which gateway to use for this route. For VXLAN routes, the
+// gateway is the remote node's IPv4VXLANTunnelAddr. If we don't know the remote node's tunnel
+// address, this function will return an empty string.
+func (c *VXLANResolver) determineGatewayForRoute(r vxlanRoute) string {
+	if ip, ok := c.nodeNameToVXLANTunnelAddr[r.node]; ok {
```
fasaxc: You could just do `return c.nodeNameToVXLANTunnelAddr[r.node]` since accessing a missing map entry returns the zero value for the type.

calc/vxlan_resolver.go
```diff
@@ -0,0 +1,476 @@
...
+		return ip
+	}
+	return ""
+}
+
+// vtepMACForHost calculates a deterministic MAC address based on the provided host.
+// The returned address matches the address assigned to the VXLAN device on that node.
+func (c *VXLANResolver) vtepMACForHost(nodename string) string {
```
fasaxc: Does this match flannel's algorithm?  It'd probably need to if we want to support migration at some point.

dataplane/linux/vxlan_mgr.go
```diff
@@ -0,0 +1,379 @@
...
+
+func (m *vxlanManager) getLocalVTEP() *proto.VXLANTunnelEndpointUpdate {
+	m.Lock()
+	defer m.Unlock()
+	return m.myVTEP
+}
+
+func (m *vxlanManager) setDirty() {
+	select {
+	case m.dirty <- true:
```
fasaxc: Why do you need this channel machinery?  I can't see anywhere that it gets used from another goroutine

routetable/route_table.go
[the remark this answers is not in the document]
fasaxc: If it's tough to do, think we can live with.  Th grace period is something like 30s.

calc/vxlan_resolver.go
```diff
@@ -0,0 +1,476 @@
...
+}
+
+// OnHostIPUpdate gets called whenever a node IP address changes. On an add/update,
+// we need to check if there are VTEPs or routes which are now valid, and trigger programming
+// of them to the data plane. On a delete, we need to withdraw any routes and VTEPs associated
+// with the node.
+func (c *VXLANResolver) OnHostIPUpdate(update api.Update) (_ bool) {
+	nodeName := update.Key.(model.HostIPKey).Hostname
+	logCxt := logrus.WithField("node", nodeName)
+	pendingSet, sentSet := c.routeSets()
```
fasaxc: Actually generating these sets will make this logic O(n^2) to add n nodes.  I'd expect it to start getting sluggish at 1000-node scale and to explode by 10k node scale.

I had in mind to scan only the relevant part of the index for each type of update, if if you get a block update, you only need to look at the routes that were in the old version of that block since a block update can't trigger routes from other blocks to be suddenly valid.  If you get a node update, you only need to scan the routes that have that node as next hop.

The approach you've got looks correct though and should work fine for 100s of nodes, so I think any changes to be more precise in the indexing should be done in a follow-up increment.

fasaxc:
I can think of some contrived ways that we could get clashing tunnel addresses (resync while two nodes get recreated).  What would actually happen in that case, what's the blast radius and would it fix itself (by triggering a panic if nothing else)?

dataplane/linux/vxlan_mgr.go
caseydavenport: Hrm, for some reason I thought `CompleteDeferredWork` was called from another goroutine... but looking at it now I don't see that it is... I'll clean it up.

calc/vxlan_resolver.go
```diff
@@ -0,0 +1,476 @@
...
+			// We need to revoke it and any routes before sending any updates.
+			logCxt.Info("Withdrawing routes and VTEP, node changed IP address")
+			sentSet.Iter(func(item interface{}) error {
+				r := item.(vxlanRoute)
+				if r.node == nodeName {
+					c.withdrawRoute(r)
+				}
+				return nil
+			})
+			c.sendVTEPRemove(nodeName)
```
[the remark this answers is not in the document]
caseydavenport: Yep. IIUC it will handle it properly the way it is currently written, will want to test though.

I think the current data plane will update its existing entries, triggering some updates routes in the route table.

calc/vxlan_resolver.go
caseydavenport: Bah, it looks like flannel doesn't use an algorithmic approach for this - it relies on the kernel to generate a MAC for the device when creating it, rather than doing deterministic calculation of it.

https://github.com/coreos/flannel/blob/master/backend/vxlan/vxlan.go#L137

calc/vxlan_resolver.go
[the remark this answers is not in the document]
caseydavenport: I think we should do this as a follow on - I'll raise something to track.

caseydavenport:
>  I can think of some contrived ways that we could get clashing tunnel addresses (resync while two nodes get recreated). What would actually happen in that case, what's the blast radius and would it fix itself (by triggering a panic if nothing else)?

Since everything is keyed by node, rather than by address, they won't end up hiding each other in the calc graph.

We'll end up with two VTEP entries which share a tunnel address, which means we'll get two ARP entries for the same IP but with different MAC addresses. One of them will be programmed second, and will win out.

Once the duplicate IPs are fixed, it should resolve itself by sending updates with the new IP down to the data plane, at which point the ARP entries won't conflict and both will be programmed. 

So, I think it should just work.

config/config_params.go
```diff
@@ -166,7 +166,7 @@ type Config struct {
 
 	VXLANEnabled        bool   `config:"bool;false"`
 	VXLANPort           int    `config:"int;0"`
-	VXLANID             int    `config:"int;0"`
+	VXLANVNI            int    `config:"int;0"`
 	VXLANMtu            int    `config:"int;1440;non-zero"`
```
fasaxc: Casing between this and the value above is inconsistent.

calc/vxlan_resolver.go
caseydavenport: @fasaxc wondering if we can merge as-is and then do a follow up to adjust this.

I'm not really sure what we can do about this - we need to teach flannel about our local VXLAN tunnel endpoints, and also learn about Flannel's in order to make this work, which means either augmenting Calico to write to flannel's API or vice-versa, I think.

fasaxc: Maybe we just start storing the MAC in the datamodel?  We already store the IP.

routetable/route_table.go
```diff
@@ -503,6 +595,101 @@ func (r *RouteTable) syncRoutesForLink(ifaceName string) error {
...
+		return err
+	}
+
+	// For each existing neighbor, if it is not present in the expected set, then remove it.
+	var updatesFailed bool
+	for _, existing := range existingNeigh {
+		if existing.Family == syscall.AF_BRIDGE {
+			// FDB entries have family set to bridge.
+			if _, ok := expectedFDBEntries[existing.IP.String()]; !ok {
+				logCxt.WithField("neighbor", existing).Warnf("Removing old neighbor entry (FDB)")
```
fasaxc: Should be `Info`?

routetable/route_table.go
```diff
@@ -503,6 +595,101 @@ func (r *RouteTable) syncRoutesForLink(ifaceName string) error {
...
+				if err := netlink.NeighDel(&existing); err != nil {
+					updatesFailed = true
+					continue
+				}
+			}
+		} else {
+			if _, ok := expectedARPEntries[existing.IP.String()]; !ok {
+				logCxt.WithField("neighbor", existing).Warnf("Removing old neighbor entry (ARP)")
```
fasaxc: Same here

PR #1991, opened 2019-04-11
Dataplane: optimize early host-endpoint blacklist rules

iaguis:
## Description

This adds support for optimizing suitable GlobalNetworkPolicy rules with XDP.
Given host support, eligible policies will be optimized automatically.

It adds a new XDP Manager that keeps the state for XDP and defines callbacks to
get information from the calculation graph. Additionally, existing managers are
extended so the XDP Manager can access their information.

To implement low-level functionality, it adds a BPF library that shells out to
the `bpftool` and `ip` tools.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```


caseydavenport:
Config docs here: https://github.com/projectcalico/calico/pull/2546
And user guide: https://docs.projectcalico.org/master/security/defend-dos-attack

PR #1993, opened 2019-04-15
Fix vxlan device recreation

caseydavenport:
Previously, we were only checking if the device needed to be recreated if it didn't already exist, which is just wrong really since if the device doesn't exist then we can create it exactly how we want it!

This moves that logic outside of the "failed to query device" branch, and instead will perform the check of every time, even if the device exists. 

- [ ] Documentation
- [ ] Backport
- [ ] Release note

```release-note
None required
```

### DIFF
--- a/dataplane/linux/vxlan_mgr.go
+++ b/dataplane/linux/vxlan_mgr.go
@@ -187,6 +187,7 @@ func (m *vxlanManager) KeepVXLANDeviceInSync(mtu int) {
 			time.Sleep(1 * time.Second)
 			continue
 		}
+		logrus.Info("VXLAN tunnel device configured")
 		time.Sleep(10 * time.Second)
 	}
 }
@@ -248,12 +249,9 @@ func (m *vxlanManager) configureVXLANDevice(mtu int, localVTEP *proto.VXLANTunne
 		}
 
 		// The device now exists - requery it to check that the link exists and is a vxlan device.
-		link, err = netlink.LinkByIndex(vxlan.Index)
+		link, err = netlink.LinkByName(m.vxlanDevice)
 		if err != nil {
-			return fmt.Errorf("can't locate created vxlan device with index %v", vxlan.Index)
-		}
-		if _, ok := link.(*netlink.Vxlan); !ok {
-			return fmt.Errorf("created vxlan device with index %v is not vxlan", vxlan.Index)
+			return fmt.Errorf("can't locate created vxlan device %v", m.vxlanDevice)
 		}
 	}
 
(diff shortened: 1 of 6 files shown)


dataplane/linux/vxlan_mgr.go
```diff
@@ -216,62 +216,61 @@ func (m *vxlanManager) getParentInterface(localVTEP *proto.VXLANTunnelEndpointUp
...
+			// Device already exists - likely a race.
+			logrus.Debug("VXLAN device already exists, likely created by someone else.")
 		} else if err != nil {
+			// Error other than "device exists" - return it.
 			return err
 		}
 
-		ifindex := vxlan.Index
+		// The device now exists - requery it to check that the link exists and is a vxlan device.
 		link, err = netlink.LinkByIndex(vxlan.Index)
```
fasaxc: Won't vxlan.Index be 0?  I think you only just created that struct above.

dataplane/linux/vxlan_mgr.go
```diff
@@ -216,62 +216,61 @@ func (m *vxlanManager) getParentInterface(localVTEP *proto.VXLANTunnelEndpointUp
...
-			// Existing device doesn't match desired configuration - delete it and recreate.
-			logrus.Warningf("%q already exists with incompatable configuration: %v; recreating device", vxlan.Name, incompat)
-			if err = netlink.LinkDel(existing); err != nil {
-				return fmt.Errorf("failed to delete interface: %v", err)
-			}
-			if err = netlink.LinkAdd(vxlan); err != nil {
-				return fmt.Errorf("failed to create vxlan interface: %v", err)
-			}
+			// Device already exists - likely a race.
+			logrus.Debug("VXLAN device already exists, likely created by someone else.")
```
fasaxc: If you hit this case, need to re-query the device?

fasaxc:
Could do with some FVs, presumably felix can create a vxlan device inside a felix FV test container so should be fairly straightforward.

fv/infrastructure/topology.go
```diff
@@ -166,6 +172,9 @@ func StartNNodeTopology(n int, opts TopologyOptions, infra DatastoreInfra) (feli
 		if opts.IPIPEnabled {
 			infra.SetExpectedIPIPTunnelAddr(felix, i, bool(n > 1))
 		}
+		if opts.VXLANEnabled {
+			infra.SetExpectedVXLANTunnelAddr(felix, i, bool(n > 1))
+		}
 
 		var w chan struct{}
 		if felix.ExpectedIPIPTunnelAddr != "" {
```
fasaxc: Should probably wait for VXLAN IP to show up in the log too.  That removed a lot of flakes from the IPIP tests.

PR #1994, opened 2019-04-16
Add calc graph FVs for VXLAN processing.

fasaxc:
## Description

Add calc graph FVs for VXLAN processing.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/calc/vxlan_resolver.go
+++ b/calc/vxlan_resolver.go
@@ -269,6 +269,7 @@ func (c *VXLANResolver) OnPoolUpdate(update api.Update) (_ bool) {
 	pendingSet, sentSet := c.routeSets()
 	if update.Value != nil && update.Value.(*model.IPPool).VXLANMode != encap.Undefined {
 		// This is an add/update of a pool with VXLAN enabled.
+		logrus.WithField("pool", k.CIDR).Info("Update of VXLAN-enabled IP pool.")
 		if curr, ok := c.vxlanPools[k.String()]; ok {
 			// We already know about this IP pool. Check to see if the CIDR has changed.
 			// While this isn't possible directly in the user-facing API, it's possible that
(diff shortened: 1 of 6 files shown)


calc/vxlan_resolver.go
```diff
@@ -351,15 +358,15 @@ func (c *VXLANResolver) routeReady(r vxlanRoute) bool {
 	logCxt := logrus.WithField("route", r)
 	gw := c.determineGatewayForRoute(r)
 	if gw == "" {
-		logCxt.Info("No gateway yet for VXLAN route, skip")
+		logCxt.Debug("No gateway yet for VXLAN route, skip")
```
fasaxc: I made these Info before but then realised they could be too spammy so backing them back down to debug.

calc/vxlan_resolver.go
```diff
@@ -127,10 +128,12 @@ func (c *VXLANResolver) OnBlockUpdate(update api.Update) (_ bool) {
 	} else {
 		// Block has been deleted. Clean up routes that were contributed by this block.
 		routes := c.blockToRoutes[key]
-		routes.Iter(func(item interface{}) error {
-			c.withdrawRoute(item.(vxlanRoute))
-			return nil
-		})
+		if routes != nil {
```
fasaxc: Route can be missing if you get a duplicate delete event, which is possible if the first delete was actually a validation error.  Not sure if it's possible end-to-end with the watcher cache layer, now but it was easier to fix than to prevent the calc graph FV suite from generating this case automatically.

calc/vxlan_resolver.go
```diff
@@ -208,7 +211,8 @@ func (c *VXLANResolver) OnHostConfigUpdate(update api.Update) (_ bool) {
 	case "IPv4VXLANTunnelAddr":
 		nodeName := update.Key.(model.HostConfigKey).Hostname
 		vtepSent := c.vtepSent(nodeName)
-		logCxt := logrus.WithField("node", nodeName)
+		logCxt := logrus.WithField("node", nodeName).WithField("value", update.Value)
```
caseydavenport: I think it's a bit weird to add the "value" to the context logger, since we augment with the old / new IP below so we'll get duplicate info in the log I think?

Maybe just log it out in the debug line itself?

PR #1995, opened 2019-04-16
Add support for borrowed IPs to VXLAN

fasaxc:
## Description

- Calculate the borrowed IPs from the IPAM block directly,
- Add calc graph tests for VXLAN.
- Fix that changing the affinity of a block wouldn't remove the old route from the cache.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```


calc/vxlan_resolver.go
```diff
@@ -432,19 +441,37 @@ func (c *VXLANResolver) routeWithinVXLANPool(r vxlanRoute) bool {
...
+		r := vxlanRoute{
+			dst:  ip.CIDRFromNetIP(alloc.Addr.IP),
+			node: alloc.Host,
+		}
+		routes[r.Key()] = r
+	}
+
 	if b.Host() == c.hostname {
 		logrus.Debug("Skipping VXLAN routes for local node")
 		return nil
```
caseydavenport: I think we need to return `routes` here instead of nil, since it might contain borrowed routes from this block.

Alternatively, we just move this check down below:

```
if b.Host() != c.hostname && b.Host != "" {
  // Add the route for the block CIDR if it is for a remote node.
}
```

fasaxc: Good catch

fasaxc: Fixed with a new calc graph test to prove it :-)

PR #2009, opened 2019-04-26
Replicate tunnel masquerade rule for VXLAN.

fasaxc:
## Description

I bet we need this same logic for VXLAN that we have for IPIP.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```


caseydavenport:
@fasaxc LGTM - could you make sure to pick this into the v3.7 branch as well (and we'll need to update the node pin)

PR #2017, opened 2019-05-02
Form Typha address string correctly with an IPv6 address

neiljerram:


daemon/daemon.go
```diff
@@ -1014,18 +1013,7 @@ func discoverTyphaAddr(configParams *config.Config) (string, error) {
 
 	// If we get here, we need to look up the Typha service using the k8s API.
 	// TODO Typha: support Typha lookup without using rest.InClusterConfig().
```
jpalardy: this TODO is done :-)

PR #2019, opened 2019-05-07
refactor repeated code into CreateDefaultProfile()

rafaelvanoni:


fv/etcd_restart_test.go
```diff
@@ -51,16 +51,7 @@ var _ = Context("etcd connection interruption", func() {
...
-		// Install a default profile that allows all ingress and egress, in the absence of any Policy.
-		defaultProfile := api.NewProfile()
-		defaultProfile.Name = "default"
-		defaultProfile.Spec.LabelsToApply = map[string]string{"default": ""}
-		defaultProfile.Spec.Egress = []api.Rule{{Action: api.Allow}}
-		defaultProfile.Spec.Ingress = []api.Rule{{Action: api.Allow}}
-		_, err := client.Profiles().Create(utils.Ctx, defaultProfile, utils.NoOptions)
-		Expect(err).NotTo(HaveOccurred())
-
+		infrastructure.CreateDefaultProfile(client, "default", map[string]string{"default": ""}, "")
```
neiljerram: In the old code here there was no Source Selector, but CreateDefaultProfile will specify a Source Selector of `""`.  Is that equivalent?

fv/named_ports_test.go
```diff
@@ -1015,14 +993,7 @@ var _ = Describe("tests with mixed TCP/UDP", func() {
...
 		felix, etcd, client = infrastructure.StartSingleNodeEtcdTopology(infrastructure.DefaultTopologyOptions())
-
-		// Create a profile that opens up traffic by default.
-		defaultProfile := api.NewProfile()
-		defaultProfile.Name = "open"
-		defaultProfile.Spec.Egress = []api.Rule{{Action: api.Allow}}
-		defaultProfile.Spec.Ingress = []api.Rule{{Action: api.Allow}}
-		_, err := client.Profiles().Create(utils.Ctx, defaultProfile, utils.NoOptions)
-		Expect(err).NotTo(HaveOccurred())
+		infrastructure.CreateDefaultProfile(client, "open", map[string]string{"default": ""}, "")
```
neiljerram: The old code here had no labels, so I think you shouldn't pass a map here.

fv/etcd_restart_test.go
rafaelvanoni: I've added a check in CreateDefaultProfile() to only assign a source selector if the given string's length is not zero.

neiljerram: Thanks, that sounds good.

fv/infrastructure/infrastructure.go
```diff
@@ -71,3 +73,23 @@ type DatastoreInfra interface {
...
+		d.Spec.LabelsToApply = labels
+	}
+	d.Spec.Egress = []api.Rule{{Action: api.Allow}}
+
+	if len(entityRuleSelector) > 0 {
+		d.Spec.Ingress = []api.Rule{{
+			Action: api.Allow,
+			Source: api.EntityRule{Selector: entityRuleSelector},
+		}}
+	}
```
neiljerram: I think we still need `d.Spec.Ingress = []api.Rule{{Action: api.Allow}}` when the selector is empty, don't we?

PR #2028, opened 2019-05-29
Add support for NATOutgoingAddress option

maxstr:
## Description
Related to https://github.com/projectcalico/libcalico-go/pull/1089.

As in that PR, this is for addressing projectcalico/calico#2222. We chose to implement this by allowing users to specify the IP address being used rather than using the BGP address by default because it seemed more flexible. For our use case we are using the Kubernetes downward API to supply this address. I've tested this by setting/unsetting FELIX_NATOUTGOINGADDRESS in a calico node container using the modified felix.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/config/config_params.go
+++ b/config/config_params.go
@@ -199,6 +199,7 @@ type Config struct {
 
 	KubeNodePortRanges []numorstring.Port `config:"portrange-list;30000:32767"`
 	NATPortRange       numorstring.Port   `config:"portrange;"`
+	NATOutgoingAddress net.IP             `config:"ipv4;"`
 
 	UsageReportingEnabled          bool          `config:"bool;true"`
 	UsageReportingInitialDelaySecs time.Duration `config:"seconds;300"`
--- a/rules/nat.go
+++ b/rules/nat.go
@@ -48,18 +48,28 @@ func (r *DefaultRuleRenderer) MakeNatOutgoingRule(protocol string, action iptabl
 func (r *DefaultRuleRenderer) NATOutgoingChain(natOutgoingActive bool, ipVersion uint8) *iptables.Chain {
 	var rules []iptables.Rule
 	if natOutgoingActive {
+		var defaultSnatRule iptables.Action = iptables.MasqAction{}
+		if r.Config.NATOutgoingAddress != nil {
+			defaultSnatRule = iptables.SNATAction{ToAddr: r.Config.NATOutgoingAddress.String()}
+		}
+
 		if r.Config.NATPortRange.MaxPort > 0 {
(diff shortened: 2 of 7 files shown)


config/config_params.go
```diff
@@ -225,6 +225,8 @@ type Config struct {
 
 	XDPEnabled        bool `config:"bool;true"`
 	GenericXDPEnabled bool `config:"bool;false"`
+
+	NATOutgoingAddress net.IP `config:"ipv4;"`
```
fasaxc: Suggest moving this up to line 200 since it naturally groups with `NATPortRange`

maxstr:
Added validation to my PR here: https://github.com/projectcalico/libcalico-go/pull/1089. Looks like that PR will need to be merged and the Glidefile updated here in order to build?

rules/nat.go
```diff
@@ -48,18 +48,28 @@ func (r *DefaultRuleRenderer) MakeNatOutgoingRule(protocol string, action iptabl
 func (r *DefaultRuleRenderer) NATOutgoingChain(natOutgoingActive bool, ipVersion uint8) *iptables.Chain {
 	var rules []iptables.Rule
 	if natOutgoingActive {
+		var defaultSnatRule iptables.Action = iptables.MasqAction{}
+		if r.Config.NATOutgoingAddress != nil {
```
steven-sheehy: Can this also check if the NATOutgoingAddress exists on an interface and fallback to masquerading if not? My use case is for floating/virtual IPs that will only exist on one of the calico nodes. But even without using a VIP it is good error checking to validate the IP exists.

maxstr: It seems like the correct place/time to check that would be on startup/reconfigure, but not sure if the correct behavior would be to error out or to warn and fallback. Happy to defer to the Calico folks on what it should be and implement it though.

neiljerram: @steven-sheehy It doesn't feel right to me to add that feature into this PR.  I'd rather merge this PR first, then that can be added as a delta if it really needs a Felix code change.

Anyway, I wonder if this can be done (in a Kubernetes context) by combining an initContainer with the downward API?  Something like:
- use the downward API to populate an env var for an initContainer
- get the initContainer to check that that address really exists
- if so, also set it as FELIX_NATOUTGOINGADDRESS for the calico/node container.

maxstr:
Rebased and cherry-picked your bump.

rules/nat.go
steven-sheehy: Ok, let's handle it in another PR.

About the initContainer, it can't be done solely at calico startup since this IP can come and go at any time and it needs calico to watch for interface changes and adjust the NAT rules.

PR #2029, opened 2019-05-31
Added code and test to enable vxlan MAC if present in node spec.

tathagatachowdhury:
## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/calc/states_for_test.go
+++ b/calc/states_for_test.go
@@ -906,6 +906,19 @@ var vxlanWithBlock = empty.withKVUpdates(
 	},
 )
 
+// Minimal VXLAN set-up with a MAC address.
+var vxlanWithMAC = vxlanWithBlock.withKVUpdates(
+	KVPair{Key: remoteHostVXLANTunnelMACConfigKey, Value: remoteHostVXLANTunnelMAC},
+).withName("VXLAN MAC").withVTEPs(
+	// VTEP for the remote node.
+	proto.VXLANTunnelEndpointUpdate{
+		Node:           remoteHostname,
+		Mac:            remoteHostVXLANTunnelMAC,
+		Ipv4Addr:       remoteHostVXLANTunnelIP,
+		ParentDeviceIp: remoteHostIP.String(),
+	},
+)
+
 // As above but with a more complex block.  The block has some allocated IPs on the same
 // node as well as one that's borrowed by a second node.  We add the extra VTEP config for the
 // other node.
--- a/calc/vxlan_resolver.go
+++ b/calc/vxlan_resolver.go
@@ -64,6 +64,7 @@ type VXLANResolver struct {
 	// block that contributed them. The following comprises the full internal data model.
 	nodeNameToVXLANTunnelAddr map[string]string
 	nodeNameToIPAddr          map[string]string
+	nodeNameToVXLANMac        map[string]string
 	blockToRoutes             map[string]set.Set
 	vxlanPools                map[string]model.IPPool
 }
@@ -74,6 +75,7 @@ func NewVXLANResolver(hostname string, callbacks PipelineCallbacks) *VXLANResolv
 		callbacks:                 callbacks,
 		nodeNameToVXLANTunnelAddr: map[string]string{},
 		nodeNameToIPAddr:          map[string]string{},
+		nodeNameToVXLANMac:        map[string]string{},
 		blockToRoutes:             map[string]set.Set{},
 		vxlanPools:                map[string]model.IPPool{},
 	}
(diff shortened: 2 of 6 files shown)


calc/vxlan_resolver.go
```diff
@@ -263,6 +265,48 @@ func (c *VXLANResolver) OnHostConfigUpdate(update api.Update) (_ bool) {
...
+					logCxt.Info("Sent VTEP to dataplane, check pending routes")
+					c.kickPendingRoutes(pendingSet)
+				}
+			}
+
+		} else {
+			logCxt.Info("Update the VTEP with the system generated MAC address and send it to dataplane")
+			_, ok := c.nodeNameToVXLANMac[nodeName]
+			if ok {
+				delete(c.nodeNameToVXLANMac, nodeName)
```
fasaxc: I think it is safe to delete from a map even if the key is not present so you can elide this check.

calc/states_for_test.go
```diff
@@ -906,6 +906,19 @@ var vxlanWithBlock = empty.withKVUpdates(
 	},
 )
 
+// Minimal VXLAN set-up with a MAC address.
+var vxlanWithMAC = empty.withKVUpdates(
```
fasaxc: Think you need to `s/empty/vxlanWithBlock/`

calc/vxlan_resolver.go
```diff
@@ -263,6 +265,48 @@ func (c *VXLANResolver) OnHostConfigUpdate(update api.Update) (_ bool) {
...
 		}
+	case "VXLANTunnelMACAddr":
+		nodeName := update.Key.(model.HostConfigKey).Hostname
+		vtepSent := c.vtepSent(nodeName)
+		logCxt := logrus.WithField("node", nodeName).WithField("value", update.Value)
+		logCxt.Debug("VXLANTunnelMACAddr update")
+		if update.Value != nil {
+			// Update for a VXLAN tunnel MAC address.
+			newMAC := update.Value.(string)
+			currMAC := c.nodeNameToVXLANMac[nodeName]
```
fasaxc: I think you should get `currMAC` by calling `vtepMACForHost()`  Then we'll avoid sending a dupe if the datastore is updated to contain the auto-generated MAC.

calc/vxlan_resolver.go
```diff
@@ -263,6 +265,48 @@ func (c *VXLANResolver) OnHostConfigUpdate(update api.Update) (_ bool) {
...
+					// If we've already handled this node, there's nothing to do. Deduplicate.
+					logCxt.Debug("Skipping duplicate tunnel MAC addr update")
+					return
+				}
+
+				// Try sending a VTEP update. If we do, this will trigger a kick of any
+				// pending routes which might now be ready to send.
+				c.nodeNameToVXLANMac[nodeName] = newMAC
+				if c.sendVTEPUpdate(nodeName) {
+					// We've successfully sent a new VTEP - check to see if any pending routes
```
fasaxc: I don't think we can actually kick any routes here since adding/removing the MAC can't make the VTEP valid/invalid.  Could probably remove this call to `kickPendingRoutes()`

calc/vxlan_resolver.go
```diff
@@ -263,6 +265,48 @@ func (c *VXLANResolver) OnHostConfigUpdate(update api.Update) (_ bool) {
...
+			_, ok := c.nodeNameToVXLANMac[nodeName]
+			if ok {
+				delete(c.nodeNameToVXLANMac, nodeName)
+			}
+
+			if c.sendVTEPUpdate(nodeName) {
+				// We've successfully sent a new VTEP - check to see if any pending routes
+				// are now ready to be programmed.
+				logCxt.Info("Sent VTEP to dataplane, check pending routes")
+				c.kickPendingRoutes(pendingSet)
```
fasaxc: I don't think we can actually kick any routes here since adding/removing the MAC can't make the VTEP valid/invalid. Could probably remove this call to kickPendingRoutes()

calc/states_for_test.go
tathagatachowdhury: Acknowledged and changed the code.

calc/vxlan_resolver.go
tathagatachowdhury: Acknowledged and changed the code.

calc/vxlan_resolver.go
tathagatachowdhury: Acknowledged and changed the code.

calc/vxlan_resolver.go
tathagatachowdhury: Acknowledged and changed the code.

calc/vxlan_resolver.go
tathagatachowdhury: Acknowledged and changed the code.

fasaxc:
Marked as Calico v3.8.0 since the libcalico-go change for this has been merged.

calc/vxlan_resolver.go
```diff
@@ -499,6 +535,12 @@ func (c *VXLANResolver) determineGatewayForRoute(r vxlanRoute) string {
 // vtepMACForHost calculates a deterministic MAC address based on the provided host.
```
fasaxc: Out of date comment

tathagatachowdhury: Updated the comment.

PR #2036, opened 2019-06-07
fv: add sockmap test

iaguis:
## Description

This PR adds a test that checks connections are accelerated via sockmap. Note sockmap tests won't run on CI because they require a very recent kernel version and access to BPF and they can't run in parallel because they interfere with each other.

It also adds some small improvements, code to clean up sockmap state after running tests, and reverts the addition of the SockmapCgroupV2Subdir parameter since we can't really use it for tests (see commit message for details).

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```


Makefile
```diff
@@ -328,10 +328,18 @@ protobuf proto/felixbackend.pb.go: proto/felixbackend.proto
...
-bpf/xdp/generated/xdp.o: bpf/xdp/filter.c
-	mkdir -p bpf/xdp/generated
+BPF_INC_FILES := bpf/include/bpf.h
+BPF_XDP_INC_FILES :=
+
+CLANG_BUILDER_STAMP := .built-bpf-clang-builder-$(BUILDARCH)
+
+$(CLANG_BUILDER_STAMP): docker-build-images/bpf-clang-builder.Dockerfile.$(BUILDARCH)
 	# the bpf object file is not arch dependent, so we can build with the current ARCH
 	docker build -t calico-build/bpf-clang -f docker-build-images/bpf-clang-builder.Dockerfile.$(BUILDARCH) .
```
fasaxc: Did you add this extra machinery because the docker command was slow?  I think you can speed it up a lot by changing `.` to `docker-build-images` so that it no longer sends the whole felix dir as build context.

fv/sockmap_test.go
```diff
@@ -0,0 +1,322 @@
...
+	"github.com/projectcalico/felix/fv/workload"
+	"github.com/projectcalico/libcalico-go/lib/apiconfig"
+)
+
+type mapEntry struct {
+	Key []string `json:"key"`
+}
+
+// Meh.
+func strSliceEqual(a, b []string) bool {
```
fasaxc: For tests, I'd probably just reach for `reflect.DeepEqual()`

fasaxc: (or arrange it so my code just returns the slice and then use `Expect(theSlice).To(Equal(...))`

fv/sockmap_test.go
```diff
@@ -0,0 +1,322 @@
...
+	})
+
+	AfterEach(func() {
+		host.Stop()
+		felix.Stop()
+		func() {
+			opts := getSockmapOpts()
+			felix, _ = infrastructure.StartSingleNodeTopology(opts, infra)
+			defer felix.Stop()
+			outputBytes, err := utils.Command(
```
fasaxc: Think you can do `felix.ExecOutput()`

fasaxc: Won't felix already be stopped/stopping due  to the call above?  Should this run before we stop the felix container?

fv/sockmap_test.go
```diff
@@ -0,0 +1,322 @@
...
+			}
+			log.Info("Cleanup finished")
+		}()
+		infra.Stop()
+	})
+
+	It("should put the IP of the host in sockmap endpoints map", func() {
+		found := false
+		maxTries := 5
+		for try := 0; try < maxTries; try++ {
```
fasaxc: Ginkgo has quite a nice way of doing retry loops.  If you made the function return the `entries` slice, you could do this:
```
Eventually(getEntries, "5s", "100ms").Should(Contain(expectedSlice))
```

fv/sockmap_test.go
```diff
@@ -0,0 +1,322 @@
...
+	return true
+}
+
+func prepend0xInPlace(hexen []string) {
+	for idx := range hexen {
+		hexen[idx] = "0x" + hexen[idx]
+	}
+}
+
+func getSockmapKeys(ip string, port int) [][]string {
```
fasaxc: How about `getExpectedSockmapKeys` (and maybe a bit of a comment explaining the format?)

fv/test-workload/test-workload.go
```diff
@@ -316,3 +337,20 @@ func writeProcSys(path, value string) error {
 	}
 	return err
 }
+
+func doIptablesSetup() error {
```
fasaxc: Worth a comment that the purpose of this is to simulate a sidecar?

fv/workload/workload.go
```diff
@@ -297,6 +299,177 @@ func (w *Workload) LatencyTo(ip, port string) time.Duration {
...
+}
+
+func startPermanentConnection(w *Workload, ip string, port, sourcePort int) (*PermanentConnection, error) {
+	// Ensure that the host has the 'test-connection' binary.
+	w.C.EnsureBinary("test-connection")
+	permConnIdx++
+	n := fmt.Sprintf("%s-pc%d", w.Name, permConnIdx)
+	loopFile := fmt.Sprintf("/tmp/%s-loop", n)
+
+	err := utils.Command("docker", "exec", w.C.Name,
```
fasaxc: We have a utility method for running commands inside the container.  `w.Exec` or `w.C.Exec`, I think

fv/workload/workload.go
```diff
@@ -297,6 +299,177 @@ func (w *Workload) LatencyTo(ip, port string) time.Duration {
...
+		fmt.Sprintf("--loop-with-file=%s", loopFile),
+	)
+	logName := fmt.Sprintf("permanent connection %s", n)
+	if err := utils.LogOutput(runCmd, logName); err != nil {
+		return nil, fmt.Errorf("failed to start output logging for %s", logName)
+	}
+	if err := runCmd.Start(); err != nil {
+		return nil, fmt.Errorf("failed to start a permanent connection: %v", err)
+	}
+	waitScript := []string{
```
fasaxc: Cleaner to do this loop in Go?

Something like this might work:
```
Eventually(func() err {return w.ExecMayFail("stat " + loopFile)}).Should(HaveOccured())
```

fv/test-connection/test-connection.go
```diff
@@ -35,11 +36,12 @@ import (
...
 
 Usage:
-  test-connection <namespace-path> <ip-address> <port> [--source-port=<source>] [--protocol=<protocol>]
+  test-connection <namespace-path> <ip-address> <port> [--source-port=<source>] [--protocol=<protocol>] [--loop-with-file=<file>]
 
 Options:
   --source-port=<source>  Source port to use for the connection [default: 0].
-  --protocol=<protocol>  Protocol to test [default: tcp].
+  --protocol=<protocol>   Protocol to test [default: tcp].
+  --loop-with-file=<file>  Whether to send messages repeatedly, file is used for synchronization
```
fasaxc: I found it a bit fiddly to unpick the loop-with-file "protocol".  Maybe you could spell it out in a comment?  I think
- File should exist before this binary is started.
- After first successful connection, this binary deletes the file.
- If the file reappears, this binary shuts down.

Makefile
krnowak: Yeah, that was the reason. Changing the directory indeed speeds that up greatly. Would you prefer to get rid of this extra machinery then?

fv/sockmap_test.go
krnowak: I rewrote the code to use more `Eventually` stuff, so I have dropped this function.

fv/sockmap_test.go
krnowak: Renamed and documented.

fv/sockmap_test.go
krnowak: We stop felix and then start a new one to clean up the old sockmap state completely. The new sockmap state will be empty, so we can easily detach the sockhash map from the cgroup manually without worrying about some weird corner cases resulting from failed test. I'll document it.

Used the `ExecOutput` function too.

fv/sockmap_test.go
krnowak: Yeah, I rewrote it to use `Eventually`.

fv/test-connection/test-connection.go
krnowak: Wrote a comment about it.

fv/test-workload/test-workload.go
krnowak: Done. Also renamed it to `doSidecarIptablesSetup`.

fv/workload/workload.go
krnowak: Done.

fv/workload/workload.go
krnowak: Thanks, did that and it seems to work fine.

PR #2037, opened 2019-06-11
Add support for setting the source hint on programmed device routes

maxstr:
## Description
Hi there, this is intended to address https://github.com/projectcalico/calico/issues/2271

It allows you to specify `DeviceRouteSourceAddress` as a felix configuration option. If you specify an IP for`DeviceRouteSourceAddress` then a source hint will be added to the device routes that felix programs with that address. This PR also correctly handles the case where the source hint needs to be added or removed and will update existing routes to either have or not have the hint.

Note that the initial route programmed by the CNI plugin will still lack the source address (I have a PR to address this that checks the felix configuration for the node, but it has not been tested as thoroughly as this one), but with this PR felix will detect the source address is missing and update it to be correct. 

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```


fasaxc:
@maxstr thanks for the PR and for including tests :-).  I gave the code a try locally and spotted a couple of issues:

- looks like the interaction with the grace period isn't quite right; we should skip the grace period if the destination is correct but the source is not
- we should also skip the conntrack cleanup in that case.

Since we're trying to get this into our impending release, I had a go at making the markups and fixing the conflicts here: https://github.com/projectcalico/felix/pull/2092 but then I realised that CLA bot isn't happy because it's no longer coming from your org.  If you agree with my changes, you could push them to your branch and we could get this merged.

maxstr:
Thanks, your changes look a lot cleaner :) After cherry-picking your commits it looks like I have to be apart of the bloomberg org now to satisfy CLA assistant, so will try and get that handled sometime today.

fasaxc:
@maxstr I think cherry-picking the merge commit has lost the merge metadata, causing conflicts.  I think doing a git merge of my changes to your branch would have done the right thing.  If you want to go back and do that, it should be a matter of
```
git checkout <your branch>
git reset 2d1a0c3283a13615ff1e070f2f1947fb00e8dd4e # your old commit
git merge ee60cff02567575b24054cad667ab7a243ce1749 # my commit
```

maxstr:
Repushed with a merge instead of the cherry picks.

kpfleming:
I will add @maxstr to the 'bloomberg' org to satisfy the CLA Assistant but this isn't the way CLA Assistant was setup to be used (we were the ones who added the org checking in the first place). The fact that Max was able to send a PR from a repository in the bloomberg org is sufficient evidence that the org's owners have granted such permission, since PRs can only be sent by people have write access on the source repository of the PR.

Please consider relaxing this 'public org membership' check as I don't see how it serves any useful purpose. Thanks.

fasaxc:
@kpfleming We use the hosted CLA assistant.   I don't think we've changed the config since it was previously working.  Perhaps adding my commits to the branch has caused this change, or perhaps they've just updated the hosted bot.

FWIW, I can't see any config option to disable that check.

maxstr:
Looks like CLA assistant is still not picking up I'm in the bloomberg organization. Want me to just reopen this PR and the other one in order to (hopefully) get it to pick it up?

fasaxc:
@maxstr did you make your org membership public? I can't see it in your list of orgs.

fasaxc:
I can override the check if needed but it'd be nice to have a proper solution going forward.

maxstr:
Ah, I thought if I could see it when I went to my profile that meant it was public. Looks like it's correct now.

fasaxc:
I've merged the libcalico-go change and updated the pin in the typha repo.  If you update the glide.yaml  to point at typha commit `adca5c7b84f5b9a84fd868073e6f6c71c45d29ec` and then run `make update-vendor`, the CI should pass.

fasaxc:
@maxstr Merged, thanks!  Please can you create a PR to document the new flag?  You'll need to modify the same two files as this PR: https://github.com/projectcalico/calico/pull/2654/files

maxstr:
Yep, will do!

maxstr:
@fasaxc Documented here: https://github.com/projectcalico/calico/pull/2779

PR #2052, opened 2019-06-24
Drop VXLAN and IPIP packets originating in pods

rafaelvanoni:


dataplane/linux/endpoint_mgr_test.go
```diff
@@ -135,11 +135,19 @@ func chainsForIfaces(ifaceMetadata []string,
 	epMarkMapper rules.EndpointMarkMapper,
 	host bool,
 	tableKind string) []*iptables.Chain {
+	const (
+		ProtoUDP  = 17
+		ProtoIPIP = 4
+		VXLANPort = 0
```
neiljerram: Why isn't this 4789 (which I think is the Felix default)?

dataplane/linux/endpoint_mgr_test.go
```diff
@@ -135,11 +135,19 @@ func chainsForIfaces(ifaceMetadata []string,
...
+		VXLANVNI  = 0
```
neiljerram: Similarly, 4096?

dataplane/linux/endpoint_mgr_test.go
neiljerram: Oh, I guess it's because you don't initialize that field in the Config struct.  I think it would be more realistic to do that initialization; or else to add a comment here to explain why it's 0.

dataplane/linux/endpoint_mgr_test.go
neiljerram: Same followup as above.

rules/endpoints.go
```diff
@@ -293,6 +293,19 @@ func (r *DefaultRuleRenderer) endpointIptablesChain(
...
+			VXLANVNI(uint32(r.Config.VXLANVNI)),
+		Action:  DropAction{},
+		Comment: "Drop VXLAN encapped packets originating in pods",
+	})
+	rules = append(rules, Rule{
+		Match:   Match().ProtocolNum(ProtoIPIP),
+		Action:  DropAction{},
+		Comment: "Drop IPinIP encapped packets originating in pods",
+	})
+
```
neiljerram: I believe these rules will be enforced both (1) on egress from a local workload, and (2) on ingress **to** a local workload.  Right?  I understand that we definitely want (1), but do we really want to enforce (2) as well?

neiljerram: What about when we go through these rules for a packet from or to a host endpoint?  Do we want to enforce them in that case?

neiljerram: (Relatedly, do we want these rules in raw and pre-DNAT endpoint chains?  raw and pre-DNAT are used for host endpoints only, but the pre-DNAT now allows the "*" host endpoint which includes local workload interfaces as well...)

rules/endpoints_test.go
```diff
@@ -27,6 +27,13 @@ import (
 )
 
 var _ = Describe("Endpoints", func() {
+	const (
+		ProtoUDP  = 17
+		ProtoIPIP = 4
+		VXLANPort = 0
+		VXLANVNI  = 0
```
neiljerram: Same points as in other test file.

rules/endpoints_test.go
```diff
@@ -63,14 +70,27 @@ var _ = Describe("Endpoints", func() {
...
+
 		Context("with normal config", func() {
 			BeforeEach(func() {
 				renderer = NewRenderer(rrConfigNormalMangleReturn)
 				epMarkMapper = NewEndpointMarkMapper(rrConfigNormalMangleReturn.IptablesMarkEndpoint,
 					rrConfigNormalMangleReturn.IptablesMarkNonCaliEndpoint)
 			})
 
-			It("Song should render a minimal workload endpoint", func() {
+			It("should render a minimal workload endpoint", func() {
```
neiljerram: :laughing:

lwr20:
I see you've added UT for the new code, but is this change worthy of an FV?  

Feels like this is the sort of thing we could easily regress without realising it, so an FV test feels like a good idea (to me at least)

dataplane/linux/endpoint_mgr_test.go
rafaelvanoni: Fixed

dataplane/linux/endpoint_mgr_test.go
rafaelvanoni: Fixed

rules/endpoints.go
rafaelvanoni: Fixed per our conversation on slack

dataplane/linux/endpoint_mgr_test.go
rafaelvanoni: Fixed

dataplane/linux/endpoint_mgr_test.go
rafaelvanoni: Fixed

rules/endpoints_test.go
rafaelvanoni: Fixed

PR #2062, opened 2019-06-25
Drop VXLAN and IPIP packets originating in pods (v3.8)

rafaelvanoni:


neiljerram:
See review comments and discussion at #2052

caseydavenport:
```
[Fail] Latency tests with initialized Felix and etcd datastore IPv4: Network sets tests with initialized Felix and etcd datastore [It] with allow-all should have good latency
```

This is a known flake, so we're OK to merge.

PR #2065, opened 2019-06-25
Drop IPIP packets originating in pods (v3.5)

rafaelvanoni:


neiljerram:
See review comments and discussion at #2052

PR #2073, opened 2019-07-08
Refactor iptables nftables workaround to depend on version.

fasaxc:
## Description

- Move detection into the feature detector.
- Only apply workaround for <=1.8.2 (fix should be in 1.8.3).
- Remove split commit 'fix', which I no longer believe to be needed.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/iptables/feature_detect.go
+++ b/iptables/feature_detect.go
@@ -16,10 +16,12 @@ package iptables
 
 import (
 	"io"
+	"os/exec"
 	"regexp"
+	"strings"
 	"sync"
 
-	version "github.com/hashicorp/go-version"
+	"github.com/hashicorp/go-version"
 	log "github.com/sirupsen/logrus"
 
 	"github.com/projectcalico/felix/versionparse"
@@ -36,6 +38,8 @@ var (
 	v1Dot6Dot0 = versionparse.MustParseVersion("1.6.0")
 	// v1Dot6Dot2 added --random-fully to MASQUERADE and the xtables lock to iptables-restore.
 	v1Dot6Dot2 = versionparse.MustParseVersion("1.6.2")
+	// v1Dot8Dot2 and earlier has a bug with -R in nftables mode.
+	v1Dot8Dot2 = versionparse.MustParseVersion("1.8.2")
 
 	// Linux kernel versions:
 	// v3Dot10Dot0 is the oldest version we support at time of writing.
@@ -52,6 +56,9 @@ type Features struct {
 	// RestoreSupportsLock is true if the iptables-restore command supports taking the xtables lock and the
 	// associated -w and -W arguments.
 	RestoreSupportsLock bool
+	// IptablesRestoreHasReplaceBug is true if iptables-restore mishandles the replace ("-R") command.
+	// When in nftables mode, v1.8.2 and before end up replacing the wrong rule.
+	IptablesRestoreHasReplaceBug bool
 }
 
 type FeatureDetector struct {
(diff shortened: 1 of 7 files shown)


iptables/feature_detect.go
```diff
@@ -92,14 +103,15 @@ func (d *FeatureDetector) RefreshFeatures() {
...
 
 	// Calculate the features.
 	features := Features{
-		SNATFullyRandom:     iptV.Compare(v1Dot6Dot0) >= 0 && kerV.Compare(v3Dot14Dot0) >= 0,
-		MASQFullyRandom:     iptV.Compare(v1Dot6Dot2) >= 0 && kerV.Compare(v3Dot14Dot0) >= 0,
-		RestoreSupportsLock: iptV.Compare(v1Dot6Dot2) >= 0,
+		SNATFullyRandom:              iptV.Compare(v1Dot6Dot0) >= 0 && kerV.Compare(v3Dot14Dot0) >= 0,
+		MASQFullyRandom:              iptV.Compare(v1Dot6Dot2) >= 0 && kerV.Compare(v3Dot14Dot0) >= 0,
+		RestoreSupportsLock:          iptV.Compare(v1Dot6Dot2) >= 0,
+		IptablesRestoreHasReplaceBug: iptMode == "nft" && iptV.Compare(v1Dot8Dot2) <= 0,
```
neiljerram: I know it's unrelated to your change, but... this code is quite hard to read, and I wonder if that's necessary.  I'd much rather read this:
```
		SNATFullyRandom:              iptV.AtLeast("1.6.0") && kerV.AtLeast("3.14.0"),
		MASQFullyRandom:              iptV.AtLeast("1.6.2") && kerV.AtLeast("3.14.0"),
		RestoreSupportsLock:          iptV.AtLeast("1.6.2"),
		IptablesRestoreHasReplaceBug: iptMode == "nft" && iptV.UpTo("1.8.2"),
```

iptables/features_detect_test.go
```diff
@@ -102,12 +102,42 @@ func TestFeatureDetection(t *testing.T) {
...
+				IptablesRestoreHasReplaceBug: false,
+			},
+		},
 	} {
 		tst := tst
 		t.Run("iptables version "+tst.iptablesVersion+" kernel "+tst.kernelVersion, func(t *testing.T) {
 			RegisterTestingT(t)
 			dataplane := newMockDataplane("filter", map[string][]string{}, "legacy")
-			featureDetector := NewFeatureDetector()
+			featureDetector := NewFeatureDetector("nft")
```
neiljerram: Hmm, it's not really right, if backend mode is passed as "nft", to have cases above saying "(legacy)", is it?  I understand the test intention, to cover all the cases as easily as possible, but I wonder if Felix should panic if it saw that for real.  Perhaps it would be better to partition the cases and run them with either `NewFeatureDetector()` or `NewFeatureDetector("nft")` as appropriate.  WDYT?

fasaxc: Agree, it's got a bit messy

iptables/feature_detect.go
fasaxc: Agree, do you know of a lib that implements a nice API like that?  I just grabbed something off the shelf.

neiljerram: I was thinking something like this:
```
func (v *Version) AtLeast(minVersion string) bool {
	return v.Compare(versionparse.MustParseVersion(minVersion) >= 0
}
```
I see that the Version type comes from github.com/hashicorp/go-version, but I think you can easily wrap that so as to permit adding another method, can't you?

neiljerram: (Also, apologies for leaving this sitting for so long!)

neiljerram:
@fasaxc Is this PR still of interest?

Issue #1796, opened 2018-04-15
Some Felix etcd config options ignored when not setting DataStoreType in the same source.


matthewdupre:
Datastore type defaults to etcdv3, but the Felix specific (not generic calicoapiconfig) suboptions get ignored silently (at all log levels) when it's not specified.

e.g. a Felix config file, or Felix environment variables setting etcd certificates are ignored, but the logs actually declare them being parsed successfully...

https://github.com/projectcalico/felix/blob/682bfdb4a26f45b7ef7628bbb925ee72edb3bfca/config/config_params.go#L368 is the culprit: it requires the datastore type to come from a Felix-specific source in order to consider other variables from those sources.

This took quite a long time to debug: the logging is not up to scratch given how complex the code is...

neiljerram:
@matthewdupre @Brian-McM in the beginning, Felix was written in Python and there was no cross-component approach to config (such as we now have via libcalico-go). For configuring Felix's data store we supported FELIX_... environment variables and felix.cfg.

Then the common libcalico-go approach was added, which gives data store configuration by environment variables with CALICO_ prefix, or with no prefix at all. Felix calls LoadConfigFromEnvironment to initialize a CalicoAPIConfig object based on those possible variables.

So we wanted the common approach to work, and we also wanted existing setups to keep working if they were using FELIX_... or felix.cfg. That's why the code is as you currently see it.

neiljerram:
The change in #2077 is definitely wrong because the default value of config.DatastoreType is "etcdv3". So, if you configure DATASTORE_TYPE=kubernetes, plus other common variables for KDD, this change will wrongly overwrite that by setting cfg.Spec.DatastoreType to etcdv3.

neiljerram:
I think we definitely shouldn't change the logic here. There isn't an easy win on that front, and we shouldn't make an already tricky situation even trickier.

What *may* be worthwhile, though, is to improve our documentation about this, and/or emit warning logs that help people see and understand what's happening when FELIX_... or felix.cfg settings are ignored.

@Brian-McM WDYT?

Brian-McM:
I'm declining that PR I made, I clearly had some misconceptions about how that configuration worked and looking at it again I see that doesn't fix anything. One way of fixing this issue though would be to not nest the "if" statements for the sub options under the check for the DatastoreType, and just overwrite the etcd options regardless of whether the DatastoreType is etcd or not. That way we can just say the old felix configuration options always overwrite the common ones, without special triggers.

PR #2077, opened 2019-07-09
Use etcd configuration set from file or the env when DatastoreType is…

Brian-McM:
… etcd but not set from a file or the env

etcd suboptions, like setting the EtcdKeyFile, set in the environment or configuration file would be ignore if the configuration file or environment didn't also set the Datastore type, even if the Datastore type was set to etcd by the calico api configuration.

This change fixes this issue https://github.com/projectcalico/felix/issues/1796. Since the datastore defaults to etcd, the etcd configuration set in felix's config shouldn't be ignored if the datastore isn't explicitly set either to etcd.

I've added a unit test to check that the etcd suboptions set by a felix config file (mimicked using the configs UpdateFrom method) are set when the felix configuration doesn't set the datatype, but the datatype is still set to etcd. I've tested that this test fails before this change, and passes after.

As for manual testing I've run felix in a docker container and modified the felix.cfg to add the EtcdEndpoints parameter but not the DataStoreType parameter. Felix fails to run if it's datastore type is etcd and it can't find the etcd endpoint, thus felix failed to start before this fix (as the etcd configuration in the felix.cfg file wasn't respected), and succeeded in starting after the change.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/config/config_params.go
+++ b/config/config_params.go
@@ -402,7 +402,7 @@ func (config *Config) DatastoreConfig() apiconfig.CalicoAPIConfig {
 
 	// Now allow FELIX_XXXYYY variables or XxxYyy config file settings to override that, in the
 	// etcd case.
-	if config.setByConfigFileOrEnvironment("DatastoreType") && config.DatastoreType == "etcdv3" {
+	if config.DatastoreType == "etcdv3" {
 		cfg.Spec.DatastoreType = apiconfig.EtcdV3
 		// Endpoints.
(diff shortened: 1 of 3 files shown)


neiljerram:
@Brian-McM please can you write more about *why* you think this is a good change to make?

Brian-McM:
@neiljerram I've edited the description with a bit on the why and some information on the testing. Please let me know if that's sufficient.

Brian-McM:
Closing this because this doesn't fix anything, it just removes a necessary check to see if the DatastoreType was set to etcd by the old felix configuration

PR #2045, opened 2019-06-17
Workaround goimports removing packr-generated import

neiljerram:
In wavetank (e.g. [1]), when building bin/calico-felix-amd64, we get this:

  ```
  github.com/projectcalico/felix/vendor/github.com/hashicorp/go-version
  github.com/projectcalico/felix/bpf/packrd
  # github.com/projectcalico/felix/bpf/packrd
  bpf/packrd/packed-packr.go:14:7: undefined: packr
  bpf/packrd/packed-packr.go:26:8: undefined: packr
  bpf/packrd/packed-packr.go:27:28: undefined: packr
  bpf/packrd/packed-packr.go:28:30: undefined: packr
  bpf/packrd/packed-packr.go:32:8: undefined: packr
  bpf/packrd/packed-packr.go:33:26: undefined: packr
  github.com/projectcalico/felix/vendor/github.com/vishvananda/netns
  ```

I also get that when running `make bin/calico-felix` or `make deb` on my
own laptop.

In semaphore (e.g. [2]), for some reason we don't get that same problem.

[1] https://wavetank.tigera.io/view/Openstack/job/openstack-felix-build-package/45/console
[2] https://semaphoreci.com/calico/felix-2/branches/master/builds/1662

When it occurs, the problem seems to be that packr2 generates this
import:

	"github.com/gobuffalo/packr/v2"

and then goimports removes that again because it thinks it isn't
needed.  But the import _is_ needed, and removing it causes the build to
fail as above.

goimports is buggy here, and is incorrectly removing a needed import.
This change works around that by naming the import, because goimports
does not remove named imports.

PR #2080, opened 2019-07-16
Update go-build version

caseydavenport:
I think this addresses the discovery in this PR: https://github.com/projectcalico/felix/pull/2045

Namely that we needed to update to a newer version of go-build. 

```
docker run -ti calico/go-build:v0.22 go version                
go version go1.12.6 linux/amd64

docker run -ti calico/go-build:v0.20 go version 
go version go1.11.2 linux/amd64

```

- [ ] Documentation
- [ ] Backport
- [ ] Release note

```release-note
None required
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -147,7 +147,7 @@ PUSH_NONMANIFEST_IMAGES=$(filter-out $(PUSH_MANIFEST_IMAGES),$(PUSH_IMAGES))
 # location of docker credentials to push manifests
 DOCKER_CONFIG ?= $(HOME)/.docker/config.json
 
-GO_BUILD_VER?=v0.20
+GO_BUILD_VER?=v0.22
 # For building, we use the go-build image for the *host* architecture, even if the target is different
 # the one for the host should contain all the necessary cross-compilation tools
 # we do not need to use the arch since go-build:v0.15 now is multi-arch manifest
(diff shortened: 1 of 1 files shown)


neiljerram:
@caseydavenport Out of interest, though, re:
```
docker run -ti calico/go-build:v0.22 go version                
go version go1.12.6 linux/amd64

docker run -ti calico/go-build:v0.20 go version 
go version go1.11.2 linux/amd64
```
I was thinking it's the version of goimports that matters most here - is that right?  Is goimports included in the versioned golang distribution as shown here?

caseydavenport:
@neiljerram yeah, good question. I assumed so because goimports doesn't seem to have a version reporting command AFAICT.

I also can't seem to see where we're pulling it in... it's not in our Dockerfile here: https://github.com/projectcalico/go-build/blob/master/Dockerfile.amd64

And, I don't see it in what appears to be the base image Dockerfile: https://github.com/docker-library/golang/blob/master/1.12/alpine3.9/Dockerfile

So, I guess I'm a little confused...

neiljerram:
@caseydavenport Curious. But if it's working, I guess we don't need to worry.  Presumably one of the other Dockerfile steps pulls it in.  (Guess `gometalinter --install`?)

caseydavenport:
>  But if it's working, I guess we don't need to worry.

Perhaps, though I've realized now that the packr file that was originally the problem no longer exists :D

Issue #1948, opened 2019-02-01
Cannot expose metrics on custom IP (but 0.0.0.0)


oblazek:
When I run calico-node (felix) and enable metrics so that they are scraped by prometheus calico-node automatically binds to 0.0.0.0 and there is no option to change it to a custom IP.

## Expected Behavior
It should be possible to specify something like `PrometheusMetricsHost` in configuration file where we would specify which IP to use for binding. Default value would still be 0.0.0.0.

## Current Behavior
Currently it binds only to 0.0.0.0

## Possible Solution
It should be possible to specify something like `PrometheusMetricsHost` in configuration file where we would specify which IP to use for binding. Default value would still be 0.0.0.0.

## Steps to Reproduce (for bugs)

1. run calico-node on any linux machine
2. check `netstat -ntl` and look for port `9091`

## Your Environment
* `calico/node` image: quay.io/calico/node:v3.3.1
* Which orchestrator are you using: docker
* Orchestrator version (if used): docker-engine 17.05.0~ce-0~ubuntu-xenial

caseydavenport:
Sounds like a good thing to do.

caseydavenport:
This has been fixed and will be in v3.9.

PR #2081, opened 2019-07-18
Configurable prometheus host.  Default value for prometheus host: empty

rene-dekker:
Make prometheus host configurable.

This change was requested in https://github.com/projectcalico/felix/issues/1948

Other related PR's including documentation:
projectcalico/libcalico-go#1106
projectcalico/libcalico-go#1108
projectcalico/calico#2728
projectcalico/typha#284
projectcalico/confd/pull/254

## Release Note
If you wish to override the default settings for the prometheus host/ip, you can do this by modifying your FelixConfiguration.

### DIFF
--- a/config/config_params.go
+++ b/config/config_params.go
@@ -46,7 +46,7 @@ var (
 	StringRegexp             = regexp.MustCompile(`^.*$`)
 	IfaceParamRegexp         = regexp.MustCompile(`^[a-zA-Z0-9:._+-]{1,15}$`)
 	// Hostname  have to be valid ipv4, ipv6 or strings up to 64 characters.
-	PrometheusHostRegexp = regexp.MustCompile(`^[a-zA-Z0-9:._+-]{1,64}$`)
+	HostAddressRegexp = regexp.MustCompile(`^[a-zA-Z0-9:._+-]{1,64}$`)
 )
 
 const (
@@ -191,9 +191,9 @@ type Config struct {
 
 	HealthEnabled                   bool   `config:"bool;false"`
 	HealthPort                      int    `config:"int(0,65535);9099"`
-	HealthHost                      string `config:"string;localhost"`
+	HealthHost                      string `config:"host-address;localhost"`
 	PrometheusMetricsEnabled        bool   `config:"bool;false"`
-	PrometheusMetricsHost           string `config:"prometheus-host;"`
+	PrometheusMetricsHost           string `config:"host-address;"`
 	PrometheusMetricsPort           int    `config:"int(0,65535);9091"`
 	PrometheusGoMetricsEnabled      bool   `config:"bool;true"`
 	PrometheusProcessMetricsEnabled bool   `config:"bool;true"`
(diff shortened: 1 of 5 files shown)


config/config_params.go
```diff
@@ -191,6 +193,7 @@ type Config struct {
 	HealthPort                      int    `config:"int(0,65535);9099"`
 	HealthHost                      string `config:"string;localhost"`
 	PrometheusMetricsEnabled        bool   `config:"bool;false"`
+	PrometheusMetricsHost           string `config:"prometheus-host;"`
```
neiljerram: Why should the validation for PrometheusMetricsHost be different from that for HealthHost?
- If there is a good reason for the different handling, please add a comment in the code about that.
- If the `prometheus-host` handling is no better than what we already do for HealthHost, please revert the prometheus-host changes and instead do the same as we do for HealthHost.
- If the `prometheus-host` handling is clearly better than what we do for other host config parms, please rename to make it non-prometheus-specific, and apply to all similar parameters.

daemon/daemon.go
```diff
@@ -541,7 +544,8 @@ func servePrometheusMetrics(configParams *config.Config) {
 			}
 		}
 		http.Handle("/metrics", promhttp.Handler())
-		err := http.ListenAndServe(fmt.Sprintf(":%v", configParams.PrometheusMetricsPort), nil)
+		err := http.ListenAndServe(fmt.Sprintf("[%v]:%v",
```
neiljerram: This is wrong for IPv4 addresses, isn't it - have you tested this?

Anyway, please use `net.JoinHostPort` instead.

config/config_params.go
rene-dekker: I see that `HealthHost` currently has no validation eventually ends up in the same net/http/server.go, but before that there is a difference in how the data ends up there:
`fmt.Sprintf("%s:%v", host, port)` in libcalico.go
vs
`fmt.Sprintf("[%s]:%v", host, port)` // in typha and felix project daemon.go

My validation makes sure that when you enter the value for prometheushost, you are not including '[]', which makes sure that ipv4, ipv6 and 'localhost' would all work. If I apply it to HealthHost, ipv6 would not work unless I modify libcalico.go accordingly. If I change the validation, I would require more logic to make sure the input value works.
Want me to create a separate PR for HealthHost in libcalico.go?

daemon/daemon.go
rene-dekker: It works for all these: `localhost`, `127.0.0.1`, `::1`.

As for net.JoinHostPort, I'll look into that, thank you.

neiljerram: I'm surprised that `localhost` and `127.0.0.1` work, but thanks for testing that it does.  Even so, I think I'd prefer net.JoinHostPort...

config/config_params.go
neiljerram: So you're saying that the validation for Prometheus is clearly better, right?  In that case I think we should apply it everywhere.

daemon/daemon.go
rene-dekker: I have modified the code accordingly and created PR's to do the same for HealthHost. See also:
https://github.com/projectcalico/typha/pull/285
https://github.com/projectcalico/libcalico-go/pull/1109/

config/config_params.go
rene-dekker: Ok, please see also: 
projectcalico/typha#285
projectcalico/libcalico-go#1109

PR #2085, opened 2019-07-23
Override etcd suboptions independent of the datastore type

Brian-McM:
There was previously an issue with the etcd suboptions not being applied from the felix configuration if the etcd datastore type wasn't set the etcd in the felix configuration as well. This is a problem because the default datastore type is etcd, so users are bothering to explicitly set the datastore type to etcd. Since the etcd suboptions only have an effect if the datastore is etcd we can just overide the suboptions with what's in the felix configuration, which fixes this issue.

## Description

This is a bug fix for issue https://github.com/projectcalico/felix/issues/1796.  This should be merged because 1) it fixes the issue where the etcd suboptions aren't applied even when the default datastore type is etcd and 2) it makes the felix configuration overrides simpler if we can assume that an etcd suboption in the felix configuration will always be applied, as opposed to having a hidden condition.

There are new unit tests testing this functionality, and I've done some manual tests by bringing up an etcd configured cluster in minikube, leaving the datastore type to it's default etcd, and applying some etcd suboptions and seeing if they were overwritten (i.e. setting the etcd address to an invalid one would make the node pod hang)

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/config/config_params.go
+++ b/config/config_params.go
@@ -232,6 +232,8 @@ type Config struct {
 	SidecarAccelerationEnabled bool `config:"bool;false"`
 	XDPEnabled                 bool `config:"bool;false"`
 	GenericXDPEnabled          bool `config:"bool;false"`
+
+	loadClientConfigFromEnvironment func() (*apiconfig.CalicoAPIConfig, error)
 }
 
 type ProtoPort struct {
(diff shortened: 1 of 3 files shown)


config/config_params.go
```diff
@@ -404,27 +404,44 @@ func (config *Config) DatastoreConfig() apiconfig.CalicoAPIConfig {
...
+	// This allows the user to rely the default DatastoreType being etcdv3 and still being able
+	// to configure the other etcdv3 options. As of the time of this code change, the etcd options
+	// have no affect if the DatastoreType is not etcdv3.
+
+	// Datastore type, either etcdv3 or kubernetes
+	if config.setByConfigFileOrEnvironment("DatastoreType") {
+		log.Infof("Overriding DatastoreType from felix config to %s", config.DatastoreType)
+		if config.DatastoreType == string(apiconfig.EtcdV3) {
+			cfg.Spec.DatastoreType = apiconfig.EtcdV3
+		} else if config.DatastoreType == string(apiconfig.Kubernetes) {
```
neiljerram: This would be a change, but I think it probably makes sense.

- The change is: if someone has `FELIX_DATASTORETYPE=kubernetes`, that would previously have been ignored, and now it will take effect.

- The concern about making sense is that there's no way with `FELIX_...` or felix.cfg of specifying the k8s-apiserver connection.

So you'd need a combination of `FELIX_DATASTORETYPE=kubernetes` and other non-FELIX settings like `CALICO_KUBECONFIG=/etc/kubeconfig` or `KUBECONFIG=/etc/kubeconfig` or `K8S_...=...`.

And, instead of `FELIX_DATASTORETYPE=kubernetes`, you could equally specify `DATASTORE_TYPE=kubernetes` or `CALICO_DATASTORE_TYPE=kubernetes`.  So you might question why `FELIX_DATASTORETYPE=kubernetes` needs to be allowed as well?

But I think the point is that someone might already have all the (non-CALICO-prefixed) `K8S....` or `KUBE...` variables set up in their environment, and then they should only need one thing to make Felix use those too, and then it's helpful if that can be any of `FELIX_DATASTORETYPE=kubernetes` or `DATASTORE_TYPE=kubernetes` or `CALICO_DATASTORE_TYPE=kubernetes`.  So, on balance, I think this is a helpful change.  WDYT?

config/config_params_test.go
```diff
@@ -427,6 +431,70 @@ var _ = Describe("DatastoreConfig tests", func() {
 			Expect(c.DatastoreConfig().Spec.K8sDisableNodePoll).To(BeTrue())
 		})
 	})
+	Describe("with the configuration set only from the common calico configuration", func() {
+		BeforeEach(func() {
+			os.Setenv("CALICO_DATASTORE_TYPE", "etcdv3")
+			os.Setenv("CALICO_ETCD_ENDPOINTS", "http://localhost:1234")
+			os.Setenv("CALICO_ETCD_KEY_FILE", testutils.TestDataFile("etcdkeyfile.key"))
+			os.Setenv("CALICO_ETCD_CERT_FILE", testutils.TestDataFile("etcdcertfile.cert"))
+			os.Setenv("CALICO_ETCD_CA_CERT_FILE", testutils.TestDataFile("etcdcacertfile.cert"))
```
neiljerram: I'm wondering why these `os.Setenv` calls work with ginkgo executing tests in an arbitrary order, and possibly in multiple parallel processes.  The concern is that they change global state and don't reset it for other test cases.

daemon/daemon.go
```diff
@@ -113,6 +113,7 @@ const (
 // main config parameters by exiting and allowing itself to be restarted by the init
 // daemon.
 func Run(configFile string) {
+	log.Infof("OSENV: %v", os.Environ())
```
neiljerram: Concerned that we might be leaking sensitive information into the log here.  Do we really need these logs?

daemon/daemon.go
```diff
@@ -186,6 +188,7 @@ configRetry:
 			continue configRetry
 		}
 		// Parse and merge the local config.
+		log.Infof("ENVRAW: %v", envConfig)
```
neiljerram: (here also)

testutils/file.go
```diff
@@ -0,0 +1,12 @@
...
+
+import (
+	"os"
+	"path"
+)
+
+func TestDataFile(name string) string {
+	dir, _ := os.Getwd()
+
+	return path.Join(dir, "testdata", name)
```
neiljerram: Would it make sense to create an empty file here, perhaps in a tmp dir, instead of checking empty files into Git?

config/config_params_test.go
neiljerram: I see the tests are written so as not to be sensitive to this.  All the same, if there's something better we can easily do, let's do that; or at least comment about this.

config/config_params.go
Brian-McM: I definitely understand the concern, but I agree that this is a helpful change. The main reason I wanted this in here though was to keep the felix config overwriting simple. Maybe somebody will break their felix by specifying kubernetes in the felix configuration, but they should know that that will take affect, as opposed to the flip side where somebody specifies the other kubernetes env variables through the calico common configuration and specifies kubernetes in the felix configuration for the datastore (maybe in the scenerio you outlined) and they have no idea why the datastore is still etcd

config/config_params_test.go
Brian-McM: I think we need to use the environment here, if you look at line 401 in config_params.go it loads the calico configuration from the environment, so this test is making sure that it actually loads them from the environment and that the DataStoreConfig returned is filled with these values. I understand the concern, and will add a comment about it, and I'm definitely open to other suggestions about how to test this.

daemon/daemon.go
Brian-McM: ug sorry, that was left over, I'll remove these

testutils/file.go
Brian-McM: We could do that but I don't necessarily think empty files in git are a problem for testing, and makes this slightly more complicated with testing (really stressing the slightly in that part). I, of course, don't have a problem if you'd prefer me creating files in the tmp directory

config/config_params_test.go
neiljerram: General point: this kind of thing comes up a lot in unit tests.  Remember that this is unit testing, and generally, with unit testing, what we're really trying to test is only the code directly in front of us in a given unit (= .go file or package).  When that code has external inputs, like `apiconfig.LoadClientConfigFromEnvironment()`, we can often control those inputs for test purposes by insinuating a shim or mock at that point.

In this specific case, what we could do is:
- add a `getConfigFromEnvironment func() (*apiconfig.CalicoAPIConfig, error)` field in `Config`
- for production, set `getConfigFromEnvironment` to `apiconfig.LoadClientConfigFromEnvironment`
- for unit testing, have a way for the UT code to provide an alternative function implementation, under the UT code's control.

That's the kind of thing that we often do.  The question is just whether it's worth it in this case; I think I'm leaning towards a yes.  WDYT?

testutils/file.go
neiljerram: OK, no worries.  I don't have a strong feeling here, and when that's the case I think it's better to defer to what you've already implemented.

config/config_params_test.go
```diff
@@ -74,8 +75,8 @@ var _ = Describe("FelixConfig vs ConfigParams parity", func() {
 	}
 
 	BeforeEach(func() {
-		fcFields = fieldsByName(v3.FelixConfigurationSpec{})
-		cpFields = fieldsByName(Config{})
+		fcFields = exportedFieldsByName(v3.FelixConfigurationSpec{})
```
Brian-McM: I felt we didn't want to compare any unexported fields, i.e. loadClientConfigFromEnvironment, as the internals of an object shouldn't should matter to the outside (I know the rules on the "internals of an object" are a bit different golang, with respect to packages, but for the config struct I think the rule holds)

neiljerram: Yes, but we already have a solution for that, i.e. `cpFieldsToIgnore`.  I think it would be much simpler just to add loadClientConfigFromEnvironment to that.

config/config_params_test.go
```diff
@@ -427,6 +435,90 @@ var _ = Describe("DatastoreConfig tests", func() {
 			Expect(c.DatastoreConfig().Spec.K8sDisableNodePoll).To(BeTrue())
 		})
 	})
+	// Note that these tests set common calico env variables, as there's no other way to test that the datastore config loads
+	// the configuration from the apiconfig and then overwrites them with the felix config. They shouldn't persist between
+	// tests, but this is worth noting that this is happening and there isn't another way to properly test this functionality
+	// without using the environment variables.
```
neiljerram: This comment isn't correct now.

config/config_params_test.go
Brian-McM: I did see this, but I added that because I couldn't see a reason why we would want to make sure that unexported fields are the same, because every time somebody adds an unexported field we would have to update this test, so this seemed simpler to me in the long run.

config/config_params_test.go
Brian-McM: No it isn't correct, thanks for pointing that out.

config/config_params_test.go
neiljerram: I'm afraid not.  It's unnecessary code churn - so will impact e.g. open source -> private merges.  And it's just _more_ code in config_parms_test.go, so makes it harder whenever someone needs to read and understand that again.  Please revert.

Brian-McM: Reverting

PR #2086, opened 2019-07-24
Fix nft mode failures due to u32 match

caseydavenport:
NFT mode doesn't support the u32 match which we use for VXLAN VNI
matching. This PR removes that additional match, making our rules a bit
more broad. I think this is OK, since we primarily use this to restrict
outgoing VXLAN from pods and we don't expect pods to be sending VXLAN
traffic direclty over the Calico network.

We also use this to allow VXLAN traffic between Calico nodes. Again, it
seems OK to allow VXLAN traffic in general since the VTEP on the other
side will only accept traffic to the right VNI anyway.

- [ ] Documentation
- [ ] Backport
- [ ] Release note

```release-note
None required
```

### DIFF
--- a/rules/endpoints.go
+++ b/rules/endpoints.go
@@ -311,8 +311,7 @@ func (r *DefaultRuleRenderer) endpointIptablesChain(
 	if dropEncap {
 		rules = append(rules, Rule{
 			Match: Match().ProtocolNum(ProtoUDP).
-				DestPorts(uint16(r.Config.VXLANPort)).
-				VXLANVNI(uint32(r.Config.VXLANVNI)),
+				DestPorts(uint16(r.Config.VXLANPort)),
 			Action:  DropAction{},
 			Comment: "Drop VXLAN encapped packets originating in pods",
 		})
--- a/rules/static.go
+++ b/rules/static.go
@@ -225,23 +225,21 @@ func (r *DefaultRuleRenderer) filterInputChain(ipVersion uint8) *Chain {
 	}
 
 	if ipVersion == 4 && r.VXLANEnabled {
-		// VXLAN is enabled, filter incoming VXLAN packets that match our VXLAN port and VNI to ensure they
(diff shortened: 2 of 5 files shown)


rules/endpoints.go
```diff
@@ -311,8 +311,7 @@ func (r *DefaultRuleRenderer) endpointIptablesChain(
 	if dropEncap {
 		rules = append(rules, Rule{
 			Match: Match().ProtocolNum(ProtoUDP).
-				DestPorts(uint16(r.Config.VXLANPort)).
-				VXLANVNI(uint32(r.Config.VXLANVNI)),
+				DestPorts(uint16(r.Config.VXLANPort)),
```
tmjd: So does this mean that pods will not be able to send traffic to any host on this UDP port?
Is that a problem? It seems like that might cause a problem occasionally.
This is on my rough understanding of what is going on here based off its usage at line 66 and the comment above that `Chain for traffic _from_ the endpoint.`

rules/static.go
```diff
@@ -232,16 +232,14 @@ func (r *DefaultRuleRenderer) filterInputChain(ipVersion uint8) *Chain {
 				Match: Match().ProtocolNum(ProtoUDP).
 					DestPorts(uint16(r.Config.VXLANPort)).
 					SourceIPSet(r.IPSetConfigV4.NameForMainIPSet(IPSetIDAllVXLANSourceNets)).
-					DestAddrType(AddrTypeLocal).
-					VXLANVNI(uint32(r.Config.VXLANVNI)), /* relies on protocol and port check */
+					DestAddrType(AddrTypeLocal),
```
tmjd: Should update the comment on line 228

rules/endpoints.go
caseydavenport: Yeah, I think that's right.

I think we're OK with this though - pods shouldn't be sending VXLAN traffic anyway. We default to the standard UDP VXLAN port https://github.com/projectcalico/felix/blob/master/config/config_params.go#L172

PR #2088, opened 2019-07-26
Delete rules using full line instead of line number for non-cali chains

lmm:
## Description

For iptables chains that we don't fully control, there is a possible race condition where deleting a rule by rule number may delete the wrong rule if the dataplane state that Felix observed has changed.

This PR addresses that by:
- Storing the full iptables line for non-Calico chains in the new table field `ruleHashToFullRule`
- Deleting non-Calico chains rules using the full rule. For example, the rule `-A FORWARD -m comment "cali1234" -j DROP` would result in `-D FORWARD -m comment "cali1234" -j DROP` (just flipping the `-A` to `-D`). Rules in Calico chains are deleted as before using the rule number.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
Fix race condition where a non-Calico rule may be deleted.
```

### DIFF
--- a/iptables/table.go
+++ b/iptables/table.go
@@ -994,10 +994,17 @@ func (t *Table) applyUpdates() error {
 		return nil // Delay clearing the set until we've programmed iptables.
 	})
 
+	newChainToFullRules := map[string][]string{}
+	for chain, rules := range t.chainToFullRules {
+		newChainToFullRules[chain] = make([]string, len(rules))
+		copy(newChainToFullRules[chain], rules)
+	}
+
 	// Now calculate iptables updates for our inserted rules, which are used to hook top-level chains.
 	t.dirtyInserts.Iter(func(item interface{}) error {
 		chainName := item.(string)
 		previousHashes := t.chainToDataplaneHashes[chainName]
+		newRules := newChainToFullRules[chainName]
 
 		// Calculate the hashes for our inserted rules.
 		newChainHashes, newRuleHashes := t.expectedHashesForInsertChain(
@@ -1014,10 +1021,22 @@ func (t *Table) applyUpdates() error {
 			if previousHashes[i] != "" {
 				line := t.renderDeleteByValueLine(chainName, i)
 				buf.WriteLine(line)
+
+				// So let's just empty out the rule
+				newRules[i] = ""
 			}
 		}
+		// Go over our slice of "new" rules and remove any rules we emptied out while deleting rules.
+		copyOfNewRules := []string{}
+		for _, rule := range newRules{
+			if rule != "" {
+				copyOfNewRules = append(copyOfNewRules, rule)
+			}
+		}
+		newRules = copyOfNewRules
 
 		rules := t.chainToInsertedRules[chainName]
+		insertRuleLines := make([]string, len(rules))
 		if t.insertMode == "insert" {
 			t.logCxt.Debug("Rendering insert rules.")
 			// Since each insert is pushed onto the top of the chain, do the inserts in
@@ -1027,17 +1046,22 @@ func (t *Table) applyUpdates() error {
 				prefixFrag := t.commentFrag(newRuleHashes[i])
 				line := rules[i].RenderInsert(chainName, prefixFrag, features)
 				buf.WriteLine(line)
+				insertRuleLines[i] = line
 			}
+			newRules = append(insertRuleLines, newRules...)
 		} else {
 			t.logCxt.Debug("Rendering append rules.")
 			for i := 0; i < len(rules); i++ {
 				prefixFrag := t.commentFrag(newRuleHashes[i])
 				line := rules[i].RenderAppend(chainName, prefixFrag, features)
 				buf.WriteLine(line)
+				insertRuleLines[i] = line
 			}
+			newRules = append(newRules, insertRuleLines...)
 		}
 
 		newHashes[chainName] = newChainHashes
+		newChainToFullRules[chainName] = newRules
 
 		return nil // Delay clearing the set until we've programmed iptables.
 	})
--- a/iptables/table.go
+++ b/iptables/table.go
@@ -1035,11 +1035,6 @@ func (t *Table) applyUpdates() error {
 			if previousHashes[i] != "" {
 				line := t.renderDeleteByValueLine(chainName, i)
 				buf.WriteLine(line)
-
-				// Empty out the rule, if it exists.
-				if i < len(newRules) {
-					newRules[i] = ""
-				}
 			}
 		}
 
--- a/iptables/table.go
+++ b/iptables/table.go
@@ -46,6 +46,9 @@ var (
 		"raw":    []string{"PREROUTING", "OUTPUT"},
 	}
 
+	// topLevelChains is the set of top-level chains we insert into.
+	topLevelChains = set.From("INPUT", "OUTPUT", "FORWARD", "PREROUTING", "POSTROUTING")
+
 	// chainCreateRegexp matches iptables-save output lines for chain forward reference lines.
 	// It captures the name of the chain.
 	chainCreateRegexp = regexp.MustCompile(`^:(\S+)`)
@@ -787,8 +790,9 @@ func (t *Table) readHashesAndRulesFrom(r io.ReadCloser) (hashes map[string][]str
 		}
 		hashes[chainName] = append(hashes[chainName], hash)
 
-		// If this is a non-cali chain, store the full rule.
-		if !strings.HasPrefix(chainName, "cali") {
+		// Either we've explicitly inserted a rule into this chain or it's one of the top-level chains that we may have hooked.
+		// Store off the complete contents of the rules so that we can generate precise deletes later if we need to.
+		if _, ok := t.chainToInsertedRules[chainName]; topLevelChains.Contains(chainName) || ok {
 			fullRule := string(line)
 			rules[chainName] = append(rules[chainName], fullRule)
 			hashesToRules[hash] = fullRule
--- a/iptables/table.go
+++ b/iptables/table.go
@@ -208,8 +208,8 @@ type Table struct {
 	// to what we calculate from chainToContents.
 	chainToDataplaneHashes map[string][]string
 
-	// chainsToFullRules contains the full rules, mapped from chain name to slices of rules in that chain.
-	chainsToFullRules map[string][]string
+	// chainToFullRules contains the full rules, mapped from chain name to slices of rules in that chain.
+	chainToFullRules map[string][]string
 
 	// hashCommentPrefix holds the prefix that we prepend to our rule-tracking hashes.
 	hashCommentPrefix string
@@ -368,7 +368,7 @@ func NewTable(
 		chainNameToChain:       map[string]*Chain{},
 		dirtyChains:            set.New(),
 		chainToDataplaneHashes: map[string][]string{},
-		chainsToFullRules:      map[string][]string{},
+		chainToFullRules:       map[string][]string{},
 		logCxt: log.WithFields(log.Fields{
 			"ipVersion": ipVersion,
 			"table":     name,
@@ -612,7 +612,7 @@ func (t *Table) loadDataplaneState() {
 
 	t.logCxt.Debug("Finished loading iptables state")
 	t.chainToDataplaneHashes = dataplaneHashes
-	t.chainsToFullRules = dataplaneRules
+	t.chainToFullRules = dataplaneRules
 	t.inSyncWithDataPlane = true
 }
 
@@ -1175,7 +1175,7 @@ func (t *Table) renderDeleteByIndexLine(chainName string, ruleNum int) string {
 // used for non-Calico chains.
 func (t *Table) renderDeleteByValueLine(chainName string, ruleNum int) string {
 	// For non-cali chains, get the rule by number but delete using the full rule instead of rule number.
-	rules, ok := t.chainsToFullRules[chainName]
+	rules, ok := t.chainToFullRules[chainName]
 	if !ok || ruleNum >= len(rules) {
 		t.logCxt.WithFields(log.Fields{
 			"chainName": chainName,
--- a/iptables/table.go
+++ b/iptables/table.go
@@ -1017,10 +1017,7 @@ func (t *Table) applyUpdates() error {
 
 		// For simplicity, if we've discovered that we're out-of-sync, remove all our
 		// rules from this chain, then re-insert/re-append them below.
-		//
-		// Remove in reverse order so that we don't disturb the rule numbers of rules we're
-		// about to remove.
-		for i := len(previousHashes) - 1; i >= 0; i-- {
+		for i := 0; i < len(previousHashes); i++ {
 			if previousHashes[i] != "" {
 				line := t.deleteRule(chainName, i)
 				buf.WriteLine(line)
--- a/iptables/table.go
+++ b/iptables/table.go
@@ -205,6 +205,12 @@ type Table struct {
 	// to what we calculate from chainToContents.
 	chainToDataplaneHashes map[string][]string
 
+	// chainsToFullRules contains the full rules, mapped from chain name to slices of rules in that chain.
+	chainsToFullRules map[string][]string
+
+	// hashToFullRules contains a mapping of rule hashes to the full rules.
+	hashToFullRules map[string]string
+
 	// hashCommentPrefix holds the prefix that we prepend to our rule-tracking hashes.
 	hashCommentPrefix string
 	// hashCommentRegexp matches the rule-tracking comment, capturing the rule hash.
@@ -362,6 +368,7 @@ func NewTable(
 		chainNameToChain:       map[string]*Chain{},
 		dirtyChains:            set.New(),
 		chainToDataplaneHashes: map[string][]string{},
+		chainsToFullRules:      map[string][]string{},
 		logCxt: log.WithFields(log.Fields{
 			"ipVersion": ipVersion,
 			"table":     name,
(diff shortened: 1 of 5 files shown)


neiljerram:
@lmm Has this been observed in practice? I'm wondering if it might already be covered by the logic we have to retry in case of error?

lmm:
> @lmm Has this been observed in practice? I'm wondering if it might already be covered by the logic we have to retry in case of error?

Yes, we've had a Calico user report this.

iptables/utils_for_test.go
```diff
@@ -330,17 +350,43 @@ func (d *restoreCmd) Run() error {
 			d.Dataplane.ChainMods.Add(chainMod{name: chainName, ruleNum: ruleNum})
 		case "-D", "--delete":
 			chainName = parts[1]
-			Expect(len(parts)).To(Equal(3), "--delete only expects two arguments")
-			ruleNum, err := strconv.Atoi(parts[2]) // 1-indexed position of rule.
-			Expect(err).NotTo(HaveOccurred())
-			ruleIdx := ruleNum - 1 // 0-indexed array index of rule.
+			// Deleting cali chains by index
+			if strings.HasPrefix(chainName, "cali") {
```
fasaxc: This is good enough but it'd be clearer if this code was structured to parser the rule and then assert on it, rather than having special cases.  I.e.
```
if <second arg is numeric> {
    <assert no more args>
    <assert this is a cali rule, with message like "delete-by-index used for non-calico chain, could cause races">
    <handle the deletion>
} else {
    <handle the deletion>
}
```

iptables/table.go
```diff
@@ -1153,8 +1185,25 @@ func (t *Table) commentFrag(hash string) string {
 	return fmt.Sprintf(`-m comment --comment "%s%s"`, t.hashCommentPrefix, hash)
 }
 
-func deleteRule(chainName string, ruleNum int) string {
-	return fmt.Sprintf("-D %s %d", chainName, ruleNum)
+func (t *Table) deleteRule(chainName string, ruleNum int) string {
```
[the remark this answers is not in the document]
lmm: Good call; fixed.

iptables/table.go
```diff
@@ -998,8 +1031,7 @@ func (t *Table) applyUpdates() error {
 		// about to remove.
```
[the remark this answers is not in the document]
lmm: Fixed, thanks

iptables/table.go
```diff
@@ -998,8 +1031,7 @@ func (t *Table) applyUpdates() error {
...
 		for i := len(previousHashes) - 1; i >= 0; i-- {
 			if previousHashes[i] != "" {
-				ruleNum := i + 1
-				line := deleteRule(chainName, ruleNum)
+				line := t.deleteRule(chainName, i)
```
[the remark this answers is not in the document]
lmm: Thanks, I went with the latter.

iptables/table.go
```diff
@@ -775,13 +786,22 @@ func (t *Table) readHashesFrom(r io.ReadCloser) (hashes map[string][]string, err
 			hash = "OLD INSERT RULE"
 		}
 		hashes[chainName] = append(hashes[chainName], hash)
+
+		// If this is a non-cali chain, store the full rule.
```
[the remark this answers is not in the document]
lmm: Thanks, I should have had this from the get-go

iptables/utils_for_test.go
lmm: That's much better; updated.

iptables/table.go
```diff
@@ -205,6 +208,12 @@ type Table struct {
 	// to what we calculate from chainToContents.
 	chainToDataplaneHashes map[string][]string
 
+	// chainToFullRules contains the full rules, mapped from chain name to slices of rules in that chain.
+	chainToFullRules map[string][]string
```
fasaxc: Worth commenting that this is partial information, limited to the chains that we expect to hook.

iptables/table.go
```diff
@@ -631,18 +649,18 @@ func (t *Table) expectedHashesForInsertChain(
 	return
 }
 
-// getHashesFromDataplane loads the current state of our table and parses out the hashes that we
+// getHashesAndRulesFromDataplane loads the current state of our table and parses out the hashes that we
 // add to rules.  It returns a map with an entry for each chain in the table.  Each entry is a slice
 // containing the hashes for the rules in that table.  Rules with no hashes are represented by
 // an empty string.
```
fasaxc: Comment out of date now.

iptables/table.go
```diff
@@ -205,6 +208,12 @@ type Table struct {
 	// to what we calculate from chainToContents.
 	chainToDataplaneHashes map[string][]string
 
+	// chainToFullRules contains the full rules, mapped from chain name to slices of rules in that chain.
+	chainToFullRules map[string][]string
+
+	// topLevelChains is the set of top-level chains we insert into.
+	topLevelChains set.Set
```
fasaxc: Just remembered `tableToKernelChains` above, I think the way we use that makes this redundant since we pre-populate `chainToInsertedRules` with empty slices for all the top-level chains!

iptables/table.go
```diff
@@ -605,6 +622,7 @@ func (t *Table) loadDataplaneState() {
 
 	t.logCxt.Debug("Finished loading iptables state")
 	t.chainToDataplaneHashes = dataplaneHashes
+	t.chainToFullRules = dataplaneRules
```
fasaxc: Think there's a problem just above here now that we're filtering out non-top-level chains.  The logic around line 604 (https://github.com/projectcalico/felix/blob/a11bdf65f2e44d202c1459aba0ac1df03e44df53/iptables/table.go#L604 scans _all_ chains for calico insertions and it may add chains to `t.dirtyInserts` that we haven't loaded the rules for. 

Wonder if we need to take another approach.  Maybe refactor `readHashesAndRulesFrom` so that it:

- always build up a `fullRules` slice (for non-calico chains)
- checks non-calico chains for inserts as it goes
- if it finds an insert, stores the `fullRules` slice in `rules`
- otherwise, throws `fullRules` away.

iptables/table_test.go
```diff
@@ -482,6 +484,59 @@ func describeEmptyDataplaneTests(dataplaneMode string) {
...
+					"-m comment --comment \"cali:42h7Q64_2XDzpwKe\" --jump ACCEPT",
+					"-m comment --comment \"cali:0sUFHicPNNqNyNx8\" --jump DROP",
+				},
+			}))
+		})
+		Describe("then truncating the chain, with the iptables changed before iptables-restore", func() {
+			BeforeEach(func() {
+				dataplane.InsertRandomRuleInForward = true
+				table.SetRuleInsertions("FORWARD", []Rule{
+					{Action: DropAction{}},
```
fasaxc: Suggest you throw a comment onto each rule you're using in these tests.  It'd be easy to have bugs where one of the old rules hung around but the new drop rule you asked for got lost.

iptables/table_test.go
```diff
@@ -739,6 +794,7 @@ func describeDirtyDataplaneTests(appendMode bool, dataplaneMode string) {
 				InsertMode:               insertMode,
 				BackendMode:              dataplaneMode,
 				LookPathOverride:         lookPathNoLegacy,
+				TopLevelChainsOverride:   set.From("INPUT", "OUTPUT", "FORWARD", "PREROUTING", "POSTROUTING", "unexpected-insert"),
```
fasaxc: I don't think it's right to override the top-level chains here; I think the test found the bug; the approach I gave you can't inserts to unexpected chains :-(

iptables/utils_for_test.go
```diff
@@ -312,11 +322,21 @@ func (d *restoreCmd) Run() error {
 			Expect(chains[chainName]).NotTo(BeNil(), "Insert to unknown chain: "+chainName)
 			chains[chainName] = append(chains[chainName], "") // Make room
 			chain := chains[chainName]
-			for i := len(chain) - 1; i > 0; i-- {
-				chain[i] = chain[i-1]
+
+			// If the first arg after the chain name is a line number, then insert by line number.
+			if lineNum, err := strconv.Atoi(parts[2]); err == nil {
+				lineNum = lineNum - 1 // 0-indexed
+				copy(chain[lineNum+1:], chain[lineNum:])
```
fasaxc: `copy` won't extend the slice so you need to append something to it first.

iptables/utils_for_test.go
```diff
@@ -312,11 +322,21 @@ func (d *restoreCmd) Run() error {
 			Expect(chains[chainName]).NotTo(BeNil(), "Insert to unknown chain: "+chainName)
 			chains[chainName] = append(chains[chainName], "") // Make room
 			chain := chains[chainName]
-			for i := len(chain) - 1; i > 0; i-- {
-				chain[i] = chain[i-1]
+
+			// If the first arg after the chain name is a line number, then insert by line number.
+			if lineNum, err := strconv.Atoi(parts[2]); err == nil {
+				lineNum = lineNum - 1 // 0-indexed
```
fasaxc: Think I've used `Num` to signify 1-indexed and `Idx` to signify 0-indexed in a lot of this code; suggest having both variables rather than decrementing this one.

iptables/utils_for_test.go
```diff
@@ -330,20 +350,47 @@ func (d *restoreCmd) Run() error {
...
-			Expect(err).NotTo(HaveOccurred())
-			ruleIdx := ruleNum - 1 // 0-indexed array index of rule.
-			chain := chains[chainName]
-			Expect(len(chain)).To(BeNumerically(">", ruleIdx), "Delete of non-existent rule")
-			for i := ruleIdx; i < len(chain)-1; i++ {
-				chain[i] = chain[i+1]
+
+			// If second arg is numeric, this is a delete by line number.
+			if ruleNum, err := strconv.Atoi(parts[2]); err == nil {
+				Expect(parts).To(HaveLen(3), "calico chain: expected a delete-by-number '--delete '")
```
fasaxc: Message is a little misleading, you haven't checked that it's  calico chain yet.  Maybe just say `"unexpected argument after rule position in --delete"`

iptables/utils_for_test.go
```diff
@@ -330,20 +350,47 @@ func (d *restoreCmd) Run() error {
...
+				Expect(chainName).To(HavePrefix("cali"), "deleting non-calico chain by number can cause races")
```
fasaxc: `s/non-calico chain/rule from non-calico chain/`

iptables/utils_for_test.go
```diff
@@ -330,20 +350,47 @@ func (d *restoreCmd) Run() error {
...
+				chains[chainName] = newChain
+				d.Dataplane.ChainMods.Add(chainMod{name: chainName, ruleNum: i})
+
 			}
-			chains[chainName] = chain[:len(chain)-1]
-			d.Dataplane.ChainMods.Add(chainMod{name: chainName, ruleNum: ruleNum})
 		case "-X", "--delete-chain":
 			chainName = parts[1]
-			Expect(len(parts)).To(Equal(2), "--delete-chain only has one argument")
+			Expect(parts).To(HaveLen(2), "--delete-chain only has one argument")
```
fasaxc: Thanks!

iptables/table.go
```diff
@@ -993,18 +1029,29 @@ func (t *Table) applyUpdates() error {
...
-		for i := len(previousHashes) - 1; i >= 0; i-- {
+		for i := 0; i < len(previousHashes); i++ {
 			if previousHashes[i] != "" {
-				ruleNum := i + 1
-				line := deleteRule(chainName, ruleNum)
+				line := t.renderDeleteByValueLine(chainName, i)
 				buf.WriteLine(line)
+
+				// Empty out the rule, if it exists.
+				if i < len(newRules) {
```
fasaxc: If we get the loading correct, I don't think this check is needed.

iptables/table.go
```diff
@@ -1014,17 +1061,22 @@ func (t *Table) applyUpdates() error {
...
 				prefixFrag := t.commentFrag(newRuleHashes[i])
 				line := rules[i].RenderAppend(chainName, prefixFrag, features)
 				buf.WriteLine(line)
+				insertRuleLines[i] = line
 			}
+			newRules = append(newRules, insertRuleLines...)
 		}
 
 		newHashes[chainName] = newChainHashes
+		newChainToFullRules[chainName] = newRules
```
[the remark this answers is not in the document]
lmm: Thanks; fixed

iptables/table.go
lmm: Fixed

iptables/table.go
lmm: Good catch, thanks

iptables/table.go
lmm: I've fixed that as suggested but we'll be slurping all the KUBE* rules. Do you think it's safe to filter out the known kube-proxy chains?

iptables/table.go
```diff
@@ -795,17 +789,25 @@ func (t *Table) readHashesAndRulesFrom(r io.ReadCloser) (hashes map[string][]str
...
 
-		// Either we've explicitly inserted a rule into this chain or it's one of the top-level chains that we may have hooked.
-		// Store off the complete contents of the rules so that we can generate precise deletes later if we need to.
-		if _, ok := t.chainToInsertedRules[chainName]; t.topLevelChains.Contains(chainName) || ok {
+		// Not our chain so cache the full rule in case we need to generate deletes later on.
+		// After scanning the input, we prune any chains of full rules that do not contain inserts.
+		if !t.ourChainsRegexp.MatchString(chainName) {
 			fullRule := string(line)
 			rules[chainName] = append(rules[chainName], fullRule)
+			nonCalicoChainHasInserts[chainName] = true
```
fasaxc: I think you need to check whether this particular rule is actually an insert or not before setting this flag.  Right now, IIUC, you'll set it for every non-Calico chain.

iptables/table.go
```diff
@@ -795,17 +789,25 @@ func (t *Table) readHashesAndRulesFrom(r io.ReadCloser) (hashes map[string][]str
...
 		}
 	}
 	if scanner.Err() != nil {
 		log.WithError(scanner.Err()).Error("Failed to read hashes from dataplane")
 		return nil, nil, scanner.Err()
 	}
+
+	// Remove full rules for the non-Calico chain if it does not have inserts.
+	for chainName, _ := range rules {
+		if hasInserts, ok := nonCalicoChainHasInserts[chainName]; ok && !hasInserts {
```
fasaxc: Can this conditional ever be true?  I think `ok` will only be true if the loop above put something in the map. However, the code never puts `false` in the map.

One of Go's idioms is to make the "zero value" for a type useful.  To that end, Go's maps return the zero value if you look up an unknown key, rather than panicking.  The zero value for bools is `false` so that means you can implement a set like this without having to check for membership:
```
set := map[string]bool{}
set["a"] = true
set["b"] = true
if set["a"] /* true */ ...
if set["c"] /* false */ ...
```
Note: we also have a `Set` utility struct in libcalico-go:
```
set := set.New()
set.Add("a")
set.Add("b")
if set.Contains("a") /*true*/
if set.Contains("c") /*true*/
```
The `Set` struct is a little more space efficient because it avoids storing a bool for each entry in the set.

iptables/table.go
fasaxc: I think you can piggy-back on the logic just above this, Just add `nonCalicoChainHasInserts[chainName] = true` to both arms of the conditional above (where we get the hash or set hash to "OLD INSERT RULE").  Since the above conditional fires for all chains, I'd rename `nonCalicoChainHasInserts` to `chainHasCalicoRule`, but the logic below will carry on working with that change.

iptables/table.go
fasaxc: Yes, that's a bit frustrating but avoiding it completely would need a big refactoring, I think.  

We could do what we for hashes: store `""` for any non-Calico rules.  At least that'd mean that we wouldn't store a copy of all of kube-proxy's rules.

iptables/utils_for_test.go
```diff
@@ -245,6 +246,15 @@ func (d *restoreCmd) Run() error {
 		log.Warn("Simulating an iptables-restore failure")
 		return errors.New("Simulated failure")
 	}
+	// Alter the dataplane before the restore is applied.
+	if d.Dataplane.InsertRandomRuleInForward {
```
fasaxc: Just spotted that we already had a hook `OnPreRestore` that lets you attach a function to call before the restore happens.  Not sure it's worth switching to that mechanism since this is nice and simple but always worth looking around the code to see if there's already a way to get what you need.

lmm: Nice, I was looking for something like that but missed it.

iptables/utils_for_test.go
```diff
@@ -336,7 +326,8 @@ func (d *restoreCmd) Run() error {
 					chain[i] = chain[i-1]
 				}
 				chain[0] = rest
-				d.Dataplane.ChainMods.Add(chainMod{name: chainName, ruleNum: 1})
+				ruleNum := d.Dataplane.ChainMods.Len() + 1
```
fasaxc: I think this is a bit of a hack; it assumes that the ChainMods set already contains the expected set of data so that we'll get the correct output.  I prefer to avoid that kind of hack in a test since, if the assumption fails, it's usually hard to track down (because you have to know that there's a short-cut being taken in order to figure out why it's giving the wrong result).

iptables/table_test.go
```diff
@@ -535,6 +545,160 @@ func describeEmptyDataplaneTests(dataplaneMode string) {
...
+			Expect(dataplane.Chains).To(Equal(map[string][]string{
+				"FORWARD": {
+					"-m comment --comment \"cali:upaItQyFMdN7MTTl\" -m comment --comment \"a drop rule\" --jump DROP",
+					"-m comment --comment \"cali:EpCg3AYNp_DftVFS\" -m comment --comment \"an accept rule\" --jump ACCEPT",
+				},
+				"INPUT":  {},
+				"OUTPUT": {},
+			}))
+
+			expectedWrites := set.New()
```
fasaxc: I think, to make the chainMods assertion useful, it needs to make sure that there aren't any unexpected writes.  (Right now, you're only verifying tha thte expected writes _were_ done, not that there weren't any others.)

However, you've already covered that base by doing this check: 

```Expect(dataplane.CmdNames).To(ConsistOf("iptables", "iptables-nft-save"))```

There can't be an unexpected write if the code didn't call `iptables-restore` at all!

Since chainMods are a bit hard to pin down for inserts, I say drop the chainMod assertions and rely on the check that iptables-restore didn't get called.

PR #2090, opened 2019-07-29
switch felix to go modules

rafaelvanoni:


fv/run-batches
```diff
@@ -29,8 +29,7 @@ for batch in ${FV_BATCHES_TO_RUN}; do
                -ginkgo.seed 1 \
                -ginkgo.randomizeAllSpecs=true \
                -ginkgo.slowSpecThreshold ${FV_SLOW_SPEC_THRESH} \
-               -ginkgo.focus="${GINKGO_FOCUS}" \
-               ${GINKGO_ARGS}
```
caseydavenport: What is this change for?

proto/felixbackend.pb.go
```diff
@@ -1,15797 +0,0 @@
-// Code generated by protoc-gen-gogo. DO NOT EDIT.
```
caseydavenport: I don't think we should be deleting this file - is there something that go modules does that causes this?

fv/run-batches
rafaelvanoni: ${GINKGO_ARGS} is set to "-mod=vendor" which isn't a valid option for the compiled fv.test

rafaelvanoni: ${GINKGO_ARGS} is set to "-mod=vendor" which isn't a valid option for the compiled fv.test

proto/felixbackend.pb.go
rafaelvanoni: Hmm.. not sure how this happened, will revert it.

Makefile
```diff
@@ -613,24 +593,24 @@ check-packr: bpf/packrd/packed-packr.go
...
 
+SOURCE_DIRS := "./bpf/ ./buildinfo/ ./calc/ ./check-licenses/ ./cmd/ ./config/ ./conntrack/ ./daemon/ ./dataplane/ ./dispatcher/ ./fv/ ./hashutils/ ./ifacemonitor/ ./ip/ ./ipsets/ ./iptables/ ./jitter/ ./k8sfv/ ./labelindex/ ./logutils/ ./markbits/ ./multidict/ ./policysync/ ./proto/ ./routetable/ ./rules/ ./statusrep/ ./stringutils/ ./testutils/ ./throttle/ ./usagerep/ ./versionparse/"
+
 # Run go fmt on all our go files.
 .PHONY: go-fmt goimports fix
 fix go-fmt goimports:
-	$(DOCKER_RUN) $(CALICO_BUILD) sh -c 'glide nv -x | \
-	      grep -v -e "^\\.$$" | \
-	      xargs goimports -w -local github.com/projectcalico/'
+	$(DOCKER_RUN) $(CALICO_BUILD) sh -c 'echo $(SOURCE_DIRS) | xargs goimports -w -local github.com/projectcalico/'
```
spikecurtis: This `SOURCE_DIRS` is fragile and creates a maintenance burden, e.g. that we have to remember to update when we add new packages.  How about

```suggestion
	$(DOCKER_RUN) $(CALICO_BUILD) sh -c 'find . -iname "*.go" ! -wholename "./vendor/*" | xargs goimports -w -local github.com/projectcalico/'
```

Makefile
```diff
@@ -613,24 +593,24 @@ check-packr: bpf/packrd/packed-packr.go
...
 	@echo
 	@echo "Felix's libcalico-go pin:"
-	@grep libcalico-go glide.lock -A 5 | grep 'version:' | head -n 1
+	@grep libcalico-go go.mod | awk '{print $$2}'
 	@echo "Typha's libcalico-go pin:"
-	@grep libcalico-go vendor/github.com/projectcalico/typha/glide.lock -A 5 | grep 'version:' | head -n 1
-	if [ "`grep libcalico-go glide.lock -A 5 | grep 'version:' | head -n 1`" != \
-	     "`grep libcalico-go vendor/github.com/projectcalico/typha/glide.lock -A 5 | grep 'version:' | head -n 1`" ]; then \
+	@grep libcalico-go vendor/github.com/projectcalico/typha/go.mod | awk '{print $$2}'
+	if [ "`grep libcalico-go vendor/github.com/projectcalico/typha/go.mod | awk '{print $$2}'`" != \
```
spikecurtis: You can access the downloaded contents of dependencies using something like

```
cat $(go list -m -f "{{.Dir}}" github.com/projectcalico/typha)/glide.lock
```

This makes it so we don't need `vendor` for this kind of thing

Or, specifically for the go.mod file

```
cat $(go list -m -f "{{.GoMod}}" github.com/projectcalico/typha)
```
```

File: fv/infrastructure/infra_k8s.go [hunk 5]
```diff
@@ -236,7 +236,7 @@ func setupK8sDatastoreInfra() (*K8sDatastoreInfra, error) {
 	log.Info("Started controller manager.")
 
 	// Copy CRD registration manifest into the API server container, and apply it.
-	err := kds.k8sApiContainer.CopyFileIntoContainer("../vendor/github.com/projectcalico/libcalico-go/test/crds.yaml", "/crds.yaml")
+	err := kds.k8sApiContainer.CopyFileIntoContainer("infrastructure/crds.yaml", "/crds.yaml")
```
spikecurtis: Not sure if you want to go down this route, but you can do something like

```
cat $(go list -m -f "{{.Dir}}" github.com/projectcalico/libcalico-go)/test/crds.yaml
```

to get this content

Makefile
```diff
@@ -268,45 +270,36 @@ build-all: $(addprefix sub-build-,$(VALIDARCHES))
...
-          sed -i "s/$$OLD_VER/$(TYPHA_VERSION)/" glide.yaml && \
-          glide up --strip-vendor || glide up --strip-vendor; \
-        fi'
+	$(DOCKER_RUN) $(CALICO_BUILD) sh -c '\
+	  echo "Updating typha to $(TYPHA_VERSION) from $(TYPHA_REPO)"; \
+	  export OLD_VER=$$(grep typha | awk '{print $2}') ;\
+	  echo "Old version: $$OLD_VER";\
+	  if [ $(TYPHA_VERSION) != $$OLD_VER ]; then \
+	  	sed -i "s/$$OLD_VER/$(TYPHA_VERSION)/" go.mod;
+	  fi'
```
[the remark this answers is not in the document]
rafaelvanoni: Fixed

Makefile
rafaelvanoni: Fixed

fv/infrastructure/infra_k8s.go
rafaelvanoni: I spoke to Casey about the same issue on a different repo and we settled on curling these files from github, so I'm adding a rule for this one instead of adding the file.

rafaelvanoni: Ok, moved to `get list` instead

PR #2098, opened 2019-08-06
Work around hpings' buggy RTT calculation.

fasaxc:
## Description

hping does this to get its timestamp:

    secs = time(NULL);
    usecs = get_usecs();

If usecs is just about to wrap when secs is read then it can record a
time almost one second in the past.  This results in adding 1s to RTTs
at random.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/fv/workload/workload.go
+++ b/fv/workload/workload.go
@@ -284,6 +284,7 @@ func (w *Workload) LatencyTo(ip, port string) (time.Duration, string) {
 
 	lines := strings.Split(out, "\n")[1:] // Skip header line
 	var rttSum time.Duration
+	var numBuggyRTTs int
 	for _, line := range lines {
 		if len(line) == 0 {
 			continue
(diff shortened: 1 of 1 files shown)


fv/latency_test.go
```diff
@@ -160,6 +160,12 @@ var _ = Context("Latency tests with initialized Felix and etcd datastore", func(
 		It("with allow-all should have good latency", func() {
 			meanRtt, out := w[0].LatencyTo(w[1].IP, w[1].DefaultPort)
 			_, err := fmt.Fprintf(resultsFile, "allow-all: %v\n", meanRtt)
+
+			if meanRtt > 10 {
+				log.Info("Sleeping.....")
+				time.Sleep(1000000 * time.Second)
+			}
+
```
neiljerram: Presumably this was debugging and should be removed now.

fasaxc: Yes, d'oh.

fasaxc:
Yes, I filed an upstream bug here: https://github.com/antirez/hping/issues/48

Once that's fixed and flows through we can remove this workaround.  I figured that building hping from source from a fork would be a bit overkill for this use case.  I'm expecting latencies to be well under 1s, it's hard to imaging how we could violate that on a single host.  I think hping is generating individual packets so we're not even going to get TCP retries.

PR #2101, opened 2019-08-07
Properly manage route with VXLANMode CrossSubnet

Brian-McM:
Essentially we want routes for pod CIDR ranges from remote nodes in the same subnet to go to the remote nodes gateway using the default interface. Some of the changes include:

1) update the VXLANResolver to allow it to listen for updates with a ResourceKey, and modify the RouteUpdate routes in CrossSubnet pools whos node is in the same subnet as the local node

2) Update the internal dataplane to retrieve the route tables from the managers so that the vxlan manager can pass the "parent" route table, once available, back to the internal dataplane to receive updates

3) update the vxlan_mgr to add non vxlan routes, when found, and add it using the localVTEPS parent interface

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
Add VXLANMode CrossSubnet so routes use the VXLAN interface only when the destination is on a node in a different subnet
```


calc/vxlan_resolver.go
```diff
@@ -159,66 +182,110 @@ func (c *VXLANResolver) OnBlockUpdate(update api.Update) (_ bool) {
...
+	if update.Value != nil {
+		node := update.Value.(*apiv3.Node)
+
+		bgp := node.Spec.BGP
+		if bgp != nil {
+			c.nodeNameToNode[nodeName] = node
+
+			ipv4, _, err := cnet.ParseCIDROrIP(bgp.IPv4Address)
+			if err != nil {
+				logCxt.Error(err.Error())
```
caseydavenport: Let's use the `WithError` functionality here, and add a contextual message.

e.g. `logCxt.WithError(error).Error("Failed to parse IPv4 BGP address")`

calc/vxlan_resolver.go
```diff
@@ -423,25 +490,80 @@ func (c *VXLANResolver) routeReady(r vxlanRoute) bool {
 		logCxt.Debug("Don't yet know the VTEP for this route")
 		return false
 	}
+
+	if _, ok := c.nodeNameToNode[c.hostname]; !ok {
```
caseydavenport: Hm, I think this won't work in the scenario where we're using an old-version of Typha with a new version of Felix, since the old version of Typha will never send through the `v3.Node` object and so this check will never pass.

calc/vxlan_resolver.go
```diff
@@ -423,25 +490,80 @@ func (c *VXLANResolver) routeReady(r vxlanRoute) bool {
...
+	return cidr, nil
+}
+
 // kickPendingRoutes loops through the provided routes to see if there are any which are now programmable,
 // and will send any that are.
 func (c *VXLANResolver) kickPendingRoutes(pendingRouteUpdates set.Set) {
 	pendingRouteUpdates.Iter(func(item interface{}) error {
 		r := item.(vxlanRoute)
-		logCxt := logrus.WithField("route", r)
+		logCxt := logrus.WithField("route", r).WithField("debug", "brian")
```
caseydavenport: Stray "brian" here :)

calc/vxlan_resolver.go
```diff
@@ -423,25 +490,80 @@ func (c *VXLANResolver) routeReady(r vxlanRoute) bool {
...
-		})
+		}
+
+		pool := c.vxlanPoolForRoute(r)
+		if pool.VXLANMode == encap.CrossSubnet {
+			logCxt.Debug("CrossSubnet")
+
+			localNodeCidr, err := c.nodeCidr(c.hostname)
+			if err != nil {
+				logCxt.WithError(err).Errorf("Couldn't retrieve local node cidr")
```
caseydavenport: I think we probably want to stop if we hit this condition. We should probably continue the iteration with the next route since the `IsNetOverlap` below might cause a problem.

calc/vxlan_resolver.go
```diff
@@ -423,25 +490,80 @@ func (c *VXLANResolver) routeReady(r vxlanRoute) bool {
...
+
+		routeUpdate := &proto.RouteUpdate{
 			Type: proto.RouteType_VXLAN,
 			Node: r.node,
 			Dst:  r.dst.String(),
 			Gw:   c.determineGatewayForRoute(r),
-		})
+		}
+
+		pool := c.vxlanPoolForRoute(r)
```
caseydavenport: Logically, it feels like this belongs in `determineGatewayForRoute`.

I think it probably makes sense to augment that function with this logic. This probably also helps with the "readiness" checks above. I think it will automatically get us the right readiness logic for cross subnet routes.

calc/vxlan_resolver.go
```diff
@@ -423,25 +490,80 @@ func (c *VXLANResolver) routeReady(r vxlanRoute) bool {
...
 	return true
 }
 
+func (c *VXLANResolver) nodeCidr(nodeName string) (*cnet.IPNet, error) {
+	logCxt := logrus.WithField("node", nodeName)
+
+	node := c.nodeNameToNode[nodeName]
+	bgp := node.Spec.BGP
+	if bgp == nil {
+		return nil, errors.New("no bgp information for node")
```
caseydavenport: I think to make this log more useful for people skimming the logs, we should say "No IPv4 address for node".

It's kinda confusing that we use a field in the BGP section in order to do VXLAN routes, since BGP isn't involved at all. But, that's where we're at. But the real thing we're looking for is the IPv4 address.

calc/vxlan_resolver.go
```diff
@@ -423,25 +490,80 @@ func (c *VXLANResolver) routeReady(r vxlanRoute) bool {
...
+	}
+
+	_, cidr, err := cnet.ParseCIDROrIP(bgp.IPv4Address)
+	if err != nil {
+		logCxt.Error(err.Error())
```
caseydavenport: Should use "WIthError" here as well.

calc/vxlan_resolver.go
```diff
@@ -75,14 +80,32 @@ func NewVXLANResolver(hostname string, callbacks PipelineCallbacks) *VXLANResolv
...
+	c.useResourceUpdates = b
+
+	return nil
+}
+
 func (c *VXLANResolver) RegisterWith(allUpdDispatcher *dispatcher.Dispatcher) {
-	allUpdDispatcher.Register(model.HostIPKey{}, c.OnHostIPUpdate)
+	c.registeredWithDispatcher = true
+	if c.useResourceUpdates {
+		allUpdDispatcher.Register(model.ResourceKey{}, c.OnResourceUpdate)
```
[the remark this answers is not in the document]
Brian-McM: I'll look into that, I didn't see anything obvious pointing to that at a first glance.

calc/vxlan_resolver.go
```diff
@@ -75,14 +80,32 @@ func NewVXLANResolver(hostname string, callbacks PipelineCallbacks) *VXLANResolv
...
 		nodeNameToIPAddr:          map[string]string{},
+		nodeNameToNode:            map[string]*apiv3.Node{},
 		nodeNameToVXLANMac:        map[string]string{},
 		blockToRoutes:             map[string]set.Set{},
 		vxlanPools:                map[string]model.IPPool{},
+		useResourceUpdates:        false,
 	}
 }
 
+func (c *VXLANResolver) SetUseResourceUpdates(b bool) error {
```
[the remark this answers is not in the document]
Brian-McM: I don't think should be allowed to change once set, but I was hesitant to modify the constructor as this is open source, and hesitant to add a differently named constructor just for this (a shame you can't overload constructors). In the past, I've implemented creation patterns such as builders in this case, but thought this wouldn't be the time to do it, so I went with this.

That's just my reasoning for this, but I'll defer to whatever style you think we should keep for this project. Would you like a modified constructor, new constructor, or something else?

calc/vxlan_resolver.go
```diff
@@ -159,66 +182,110 @@ func (c *VXLANResolver) OnBlockUpdate(update api.Update) (_ bool) {
 	return
 }
 
+func (c *VXLANResolver) OnResourceUpdate(update api.Update) (_ bool) {
+	resourceKey := update.Key.(model.ResourceKey)
+	if resourceKey.Kind != "node" {
```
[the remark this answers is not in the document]
Brian-McM: Will do!

calc/vxlan_resolver.go
```diff
@@ -159,66 +182,110 @@ func (c *VXLANResolver) OnBlockUpdate(update api.Update) (_ bool) {
...
+			if err != nil {
+				logCxt.Error(err.Error())
+				return
+			}
+
+			c.onNodeIPUpdate(nodeName, ipv4.String())
+		}
+	} else {
+		delete(c.nodeNameToNode, nodeName)
+
```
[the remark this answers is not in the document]
Brian-McM: gotcha, I'll try to conform to the style you have laid out. thanks for the comment!

calc/vxlan_resolver.go
Brian-McM: ooops

calc/vxlan_resolver.go
Brian-McM: good point, I'll update it

calc/vxlan_resolver.go
Brian-McM: right, good point. I'll update

calc/vxlan_resolver.go
Brian-McM: yes I should, I'll update

calc/vxlan_resolver.go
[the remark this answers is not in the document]
caseydavenport: right. yeah, we don't make any promises about the compatibility of the constructor here - this isn't meant to be a library for external users.

I say just modify the constructor. Anyone using it (and I doubt there is) should pin their versions anyway, and then fix their code when they pull in the new version.

calc/vxlan_resolver.go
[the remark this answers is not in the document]
caseydavenport: Yeah, I'm guessing we don't and we'd need to implement it.

calc/vxlan_resolver.go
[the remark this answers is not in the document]
Brian-McM: Gotcha, I'll change that and maybe change that in other places I used this pattern as well

calc/vxlan_resolver.go
Brian-McM: ah right, I meant to check what notifications we're accepting before doing this. This is only meant to happen if we have the resource updates flag set

calc/vxlan_resolver.go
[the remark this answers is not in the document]
Brian-McM: It doesn't look like the dispatcher has any logic specific to that.

calc/vxlan_resolver.go
Brian-McM: definitely, a little oversight on my part here

daemon/daemon.go
```diff
@@ -369,6 +369,9 @@ configRetry:
 	} else {
 		// Use the syncer locally.
 		syncer = felixsyncer.New(backendClient, datastoreConfig.Spec, syncerToValidator)
+
+		log.Info("using resource updates where applicable")
+		configParams.SetUseResourceUpdates(true)
```
caseydavenport: I think same here - in general we don't need to use Setters / Getters since configParams isn't a public API.

caseydavenport: Is the plan to do the Typha support for this in a follow-on PR?

dataplane/linux/vxlan_mgr.go
```diff
@@ -175,10 +179,12 @@ func (m *vxlanManager) CompleteDeferredWork() error {
 
 	if m.routesDirty {
 		// Iterate through all of our L3 routes and send them through to the route table.
-		var routes []routetable.Target
+		var vxlanRoutes []routetable.Target
+		var defaultRoutes []routetable.Target
 		for _, r := range m.routesByDest {
 			logCtx := logrus.WithField("route", r)
 			cidr, err := ip.CIDRFromString(r.Dst)
+
```
caseydavenport: Here's another blank line we should get rid of. In general it's nice to keep the handling of the error within a block with the call that generated the error.

dataplane/linux/vxlan_mgr.go
```diff
@@ -194,16 +200,52 @@ func (m *vxlanManager) CompleteDeferredWork() error {
...
+					logrus.WithField("debug", "brian").Errorf("error getting parent interface: %s", err.Error())
+				} else {
+					ifName := link.Attrs().Name
+					logrus.WithField("debug", "brian").Debugf("type: %s", ifName)
+
+					logrus.WithField("debug", "brian").WithField("route", defaultRoutes[0]).Info("in thingy here setting route")
+					m.routeTable.SetRoutes(ifName, defaultRoutes)
+				}
+			}
+
```
caseydavenport: nit: extra blank lines here

dataplane/linux/vxlan_mgr.go
```diff
@@ -194,16 +200,52 @@ func (m *vxlanManager) CompleteDeferredWork() error {
...
+		if len(defaultRoutes) > 0 {
+			localInterface := m.getLocalVTEP()
+			if localInterface != nil {
+				link, err := m.getParentInterface(localInterface)
+				if err != nil {
+					logrus.WithField("debug", "brian").Errorf("error getting parent interface: %s", err.Error())
+				} else {
+					ifName := link.Attrs().Name
+					logrus.WithField("debug", "brian").Debugf("type: %s", ifName)
```
caseydavenport: extra "brian"

dataplane/linux/vxlan_mgr.go
```diff
@@ -194,16 +200,52 @@ func (m *vxlanManager) CompleteDeferredWork() error {
...
 		logrus.Info("VXLAN Manager completed deferred work")
-		m.routeTable.SetRoutes(m.vxlanDevice, routes)
+		if len(vxlanRoutes) > 0 {
+			logrus.WithField("vxlanroutes", vxlanRoutes).Debug("VXLAN manager sending VXLAN L3 updates")
+			m.routeTable.SetRoutes(m.vxlanDevice, vxlanRoutes)
+		}
+
+		if len(defaultRoutes) > 0 {
+			localInterface := m.getLocalVTEP()
+			if localInterface != nil {
```
caseydavenport: Do we need to handle the case where the localInterface is nil?

routetable/route_table.go
```diff
@@ -37,6 +37,13 @@ const (
 	maxConnFailures    = 3
 )
 
+type RTProtocol int
+
+const (
+	RTProtocolUnspec       RTProtocol = 0
+	RTProtocolVXLANManaged RTProtocol = 0x50
```
caseydavenport: I think we'll want a way to configure the value for this, since 0x50 might already be taken on some systems. 

Probably good enough to add a new FelixConfiguration option. That will require some changes to this PR as well as a new one in libcalico to make it configurable through the datastore.

routetable/route_table.go
```diff
@@ -286,7 +314,9 @@ func (r *RouteTable) Apply() error {
 				continue
 			}
 			ifaceName := attrs.Name
-			if r.ifacePrefixRegexp.MatchString(ifaceName) {
+
+			//set the iface to dirty if either the rtProtocolFilter is specified OR the iface is managed by this RouteTable
```
caseydavenport: Don't we only want to do this if the protocol filter is specified _and_ matches the filter? Or am I misunderstanding?


Also nit: I'd add a space after the comment and use sentence case.

e.g. `// Set the iface` rather than `//set the iface`

routetable/route_table.go
```diff
@@ -552,11 +613,16 @@ func (r *RouteTable) syncRoutesForLink(ifaceName string) error {
...
+				logCxt.Infof("setting route protocol to %d", target.Proto)
+				route.Protocol = int(target.Proto)
+			}
+
 			if target.GW != nil {
 				route.Gw = target.GW.AsNetIP()
 			}
 
-			if target.Type == TargetTypeVXLAN {
+			if target.Type == TargetTypeVXLAN || target.Type == TargetTypeDefault {
```
caseydavenport: Hm, I'm not sure we need to set these for non-VXLAN routes. Have you tried it?

daemon/daemon.go
Brian-McM: Yes, typha support will have to come after this

routetable/route_table.go
Brian-McM: I did and it wasn't working, but it may have been something else causing the problem. I'll try again.

routetable/route_table.go
Brian-McM: OK I'll add that

dataplane/linux/vxlan_mgr.go
Brian-McM: Yes, at least log something

calc/vxlan_resolver.go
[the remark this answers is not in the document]
caseydavenport: Probably fine for now since we only expect a single type of Resource. When we add a second, we'll need to add filtering.

dataplane/linux/vxlan_mgr.go
```diff
@@ -194,16 +205,49 @@ func (m *vxlanManager) CompleteDeferredWork() error {
...
+
+				vxlanRoutes = append(vxlanRoutes, vxlanRoute)
+				logCtx.WithField("route", vxlanRoute).Debug("adding vxlan route to list for addition")
+			}
 		}
 
-		logrus.WithField("routes", routes).Debug("VXLAN manager sending L3 updates")
 		logrus.Info("VXLAN Manager completed deferred work")
-		m.routeTable.SetRoutes(m.vxlanDevice, routes)
+		if len(vxlanRoutes) > 0 {
```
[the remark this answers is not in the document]
Brian-McM: I'm not sure what we're setting though if there's not routes? I think this caused an error if I didn't check this, pass an empty list, but I'll think about this some more

daemon/daemon.go
Brian-McM: I realised I didn't comment on your first comment, but this is a bit different because we don't know if we want to use the resource updates until this point

caseydavenport: @Brian-McM have you started a PR for the Typha changes?

dataplane/linux/vxlan_mgr.go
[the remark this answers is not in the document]
caseydavenport: Yeah, I think we need to set the routes even if they are length 0.

daemon/daemon.go
Brian-McM: I haven't, I've made a task for it and Raphael has volunteered to help with it

dataplane/linux/int_dataplane.go
```diff
@@ -258,7 +260,7 @@ func NewIntDataplaneDriver(config Config) *InternalDataplane {
 		applyThrottle:     throttle.New(10),
 	}
 	dp.applyThrottle.Refill() // Allow the first apply() immediately.
-
+	dp.hostname = config.Hostname
```
caseydavenport: Do we need this change? I can't seem to see where we use this, but could be missing it.

dataplane/linux/int_dataplane.go
```diff
@@ -339,12 +341,10 @@ func NewIntDataplaneDriver(config Config) *InternalDataplane {
...
 	dp.ipSets = append(dp.ipSets, ipSetsV4)
 
-	routeTableV4 := routetable.New(config.RulesConfig.WorkloadIfacePrefixes, 4, false, config.NetlinkTimeout, config.DeviceRouteSourceAddress)
-	dp.routeTables = append(dp.routeTables, routeTableV4)
-
 	if config.RulesConfig.VXLANEnabled {
-		routeTableVXLAN := routetable.New([]string{"vxlan.calico"}, 4, true, config.NetlinkTimeout, config.DeviceRouteSourceAddress)
-		dp.routeTables = append(dp.routeTables, routeTableVXLAN)
+		routeTableVXLAN := routetable.New([]string{"vxlan.calico"}, 4, true, config.NetlinkTimeout, config.DeviceRouteSourceAddress,
+			config.DeviceRouteProtocol, config.RemoveExternalRoutes)
```
caseydavenport: I don't think we want to drive this based on the config parameters - we _never_ want to remove external routes on eth0, for example.

Brian-McM: Yes, I'll update this.

dataplane/linux/int_dataplane.go
Brian-McM: No we don't, I'll remove it

routetable/route_table.go
Brian-McM: I had to keep this in because it would break with it, the routes couldn't be added

dataplane/linux/int_dataplane.go
```diff
@@ -341,23 +338,17 @@ func NewIntDataplaneDriver(config Config) *InternalDataplane {
...
 		vxlanManager := newVXLANManager(
 			ipSetsV4,
-			config.MaxIPSetSize,
-			config.Hostname,
 			routeTableVXLAN,
 			"vxlan.calico",
-			config.RulesConfig.VXLANVNI,
-			config.RulesConfig.VXLANPort,
-			config.ExternalNodesCidrs,
+			config,
```
caseydavenport: Was switching to just pass the config a cleanup? Or is there another reason?

It's nice to be explicit in the arguments here rather than passing the entire config struct. While it _does_ mean we have a longer constructor, it also means we know exactly what the VXLAN manager requires in order to operate. It's a fairly minor point though - WDYT?

dataplane/linux/int_dataplane.go
```diff
@@ -571,7 +564,26 @@ type Manager interface {
...
+		rts = append(rts, mrts.GetRouteTables()...)
+	}
+
+	return rts
+}
+
 func (d *InternalDataplane) RegisterManager(mgr Manager) {
+	switch mgr.(type) {
+	case ManagerWithRouteTables:
+		log.WithField("manager", mgr).Debug("found ManagerWithRouteTables")
```
caseydavenport: nit: I'd say "registering ManagerWithRouteTables" rather than "found"

dataplane/linux/vxlan_mgr.go
```diff
@@ -59,47 +74,83 @@ type vxlanManager struct {
...
+	nlHandle, _ := netlink.NewHandle()
+
+	return newVXLANManagerWithShims(
+		ipsetsDataplane,
+		rt,
+		deviceName,
+		dpConfig,
+		nlHandle,
+		func(interfacePrefixes []string, ipVersion uint8, vxlan bool, netlinkTimeout time.Duration,
+			deviceRouteSourceAddress net.IP, deviceRouteProtocol int, removeExternalRoutes bool) routeTable {
```
caseydavenport: Do we need the extra layer of indirection here by wrapping in another function? Couldn't we just do this?

```
parentRTConstruct: routetable.New,
```

dataplane/linux/int_dataplane.go
caseydavenport: Yeah, I see that you pass it down pretty far. I guess this is probably the right change to make.

dataplane/linux/vxlan_mgr.go
```diff
@@ -137,6 +188,20 @@ func (m *vxlanManager) getLocalVTEP() *proto.VXLANTunnelEndpointUpdate {
...
 }
 
+func (m *vxlanManager) getLocalVTEPParent() (netlink.Link, error) {
+	return m.getParentInterface(m.getLocalVTEP())
+}
+
+func (m *vxlanManager) GetRouteTables() []routeTable {
+	rts := []routeTable{m.routeTable}
+
+	if m.parentRouteTable != nil {
```
caseydavenport: I suggest adding a comment here explaining why we have this nil check - it's somewhat abnormal so worth explaining for future readers.

dataplane/linux/int_dataplane.go
Brian-McM: Since the parent route table is now created in the vxlan manager we'd needed more information from the config, so I just added the config instead of adding more parameters, then realised there was no point in passing both the config and the explicit parameters from the config, so I just removed the other ones.

I think it's OK sometimes to have longer constructors, but as a preference I don't like them, I found it's more difficult to code with them, and I prefer hiding bigger constructors behind builder  patterns or allow config parameters. This said, I will defer to whatever has been common / agreed upon in the core team as this is more coding style specific.

dataplane/linux/vxlan_mgr.go
```diff
@@ -175,7 +240,8 @@ func (m *vxlanManager) CompleteDeferredWork() error {
 
 	if m.routesDirty {
 		// Iterate through all of our L3 routes and send them through to the route table.
-		var routes []routetable.Target
+		var vxlanRoutes []routetable.Target
+		var parentRoutes []routetable.Target
```
caseydavenport: nit on naming - rather than calling these "parent routes" I'd suggest something like "noEncapRoutes" - I think that will be more clear to readers of this code and more accurately describes what the routes are.

Similarly for other places you're using "parent" right now - e.g., `parentRouteTable` is probably better called `noEncapRouteTable`

caseydavenport: Or even better you could call them `defaultRoutes` since that aligns with the route type.

dataplane/linux/vxlan_mgr.go
Brian-McM: unfortunately not, and the reasoning being is because golang sucks with interfacing. The constructor returns a routetable.RouteTable instance, which technically does not much the func definition of returning a routeTable instance. Can't wait for go generics

dataplane/linux/vxlan_mgr.go
Brian-McM: OK will do.

dataplane/linux/vxlan_mgr.go
Brian-McM: I'll update that.

dataplane/linux/vxlan_mgr.go
```diff
@@ -194,16 +260,44 @@ func (m *vxlanManager) CompleteDeferredWork() error {
...
+			}
+		}
+
+		logrus.WithField("vxlanroutes", vxlanRoutes).Debug("VXLAN manager sending VXLAN L3 updates")
+		m.routeTable.SetRoutes(m.vxlanDevice, vxlanRoutes)
+
+		if m.parentRouteTable != nil {
+			if parentDevice, err := m.getLocalVTEPParent(); err == nil {
+				ifName := parentDevice.Attrs().Name
+				log.WithField("link", parentDevice).WithField("routes", parentRoutes).Debug("setting default routes")
```
caseydavenport: I think it would be nice to make this log symmetric with the "VXLAN manager sending VXLAN L3 updates" one above.

e.g. "VXLAN manager sending unencapsulated L3 updates"

dataplane/linux/vxlan_mgr.go
```diff
@@ -212,7 +306,7 @@ func (m *vxlanManager) CompleteDeferredWork() error {
 
 // KeepVXLANDeviceInSync is a goroutine that configures the VXLAN tunnel device, then periodically
 // checks that it is still correctly configured.
-func (m *vxlanManager) KeepVXLANDeviceInSync(mtu int) {
+func (m *vxlanManager) KeepVXLANDeviceInSync(mtu, wait int) {
```
caseydavenport: I'd recommend passing a `time.Duration` here - it's a more obvious API for callers. Then they don't need to look into this function to see how the `wait` is interpreted (is it seconds? minutes? something else?)

dataplane/linux/vxlan_mgr.go
```diff
@@ -221,25 +315,38 @@ func (m *vxlanManager) KeepVXLANDeviceInSync(mtu int) {
 			time.Sleep(1 * time.Second)
 			continue
 		}
+
+		if parent, err := m.getLocalVTEPParent(); err != nil {
+			logrus.WithError(err).Warn("Failed configure VXLAN tunnel device, retrying...")
+			time.Sleep(1 * time.Second)
+			continue
+		} else {
+			if m.parentRouteTable == nil {
```
caseydavenport: I think we're working with this parentRouteTable field across multiple goroutines which means we should probably add some mutexes for safety.

An easy pattern for this looks like:

```
func getRouteTable() {
  m.Lock()
  defer m.Unlock()
  return m.parentRouteTable
}
```

Brian-McM: right, I'll update that

dataplane/linux/vxlan_mgr_test.go
```diff
@@ -0,0 +1,203 @@
...
+
+	"github.com/projectcalico/felix/routetable"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"github.com/projectcalico/felix/proto"
+	"github.com/vishvananda/netlink"
+)
+
+type mockVXLANDataplan struct {
```
caseydavenport: Missing an `e` at the end here

dataplane/linux/vxlan_mgr_test.go
```diff
@@ -0,0 +1,203 @@
...
+	links []netlink.Link
+}
+
+func (mockVXLAND *mockVXLANDataplan) LinkByName(name string) (netlink.Link, error) {
```
caseydavenport: nit: generally it's common practice to use shorter names for structures in bound functions like this.

So, it might be more idiomatic to do `(m *mockVXLANDataplane)`

dataplane/linux/vxlan_mgr.go
```diff
@@ -34,12 +35,26 @@ import (
 	"github.com/projectcalico/felix/routetable"
 )
 
+// added so that we can shim netlink for tests
+type vxlanDataplane interface {
```
caseydavenport: None of these seem like VXLAN specific functions - I might call this interface something to do with netlink, since that is what it seems to be wrapping. e.g., `type netlinkHandle interface`

dataplane/linux/vxlan_mgr.go
```diff
@@ -59,47 +74,83 @@ type vxlanManager struct {
...
+		func(interfacePrefixes []string, ipVersion uint8, vxlan bool, netlinkTimeout time.Duration,
+			deviceRouteSourceAddress net.IP, deviceRouteProtocol int, removeExternalRoutes bool) routeTable {
+			return routetable.New(interfacePrefixes, ipVersion, vxlan, netlinkTimeout,
+				deviceRouteSourceAddress, deviceRouteProtocol, removeExternalRoutes)
+		},
+	)
+}
+
+func newVXLANManagerWithShims(
+	ipsetsDataplane ipsetsDataplane,
```
caseydavenport: So, do we use this `ipsetsDataplane` anywhere? Is this just a remnant from when you were looking at using ipsets here?

dataplane/linux/vxlan_mgr_test.go
```diff
@@ -0,0 +1,203 @@
...
+			ParentDeviceIp: "172.0.12.1",
+		})
+
+		localVTEP := manager.getLocalVTEP()
+		Expect(localVTEP).NotTo(BeNil())
+
+		manager.parentRouteTable = prt
+
+		err := manager.configureVXLANDevice(50, localVTEP)
+		Expect(err).To(BeNil())
```
caseydavenport: There's a slightly nicer `Expect(err).NotTo(HaveOccurred())` you can use here.

dataplane/linux/vxlan_mgr_test.go
```diff
@@ -0,0 +1,203 @@
...
+		Expect(manager.parentRouteTable).NotTo(BeNil())
+		parent, err := manager.getLocalVTEPParent()
+
+		Expect(parent).NotTo(BeNil())
+		Expect(err).To(BeNil())
+
+		manager.OnUpdate(&proto.RouteUpdate{
+			Type: proto.RouteType_DEFAULT,
+			Node: "node2",
+			Dst:  "172.0.0.1/32",
```
caseydavenport: Could you make the dest something other than a /32? Or just add a second route that is a larger network? That is more realistically what our IPAM will generate.

dataplane/linux/vxlan_mgr_test.go
```diff
@@ -0,0 +1,203 @@
...
+			Type: proto.RouteType_VXLAN,
+			Node: "node2",
+			Dst:  "172.0.0.2/32",
+			Gw:   "172.8.8.8/32",
+		})
+
+		err = manager.CompleteDeferredWork()
+
+		Expect(err).To(BeNil())
+		Expect(rt.currentRoutes["vxlan.calico"]).NotTo(BeNil())
```
caseydavenport: Might be worth asserting that these are in fact `nil` before calling the deferred work - then you can be confident it was the deferred work execution that caused this change.

dataplane/linux/vxlan_mgr_test.go
```diff
@@ -0,0 +1,203 @@
...
+		Expect(manager.routesDirty).To(BeTrue())
+
+		manager.OnUpdate(&proto.VXLANTunnelEndpointUpdate{
+			Node:           "node1",
+			Mac:            "00:0a:74:9d:68:16",
+			Ipv4Addr:       "10.0.0.0",
+			ParentDeviceIp: "172.0.0.2",
+		})
+
+		time.Sleep(2 * time.Second)
```
caseydavenport: Is this sleep actually required?

fv/infrastructure/infra_etcd.go
```diff
@@ -97,7 +97,7 @@ func (eds *EtcdDatastoreInfra) AddNode(felix *Felix, idx int, needBGP bool) {
 	felixNode.Spec.IPv4VXLANTunnelAddr = felix.ExpectedVXLANTunnelAddr
 	if needBGP {
 		felixNode.Spec.BGP = &api.NodeBGPSpec{
-			IPv4Address:        felix.IP,
+			IPv4Address:        felix.IP + "/24",
```
caseydavenport: I'm guessing this is to trick the FVs into thinking its in the same subnet? 

Do we know what the actual subnet is? It would be much better if we could just use that rather than pretending it is a /24 when it might not actually be.

dataplane/linux/vxlan_mgr.go
Brian-McM: I just did this because there's another one for the IPIP stuff. I'll rename this though

fv/vxlan_test.go
```diff
@@ -25,351 +25,356 @@ import (
...
 	"github.com/projectcalico/libcalico-go/lib/apiconfig"
 	api "github.com/projectcalico/libcalico-go/lib/apis/v3"
 	client "github.com/projectcalico/libcalico-go/lib/clientv3"
+	"github.com/projectcalico/libcalico-go/lib/ipam"
+	"github.com/projectcalico/libcalico-go/lib/net"
 	"github.com/projectcalico/libcalico-go/lib/options"
 )
 
 var _ = infrastructure.DatastoreDescribe("VXLAN topology before adding host IPs to IP sets", []apiconfig.DatastoreType{apiconfig.EtcdV3, apiconfig.Kubernetes}, func(getInfra infrastructure.InfraFactory) {
+	for _, vxlanM := range []api.VXLANMode{api.VXLANModeAlways, api.VXLANModeCrossSubnet} {
```
caseydavenport: This is pretty hard to review - did you change any tests within this? Or just run them for both modes now?

dataplane/linux/vxlan_mgr.go
Brian-McM: line 252, m.ipsetsDataplane.AddOrReplaceIPSet(m.ipSetMetadata, allowedVXLANSources)

routetable/route_table.go
```diff
@@ -65,7 +65,8 @@ func init() {
 type TargetType string
 
 const (
-	TargetTypeVXLAN TargetType = "vxlan"
+	TargetTypeVXLAN       TargetType = "vxlan"
+	TargetTypeVXLANParent TargetType = "default"
```
caseydavenport: As I suggested earlier I think we should align the names here better so the variable name matches the value. It will be easier to follow that way. Probably remove VXLAN and Parent and go with something generic.

e.g. `TargetTypeDefault = "default"`

dataplane/linux/vxlan_mgr_test.go
Brian-McM: Yes, because the KeepInSync runs on 1 second intervals and that's what updates the noEncap table

fv/infrastructure/infra_etcd.go
Brian-McM: I'm not sure, I'm still a little fuzzy on how some of this stuff works for the fv's

fv/vxlan_test.go
Brian-McM: No changes, they're just run in both modes now

dataplane/linux/int_dataplane.go
caseydavenport: > but as a preference I don't like them, I found it's more difficult to code with them, and I prefer hiding bigger constructors behind builder patterns or allow config parameters

Yep, I agree completely. My though was more around passing a more targeted subset of the config. e.g, by defining a new `VXLANManagerConfig` struct or similar. 

But, I'm happy with the current approach after reading a bit further on.

fv/infrastructure/infra_etcd.go
caseydavenport: Looks like the IP address is set from here:
- https://github.com/projectcalico/felix/blob/9891c8b588f38e19634f75e0dbb1174a6968a74e///fv/containers/containers.go#L193
- https://github.com/projectcalico/felix/blob/9891c8b588f38e19634f75e0dbb1174a6968a74e///fv/containers/containers.go#L297:21

We could add a new field, e.g. `IPPrefixLen` and add it to the `Container` Struct and then populate it from `docker inspect` output for the container (which has a field with the same name). 

Looking at Docker inspect output should show you what that the network looks like for your containers.

e.g., here's the network section for a container I'm running locally:

```
           "Networks": {
                "operator-test-cluster": {
                    "IPAMConfig": null,
                    "Links": null,
                    "Aliases": [
                        "k3d-operator-test-cluster-worker-0",
                        "31e922449d40"
                    ],
                    "NetworkID": "43f081b512f3e9ee012b2bba17384e983085665ff57629ced783861ebd00e801",
                    "EndpointID": "cdcb6bda08058d84cfe82483d0a74ab0d977812dcce29edafcb23c401116aff9",
                    "Gateway": "172.29.0.1",
                    "IPAddress": "172.29.0.3",
                    "IPPrefixLen": 16,
                    "IPv6Gateway": "",
                    "GlobalIPv6Address": "",
                    "GlobalIPv6PrefixLen": 0,
                    "MacAddress": "02:42:ac:1d:00:03",
                    "DriverOpts": null
                }
            }
```

PR #2107, opened 2019-08-12
Have node pass buildinfo values as arguments rather than compiled in through LDFLAGS

rafaelvanoni:


go.mod
```diff
@@ -57,6 +57,7 @@ require (
 	github.com/projectcalico/go-yaml v0.0.0-20161201183616-955bc3e451ef // indirect
 	github.com/projectcalico/go-yaml-wrapper v0.0.0-20161127220527-598e54215bee // indirect
 	github.com/projectcalico/libcalico-go v0.0.0-20190802144109-bf00054ed881
+	github.com/projectcalico/node v3.8.1+incompatible // indirect
```
spikecurtis: why did we pick up a dependency on node?

daemon/daemon.go
```diff
@@ -131,11 +130,17 @@ func Run(configFile string) {
...
 	}
 
-	buildInfoLogCxt := log.WithFields(log.Fields{
+	if len(buildinfo.GitVersion) == 0 && len(gitVersion) != 0 {
+		buildinfo.GitVersion = gitVersion
+		buildinfo.BuildDate = buildDate
+		buildinfo.GitRevision = gitRevision
+	}
+
+	buildInfoLogCxt := log.WithFields(log.fields{
```
spikecurtis:
```suggestion
	buildInfoLogCxt := log.WithFields(log.Fields{
```

go.mod
rafaelvanoni: Might have been leftover from an earlier change I made, I removed these and go didn't re-add them.

go.sum
```diff
@@ -86,6 +86,7 @@ github.com/jtolds/gls v4.20.0+incompatible h1:xdiiI2gbIgH/gLH7ADydsJ1uDOEzR8yvV7
 github.com/jtolds/gls v4.20.0+incompatible/go.mod h1:QJZ7F/aHp+rZTRtaJ1ow/lLfFfVYBRgL+9YlvaHOwJU=
 github.com/kardianos/osext v0.0.0-20190222173326-2bc1f35cddc0 h1:iQTw/8FWTuc7uiaSepXwyf3o52HaUYcV+Tu66S3F5GA=
 github.com/kardianos/osext v0.0.0-20190222173326-2bc1f35cddc0/go.mod h1:1NbS8ALrpOvjt0rHPNLyCIeMtbizbir8U//inJ+zuB8=
+github.com/kelseyhightower/confd v0.16.0 h1:cZTibejsLOJ+FsyXQNmwbnq+arQJooKh1NH/+VqN40o=
```
caseydavenport: Do we know where this dependency on confd came from?

spikecurtis: Packages that appear in `go.sum` aren't dependencies, per se. It's just a list of cryptographic checksums.  Presumably this showed up because Rafael had confd as a dependency, and then removed it.

rafaelvanoni: I'm fairly sure this was goimports, let me put up a PR to revert this.

PR #2118, opened 2019-08-20
Only remove interface routes that were previously added by Felix

Marlinc:
## Description

The Felix agent is currently removing routes added by others to the
host side interface. This can be problematic for external
applications that want to also other things into pods/containers.

This change specifies a specific route protocol that will be added
to routes inserted by Felix. When removing routes it will only
remove routes that have that protocol.

Config changes in projectcalico/libcalico-go#1119

## Todos
- [x] Documentation: projectcalico#2800
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```


Brian-McM:
@Marlinc I've added some similar functionality in this PR to for the VXLAN CrossSubnet option https://github.com/projectcalico/felix/pull/2101, so maybe we should coordinate on this.

I had to modify the route table to allow for a protocol to be specified so that I could manage routes in the default interface with the vxlan manager. One problem that may occur here is with the L2 route sync. If an external application adds to the ARP table on the calico managed VETH interface, the L2 sync will remove those entries.

Another problem that we both have is for upgrading felix to this new route protocol "filtering", did you have any thoughts on how the upgrade would work?

Brian-McM:
I think you can ignore my previous comments, the L2 Route managing isn't done for VETH management, and I see you're checking the protocols in the interface and adding a an error to routeProblems if there's a mismatch.

Marlinc:
They thjere @Brian-McM,

Yes, I'm adding a routeproblem so that it will be deleted and re-added with the correct protocol. Just like was implemented for `deviceRouteSourceAddress`.

In my case I'm not making any ARP changes to the VETH interface on the host side (or pod for that mather). I'm only adding a route 'L2' route indicating that a particular IPv6 is available at the VETH interface. That way the traffic for that IP will start flowing to the pods side of the VETH interface.

In terms of upgrades, the defaults in this changeset should be completely compatible with the old behaviour as there is no change. It just adds some extra options that can be used to change behaviour. In case of changes to the protocol set by Felix it will mark the route as problematic and delete/recreate as described above.

caseydavenport:
@Marlinc thanks so much for the PR!

Would you mind raising a PR to document the two new configuration options? Docs are maintained here: https://github.com/projectcalico/calico/tree/master/master

Marlinc:
Hey @caseydavenport this one already exists: projectcalico/calico#2800

caseydavenport:
Awesome, thanks!

PR #2119, opened 2019-08-20
Change libcalico-go volume mount to be /go/src/github.com/projectcali…

Brian-McM:
…co/libcalico-go instead of github.com/projectcalico/libcalico-go

This needs to be updated since the move to go modules

## Release Note

```release-note
None required
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -212,7 +212,7 @@ EXTRA_DOCKER_ARGS	+= -e GO111MODULE=on
 
 # Allow libcalico-go and the ssh auth sock to be mapped into the build container.
 ifdef LIBCALICOGO_PATH
-  EXTRA_DOCKER_ARGS += -v $(LIBCALICOGO_PATH):/github.com/projectcalico/libcalico-go:ro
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -212,7 +212,7 @@ EXTRA_DOCKER_ARGS	+= -e GO111MODULE=on
 
 # Allow libcalico-go and the ssh auth sock to be mapped into the build container.
 ifdef LIBCALICOGO_PATH
-  EXTRA_DOCKER_ARGS += -v $(LIBCALICOGO_PATH):/github.com/projectcalico/libcalico-go:ro
+  EXTRA_DOCKER_ARGS += -v $(LIBCALICOGO_PATH):/go/src/github.com/projectcalico/libcalico-go:ro
```
rafaelvanoni: Please use `-v ....:/go/src/$(PACKAGE_NAME):rw \` to follow what the rest of the file does.

Brian-McM: I may be missing something but PACKAGE_NAME name is github.com/projectcalico/felix, and it needs to be github.com/projectcalico/libcalico-go here

rafaelvanoni: Ah, my bad.. was looking at wrong repo. LGTM

PR #2122, opened 2019-08-22
Pass in user.name and user.email instead of --author in git-commit, and define GOPATH for go modules cachine when it is not specified by the environment

rafaelvanoni:


Makefile
```diff
@@ -218,19 +218,20 @@ ifdef SSH_AUTH_SOCK
...
-ifndef CIRCLECI
 	# If the environment is using multiple comma-separated directories for gopath, use the first one, as that
 	# is the default one used by go modules.
-	LOCAL_GOPATH = $(shell echo $(GOPATH) | cut -d':' -f1)
-	EXTRA_DOCKER_ARGS += -v $(LOCAL_GOPATH)/pkg/mod:/go/pkg/mod:rw
-endif
+	GOMOD_CACHE = $(shell echo $(GOPATH) | cut -d':' -f1)/pkg/mod
+else
+	# If gopath is empty, default to $(HOME)/go.
+	GOMOD_CACHE = $(HOME)/go/pkg/mod
```
neiljerram: But if the user doesn't have GOPATH, I guess that that path won't exist?

neiljerram: I feels like it would be more right only to extend EXTRA_DOCKER_ARGS when GOPATH exists.

Makefile
```diff
@@ -299,7 +300,7 @@ git-status:
 	git status --porcelain
 
 git-commit:
-	git diff-index --quiet HEAD || git commit -m "Semaphore Automatic Update" --author "Semaphore Automatic Update <marvin@tigera.io>" go.mod go.sum
+	git diff-index --quiet HEAD || git commit -m "Semaphore Automatic Update" -c user.name="Semaphore Automatic Update" -c user.email="<marvin@tigera.io>" go.mod go.sum
```
neiljerram: What difference does this make?

Makefile
rafaelvanoni: There's some code in this PR to create that directory, so this should be fine.

Makefile
rafaelvanoni: Turns out git requires a user to be configured, which then can specify a different author. But the former is still needed.

Makefile
neiljerram: Sorry, I see that now, so great.

Makefile
neiljerram: I think it's worth documenting that, so that it doesn't get reverted, when some other motivation crops up, just because someone didn't understand why it was there.  In order of preference:

1. In a comment just above that line.
2. In the Git commit message.
3. In the PR description.

WDYT?

rafaelvanoni:
> I think it's worth documenting that, so that it doesn't get reverted, when some other motivation crops up, just because someone didn't understand why it was there. In order of preference:
> 
>     In a comment just above that line.
>     In the Git commit message.
>     In the PR description.
> 
> WDYT?
> 

Don't really need a comment, the tests will fail in CI if someone removes these options. It passed on my machine because I already had these configured in my git config file.

Makefile
rafaelvanoni: Don't really need a comment, the tests will fail in CI if someone removes these options. It passed on my machine because I already had these configured in my git config file.

neiljerram: Ah, OK then.  Thanks for explaining!

PR #2127, opened 2019-08-29
Handle VXLAN mode changes

caseydavenport:
- [ ] Documentation
- [ ] Backport
- [ ] Release note

```release-note
None required
```

### DIFF
--- a/calc/vxlan_resolver.go
+++ b/calc/vxlan_resolver.go
@@ -379,21 +379,29 @@ func (c *VXLANResolver) OnPoolUpdate(update api.Update) (_ bool) {
 		// This is an add/update of a pool with VXLAN enabled.
 		logrus.WithField("pool", k.CIDR).Info("Update of VXLAN-enabled IP pool.")
 		if curr, ok := c.vxlanPools[k.String()]; ok {
-			// We already know about this IP pool. Check to see if the CIDR has changed.
+			// We already know about the IP pool. Check to see if any fields have changed.
+			vxlanModeChanged := curr.VXLANMode != update.Value.(*model.IPPool).VXLANMode
+			// Check to see if the CIDR has changed.
 			// While this isn't possible directly in the user-facing API, it's possible that
 			// we see a delete/recreate as an update over the Syncer in rare cases.
-			if curr.CIDR.String() == update.Value.(*model.IPPool).CIDR.String() {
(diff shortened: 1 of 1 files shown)


calc/vxlan_resolver.go
```diff
@@ -379,21 +379,29 @@ func (c *VXLANResolver) OnPoolUpdate(update api.Update) (_ bool) {
...
-			// If the CIDR has changed, treat this as a delete followed by a re-create
+			// The pool has changed, treat this as a delete followed by a re-create
 			// with the new CIDR. Iterate through sent routes and withdraw any within
-			// the old CIDR.
+			// the old pool's CIDR. We'll kick the pending set below to trigger any updates.
 			delete(c.vxlanPools, k.String())
 			sentSet.Iter(func(item interface{}) error {
 				r := item.(vxlanRoute)
-				if !c.containsRoute(curr, r) {
+				if c.containsRoute(curr, r) {
```
neiljerram: Removing the `!` here looks wrong.  Was that intended?

neiljerram: If I'm reading it correctly, the logic was:
```
  For each route that is currently programmed (`sentSet`):
    If that route isn't within the old IPPool CIDR:
      Withdraw that route
```
Which I admit doesn't sound quite right.  When I first read it earlier, I thought `curr` was the new IPPool state, so then it would be this, which I think _does_ make sense:
```
  For each route that is currently programmed (`sentSet`):
    If that route isn't within the new IPPool CIDR:
      Withdraw that route
```

So now your proposed logic appears to be:
```
  For each route that is currently programmed (`sentSet`):
    If that route _is_ within the old IPPool CIDR:
      Withdraw that route
      Plan to add that route again
```

Why would that be right?  Or have I just misinterpreted the logic?

neiljerram: @Brian-McM Thanks for your comment, which I only just read.  Could you explain a little further:

- why remove and re-add routes that are still needed?

- why not remote existing routes that are outside the IP Pool's new CIDR?

I don't really know VXLAN, and it's probably just my ignorance speaking here - but interested if the explanations are simple!

neiljerram: @Brian-McM Thanks for explaining, it all makes sense now.

PR #2128, opened 2019-08-29
Use node resource updates with Typha

lmm:
## Description

Related to: https://github.com/projectcalico/typha/pull/298

This updates the Typha handling to check whether the Typha supports node resource updates. If it does, then we update configParams so the VXLAN resolver knows to properly handle cross-subnet routes.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/daemon/daemon.go
+++ b/daemon/daemon.go
@@ -410,14 +410,13 @@ configRetry:
 			}
 		}
 
-		supportsNodeResources, err := typhaConnection.SupportsNodeResourceUpdates(10 * time.Second)
+		supportsNodeResourceUpdates, err := typhaConnection.SupportsNodeResourceUpdates(10 * time.Second)
 		if err != nil {
 			log.WithError(err).Error("Did not get hello message from Typha in time, assuming it does not support node resource updates")
+			return
 		}
-		if supportsNodeResources {
-			log.Info("Using node resource updates with Typha that supports it")
-			configParams.SetUseNodeResourceUpdates(true)
-		}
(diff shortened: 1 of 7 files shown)


daemon/daemon.go
```diff
@@ -378,10 +378,53 @@ configRetry:
...
+				log.WithError(err).Fatal("Failed to connect to Typha")
+			} else {
+				log.Info("Connected to Typha after retries.")
+				healthAggregator.Report(healthName, &health.HealthReport{Live: true, Ready: true})
+			}
+		}
+
+		supportsNodeResources, err := typhaConnection.SupportsNodeResourceUpdates(10 * time.Second)
+		if err != nil {
+			log.WithError(err).Error("Did not get hello message from Typha in time, assuming it does not support node resource updates")
```
caseydavenport: I wonder if we should just abort altogether if we hit this (or, maybe retry?) - it's probably worse to continue on with a bad configuration than to just exit. WDYT?

daemon/daemon.go
```diff
@@ -378,10 +378,53 @@ configRetry:
...
+		}
+		if supportsNodeResources {
+			log.Info("Using node resource updates with Typha that supports it")
+			configParams.SetUseNodeResourceUpdates(true)
```
caseydavenport: We could probably just do `configParams.SetUseNodeResourceUpdates(supportsNodeResourceUpdates)`

daemon/daemon.go
lmm: Yup, fixed

PR #2130, opened 2019-08-30
Re-enable golangci-lint

rafaelvanoni:


calc/vxlan_resolver.go
```diff
@@ -693,10 +693,12 @@ func (c *VXLANResolver) vtepMACForHost(nodename string) string {
 	}
 
 	hasher := sha1.New()
-	hasher.Write([]byte(nodename))
+	_, err := hasher.Write([]byte(nodename))
+	if err != nil {
+		logrus.WithField("node", nodename).Debug("Failed to write hash for node")
```
fasaxc: Remember to do `logrus.WithError(err)` to actually log out the error.  Also, you should log this at Error (or maybe even Panic) since hasher.Write shouldn't ever return an error (IIRC it's only defined that way to allow it to implement the Writer interface).

daemon/daemon.go
```diff
@@ -193,15 +183,15 @@ configRetry:
 			continue configRetry
 		}
 		// Parse and merge the local config.
-		configParams.UpdateFrom(envConfig, config.EnvironmentVariable)
-		if configParams.Err != nil {
+		_, err = configParams.UpdateFrom(envConfig, config.EnvironmentVariable)
+		if err != nil || configParams.Err != nil {
```
fasaxc: This code was using the design pattern used by the `bufio.Scanner`, where it collects its errors so the calling code doesn't need so much boilerplate. After your changes, we now have the error collection code and the boilerplate, which makes this code worse.

Think we should do one of these things:

- Stick with the pattern and move it closer to the Scanner API so that the linter doesn't complain (i.e. stop returning the error from each `Update...()` call.  `UpdateFrom(...) bool` and `Err() error`
- Ditch `configParams.Err` and just return the error and live with the boilerplate.

dataplane/external/ext_dataplane.go
```diff
@@ -33,6 +34,9 @@ import (
 // StartExtDataplaneDriver starts the given driver as a child process and returns a
 // connection to it along with the command itself so that it may be monitored.
 func StartExtDataplaneDriver(driverFilename string) (*extDataplaneConn, *exec.Cmd) {
+	var wg sync.WaitGroup
+	var errStdout, errStderr error
```
fasaxc: Please move these to where they're needed.  Minimising the scope of variables is best practice (and we don't need to worry about archaic C compilers since we're not writing C)

dataplane/external/ext_dataplane.go
```diff
@@ -63,17 +76,26 @@ func StartExtDataplaneDriver(driverFilename string) (*extDataplaneConn, *exec.Cm
...
-		cmd.Process.Kill()
+		_ = cmd.Process.Kill()
 		log.WithError(err).Fatal("Failed to close parent's copy of pipe")
 	}
 	dataplaneConnection := &extDataplaneConn{
 		toDataplane:   toDriverW,
 		fromDataplane: fromDriverR,
 	}
+
+	wg.Wait()
```
fasaxc: This is wrong, this function didn't wait for the dataplane driver to finish before.  That's why we started background goroutines to copy the output to the log.

fv/containers/containers.go
```diff
@@ -403,7 +407,10 @@ func (c *Container) EnsureBinary(name string) {
 	defer c.mutex.Unlock()
 
 	if !c.binaries.Contains(name) {
-		utils.Command("docker", "cp", "../bin/"+name, c.Name+":/"+name).Run()
+		err := utils.Command("docker", "cp", "../bin/"+name, c.Name+":/"+name).Run()
+		if err != nil {
+			log.WithField("name", name).Info("Failed to run 'docker cp' command")
```
fasaxc: Log at `Error()`?

hashutils/id.go
```diff
@@ -33,7 +33,10 @@ func GetLengthLimitedID(fixedPrefix, suffix string, maxLength int) string {
 		// start with the character that we use to denote a shortened string, which could
 		// result in a clash.  Hash the value and truncate...
 		hasher := sha256.New()
-		hasher.Write([]byte(suffix))
+		_, err := hasher.Write([]byte(suffix))
+		if err != nil {
+			return ""
```
fasaxc: Should probably panic here.  I think hashers are contracted not to return errors (and returning "" doesn't handle the error)

iptables/rules.go
```diff
@@ -92,15 +92,37 @@ func (c *Chain) RuleHashes(features *Features) []string {
 	// First hash the chain name so that identical rules in different chains will get different
 	// hashes.
 	s := sha256.New224()
-	s.Write([]byte(c.Name))
+	_, err := s.Write([]byte(c.Name))
+	if err != nil {
+		log.WithFields(log.Fields{
+			"chain": c.Name,
+		}).Debug("failed to write hash")
```
fasaxc: Again, should probably panic here, this shouldn't fail and it'll break everything downstream if it does

iptables/rules.go
```diff
@@ -92,15 +92,37 @@ func (c *Chain) RuleHashes(features *Features) []string {
...
 		// the rules before it affect its hash.
 		s.Reset()
-		s.Write(hash)
+		_, err = s.Write(hash)
+		if err != nil {
+			log.WithFields(log.Fields{
+				"action":   rule.Action,
+				"position": ii,
+				"chain":    c.Name,
+			}).Debug("failed to write hash")
```
fasaxc: Same here and below

k8sfv/calc_graph.go
```diff
@@ -85,7 +85,7 @@ func rotateLabels(clientset *kubernetes.Clientset, nsPrefix string) error {
 					panic(err)
 				}
 				ns_in.ObjectMeta.Labels["maturity"] = changeTo[ii]
-				ns_out, err := clientset.CoreV1().Namespaces().Update(ns_in)
+				ns_out, _ := clientset.CoreV1().Namespaces().Update(ns_in)
```
fasaxc: Probably worth at least logging this error.  I don't know this code well enough to know if we should stop after an error

calc/vxlan_resolver.go
rafaelvanoni: Fixed

daemon/daemon.go
rafaelvanoni: The second option would allow us to have more specific error message, which is always nice to have when debugging something like this. So I'm going with that one, but let me know if you prefer otherwise.

dataplane/external/ext_dataplane.go
rafaelvanoni: Fixed

dataplane/external/ext_dataplane.go
rafaelvanoni: Good point, fixed

fv/containers/containers.go
rafaelvanoni: Fixed

hashutils/id.go
rafaelvanoni: Yeah, I wasn't sure what to do here. Thanks.
Fixed

iptables/rules.go
rafaelvanoni: Fixed

iptables/rules.go
rafaelvanoni: Fixed

k8sfv/calc_graph.go
rafaelvanoni: Fixed

dataplane/linux/ipip_mgr.go
```diff
@@ -224,9 +224,7 @@ func (m *ipipManager) CompleteDeferredWork() error {
 		for _, ip := range m.activeHostnameToIP {
 			members = append(members, ip)
 		}
-		for _, ip := range m.externalNodeCIDRs {
-			members = append(members, ip)
-		}
+		members = append(members, m.externalNodeCIDRs...)
```
fasaxc: Same change just above?

dataplane/mock/mock_dataplane.go
```diff
@@ -182,18 +182,16 @@ func (d *MockDataplane) NumEventsRecorded() int {
 }
 
 func copyPolOrder(in map[string][]TierInfo) map[string][]TierInfo {
-	copy := map[string][]TierInfo{}
+	local_copy := map[string][]TierInfo{}
```
fasaxc: `s/local_copy/localCopy/`

hashutils/id.go
```diff
@@ -33,7 +35,10 @@ func GetLengthLimitedID(fixedPrefix, suffix string, maxLength int) string {
 		// start with the character that we use to denote a shortened string, which could
 		// result in a clash.  Hash the value and truncate...
 		hasher := sha256.New()
-		hasher.Write([]byte(suffix))
+		_, err := hasher.Write([]byte(suffix))
+		if err != nil {
+			log.Panic("failed to write hash suffix")
```
fasaxc: Nit: this log message is a bit sloppy.  You're not writing a "hash suffix", you're writing the suffix to the hash.  Also, you should log the error and use sentence-case for log messages (since they stand alone in the output, unlike `error`s, which are intended to be logged as _part_ of a message:
```
log.WithError(err).Panic("Failed to write suffix to hash.")
```

iptables/rules.go
```diff
@@ -92,15 +92,37 @@ func (c *Chain) RuleHashes(features *Features) []string {
 	// First hash the chain name so that identical rules in different chains will get different
 	// hashes.
 	s := sha256.New224()
-	s.Write([]byte(c.Name))
+	_, err := s.Write([]byte(c.Name))
+	if err != nil {
+		log.WithFields(log.Fields{
+			"chain": c.Name,
+		}).Panic("failed to write hash")
```
fasaxc: Same here

iptables/rules.go
```diff
@@ -92,15 +92,37 @@ func (c *Chain) RuleHashes(features *Features) []string {
...
 		// the rules before it affect its hash.
 		s.Reset()
-		s.Write(hash)
+		_, err = s.Write(hash)
+		if err != nil {
+			log.WithFields(log.Fields{
+				"action":   rule.Action,
+				"position": ii,
+				"chain":    c.Name,
+			}).Panic("failed to write hash")
```
fasaxc: And here.

iptables/rules.go
```diff
@@ -92,15 +92,37 @@ func (c *Chain) RuleHashes(features *Features) []string {
...
 		ruleForHashing := rule.RenderAppend(c.Name, "HASH", features)
-		s.Write([]byte(ruleForHashing))
+		_, err = s.Write([]byte(ruleForHashing))
+		if err != nil {
+			log.WithFields(log.Fields{
+				"ruleFragment": ruleForHashing,
+				"action":       rule.Action,
+				"position":     ii,
+				"chain":        c.Name,
+			}).Panic("failed to write rule for hashing")
```
fasaxc: And here.

k8sfv/k8sfv.go
```diff
@@ -167,7 +167,8 @@ func initialize(k8sServerEndpoint string) (clientset *kubernetes.Clientset) {
 			return
 		}
 
-		ctx, _ := context.WithTimeout(context.Background(), 10*time.Second)
+		ctx, cancelfun := context.WithTimeout(context.Background(), 10*time.Second)
```
fasaxc: Nit: use camelCase for locals `cancelFunc`

usagerep/usagerep_test.go
```diff
@@ -52,7 +52,8 @@ var _ = Describe("UsageReporter with mocked URL and short interval", func() {
 		httpHandler = &requestRecorder{}
 		go func() {
 			defer GinkgoRecover()
-			http.Serve(tcpListener, httpHandler)
+			// TODO: Investigate why this call sometimes returns an error.
+			_ = http.Serve(tcpListener, httpHandler)
```
fasaxc: You can start by logging out the error

daemon/daemon.go
```diff
@@ -238,14 +228,19 @@ configRetry:
 				time.Sleep(1 * time.Second)
 				continue configRetry
 			}
-			configParams.UpdateFrom(globalConfig, config.DatastoreGlobal)
-			configParams.UpdateFrom(hostConfig, config.DatastorePerHost)
+			_, err = configParams.UpdateFrom(globalConfig, config.DatastoreGlobal)
+			if err != nil {
+				log.WithError(err).Error("Failed update global config from datastore")
```
fasaxc: Now you're no longer checking `configParams.Err` below, this code needs to do the sleep and continue

daemon/daemon.go
```diff
@@ -238,14 +228,19 @@ configRetry:
...
+			}
+			_, err = configParams.UpdateFrom(hostConfig, config.DatastorePerHost)
+			if err != nil {
+				log.WithError(err).Error("Failed update host config from datastore")
```
fasaxc: Same here

Makefile
```diff
@@ -520,12 +520,9 @@ sub-tag-images-%:
 static-checks:
 	$(MAKE) check-typha-pins golangci-lint
 
-# TODO: re-enable these linters !
-LINT_ARGS := --disable staticcheck,ineffassign,gosimple,govet,deadcode,errcheck,unused,varcheck,structcheck
-
 .PHONY: golangci-lint
 golangci-lint: $(GENERATED_FILES)
-	$(DOCKER_RUN) $(CALICO_BUILD) golangci-lint run --deadline 5m $(LINT_ARGS)
+	$(DOCKER_RUN) $(CALICO_BUILD) golangci-lint run --deadline 5m --max-issues-per-linter 0 --max-same-issues 0 --disable govet
```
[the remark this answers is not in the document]
rafaelvanoni: The default is to only report the first 50 issues per linter and the first 3 issues from the same linter. It's just a nice to have so we know how much work is still left.

bpf/bpf.go
```diff
@@ -354,13 +354,10 @@ type BPFDataplane interface {
 	RemoveXDP(ifName string, mode XDPMode) error
 	UpdateCIDRMap(ifName string, family IPFamily, ip net.IP, mask int, refCount uint32) error
 	UpdateFailsafeMap(proto uint8, port uint16) error
-	loadXDPRaw(objPath, ifName string, mode XDPMode, mapArgs []string) error
```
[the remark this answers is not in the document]
rafaelvanoni: Sure

bpf/mock_bpf_lib.go
```diff
@@ -572,7 +572,10 @@ func NewMockSockMap(mapID int) SockMap {
 
 func GetMockXDPTag(bytes []byte) string {
 	h := sha1.New()
-	h.Write(bytes)
+	_, err := h.Write(bytes)
+	if err != nil {
+		return ""
```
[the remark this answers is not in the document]
rafaelvanoni: Fixed

dataplane/linux/ipip_mgr.go
rafaelvanoni: Fixed

dataplane/mock/mock_dataplane.go
rafaelvanoni: Fixed

hashutils/id.go
rafaelvanoni: Fixed

iptables/rules.go
rafaelvanoni: Fixed

iptables/rules.go
rafaelvanoni: Fix

iptables/rules.go
rafaelvanoni: Fixed

k8sfv/k8sfv.go
rafaelvanoni: Fixed

usagerep/usagerep_test.go
rafaelvanoni: Fixed

daemon/daemon.go
rafaelvanoni: Fixed

daemon/daemon.go
rafaelvanoni: Fixed

dataplane/linux/ipip_mgr.go
rafaelvanoni: Actually, `m.activeHostnameToIP` is a map.. I'm not sure how to use it in an `append()` call (and neither does any of the folks around here I could find atm)

dataplane/mock/mock_dataplane.go
```diff
@@ -203,11 +201,11 @@ func (d *MockDataplane) Config() map[string]string {
 	if d.config == nil {
 		return nil
 	}
-	copy := map[string]string{}
+	local_copy := map[string]string{}
```
fasaxc: Should be `localCopy`

fv/test-workload/test-workload.go
```diff
@@ -171,10 +171,19 @@ func main() {
 					return
 				}
 				err = utils.RunCommand("ip", "-6", "route", "add", "default", "via", hostIPv6Addr.String(), "dev", "eth0")
+				if err != nil {
```
fasaxc: Should probably log and carry on in this file, it's just collecting diags and there's no reason to skip later diags because one  failed.

dataplane/mock/mock_dataplane.go
rafaelvanoni: Fixed

fv/test-workload/test-workload.go
rafaelvanoni: Fixed

fv/test-workload/test-workload.go
```diff
@@ -171,10 +171,19 @@ func main() {
...
+					log.WithField("hostIP", hostIPv6Addr.String()).WithError(err).Info("Failed to add IPv6 route to eth0.")
+				}
 
 				// Output the routing table to the log for diagnostic purposes.
-				utils.RunCommand("ip", "-6", "route")
-				utils.RunCommand("ip", "-6", "addr")
+				err = utils.RunCommand("ip", "-6", "route")
+				if err != nil {
```
fasaxc: Sorry meant this one and the couple below as well.

rafaelvanoni: Fixed

fv/test-workload/test-workload.go
```diff
@@ -155,56 +155,66 @@ func main() {
 				// Without these sysctls enabled, interfaces will come up but they won't get a link local IPv6 address,
 				// which is required to add the default IPv6 route.
 				if err = writeProcSys("/proc/sys/net/ipv6/conf/all/disable_ipv6", "0"); err != nil {
+					log.WithError(err).Error("Failed to disable IPV6.")
```
fasaxc: this is enabling IPv6; it's setting the disable flag to 0

rafaelvanoni: Fixed

PR #2139, opened 2019-09-09
Fix LOCAL_BUILD

neiljerram:
The libcalico-go mount must be added to EXTRA_DOCKER_ARGS before the
definition of DOCKER_RUN, as the latter is an immediate
definition (i.e. with := instead of =).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -231,6 +231,10 @@ endif
 
 EXTRA_DOCKER_ARGS += -v $(GOMOD_CACHE):/go/pkg/mod:rw
 
+ifdef LOCAL_BUILD
+EXTRA_DOCKER_ARGS+=-v $(CURDIR)/../libcalico-go:/go/src/github.com/projectcalico/libcalico-go:rw
+endif
+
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -231,6 +231,10 @@ endif
 
 EXTRA_DOCKER_ARGS += -v $(GOMOD_CACHE):/go/pkg/mod:rw
 
+ifdef LOCAL_BUILD
+EXTRA_DOCKER_ARGS+=-v $(CURDIR)/../libcalico-go:/go/src/github.com/projectcalico/libcalico-go:rw
```
caseydavenport: Does this actually need write permissions? That feels kinda scary to me - will go modules overwrite your local libcalico checkout?

caseydavenport: @rafaelvanoni

neiljerram: @rafaelvanoni Do you know the answer to @caseydavenport 's question?  (My change here just moves the existing mount a bit earlier in the Makefile, so I haven't originated the `:rw`.)

caseydavenport:
I think this should just be merged since it doesn't introduce the write privileges, but would appreciate an answer from @rafaelvanoni anyway :)

PR #2144, opened 2019-09-10
Reduce severity of BPF log levels

tmjd:
## Description
Since we fallback to iptables if BPF is not able to be applied so reduce the log level when errors happen when applying.

## Release Note

```release-note
None required
```


dataplane/linux/xdp_state.go
```diff
@@ -2155,8 +2155,8 @@ type xdpMemberCache struct {
 
 func newXDPMemberCache(family bpf.IPFamily, bpfLib bpf.BPFDataplane) *xdpMemberCache {
 	return &xdpMemberCache{
-		family:                 family,
-		cache:                  make(map[string]map[bpf.CIDRMapKey]uint32),
+		family: family,
```
tmjd: I don't know why my go fmt kept doing this.

caseydavenport: Weird. What version of go do you have?

tmjd: 1.11.2, I should update.

PR #2161, opened 2019-10-12
Add support for REJECT instead of DROP

jackkleeman:
New feature, fixes #1859, allowing you to both specify reject actions in your policies, and also set drops that occur when no policies or profiles match to be REJECT actions. This is useful in cases where you want to notify callers that the path is disallowed instead of just letting their requests time out. Affects felix only but we will want to add the new config flag in libcalico too. Not sure what testing is needed as theres a fairly limited change surface, but can add some more unit tests if needed for the reject action.

Will add documentation if we are happy with this change

## Todos
- [x] Integration tests (delete as appropriate) not needed
- [ ] Documentation
- [ ] Backport
- [x] Release note

## Release Note
```release-note
Add support for reject action which maps to iptables `REJECT` and add a config flag which means that packets which don't pass policies or profiles are rejected instead of dropped
```


bcavns01:
Would love to see this merged in.  I've got another use case for IptablesNotPassedAction, but I'm hesitant to open a PR while this one is still open.

tmjd:
@bcavns01 could you please create an issue and layout your use case as it would be good to understand it as well as this one when adding a feature like this?

tmjd:
@jackkleeman So as I understand this it is doing 2 things:
1. Adding a new action of reject that can be used in Calico (Global)NetworkPolicy
1. Adding a config to change the DROP at the end of an unmatched chain to REJECT
Would it make more sense to, instead of being able to specify the action (1), just have any drop actions for (Global)NetworkPolicy to use the 2 config action?
I'm curious if you have a specific use case where all DROPs should not just be replaced with REJECTs.
To be clear, I'm not asking for any changes on this PR (yet) but asking if you have a specific use case where you need DROP for the other NetworkPolicies but change the default to REJECT.

jackkleeman:
I guess we could have a config flag that changes all denied packets to actually be implemented by REJECT. It seems to me that it would be cool to be able to write some policies that reject, and some that don't. The thing we could maybe not do is 2, because I guess you can always write a default deny policy with a REJECT action. That might simplify this change a little

tmjd:
Sorry for the delay in responding here. I've chatted with a few people about this and I'd like to suggest some changes to this PR.

It seems to make most sense to add an option to change the behavior for all places that we drop packets instead of only doing it at the end of an unmatched change. 
Also without a good use case for adding a new action to the (Global)NetworkPolicy we should not add that at this time.
Another minor point is it would make be a more appropriate name (with the above suggested changes) for the Config option to be DropActionOverride instead of IptablesNotPassedAction.

WDYT?

jackkleeman:
No problem @tmjd I can do all of that in the next few days hopefully

caseydavenport:
@jackkleeman @tmjd looks like this PR has some conflicts. Is this still in progress?

jackkleeman:
It's just been waiting on review. I was just trying to pick out an old ticket so theres no urgency for me - is this still desirable?

fasaxc:
@jackkleeman CI failed with a build error.  Maybe a mismerge?

jackkleeman:
Sadly I have now moved job and I can't make new contributions without permission - feel free to take what I've already contributed into a new PR!

PR #2162, opened 2019-10-16
Fix memory leak in endpoint_mgr

alexhexabeam:
Backport of https://github.com/projectcalico/felix/pull/2026/files

## Description
Bug fix for memory leak in endpoint_mgr.go

We detected a leak in our calico-node 2.6.12 containers, and we cannot upgrade to the 3.x versions since we require libnetwork support, hence the backport.

This change in Felix reduced the leak by 88% in our testing by continuously spawning and removing containers on the network. I have not found the remaining 12% leak, and it is small enough to possibly be within the margin of error.

Manual testing (in addition to passing unit tests):
```
while true; do
    sudo docker run --rm -d --net exabeam-calico exabeam-mongo sleep 1
    sleep 0.1
done
```

(replace the container ID with the calico-node container)
```
while true; do
    grep "^rss " /sys/fs/cgroup/memory/docker/26728cc58950cecea5d7455c9171df64ec06d0d1eb10e0923ba27e70180568b5/memory.stat >> ~/calico-node_rss
    sleep 1
done
```

We compared the version with the fix and without it, and we saw an 88% reduction in the memory growth. I have not found the remaining 12%, and it is small enough that it may be within the margin of error.

## Todos
- [x] Documentation
- [x] Backport
- [X] Release note

## Release Note

```release-note
Fixed slow memory leak that was noticed with lots of workload churn.
```

### DIFF
--- a/intdataplane/endpoint_mgr.go
+++ b/intdataplane/endpoint_mgr.go
@@ -464,6 +464,7 @@ func (m *endpointManager) resolveWorkloadEndpoints() {
 		} else {
 			logCxt.Info("Workload removed, deleting its chains.")
 			m.filterTable.RemoveChains(m.activeWlIDToChains[id])
+			delete(m.activeWlIDToChains, id)
 			if oldWorkload != nil {
 				// Remove any routes from the routing table.  The RouteTable will
 				// remove any conntrack entries as a side-effect.


alexhexabeam:
I don't think this needs any documentation.

caseydavenport:
Agreed. @alexhexabeam we have stopped cutting v2.6 releases, so while I'm happy to merge this PR it won't show up into a v2.6 release. 

Thanks for the contribution. If you're interested, I know there are some folks using the libnetwork plugin with Calico v3, however the core team doesn't have the bandwidth to spend time on it any more so we're relying on external contributors to submit fixes and perform testing.

PR #2167, opened 2019-10-23
Handle multiple workload endpoints with the same interface name

neiljerram:
## Release Note
```release-note
Fix Felix panic when there are transiently multiple workload endpoints with the same interface name.
```

### DIFF
--- a/dataplane/linux/endpoint_mgr.go
+++ b/dataplane/linux/endpoint_mgr.go
@@ -31,6 +31,7 @@ import (
 	"github.com/projectcalico/felix/proto"
 	"github.com/projectcalico/felix/routetable"
 	"github.com/projectcalico/felix/rules"
+	"github.com/projectcalico/libcalico-go/lib/names"
 	"github.com/projectcalico/libcalico-go/lib/set"
 )
 
@@ -147,6 +148,10 @@ type endpointManager struct {
 	activeWlDispatchChains     map[string]*iptables.Chain
 	activeEPMarkDispatchChains map[string]*iptables.Chain
 
+	// Workload endpoints that would be locally active but are 'shadowed' by other endpoints
+	// with the same interface name.
+	shadowedWlEndpoints map[proto.WorkloadEndpointID]*proto.WorkloadEndpoint
+
 	// wlIfaceNamesToReconfigure contains names of workload interfaces that need to have
 	// their configuration (sysctls etc.) refreshed.
 	wlIfaceNamesToReconfigure set.Set
@@ -256,6 +261,8 @@ func newEndpointManagerWithShims(
 		activeWlEndpoints:     map[proto.WorkloadEndpointID]*proto.WorkloadEndpoint{},
 		activeWlIfaceNameToID: map[string]proto.WorkloadEndpointID{},
 		activeWlIDToChains:    map[proto.WorkloadEndpointID][]*iptables.Chain{},
+
+		shadowedWlEndpoints: map[proto.WorkloadEndpointID]*proto.WorkloadEndpoint{},
 
 		wlIfaceNamesToReconfigure: set.New(),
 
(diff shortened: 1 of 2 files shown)


dataplane/linux/endpoint_mgr.go
```diff
@@ -617,6 +674,20 @@ func (m *endpointManager) resolveWorkloadEndpoints() {
 	})
 }
 
+func wlIdToUniqueName(id *proto.WorkloadEndpointID) string {
+	ids := names.WorkloadEndpointIdentifiers{
+		// Passing "default" for the Orchestrator field tells CalculateWorkloadEndpointName
+		// to construct a name based only on the following fields, which is what we want.
```
fasaxc: Couldn't you have two clashing WEPs that differ in their orchestrator value?

dataplane/linux/endpoint_mgr_test.go
```diff
@@ -1373,6 +1373,118 @@ func endpointManagerTests(ipVersion uint8) func() {
...
+									Tiers:      tiers,
+									Ipv4Nets:   []string{"10.0.240.2/24"},
+									Ipv6Nets:   []string{"2001:db8:2::2/128"},
+								},
+							})
+							err := epMgr.CompleteDeferredWork()
+							Expect(err).ToNot(HaveOccurred())
+						})
+
+						It("should have expected chains", expectWlChainsFor("cali12345-ab_policy1"))
```
fasaxc: Can you easily make this test distinguish between the different endpoints?  For example, by giving the clashing endpoints a different list of profiles?

dataplane/linux/endpoint_mgr.go
```diff
@@ -489,111 +496,161 @@ func (m *endpointManager) resolveWorkloadEndpoints() {
...
-				for ii, natInfo := range natInfos {
-					ipStrings[len(old)+ii] = natInfo.ExtIp + addrSuffix
+	removeActiveWorkload := func(logCxt *log.Entry, oldWorkload *proto.WorkloadEndpoint, id proto.WorkloadEndpointID) {
+		m.callbacks.InvokeRemoveWorkload(oldWorkload)
+		m.filterTable.RemoveChains(m.activeWlIDToChains[id])
+		delete(m.activeWlIDToChains, id)
+		if oldWorkload != nil {
+			m.epMarkMapper.ReleaseEndpointMark(oldWorkload.Name)
+			// Remove any routes from the routing table.  The RouteTable will remove any
+			// conntrack entries as a side-effect.
```
fasaxc: If the interface is being recreated (as we thought was happening in OpenStack), is there any chance that this conntrack side-effect will cause problems?

dataplane/linux/endpoint_mgr.go
neiljerram: Yes, maybe that's possible.  I guess I could pass `"default"+id.OrchestratorId` here.  Or I could roll my own unique name function here.  WDYT?

dataplane/linux/endpoint_mgr_test.go
neiljerram: Yes, see next commit.

dataplane/linux/endpoint_mgr.go
neiljerram: Well you are Mr Conntrack :smile: , but

- I guess it would be observable if someone was explicitly monitoring or depending on conntrack entries.

- Otherwise I would guess no, and that the worst case is going through iptables and NAT again?  (But if you go through NAT again, can you get a different result?)

- I guess we need to understand the customer scenario better, to work out if this matters.

fasaxc: If you remove the conntrack entry that usually breaks any in-progress connections.  I think subsequent packets will hit the "conntrack invalid" drop rule so unfortunately it doesn't recover.

Probably OK if this really is a transient misconfiguration during a VM reboot or something (since the old connections will be gone anyway).  However, if Openstack just cycles the WEP for no reason, we might break some connections here.

dataplane/linux/endpoint_mgr.go
neiljerram: See next commit.  I've written `wlIdsAscending` instead, which doesn't rely on `WorkloadEndpointIdentifiers` from libcalico-go.

dataplane/linux/endpoint_mgr.go
neiljerram: We've agreed for now to assume that it's OK to risk impacting existing connections - on the basis that the scenario is probably one of guest OS restart.

PR #2174, opened 2019-11-07
Allow ipV6LinkLocalCIDR route to be programmed other than by us

neiljerram:
This was a regression accidentally introduced by #2118 . The IPv6 link local route is critical to IPv6 operation, so we mustn't remove it.


neiljerram:
Change already reviewed at #2173

PR #2177, opened 2019-11-08
Switch Docker base image to ubi

lmm:
## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```


lmm:
Is it kosher to have the LICENSE file in the root of the repo be a symlink?

lmm:
Unfortunately, this PR as-is will not pass CI since the kernel version in the ubi base images is too old for the XDP tests

> XDP acceleration not supported: kernel is too old (have: 4.4.0 but want at least: 4.16.0)

fasaxc:
@lmm Those messages about XDP are skips, not failures.  This line at the bottom of the log tells you which test(s) failed:
```
[Fail] Latency tests with initialized Felix and etcd datastore [BeforeEach] IPv4: Network sets tests with initialized Felix and etcd datastore with all() source selector with 10k IPs in an IP set should have good latency 
```
However, there are much better diags further up.  The actual `error` was:
```
------------------------------
• Failure in Spec Setup (BeforeEach) [4.686 seconds]
Latency tests with initialized Felix and etcd datastore
/go/src/github.com/projectcalico/felix/fv/latency_test.go:52
  IPv4: Network sets tests with initialized Felix and etcd datastore [BeforeEach]
  /go/src/github.com/projectcalico/felix/fv/latency_test.go:219
    with all() source selector
    /go/src/github.com/projectcalico/felix/fv/latency_test.go:167
      with 10k IPs in an IP set
      /go/src/github.com/projectcalico/felix/fv/latency_test.go:183
        should have good latency
        /go/src/github.com/projectcalico/felix/fv/latency_test.go:203

        Unexpected error:
            <*exec.ExitError | 0xc0019369a0>: {
                ProcessState: {
                    pid: 18122,
                    status: 32256,
                    rusage: {
...
                    },
                },
                Stderr: nil,
            }
            exit status 126
        occurred

        /go/src/github.com/projectcalico/felix/fv/utils/utils.go:85
```
`exec.ExitError` has a bit of a useless string representation but at least you get the real source of the error:
```
        /go/src/github.com/projectcalico/felix/fv/utils/utils.go:85
```
And, taking a look [there](https://github.com/projectcalico/felix/blob/master/fv/utils/utils.go#L85), the error is logged much more usefully just above.  Searching the log for "Command failed" then gives:
```
2019-11-08 19:52:57.562 [WARNING][12081] utils.go 82: Command failed args=[]string{"exec", "felix-12081-202-felixfv", "apt-get", "install", "-y", "hping3"} command="docker" error=exit status 126 output="OCI runtime exec failed: exec failed: container_linux.go:348: starting container process caused \"exec: \\\"apt-get\\\": executable file not found in $PATH\": unknown\r\n"
```
So the real problem is trying to run `docker exec felix-12081-202-felixfv apt-get install -y hping3` and getting `executable file not found in $PATH`, which makes sense!

fasaxc:
> Is it kosher to have the LICENSE file in the root of the repo be a symlink?

Think I'd prefer keeping it where it was and copying it into the staging directory when needed for the docker build.

How does the license relate to the contents of the docker image?  It's only _our_ code that's Apache.

lmm:
Thanks @fasaxc for digging into this! I had meant to follow up on my comment.

> Think I'd prefer keeping it where it was and copying it into the staging directory when needed for the docker build. Reason is just that I can see it getting moved or deleted if we refactor the docker build and I don't think that should affect the main LICENSE file.

Makes sense. I've removed the symlink to the LICENSE in other PRs as well. I'll update this PR.

> How does the license relate to the contents of the docker image? It's only our code that's Apache.

We're required to have a license in `/licenses` in the container image. We may or may not require 3rd party licenses too - I'm following up on that now.

caseydavenport:
Felix doesn't need to be certified - it's not a user-facing deliverable, it's only used in our tests ( we don't ship a calico/felix image).

The main reason for this is to make sure we're using the same package versions / dependencies in the felix tests as we do in the node image that we ship.

fasaxc:
@lmm are you still working on this?  As @caseydavenport says, it would be good to be using the same packages in felix and node.

Note, since BPF went in, we now have seom GPL-licensed parts too: the BGP programs themselves.

lmm:
@fasaxc thanks for the nudge, yeah I should get this updated and merged. I'll update this PR today or tomorrow

lmm:
Closed in favor of https://github.com/projectcalico/felix/pull/2264

PR #2178, opened 2019-11-08
Move common Makefile targets to Makefile.common

rafaelvanoni:


neiljerram:
@rafaelvanoni Very interesting, but what's the overall plan here?

rafaelvanoni:
> @rafaelvanoni Very interesting, but what's the overall plan here?

Hey Neil, this is an attempt at a common Makefile that the other repos can include. I'm just opening these PRs so CI can have a go at it before requesting reviews.

neiljerram:
> a common Makefile that the other repos can include

Right, but I'm curious what design you have in mind for that?
- A repo that's included as a submodule?
- Some kind of machinery that periodically checks if Makefile.common is the same in all repos?
- Probably other possible solutions!

rafaelvanoni:
> Right, but I'm curious what design you have in mind for that?
> 
>     * A repo that's included as a submodule?
> 
>     * Some kind of machinery that periodically checks if Makefile.common is the same in all repos?
> 
>     * Probably other possible solutions!

For now just what these changes are implementing, a separate repo for a common Makefile (and possible other files) that is downloaded and included at the top of other Makefiles. As long as the download URL is pointing at master or a branch, we shouldn't need additional verification.

ozdanborne:
I am very excited to see us try out this idea. 

- I think Neil is right, it would be nice to have a plan about how repositories will pull in this Makefile library, how they'll get updates to it, etc. This included Makefile is sort of like an API that we need to make sure we not only stay compatible with but also keep up to date. We need to avoid a scenario where building an old commit fails because it was using an older include at the time but we can't figure out which one it was using.
- Any idea how we'll ensure that changes to the included Makefile don't break any of the repos that include it?

rafaelvanoni:
> I am very excited to see us try out this idea.
> 
>     * I think Neil is right, it would be nice to have a plan about how repositories will pull in this Makefile library, how they'll get updates to it, etc. This included Makefile is sort of like an API that we need to make sure we not only stay compatible with but also keep up to date. We need to avoid a scenario where building an old commit fails because it was using an older include at the time but we can't figure out which one it was using.

I could include a `git ls-remote` at the top of the `wget` call so build logs will identify the Makefile.common commit it's using. Would that be sufficient?

>     * Any idea how we'll ensure that changes to the included Makefile don't break any of the repos that include it?

No easy way of doing that with various repos including it. We have to be somewhat conservative about changes to Makefile.common.

Makefile
```diff
@@ -41,7 +41,12 @@
...
 ###############################################################################
 
-get_common:=$(shell wget -nv https://raw.githubusercontent.com/rafaelvanoni/common/master/Makefile.common -O Makefile.common)
+GO_BUILD_VER?=v0.26
+
+MAKE_BRANCH?=$(GO_BUILD_VER)
+MAKE_REPO?=https://raw.githubusercontent.com/projectcalico/go-build/$(MAKE_BRANCH)/Makefile.common
+
+get_common:=$(shell wget -nv $(MAKE_REPO) -O Makefile.common)
 include Makefile.common
```
fasaxc: This will get the common Makefile every time.  Make has a nice way of handling this, you can include a file that itself is built by a make target.  You can do something like this to make sure that it only grabs the file if it hasn't already been downloaded:
```
include Makefile.common
MAKE_BRANCH?=$(GO_BUILD_VER)
Makefile.common: Makefile.common.$(MAKE_BRANCH)
	cp "$<" "$@"
Makefile.common.$(MAKE_BRANCH):
	# Clean up any files downloaded from other branches so they don't accumulate.
	rm -f Makefile.common.*
	wget -nv https://raw.githubusercontent.com/projectcalico/go-build/$(MAKE_BRANCH)/Makefile.common -O "$@"
```
The intermediate step to create `Makefile.common.$(MAKE_BRANCH)` ensures that we re-get the file whenever `$(MAKE_BRANCH)` changes.

rafaelvanoni: I added `-nc` to the wget call, it has the same effect.

tmjd: I think @fasaxc 's suggestion will also ensure the file is downloaded if the GO_BUILD_VER/MAKE_BRANCH changes but I don't think it would be with the current wget.

rafaelvanoni: Fixed

Makefile
```diff
@@ -481,73 +290,10 @@ ifeq ($(ARCH),amd64)
...
 # Static checks
 ###############################################################################
-.PHONY: static-checks
-static-checks:
-	$(MAKE) check-typha-pins golangci-lint check-packr
-
-LINT_ARGS := --deadline 5m --max-issues-per-linter 0 --max-same-issues 0
-ifneq ($(CI),)
-	# govet uses too much memory for CI
-	LINT_ARGS += --disable govet
```
tmjd: It looks like we won't have this anymore, is that going to be a problem?

rafaelvanoni: No, this was fixed in a more recent build of golangci-ling.

PR #2179, opened 2019-11-11
Fix that Typha was failing to start in k8sfv tests.

fasaxc:
## Description

AFAICT, Typha has been failing to start in the k8sfv framework since the change to use scratch.  

* The tests that we run on a regular basis were only examining Felix's memory usage so they didn't spot that Felix wasn't actually in-sync with the datastore.  
* Felix now waits for some time before exiting if Typha isn't contactable (to prevent log spam and thundering herd), making the failures less frequent.
* Some of the time, the tests would pass because the whole test ran before Felix decided to exit.

This change:

- Fixes the underlying problem (removes the `sh` wrapper around the calico-typha command).
- Adds a bunch of diags.
- Adds a check that Felix is in-sync before each test (by checking its prometheus metric for datastore state).

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/k8sfv/k8sfv.go
+++ b/k8sfv/k8sfv.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2017 Tigera, Inc. All rights reserved.
+// Copyright (c) 2017,2019 Tigera, Inc. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -211,6 +211,7 @@ func cleanupAll(clientset *kubernetes.Clientset, nsPrefix string) {
 
 func panicIfError(err error) {
 	if err != nil {
+		log.WithError(err).Error("About to panic...")
 		panic(err)
 	}
 }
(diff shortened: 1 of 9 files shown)


k8sfv/k8sfv.go
```diff
@@ -211,6 +211,7 @@ func cleanupAll(clientset *kubernetes.Clientset, nsPrefix string) {
 
 func panicIfError(err error) {
 	if err != nil {
+		log.WithError(err).Error("About to panic...")
```
neiljerram: Is it useful to keep this?  (Just checking.)

k8sfv/leastsquares_test.go
```diff
@@ -22,6 +22,9 @@ import (
 )
 
 var _ = Context("least squares", func() {
+	BeforeEach(func() {
+		expectFelixReady()
```
neiljerram: These particular tests don't involve Felix, I believe.  So I don't think they should check that Felix is ready.

k8sfv/k8sfv.go
fasaxc: Might be useful in case the panic gets swallowed and logged out-of-order by Ginkgo?

PR #2182, opened 2019-11-15
Disable node CIDR allocation and defer cleanup

fasaxc:
## Description

* Enabling CIDR allocation makes the controller manager stall for 10s.
* Deferring cleanup saves 2-3s when running a single test for debugging.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/fv/infrastructure/infra_k8s.go
+++ b/fv/infrastructure/infra_k8s.go
@@ -57,6 +57,10 @@ type K8sDatastoreInfra struct {
 	BadEndpoint string
 
 	CertFileName string
+
+	// needsCleanup is set when we're told to Stop() in order to trigger deferred cleanup
+	// before the next test.  (If there is no next test, we'll skip the cleanup.)
+	needsCleanup bool
 }
 
 var (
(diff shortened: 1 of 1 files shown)


fv/infrastructure/infra_k8s.go
```diff
@@ -365,6 +378,17 @@ func (kds *K8sDatastoreInfra) EnsureReady() {
 }
 
 func (kds *K8sDatastoreInfra) Stop() {
+	// We don't actually stop the infra between tests because it's too expensive.
+	// Instead, mark all our resources for cleanup, but defer the cleanup until the
+	// start of the next test (this allows us to skip the cleanup if we happen to
+	// be the last test to run, which is a big win when manually running a single
+	// test for debugging.)
```
neiljerram: Do we still need this, given the other speedups you've made today?  I'm a bit worried about the complexity, unless it's still a big win to do this as well.

fasaxc: With the other fixes, my test run is down to about 25s; this shaves another 2-3s off, which feels worth it to me.  Another 2-3 similar optimisations and the FVs would be downright snappy :-)

neiljerram: OK, fair enough!

PR #2183, opened 2019-11-15
Fix incremental builds.  Avoid depending on phony targtets.

fasaxc:
## Description

- Replace remote-deps phony with a list of files.
- ifdef out the local build dependency if it's not in use.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,6 +40,10 @@
 #	   +--------------+
 #
 ###############################################################################
+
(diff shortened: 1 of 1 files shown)


tmjd:
@rafaelvanoni should the local build changes be copied to other repos?

PR #2184, opened 2019-11-25
Makefile.common.BRANCH should not depend on wget binary

neiljerram:
In general, a newer wget does not mean that we should re-download
Makefile.common.BRANCH.

Specifically, in an environment without /usr/bin/wget, such as the
launchpad build environment, the absence of /usr/bin/wget should not
cause any build to fail, as is currently the case:

    make[1]: *** No rule to make target '/usr/bin/wget', needed by
    'Makefile.common.v0.27'.  Stop.

(For example:
https://launchpad.net/~project-calico/+archive/ubuntu/master/+build/18156203)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,7 +41,7 @@ WGET?=/usr/bin/wget
 
 Makefile.common: Makefile.common.$(MAKE_BRANCH)
 	cp "$<" "$@"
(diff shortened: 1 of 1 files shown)


neiljerram:
Superseded by #2185

PR #2185, opened 2019-11-25
Switch Makefile.common from wget to curl, remove wget dependency

rafaelvanoni:


rafaelvanoni:
> LGTM, but in case this hits CI trouble, can we merge #2184 first and then do this one as a delta?

Sure, just approved #2184

neiljerram:
> Sure, just approved #2184

And then #2184 fails CI.... :smile:

neiljerram:
And this one has passed, so I guess you're good to merge this and close #2184

PR #2189, opened 2019-12-11
Use iptables for ipv4 RPF check

### LINKED ISSUE #2082: linux/dataplane: rp_filter default value
Linux dataplane logic currently contains the following: https://github.com/projectcalico/felix/blob/56335941df75e6f88a4105cef2c0ab0902d2f588/dataplane/linux/int_dataplane.go#L999-L1000

However, starting with systemd v240, the default for systemd-based distributions is `rp_filter = 2`, see https://github.com/systemd/systemd/blob/v240/sysctl.d/50-default.conf#L24-L25.

lmm:
## Description

This PR replaces the ipv4 RPF sysctl handling with iptables rules.

Fixes #2082

### Pending
I'll remove `IgnoreLooseRPF` in a separate set of PRs (will need to remove from libcalico-go, then from felix).
## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```


rules/static.go
```diff
@@ -827,19 +827,20 @@ func (r *DefaultRuleRenderer) StaticRawPreroutingChain(ipVersion uint8) *Chain {
...
-		})
-	}
+	// Apply strict RPF check to packets from workload interfaces.  This prevents
+	// workloads from spoofing their IPs.  Note: non-privileged containers can't
+	// usually spoof but privileged containers and VMs can.
+	// But we also need to allow DHCP discover requests to be sent from the host.
+	// These DHCP requests will have source IP "0.0.0.0" and would otherwise be
+	// dropped by our RPF check rule.
+	rules = append(rules, Rule{
+		Match:  Match().SourceNet("0.0.0.0/0").DestNet("255.255.255.255").Protocol("udp").SourcePorts(68).DestPorts(67),
```
lmm: The FV's are failing because the iptables is complaining: `host/network `255.255.255.255' not found`
Does anyone see what I'm missing?

lmm: Nevermind, this isn't needed - we already have exceptions for OpenStack DHCP

fv/spoof_test.go
```diff
@@ -0,0 +1,98 @@
...
+			w[ii] = workload.Run(felixes[ii], wName, "default", wIP, "8055", protocol)
+			w[ii].ConfigureInDatastore(infra)
+		}
+
+		cc = &workload.ConnectivityChecker{Protocol: protocol}
+	}
+
+	It("should drop udp traffic that has had its IP spoofed", func() {
+		setupWorkloadsAndConnectivityChecker("udp")
+		felixes[0].Exec("iptables", "-t", "nat", "-A", "POSTROUTING", "-p", "udp", "-j", "SNAT", "--to-source", "10.65.3.2")
```
neiljerram: Isn't this a bit different from the *pod* sending from an IP that isn't its own?

neiljerram:
Also just noticed there's an `IgnoreLooseRPF` config field - what are we doing with that?

lmm:
> Also just noticed there's an IgnoreLooseRPF config field - what are we doing with that?

I'm going to get PRs to remove it from libcalico-go, then from felix.

fv/spoof_test.go
lmm: Right, yeah this isn't quite the same. I took this from the OpenStack tests which are going from VM to VM.

lmm:
> > Also just noticed there's an IgnoreLooseRPF config field - what are we doing with that?
> 
> I'm going to get PRs to remove it from libcalico-go, then from felix.

Actually, I think it's easier to just remove it now and add a temporary workaround to the tests that depend on the field of FelixConfiguration: 7a9408556460b2420294493c7531873858b10882

fv/test-connection/test-connection.go
```diff
@@ -78,7 +82,15 @@ func main() {
 	ipAddress := arguments["<ip-address>"].(string)
 	port := arguments["<port>"].(string)
 	sourcePort := arguments["--source-port"].(string)
-	log.Infof("Test connection from %v:%v to IP %v port %v", namespacePath, sourcePort, ipAddress, port)
+	// Set default for source IP.
+	sourceIpAddress := defaultSourceIP
+	if strings.Contains(ipAddress, ":") {
+		sourceIpAddress = "::"
+	}
+	if srcIP, ok := arguments["--source-ip"].(string); ok {
```
spikecurtis: I *think* that because you have `[default 0.0.0.0]` in the usage, you'll get `0.0.0.0` for `arguments["--source-ip"]` even if it wasn't on the command line.  That means this if-test always fires, and you klobber the IPv6 default of `::`.

However, I'm also pretty sure that you can use `::` as the source address even if you're connecting to an IPv4 remote and Linux will pick a local IPv4 address on the correct interface automatically.  So the upshot is that you can simplify all this defaulting code by setting `[default: ::]` in the usage and then just

```
sourceIpAddress := arguments["--source-ip"].(string)
```

like most of the other args do.

fv/test-connection/test-connection.go
```diff
@@ -119,7 +140,16 @@ func main() {
 	}
 }
 
-func tryConnect(ipAddress, port, sourcePort, protocol, loopFile string) error {
+func maybeAddInterface(sourceIP string) error {
```
spikecurtis: this function is poorly named, it should be `maybeAddAddr` (or `maybeAddAddrToInterface` if you want)

fv/test-connection/test-connection.go
```diff
@@ -119,7 +140,16 @@ func main() {
...
+	var err error
+	if sourceIP != defaultSourceIP {
+		cmd := exec.Command("ip", "addr", "add", sourceIP+"/32", "dev", "eth0")
```
spikecurtis: This won't work for IPv6 since you hardcode `/32`

fv/spoof_test.go
```diff
@@ -0,0 +1,111 @@
...
+		It("should drop udp traffic that has had its IP spoofed", func() {
+			setupWorkloadsAndConnectivityChecker("udp")
+			// Setup a spoofed workload. w[0] has the IP 10.65.0.2.
+			// Make it use 10.65.2.2 to test connections instead.
+			spoofed := &workload.SpoofedWorkload{
+				Workload:        w[0],
+				SpoofedSourceIP: "10.65.2.2",
+			}
+			cc.ExpectNone(spoofed, w[1])
+			cc.ExpectSome(w[1], w[0])
```
spikecurtis: I'd have expected `cc.ExpectSome(w[2], w[1])` --- that is to say, in this test w[0] is attempting to spoof w[2]'s IP address, so we should verify that the _real_ w[2] can access w[1].

fv/spoof_test.go
```diff
@@ -0,0 +1,111 @@
...
+			cc.CheckConnectivity()
+		})
```
spikecurtis: I'd like to see a test where a workload "Spoofs" its own, legit IP address and succeeds. Otherwise we could have a bug where the spoofing code in `test-connection` is bugged but we never notice because we only ever expect spoofed connections to fail.

fv/spoof_test.go
```diff
@@ -0,0 +1,111 @@
...
+var _ = infrastructure.DatastoreDescribe("spoof tests", []apiconfig.DatastoreType{apiconfig.EtcdV3}, func(getInfra infrastructure.InfraFactory) {
+
+	var (
+		infra   infrastructure.DatastoreInfra
+		felixes []*infrastructure.Felix
+		w       [3]*workload.Workload
+		cc      *workload.ConnectivityChecker
+	)
+
+	Context("using IPv4", func() {
```
spikecurtis: no IPv6 equivalents?

fv/spoof_test.go.temp
```diff
@@ -0,0 +1,193 @@
+// +build fvtests
```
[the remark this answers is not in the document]
lmm: Thanks, bcc7757335706334998224fd0eecc393c165daf3

fv/test-connection/test-connection.go
lmm: Thanks, I fixed this up a bit: bb800ef41057daaa8c7e9403fc2b67e73d63e4c9

I tried using `::` as the default but I ended up getting an error in `net.ResolveTCPAddr`:

```
level=info msg=\"Resolved tcp addr\" addr= resolved=\":0\"\npanic: address 10.65.0.2:8055: unexpected address type
```

It looks like the local address (`::`) for some reason was not parsable as an addr? https://github.com/golang/go/blob/master/src/net/dial.go#L592
I'm not really sure what's going on but I can go back and take a look later.

fv/test-connection/test-connection.go
lmm: 46acbf95f898593f230a8cb4762973055aecb9af

fv/test-connection/test-connection.go
lmm: Thanks, 46acbf95f898593f230a8cb4762973055aecb9af

fv/spoof_test.go
```diff
@@ -0,0 +1,196 @@
...
+var _ = infrastructure.DatastoreDescribe("spoof tests", []apiconfig.DatastoreType{apiconfig.EtcdV3}, func(getInfra infrastructure.InfraFactory) {
+
+	var (
+		infra   infrastructure.DatastoreInfra
+		felixes []*infrastructure.Felix
+		w       [3]*workload.Workload
+		cc      *workload.ConnectivityChecker
+	)
+
+	/*
```
neiljerram: Presumably this commenting shouldn't be here?

fv/spoof_test.go
```diff
@@ -0,0 +1,196 @@
...
+				wIP := fmt.Sprintf("fdc6:3dbc:e983:cbc0::%d", ii)
+				wName := fmt.Sprintf("w%d", ii)
+				w[ii] = workload.Run(felixes[ii], wName, "default", wIP, "8055", protocol)
+				w[ii].ConfigureInDatastore(infra)
+			}
+
+			cc = &workload.ConnectivityChecker{Protocol: protocol}
+		}
+
+		/*
```
neiljerram: Similarly here.

fv/spoof_test.go
lmm: Added

fv/spoof_test.go
lmm: Agreed, added

fv/spoof_test.go
lmm: Fixed with the latest push

fv/spoof_test.go
lmm: Updated 🙂

fv/spoof_test.go
```diff
@@ -0,0 +1,162 @@
...
+			infra, err = infrastructure.GetEtcdDatastoreInfra()
+			Expect(err).NotTo(HaveOccurred())
+			opts := infrastructure.DefaultTopologyOptions()
+			// For IPv6 we'll setup our workloads on a single felix.
+			felixes, _ = infrastructure.StartNNodeTopology(1, opts, infra)
+			// Install a default profile allowing all ingress and egress,
+			// in the absence of policy.
+			infra.AddDefaultAllow()
+
+			createTopology(true)
```
neiljerram: a little tidy up here - not sure createTopology is used here.

fv/spoof_test.go
```diff
@@ -0,0 +1,162 @@
...
+		spoofTests()
+	})
+
+	Context("IPv6", func() {
+		BeforeEach(func() {
+			var err error
+			infra, err = infrastructure.GetEtcdDatastoreInfra()
+			Expect(err).NotTo(HaveOccurred())
+			opts := infrastructure.DefaultTopologyOptions()
+			// For IPv6 we'll setup our workloads on a single felix.
```
neiljerram: Maybe beef up this single Felix comment a bit more.

PR #2191, opened 2019-12-13
Use go-build v0.30, to revert to go 1.12

neiljerram:
Because of https://github.com/golang/go/issues/34773


tmjd:
Merging because CLA seems to be confused.

PR #2192, opened 2019-12-13
SCTP ports & FVs

spikecurtis:
Signed-off-by: Spike Curtis <spike@tigera.io>

## Description

Adds support for SCTP ports in both policy and workload endpoint named ports.

## Todos
- [x] Integration tests (delete as appropriate) Done
- [ ] Documentation
- [x] Backport not needed
- [x] Release note

## Release Note

```release-note
Support for SCTP ports
```

### DIFF
--- a/fv/test-connection/test-connection.go
+++ b/fv/test-connection/test-connection.go
@@ -20,13 +20,17 @@ import (
 	"fmt"
 	"net"
 	"os"
+	"strconv"
 	"strings"
+	"syscall"
 	"time"
 
 	"github.com/containernetworking/cni/pkg/ns"
 	docopt "github.com/docopt/docopt-go"
+	"github.com/ishidawataru/sctp"
 	reuse "github.com/libp2p/go-reuseport"
 	log "github.com/sirupsen/logrus"
+	"golang.org/x/sys/unix"
 
 	uuid "github.com/satori/go.uuid"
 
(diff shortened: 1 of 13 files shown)


fv/test-workload/test-workload.go
```diff
@@ -302,6 +305,7 @@ func main() {
 			}
```
lmm: I can't figure out how to add review comments outside of the diff here... but, just above here, can we update the comment: `// Listen on each port for either TCP or UDP.` ?

fv/test-connection/test-connection.go
```diff
@@ -126,7 +130,10 @@ func tryConnect(ipAddress, port, sourcePort, protocol, loopFile string) error {
 	testMessage := "hello," + uid
 
 	// The reuse library implements a version of net.Dialer that can reuse UDP/TCP ports, which we
-	// need in order to make connection retries work.
+	// need in order to make connection retries work. Since we specify the source port
+	// rather than use an ephemeral port, if the SO_REUSEADDR and SO_REUSEPORT options
+	// are not set, when we make another call to this program, the original port is in post-close
+	// wait state and bind fails.
```
lmm: I don't know much about the reuse library or sockets but it looks like this comment might just apply to sctp, and if so this would fit better where we're doing the sctp dialing.

spikecurtis: The reuse library is for TCP/UDP, but the bit about the SO_REUSE* socket options applies to all connection types.

spikecurtis: I've updated this comment to hopefully make this more clear

spikecurtis:
@lmm thanks for the comments. I believe I have addressed them all and this is ready for another review

PR #2195, opened 2019-12-18
Attempt to modprobe nf_conntrack_proto_sctp

spikecurtis:
Signed-off-by: Spike Curtis <spike@tigera.io>

## Description

On some kernels, nf_conntrack_proto_sctp is a loadable kernel module, without which, SCTP connections will fail due to conntrack marking all SCTP packets invalid.  This fix causes Felix to attempt to load the module via `modprobe`.  `modprobe` should be available in the Felix container since `iptables` needs it to load kernel modules as needed (note that this is a conntrack kernel module, so iptables will not automatically load it for us).

## Todos
- [x] Integration tests (delete as appropriate) already covered
- [ ] Documentation TBD
- [ ] Backport TBD
- [x] Release note

## Release Note

```release-note
Felix will attempt to modprobe nf_conntrack_proto_sctp
```

### DIFF
--- a/dataplane/linux/modprobe.go
+++ b/dataplane/linux/modprobe.go
@@ -1,0 +1,57 @@
+// Copyright (c) 2019 Tigera, Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package intdataplane
+
+import "os/exec"
+
+const (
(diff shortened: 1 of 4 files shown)


dataplane/linux/modprobe.go
```diff
@@ -0,0 +1,40 @@
...
+	// Kernel module needed for SCTP protocol support on some kernels
+	moduleConntrackSCTP = "nf_conntrack_proto_sctp"
+)
+
+type modProbe struct {
+	module string
+	newCmd iptables.CmdFactory
+}
+
+func newModProbe(module string, newCmd iptables.CmdFactory) modProbe {
```
fasaxc: I think the go idiom would be to define a smaller interface here since you're not using most of the methods of the iptables command interface

spikecurtis: "A little bit of copying is better than a lot of dependency..."

I've defined a new, smaller interface in modprobe.go

spikecurtis:
We have a test of the modprobe function in that Semaphore will fail on the SCTP FVs without the modprobe.  It's currently in the Semaphore build commands and I will remove it once this is merged.

Longer term, Semaphore might eventually move to a newer kernel that doesn't need the modprobe, in which case it might be nice to have SCTP E2Es that we can run on some older kernels, but not planning to build any today.

PR #2215, opened 2020-02-06
Report liveness from various loops.

fasaxc:
## Description

In corner cases, the calculation graph and internal dataplane can take a long time to process some updates. Report liveness from their various loops.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/calc/async_calc_graph.go
+++ b/calc/async_calc_graph.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2016-2018 Tigera, Inc. All rights reserved.
+// Copyright (c) 2016-2020 Tigera, Inc. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -21,10 +21,11 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	log "github.com/sirupsen/logrus"
 
-	"github.com/projectcalico/felix/config"
-	"github.com/projectcalico/felix/proto"
 	"github.com/projectcalico/libcalico-go/lib/backend/api"
 	"github.com/projectcalico/libcalico-go/lib/health"
+
+	"github.com/projectcalico/felix/config"
+	"github.com/projectcalico/felix/proto"
 )
 
 const (
@@ -83,7 +84,6 @@ type AsyncCalcGraph struct {
 	dirty            bool
 
 	debugHangC <-chan time.Time
-	lastReport time.Time
 }
 
 const (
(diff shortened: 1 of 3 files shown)


calc/async_calc_graph.go
```diff
@@ -149,6 +150,9 @@ func (acg *AsyncCalcGraph) loop() {
 					typeName := reflect.TypeOf(upd.Key).Name()
 					count := countUpdatesProcessed.WithLabelValues(typeName)
 					count.Inc()
+					if time.Since(acg.lastReport) > healthInterval {
```
neiljerram: Is the elapsed time check really useful here?  It would be simpler to call `acg.reportHealth` unconditionally, so should we just do that?

dataplane/linux/int_dataplane.go
```diff
@@ -268,6 +269,7 @@ func NewIntDataplaneDriver(config Config) *InternalDataplane {
 		LockProbeInterval:     config.IptablesLockProbeInterval,
 		BackendMode:           backendMode,
 		LookPathOverride:      config.LookPathOverride,
+		OnStillAlive:          dp.reportHealth,
```
neiljerram: Just a language point check: is `dp.reportHealth` valid as soon as `dp` has been constructed?  (I'm not sure why it wouldn't be, but just being paranoid.)

dataplane/linux/int_dataplane.go
```diff
@@ -1077,6 +1081,7 @@ func (d *InternalDataplane) apply() {
 		ipSetsWG.Add(1)
 		go func(ipSets *ipsets.IPSets) {
 			ipSets.ApplyUpdates()
+			d.reportHealth()
```
neiljerram: Will this be thread-safe?
At this level:
```
	if d.config.HealthAggregator != nil {
		d.config.HealthAggregator.Report(
			healthName,
			&health.HealthReport{Live: true, Ready: d.doneFirstApply},
		)
	}
```
`d.config.HealthAggregator` should be OK because only set at start of day.  But maybe `d.doneFirstApply` could be changing in another goroutine, and we might need to delve further inside `d.config.HealthAggregator.Report`.

neiljerram: Similarly the next 3 changes are all in goroutines.

calc/async_calc_graph.go
fasaxc: Thinking was that it might cause a lot of log spam from the health aggregator?

dataplane/linux/int_dataplane.go
fasaxc: I think it's correct but I wish it was a bit simpler:

- `d.config.HealthAggregator` is start-of day config
- `doneFirstApply` only changes on the main goroutine and the goroutines from which I'm calling `reportHealth()` are short-lived: the main goroutine starts them and then waits for them to finish before continuing.
- `Report()` itself uses a mutex to protect the HealthAggregator.

dataplane/linux/int_dataplane.go
fasaxc: Yes, I think that's fine in go.

calc/async_calc_graph.go
neiljerram: I think the aggregator report-time logging is all debug, and it might be better if needed to handle any log rate limiting inside the aggregator?  I think I'd prefer to err initially on the side of over-reporting from Felix.

dataplane/linux/int_dataplane.go
neiljerram: OK, thanks for the reassurance.

dataplane/linux/int_dataplane.go
neiljerram: Thanks.

calc/async_calc_graph.go
fasaxc: Benchmarked the report function at 500ns; that's probably OK.  Removing the logCtx from the report function knocks it down to 90ns in case we need to optimise that in future.

fasaxc:
Should we also simply bump up the liveness timeout?

calc/async_calc_graph.go
```diff
@@ -141,14 +142,18 @@ func (acg *AsyncCalcGraph) loop() {
 			case []api.Update:
 				// Update; send it to the dispatcher.
 				log.Debug("Pulled []KVPair off channel")
-				updStartTime := time.Now()
-				acg.AllUpdDispatcher.OnUpdates(update)
-				summaryUpdateTime.Observe(time.Since(updStartTime).Seconds())
-				// Record stats for the number of messages processed.
-				for _, upd := range update {
+				for i, upd := range update {
```
fasaxc: @neiljerram I realised I had put the call to reportHealth() in the wrong place.  Needed to split up the calls to OnUpdates instead.

PR #2220, opened 2020-02-25
Felix doesn't push all routes to dataplane when using VXLAN with Cros…

rafaelvanoni:
Proposed fix for https://github.com/projectcalico/calico/issues/3231 as suggested in comments.

### DIFF
--- a/calc/states_for_test.go
+++ b/calc/states_for_test.go
@@ -930,7 +930,7 @@ var vxlanWithMAC = vxlanWithBlock.withKVUpdates(
 // other node.
 var vxlanWithBlockAndBorrows = vxlanWithBlock.withKVUpdates(
 	KVPair{Key: remoteIPAMBlockKey, Value: &remoteIPAMBlockWithBorrows},
-	KVPair{Key: remoteHost2IPKey, Value: &remoteHost2IP},
+	KVPair{Key: remoteHostIP2Key, Value: &remoteHostIP2},
 	KVPair{Key: remoteHost2VXLANTunnelConfigKey, Value: remoteHost2VXLANTunnelIP},
 ).withName("VXLAN borrow").withVTEPs(
 	proto.VXLANTunnelEndpointUpdate{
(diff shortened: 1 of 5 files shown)


caseydavenport:
@rafaelvanoni nice, this looks like the right fix from what I can tell. I think we'll also want an FV or UT for the scenario as well.

We've got some good control over ordering of events in the calc FV tests, so that's probably where I'd put these. Here's where the other VXLAN calc tests are: https://github.com/projectcalico/felix/blob/master/calc/calc_graph_fv_test.go#L246

calc/states_for_test.go
```diff
@@ -1152,6 +1152,43 @@ var vxlanTunnelIPDelete = vxlanWithBlock.withKVUpdates(
...
+	},
+).withRoutes(
+	proto.RouteUpdate{
+		Type: proto.RouteType_WORKLOADS_NODE,
+		Dst:  "10.0.1.0/29",
+		Node: remoteHostname,
+		Gw:   remoteHostIP.String(),
+	},
+	proto.RouteUpdate{
+		Type: proto.RouteType_VXLAN,
```
caseydavenport: Hm, I think for cross-subnet we want this to be NO_ENCAP. 

Are the two test hosts in the same subnet?

calc/states_for_test.go
```diff
@@ -1152,6 +1152,43 @@ var vxlanTunnelIPDelete = vxlanWithBlock.withKVUpdates(
 	},
 ).withVTEPs()
 
+// Two nodes with VXLAN and CrossSubnet.
+var vxlanWithCrossSubnet = empty.withKVUpdates(
+	KVPair{Key: ipPoolKey, Value: &ipPoolWithVXLANCrossSubnet},
+	KVPair{Key: localIPAMBlockKey, Value: &localIPAMBlock},
+	KVPair{Key: localHostIPKey, Value: &localHostIP},
```
caseydavenport: I think we want two states in order to simulate the original issue:

State 1:
 - Remote node IP and tunnel address
 - Remote node IPAM block
 - IP pool

State 2 (in addition to state 1):
- Local node IP and tunnel address

The way it's formulated right now, the test sends all of these in as a single state transition (vxlanWithCrossSubnet includes all of the objects in both state1 and state1), and so we can't guarantee the ordering in the original issue.

calc/states_for_test.go
rafaelvanoni: They're in different ones, I added the CIDRs as comments next to the test to make this clearer.

calc/states_for_test.go
rafaelvanoni: Ok, updated.

PR #2225, opened 2020-03-03
Fix misspellings reported by goreport

neiljerram:


neiljerram:
I've reported the remaining false positives at https://github.com/client9/misspell/issues/170

PR #2228, opened 2020-03-05
Normal policy support for host-* endpoints

lmm:
## Description

`*` hostendpoints currently only support pre-DNAT global network policy. This PR will (eventually) add normal policy support for `*` hostendpoints.

Some notes on semantics:
* If a `*` hostendpoint and a "named" (e.g. `eth0`) hostendpoint exist, the named hostendpoint will be used. (This is existing behaviour and isn't changing here.)
* Host-> pod traffic is always allowed. With this change, we would continue to have default-allow semantics. (**This desired behaviour to be confirmed**)

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
all-interfaces host endpoints now supports normal network policy in addition to pre-dnat policy
```

### DIFF
--- a/fv/apply_on_forward_test.go
+++ b/fv/apply_on_forward_test.go
@@ -120,7 +120,7 @@ var _ = infrastructure.DatastoreDescribe("apply on forward tests; with 2 nodes",
 
 	// The following tests verify that a HostEndpoint does not block forwarded traffic
 	// when there is no applyOnForward policy that applies to that HostEndpoint.  We
-	// create a HostEndpoint for eth0 on two hosts (A and B) and then test two cases:
+	// create a HostEndpoint two hosts (A and B) and then test two cases:
 	//
 	// 1. Workload on host A -> Workload on host B.  In this case, the traffic is
 	// forwarded on both hosts.
@@ -139,7 +139,7 @@ var _ = infrastructure.DatastoreDescribe("apply on forward tests; with 2 nodes",
 			cancel context.CancelFunc
 		)
 
-		Context("with named hostendpoints", func() {
+		Context("with named host endpoints on eth0", func() {
 			BeforeEach(func() {
 				addAllowAllToHostEndpoints()
 
@@ -246,15 +246,15 @@ var _ = infrastructure.DatastoreDescribe("apply on forward tests; with 2 nodes",
 
 				// Forwarded traffic is unaffected by the deny-all policy.
 				cc.ExpectSome(w[0], w[1])
-				//cc.ExpectSome(w[1], w[0])
+				cc.ExpectSome(w[1], w[0])
 
 				// Traffic originating/terminating on the host is policed by
 				// the deny-all policy.
-				//cc.ExpectNone(w[0], hostW[1])
-				//cc.ExpectNone(w[1], hostW[0])
-				//cc.ExpectNone(hostW[0], w[1])
-				//cc.ExpectNone(hostW[1], w[0])
-				//cc.ExpectNone(hostW[0], hostW[1])
+				cc.ExpectNone(w[0], hostW[1])
+				cc.ExpectNone(w[1], hostW[0])
+				cc.ExpectNone(hostW[0], w[1])
+				cc.ExpectNone(hostW[1], w[0])
+				cc.ExpectNone(hostW[0], hostW[1])
 				cc.CheckConnectivity()
 			})
 		})
--- a/fv/apply_on_forward_test.go
+++ b/fv/apply_on_forward_test.go
@@ -170,31 +170,6 @@ var _ = infrastructure.DatastoreDescribe("apply on forward tests; with 2 nodes",
 			})
 
 			expectWorkloadToWorkloadAndHostConnectivity()
-
-			It("should not block forwarded traffic if normal policy selects the hostendpoint", func() {
-				policy := api.NewGlobalNetworkPolicy()
-				policy.Name = "deny-normal"
-				order := float64(1)
-				policy.Spec.Order = &order
-				policy.Spec.Selector = "host-endpoint=='true'"
-				policy.Spec.Ingress = []api.Rule{{Action: api.Deny}}
-				policy.Spec.Egress = []api.Rule{{Action: api.Deny}}
-				_, err := client.GlobalNetworkPolicies().Create(utils.Ctx, policy, utils.NoOptions)
-				Expect(err).NotTo(HaveOccurred())
-
-				// Forwarded traffic is unaffected by the deny-all policy.
-				cc.ExpectSome(w[0], w[1])
-				cc.ExpectSome(w[1], w[0])
-
-				// Traffic originating/terminating on the host is policed by
-				// the deny-all policy.
-				cc.ExpectNone(w[0], hostW[1])
-				cc.ExpectNone(w[1], hostW[0])
-				cc.ExpectNone(hostW[0], w[1])
-				cc.ExpectNone(hostW[1], w[0])
-				cc.ExpectNone(hostW[0], hostW[1])
-				cc.CheckConnectivity()
-			})
 		})
 
 		Context("with all-interfaces host endpoints", func() {
@@ -230,31 +205,6 @@ var _ = infrastructure.DatastoreDescribe("apply on forward tests; with 2 nodes",
 			})
 
 			expectWorkloadToWorkloadAndHostConnectivity()
-
-			It("should not block forwarded traffic if normal policy selects the hostendpoint", func() {
-				order := float64(1)
-				policy := api.NewGlobalNetworkPolicy()
-				policy.Name = "deny-normal"
-				policy.Spec.Selector = "host-endpoint=='true'"
-				policy.Spec.Order = &order
-				policy.Spec.Ingress = []api.Rule{{Action: api.Deny}}
-				policy.Spec.Egress = []api.Rule{{Action: api.Deny}}
-				_, err := client.GlobalNetworkPolicies().Create(utils.Ctx, policy, utils.NoOptions)
-				Expect(err).NotTo(HaveOccurred())
-
-				// Forwarded traffic is unaffected by the deny-all policy.
-				cc.ExpectSome(w[0], w[1])
-				cc.ExpectSome(w[1], w[0])
-
-				// Traffic originating/terminating on the host is policed by
-				// the deny-all policy.
-				cc.ExpectNone(w[0], hostW[1])
-				cc.ExpectNone(w[1], hostW[0])
-				cc.ExpectNone(hostW[0], w[1])
-				cc.ExpectNone(hostW[1], w[0])
-				cc.ExpectNone(hostW[0], hostW[1])
-				cc.CheckConnectivity()
-			})
 		})
 	})
 })
--- a/fv/apply_on_forward_test.go
+++ b/fv/apply_on_forward_test.go
@@ -55,7 +55,7 @@ var _ = infrastructure.DatastoreDescribe("apply on forward tests; with 2 nodes",
 		felixes, client = infrastructure.StartNNodeTopology(2, options, infra)
 
 		// Install a default profile that allows all ingress and egress, in the absence of any Policy.
-		_ = infra.AddDefaultAllow()
+		infra.AddDefaultAllow()
 
 		// Create workloads, using that profile.  One on each "host".
 		for ii := range w {
--- a/fv/apply_on_forward_test.go
+++ b/fv/apply_on_forward_test.go
@@ -174,8 +174,7 @@ var _ = infrastructure.DatastoreDescribe("apply on forward tests; with 2 nodes",
 			It("should not block forwarded traffic if normal policy selects the hostendpoint", func() {
 				policy := api.NewGlobalNetworkPolicy()
 				policy.Name = "deny-normal"
-				var order float64
-				order = 1
+				order := float64(1)
 				policy.Spec.Order = &order
 				policy.Spec.Selector = "host-endpoint=='true'"
 				policy.Spec.Ingress = []api.Rule{{Action: api.Deny}}
@@ -233,8 +232,7 @@ var _ = infrastructure.DatastoreDescribe("apply on forward tests; with 2 nodes",
 			expectWorkloadToWorkloadAndHostConnectivity()
 
 			It("should not block forwarded traffic if normal policy selects the hostendpoint", func() {
-				var order float64
-				order = 1
+				order := float64(1)
 				policy := api.NewGlobalNetworkPolicy()
 				policy.Name = "deny-normal"
 				policy.Spec.Selector = "host-endpoint=='true'"
--- a/fv/apply_on_forward_test.go
+++ b/fv/apply_on_forward_test.go
@@ -55,7 +55,7 @@ var _ = infrastructure.DatastoreDescribe("apply on forward tests; with 2 nodes",
 		felixes, client = infrastructure.StartNNodeTopology(2, options, infra)
 
 		// Install a default profile that allows all ingress and egress, in the absence of any Policy.
-		infra.AddDefaultAllow()
+		_ = infra.AddDefaultAllow()
 
 		// Create workloads, using that profile.  One on each "host".
 		for ii := range w {
--- a/fv/apply_on_forward_test.go
+++ b/fv/apply_on_forward_test.go
@@ -29,6 +29,7 @@ import (
 	"github.com/projectcalico/felix/fv/infrastructure"
 	"github.com/projectcalico/felix/fv/utils"
 	"github.com/projectcalico/felix/fv/workload"
+	"github.com/projectcalico/felix/rules"
 	"github.com/projectcalico/libcalico-go/lib/apiconfig"
 	api "github.com/projectcalico/libcalico-go/lib/apis/v3"
 	client "github.com/projectcalico/libcalico-go/lib/clientv3"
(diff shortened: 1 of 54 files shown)


rules/dispatch.go
```diff
@@ -124,15 +125,17 @@ func (r *DefaultRuleRenderer) hostDispatchChains(
...
+		// interface in the `endpoints` map.
 		fromEndRules = []Rule{
 			Rule{
-				Action: GotoAction{Target: defaultFromChainName},
+				Action: GotoAction{Target: defaultChainName},
+			},
+		}
+		toEndRules = []Rule{
+			Rule{
+				Action: GotoAction{Target: defaultChainName},
```
lmm: @neiljerram we chatted about adding `ToHostDispatchChains` but I wasn't sure if that was necessary. I'm probably wrong but doesn't just adding `toEndRules` with the all-interfaces endpoint chain as the end rule work?

rules/dispatch_test.go
```diff
@@ -685,7 +685,7 @@ var _ = Describe("Dispatch chains", func() {
 				func(names []string, expectedChains []*iptables.Chain) {
 					input := convertToInput(names, expectedChains)
 					// Note: order of chains and rules should be deterministic.
-					Expect(renderer.HostDispatchChains(input, true)).To(Equal(expectedChains))
+					Expect(renderer.HostDispatchChains(input, "", true)).To(Equal(expectedChains))
```
fasaxc: Should have next tests here to verify the non `""` case

rules/endpoints.go
```diff
@@ -270,6 +280,7 @@ func (r *DefaultRuleRenderer) endpointIptablesChain(
 	adminUp bool,
 	allowAction Action,
 	dropEncap bool,
+	defaultDropWhenNoPolicy bool,
```
fasaxc: This method is getting unwieldy with the number of flag options.  Using an args struct might be cleaner at this point

rules/endpoints.go
```diff
@@ -403,16 +414,31 @@ func (r *DefaultRuleRenderer) endpointIptablesChain(
...
-			Action:  DropAction{},
-			Comment: []string{"Drop if no profiles matched"},
-		})
+		// For untracked and pre-DNAT rules, we don't do that because there may be tracked
+		// rules still to be applied to the packet in the filter table.
+		//
+		// For the wildcard HEP, we allow when there is no policy, because wildcard HEPs
+		// were previously implemented only for pre-DNAT policy and so had no interaction
+		// with normal policy.
+		if defaultDropWhenNoPolicy {
```
fasaxc: Traffic will fall through to these rules if there is policy but it doesn't match.  Is that what we want?  I was expecting that we'd only want to do default-allow if there is no policy at all.

fv/vxlan_test.go
```diff
@@ -218,6 +218,14 @@ var _ = infrastructure.DatastoreDescribe("VXLAN topology before adding host IPs
 					policy.Spec.Selector = "has(host-endpoint)"
 					_, err = client.GlobalNetworkPolicies().Create(utils.Ctx, policy, utils.NoOptions)
 					Expect(err).NotTo(HaveOccurred())
+
```
fasaxc: Why these changes to existing tests?

rules/endpoints.go
fasaxc: (But please check my expectation since this epic has been pretty fast-moving so I may be out-of-date!)

fv/vxlan_test.go
neiljerram: See the full discussion about this in Slack at https://tigera.slack.com/archives/GUEHC5CKA/p1584559861114600 , which I kicked off by writing...

"Think I just found an issue of incompatibility w.r.t. existing behaviour; exposed by FV tests at https://github.com/projectcalico/felix/blob/master/fv/vxlan_test.go#L189 and https://github.com/projectcalico/felix/blob/master/fv/ipip_test.go#L176.  These tests have:

- a normal GNP policy defined, to allow egress to a particular IP, that will apply to host-endpoint=='true'
- WHEP defined on each node, with host-endpoint: true
- pre-DNAT ingress allow-all policy defined, applying to has(host-endpoint). 

Current behaviour:

- normal policy not implemented for WHEPs, so the normal GNP policy has no effect, and traffic between hosts is allowed.

Behaviour with WIP:

- the normal GNP policy now blocks all egress traffic from host to {other host or workload on other host}

1. Do you agree that this is the correct new behaviour, according to what we have so far agreed?
2. Should we perhaps specify something more explicit, to control whether existing normal and AOF policy starts applying to a WHEP?
"

Only @spikecurtis replied, and his view was that new behaviour is correct, and that we shouldn't add any knobs for precise back-compatibility with existing configuration.

rules/dispatch_test.go
neiljerram: Presume you mean **new** tests?

rules/endpoints.go
neiljerram: I think your expectation is right.  But I think your interpretation of what the code does is wrong, because of "Drop if no policies passed packet" just above (https://github.com/projectcalico/felix/pull/2228/files#diff-0370ffa1b8a68941ac13794784606c0aL374).

neiljerram: (It's a separate point, but note that we rather awkwardly have two different approaches for the no-policy-at-all case.  For some types of policy it's handled by not calling into the endpoint chain at all.  For others it's handled by code like here and at https://github.com/projectcalico/felix/pull/2228/files#diff-0370ffa1b8a68941ac13794784606c0aL378.)

neiljerram: ((To be clear, that awkwardness ^ was not introduced by this PR.))

fasaxc: OK, I see that rule now; but what about profiles?  Should we apply the same logic there?  I.e. if there are some profiles then should that imply default drop?

rules/dispatch.go
```diff
@@ -169,8 +170,8 @@ func (r *DefaultRuleRenderer) hostDispatchChains(
 			"",
 			ChainDispatchFromHostEndpoint,
 			"",
-			fromEndRules,
-			toEndRules,
+			fromEndForwardRules,
+			toEndForwardRules,
```
neiljerram: I think this specific change is wrong, because we use `HostFromEndpointPfx` in `HostEndpointToMangleChains`, for pre-DNAT policy, not `HostFromEndpointForwardPfx`.  But the rest of this commit looks spot on.  (I missed that a single HostDispatchChains call is responsible for creating the dispatch chains for both normal and AOF policy.)

rules/endpoints.go
lmm: Is something like this what you had in mind? Maybe there's a better way of doing this
```
func (r *DefaultRuleRenderer) HostEndpointToRawChains(
	ifaceName string,
	ingressPolicyNames []string,
	egressPolicyNames []string,
) []*Chain {
	log.WithField("ifaceName", ifaceName).Debug("Rendering raw (untracked) host endpoint chain.")

	toOptions := &endpointIptablesChainOptions{
		ifaceName,
		PolicyOutboundPfx,
		ProfileOutboundPfx,
		HostToEndpointPfx,
		ChainFailsafeOut,
		chainTypeUntracked,
		true, // Host endpoints are always admin up.
		AcceptAction{},
		dontDropEncap,
		false,
	}
	// Chain for traffic _to_ the endpoint.
	chainTrafficToEndpoint := r.endpointIptablesChain(
		egressPolicyNames,
		nil, // We don't render profiles into the raw table.
		toOptions,
	)

	fromOptions := &endpointIptablesChainOptions{
		ifaceName,
		PolicyInboundPfx,
		ProfileInboundPfx,
		HostFromEndpointPfx,
		ChainFailsafeIn,
		chainTypeUntracked,
		true, // Host endpoints are always admin up.
		AcceptAction{},
		dontDropEncap,
		false,
	}
	// Chain for traffic _from_ the endpoint.
	chainTrafficFromEndpoint := r.endpointIptablesChain(
		ingressPolicyNames,
		nil, // We don't render profiles into the raw table.
		fromOptions,
	)

	return []*Chain{
		chainTrafficToEndpoint,
		chainTrafficFromEndpoint,
	}
}
```

rules/endpoints.go
lmm: > OK, I see that rule now; but what about profiles? Should we apply the same logic there? I.e. if there are some profiles then should that imply default drop?

It seems more consistent and I think this still works with pre-dnat policy (at the very least, the existing pre-dnat tests pass). @neiljerram wdyt?

lmm: E.g., it seems that in addition to having to provide egress normal policy rules like we did https://github.com/projectcalico/felix/pull/2228/commits/7d84dd5fe8bd61d1e1455e9342750a2d2b673cd6, if we were to drop if no profiles applied as proposed, we'd also need to provide ingress policy rules on the destination host.

fasaxc: I suppose what I'm getting at is that there's a difference between:

> there are no profiles

and

> there are profiles but none of them matched the packet

I think this treats those equally, whereas, for policies, we have a default drop if there are policies but none of the match.

rules/endpoints.go
fasaxc: Yes, except the point was to use the names of the struct fields to make it clear what all the args are (and to allow for omitting args that have a sensible default)

lmm: d064081b976f9713133e160f4477926b00d0e461

rules/endpoints.go
lmm: IIUC, do you mean a change like this?
```
@@ -440,7 +440,7 @@ func (r *DefaultRuleRenderer) endpointIptablesChain(
                // with normal policy.
                if options.defaultDropWhenNoPolicy {
                        rules = append(rules, Rule{
-                               Match:   Match(),
+                               Match:   Match().MarkClear(r.IptablesMarkPass),
                                Action:  DropAction{},
                                Comment: []string{"Drop if no profiles matched"},
                        })
```

master doesn't make this distinction either: https://github.com/projectcalico/felix/blob/master/rules/endpoints.go#L411-L414

fasaxc: > master doesn't make this distinction either

Right, master drops in both cases.  Now you allow in both cases for a WHEP.  But for policies, we allow in one case and drop in the other.  I'm wondering if that's right, or if we should allow in the "no profiles" case and drop in the "some profiles but none allowed the packet" case

fv/vxlan_test.go
```diff
@@ -207,27 +205,66 @@ var _ = infrastructure.DatastoreDescribe("VXLAN topology before adding host IPs
...
+					cc.ExpectSome(felixes[0], w[0])
+					cc.ExpectSome(felixes[1], w[1])
+
+					// But the rules to allow VXLAN between our hosts let the workload traffic through.
 					cc.ExpectSome(w[0], w[1])
 					cc.ExpectSome(w[1], w[0])
 					cc.CheckConnectivity()
 				})
+
+				/*
```
fasaxc, line 229: Commented out block?

rules/endpoints.go
```diff
@@ -270,6 +279,7 @@ func (r *DefaultRuleRenderer) endpointIptablesChain(
 	adminUp bool,
 	allowAction Action,
 	dropEncap bool,
+	_ bool,
```
fasaxc, line 282: ??

rules/endpoints.go
lmm: I removed this refactoring since we ended up not using the extra function arg.

fv/apply_on_forward_test.go
```diff
@@ -95,17 +96,31 @@ var _ = infrastructure.DatastoreDescribe("apply on forward tests; with 2 nodes",
...
+		It("should have workload to workload/host connectivity", func() {
+			cc.ExpectSome(w[0], w[1])
+			cc.ExpectSome(w[1], w[0])
+			cc.ExpectSome(w[0], hostW[1])
+			cc.ExpectSome(w[1], hostW[0])
+			cc.CheckConnectivity()
+		})
+	}
+
+	expectWorkloadToWorkloadAndHostConnectivity()
```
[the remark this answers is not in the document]
lmm: 5e8d52fef3e91829db315cc9151d63c1d9cff5b2

fv/host_port_test.go
```diff
@@ -48,21 +48,24 @@ func MetricsPortReachable(felix *infrastructure.Felix) bool {
...
 //
 // - When there is no policy or endpoint configuration, the port should be reachable.
 //
 // - When there is a local workload endpoint, the port should be reachable.  (Existence of workload
 //   endpoints should make no difference to reachability to ports on the host itself.)
 //
-// - When a host endpoint is configured for the host's interface (eth0), but not yet any policy, the
-//   port should be unreachable.
+// - When a host endpoint is configured for the host's interface (eth0) or for
+//   all-interfaces, but not yet any policy, the port should be unreachable.
```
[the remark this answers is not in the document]
lmm: Reverted in 38902a1

fv/host_port_test.go
```diff
@@ -104,35 +109,43 @@ var _ = infrastructure.DatastoreDescribe("with initialized Felix", []apiconfig.D
...
-			hostEp := api.NewHostEndpoint()
-			hostEp.Name = "host-endpoint-1"
-			hostEp.Labels = map[string]string{"host-endpoint": "true"}
-			hostEp.Spec.Node = felix.Hostname
-			hostEp.Spec.InterfaceName = "eth0"
-			_, err = client.HostEndpoints().Create(utils.Ctx, hostEp, utils.NoOptions)
-			Expect(err).NotTo(HaveOccurred())
-		})
-
+	runTest := func() {
```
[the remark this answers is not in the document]
lmm: d8371d3

fv/host_port_test.go
```diff
@@ -104,35 +109,43 @@ var _ = infrastructure.DatastoreDescribe("with initialized Felix", []apiconfig.D
...
 		It("port should not be reachable", func() {
 			Eventually(metricsPortReachable, "10s", "1s").Should(BeFalse())
 		})
 
 		Context("with pre-DNAT policy defined", func() {
+			policy := api.NewGlobalNetworkPolicy()
```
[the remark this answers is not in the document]
lmm: d8371d3

fv/host_port_test.go
```diff
@@ -104,35 +109,43 @@ var _ = infrastructure.DatastoreDescribe("with initialized Felix", []apiconfig.D
...
+
+			It("should not be able to reach the metrics port with an allow policy on another port", func() {
 				// Ensure the HostEndpoint has taken effect and is blocking traffic
 				Eventually(metricsPortReachable, "10s", "1s").Should(BeFalse())
-				policy := api.NewGlobalNetworkPolicy()
-				policy.Name = "pre-dnat-policy-1"
+			})
+
+			It("should be able to reach the metrics port once a policy allows that port", func() {
+				policy = api.NewGlobalNetworkPolicy()
```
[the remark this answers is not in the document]
lmm: Also in d8371d3

fv/host_port_test.go
```diff
@@ -144,11 +157,38 @@ var _ = infrastructure.DatastoreDescribe("with initialized Felix", []apiconfig.D
...
+	})
+
+	Context("with all-interfaces host endpoint defined", func() {
+		BeforeEach(func() {
+			hostEp := api.NewHostEndpoint()
+			hostEp.Name = "all-interfaces-hostendpoint"
+			hostEp.Labels = map[string]string{"host-endpoint": "true"}
+			hostEp.Spec.Node = felix.Hostname
+			hostEp.Spec.InterfaceName = "*"
+			hostEp.Spec.ExpectedIPs = []string{felix.IP}
```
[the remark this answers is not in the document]
lmm: Yeah that wasn't necessary... 3c06b5dd3521d7097330cfdddd80afca2d15f9d5

fv/hostendpoints_test.go
```diff
@@ -0,0 +1,343 @@
...
+			felix.Stop()
+		}
+
+		if CurrentGinkgoTestDescription().Failed {
+			infra.DumpErrorData()
+		}
+		infra.Stop()
+	})
+
+	allowHostToHostTraffic := func() {
```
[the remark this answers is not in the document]
lmm: Agreed 5c5eb88e9628a1a38775f73d6a760a5bab7eb770

fv/hostendpoints_test.go
```diff
@@ -0,0 +1,343 @@
...
+		It("should block all traffic except pod-to-pod and host-to-own-pod traffic", func() {
+			denyHostToHostTraffic()
+			denyHostToOtherPodTraffic()
+			allowPodToPodTraffic()
+			allowHostToOwnPodTraffic()
+			cc.CheckConnectivity()
+		})
+
+		It("should allow felixes[0] => felixes[1] traffic if ingress and egress policies are in place", func() {
+			// Create a policy selecting felix[1] that allows egress.
```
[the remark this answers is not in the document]
lmm: Fixed in 5c5eb88e9628a1a38775f73d6a760a5bab7eb770

fv/hostendpoints_test.go
```diff
@@ -0,0 +1,343 @@
...
+	})
+
+	Context("with no policies and an allow-all profile on the host endpoints", func() {
+		BeforeEach(func() {
+			// Install a default profile that allows all pod ingress and egress, in the absence of any policy.
+			infra.AddDefaultAllow()
+
+			// Determine "default" profile name.
+			defaultProfileName := "default"
+			if infrastructure.K8sInfra != nil {
```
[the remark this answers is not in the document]
lmm: 85a37238619ebf66aff593d4ecad7996027e00ea

fv/vxlan_test.go
lmm: Ugh... fixed here: 163b9f57cb0954ad1d8d7f17ce69171926a1c93b

rules/endpoints.go
lmm: Thanks, fixed f06eeab5a53d997895a877851f3c350b49ef9f8f

fv/ipip_test.go
```diff
@@ -57,7 +57,7 @@ var _ = infrastructure.DatastoreDescribe("IPIP topology before adding host IPs t
 		felixes, client = infrastructure.StartNNodeTopology(2, infrastructure.DefaultTopologyOptions(), infra)
 
 		// Install a default profile that allows all ingress and egress, in the absence of any Policy.
-		infra.AddDefaultAllow()
+		_ = infra.AddDefaultAllow()
```
fasaxc, line 60: Don't think you need to add `_ = ` everywhere, does the linter complain?  Normally, I'd only do this if ignoring an error.

lmm: Thanks, for some reason I thought that would have been a lint error.

rules/dispatch.go
lmm: Thanks for spotting this!

rules/endpoints.go
lmm: We resolved this in Slack. WHEPs will behave similarly to regular named HEPs regarding profiles.

PR #2229, opened 2020-03-10
BPF: MTU fix for 3.13

tomastigera:
## Description

When a pod responds to traffic initiated through a nodeport (thus encaped in vxlan) and if that packet is non-GSO and close to mtu size so that we cannot grow it to add the encap header, we return an ICMP response DNF MTU too big to let the client adjust its path mtu.

(a) the response has a wrong dest MAC and is dropped by the `ip_rcv_core()` in the pod - we need to swap the MACs

(b) we must check the skb->len > eth + mtu

(c) we must respond with the right mtu in the ICMP - revert previous changes and adjust to the pods 1410 mtu setting

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/bpf-gpl/icmp.h
+++ b/bpf-gpl/icmp.h
@@ -130,7 +130,7 @@ static CALI_BPF_INLINE int icmp_v4_reply(struct __sk_buff *skb,
 #ifdef CALI_PARANOID
 	/* XXX verify that ip_orig.daddr is always the node's IP
 	 *
-	 * we only call this function because of NodePOrt encap
+	 * we only call this function because of NodePort encap
 	 */
 	if (ip_orig.daddr != cali_host_ip()) {
 		CALI_DEBUG("ICMP v4 reply: ip_orig.daddr != cali_host_ip() 0x%x\n", ip_orig.daddr);
@@ -180,7 +180,6 @@ static CALI_BPF_INLINE int icmp_v4_too_big(struct __sk_buff *skb)
 		__be16  unused;
 		__be16  mtu;
 	} frag = {
-		// ICMP MTU ignores the ethernet header.
 		.mtu = host_to_be16(TUNNEL_MTU),
 	};
 
--- a/bpf-gpl/icmp.h
+++ b/bpf-gpl/icmp.h
@@ -130,7 +130,7 @@ static CALI_BPF_INLINE int icmp_v4_reply(struct __sk_buff *skb,
 #ifdef CALI_PARANOID
 	/* XXX verify that ip_orig.daddr is always the node's IP
 	 *
-	 * we only call this function because of NodePort encap
+	 * we only call this function because of NodePOrt encap
 	 */
 	if (ip_orig.daddr != cali_host_ip()) {
 		CALI_DEBUG("ICMP v4 reply: ip_orig.daddr != cali_host_ip() 0x%x\n", ip_orig.daddr);
@@ -180,8 +180,8 @@ static CALI_BPF_INLINE int icmp_v4_too_big(struct __sk_buff *skb)
 		__be16  unused;
 		__be16  mtu;
 	} frag = {
-		// ICMP MTU refers to the IP packet size.
-		.mtu = host_to_be16(TNNL_INNER_IP_MTU),
+		// ICMP MTU ignores the ethernet header.
+		.mtu = host_to_be16(TUNNEL_MTU),
 	};
 
 	CALI_DEBUG("Sending ICMP too big mtu=%d\n", be16_to_host(frag.mtu));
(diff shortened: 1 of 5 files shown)


neiljerram:
@tomastigera Can folk outside Tigera read our JIRA issues like https://tigera.atlassian.net/browse/BPF-350 ?  I would guess not, and so to provide a complete picture here we probably need to copy information from the JIRA issue to here.

bpf-gpl/tc.c
```diff
@@ -178,6 +178,20 @@ static CALI_BPF_INLINE int forward_or_drop(struct __sk_buff *skb,
...
+		/* Revalidate the access to the packet */
+		if ((void *)(long)skb->data + sizeof(struct ethhdr) > (void *)(long)skb->data_end) {
+			reason = CALI_REASON_SHORT;
+			goto deny;
+		}
+
+		/* Swap the MACs as we are turning it back */
+		struct ethhdr *eth_hdr = (void *)(long)skb->data;
+		unsigned char mac[ETH_ALEN];
+		__builtin_memcpy(mac, &eth_hdr->h_source, ETH_ALEN);
```
neiljerram: This needs to be from `h_dest`, not `h_source`.

(DST MAC -> buffer, SOURCE MAC -> DST MAC, buffer -> SOURCE MAC)

bpf/ut/icmp_too_big_test.go
```diff
@@ -43,7 +43,7 @@ func TestICMPTooBig(t *testing.T) {
 		pktR := gopacket.NewPacket(res.dataOut, layers.LayerTypeEthernet, gopacket.Default)
 		fmt.Printf("pktR = %+v\n", pktR)
 
-		checkICMPTooBig(pktR, ipv4, udp, natTunnelMTU-ethernetHeaderSize-20 /* compiled as WEP */)
+		checkICMPTooBig(pktR, ipv4, udp, natTunnelMTU-50 /* compiled as WEP */)
```
neiljerram: Please can you add commenting here to explain what this test is testing, and in particular the magic numbers 64 and 50?

Also, do we have a loop somewhere that tests all of the possible UDP packet sizes?  Can we add that, if we don't have it?

bpf-gpl/nat.h
```diff
@@ -361,7 +365,7 @@ static CALI_BPF_INLINE int vxlan_v4_encap(struct __sk_buff *skb,  __be32 ip_src,
 	eth_inner = (void *)(vxlan+1);
 	ip_inner = (void*)(eth_inner+1);
 
-	/* Copy the original IP header. Since it is already DNATed, the dest IP is
+	/* Copy the original IP header. Since it is aready DNATed, the dest IP is
```
neiljerram: New typo here; please revert.

bpf-gpl/tc.c
tomastigera: yes

bpf-gpl/icmp.h
```diff
@@ -180,8 +180,8 @@ static CALI_BPF_INLINE int icmp_v4_too_big(struct __sk_buff *skb)
 		__be16  unused;
 		__be16  mtu;
 	} frag = {
-		// ICMP MTU refers to the IP packet size.
-		.mtu = host_to_be16(TNNL_INNER_IP_MTU),
+		// ICMP MTU ignores the ethernet header.
```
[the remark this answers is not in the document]
tomastigera: moved the comment to the test in `vxlan_v4_encap_too_big()` where it is more relevant.

bpf/ut/icmp_too_big_test.go
tomastigera: This only tests that `icmp_v4_too_big()` can convert a packet into a ICMP response.

There is another test that actually tests that we check if packet is oversized before we encap in `TestNATNodePort()` in `bpf/ut/nat_test.go`. Note that we are not able to test the (non-)gso scenarious at this level - all packets are non-gso.

Added comment a where the -50 comes from

bpf-gpl/nat.h
```diff
@@ -444,9 +448,14 @@ static CALI_BPF_INLINE int is_vxlan_tunnel(struct iphdr *ip)
...
-	__u32 mtu = TNNL_INNER_ETH_MTU;
+	__u32 mtu = TUNNEL_MTU;
 
-	if (skb->len > mtu) {
+	/* RFC-1191: MTU is the size in octets of the largest datagram that
+	 * could be forwarded, along the path of the original datagram, without
+	 * being fragmented at this router.  The size includes the IP header and
+	 * IP data, and does not include any lower-level headers.
+	 */
+	if (skb->len - sizeof(struct ethhdr) > mtu) {
```
neiljerram: Paranoia: can an SKB have len less than sizeof(struct ethhdr)?  If so it would be safer to write `skb->len > mtu + sizeof(struct ethhdr)` to avoid a possible underflow.

tomastigera: ack

neiljerram:
Many thanks @tomastigera .  I'll merge as soon as the CI completes.

PR #2231, opened 2020-03-11
Allow DHCP from the workload, on kernels where rp_filter doesn't already

neiljerram:


neiljerram:
Successful test on CentOS at https://wavetank.tigera.io/view/Openstack/job/os-ms-smoke-openstack-crocky/21/

rules/static.go
```diff
@@ -827,6 +827,28 @@ func (r *DefaultRuleRenderer) StaticRawPreroutingChain(ipVersion uint8) *Chain {
...
+		if ipVersion == 6 {
+			dhcpSrcPort = uint16(546)
+			dhcpDestPort = uint16(547)
+		}
+		rules = append(rules,
+			Rule{
+				Match: Match().
+					Protocol("udp").
+					SourcePorts(dhcpSrcPort).
+					DestPorts(dhcpDestPort),
```
fasaxc: Think this rule could allow a workload to spoof its IP as long as it's sending traffic using the DHCP port.  That wasn't the case with the sysctl version because we'd only whitelist DHCP after the RPF check.

Should we add a source = 0.0.0.0 check?

neiljerram:
New successful OpenStack smoke tests at https://wavetank.tigera.io/view/Openstack/job/os-ms-smoke-openstack-crocky/23/console and https://wavetank.tigera.io/view/Openstack/job/os-ms-smoke-openstack-uqueens/123/console

PR #2233, opened 2020-03-11
bpf: fix FIB lookup for NP tunnels and ICMP error replies

tomastigera:
This reverts commit 779dacf2f2161fdaa5971939738058b1e1cf0546.

The verted change causes FIB to fail and thus the packet must take the
slow path through Linux ip stack. Linux delivers the packet correctly
as the packet has the right addresses in the right places.

Fix the state on the ICMP path.

Fix approval of the  ICMP error replies as related traffic.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/bpf-gpl/jump.h
+++ b/bpf-gpl/jump.h
@@ -64,4 +64,11 @@ struct bpf_map_def_extended __attribute__((section("maps"))) cali_jump = {
 #endif
 };
 
+static CALI_BPF_INLINE void tc_state_fill_from_iphdr(struct cali_tc_state *state, struct iphdr *ip)
+{
+	state->ip_src = ip->saddr;
+	state->ip_dst = ip->daddr;
+	state->ip_proto = ip->protocol;
+}
+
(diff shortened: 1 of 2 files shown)


fasaxc:
@tomastigera can you point me at where the swap happens in the ICMP case?  Does the helper function do the src/dst swap or does it happen elsewhere in our code?  I can't see anywhere that _we_ do the swap without this change.  Just want to make sure I fully understand what's going on here.

tomastigera:
>  I can't see anywhere that _we_ do the swap without this change.

Yeah, it is incorrect on the icmp error reply path. Good catch! Pushed a fix for that with some more issues addressed.

caseydavenport:
@tomastigera could you add a release note to the `release-note` section with an appropriately user-facing description of this fix? Thanks!

fasaxc:
Removed the cherry-pick label; this is already in v3.14; cherry pick was a potential backport to v3.13 but I don't think we need that.

PR #2238, opened 2020-03-25
BPF: check the VNI of the VXLAN NP tunnel

tomastigera:
## Description

To coexist with other VXLAN overlays, when we use the same port (we use the standard port by default), check the VNI to see if it is for us. If not, let the packets be processed by policy etc. as any other traffic.
    
Our default VNI is 0xca11c0 :)

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/bpf-gpl/nat.h
+++ b/bpf-gpl/nat.h
@@ -33,6 +33,10 @@
 #define CALI_VXLAN_PORT 4789 /* IANA VXLAN port */
 #endif
 
+#ifndef CALI_VXLAN_VNI
+#define CALI_VXLAN_VNI 0xca11c0
+#endif
+
 #define dnat_should_encap() (CALI_F_FROM_HEP && !CALI_F_TUNNEL)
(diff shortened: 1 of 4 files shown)


bpf-gpl/nat.h
```diff
@@ -387,8 +391,8 @@ static CALI_BPF_INLINE int vxlan_v4_encap(struct __sk_buff *skb,  __be32 ip_src,
 	udp->source = udp->dest = host_to_be16(CALI_VXLAN_PORT);
 	udp->len = host_to_be16(be16_to_host(ip->tot_len) - sizeof(struct iphdr));
 
-	/* set the I flag to make the VNI valid, keep the VNI 0-ed */
```
tomastigera: I will return the first part of the comment

tomastigera:
@fasaxc pushed the comment update and squashed with the forgotten WIP commit, one commit only now

PR #2247, opened 2020-04-02
ICMP errors generated by tracked flows treated as related traffic

tomastigera:
## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
The BPF dataplane now handles ICMP error messages as "related" traffic so that they follow the same path back through the dataplane as the packet they respond to. This improves compatibility with path MTU detection and other non-mainline traffic.
```

### DIFF
--- a/bpf-gpl/tc.c
+++ b/bpf-gpl/tc.c
@@ -690,7 +690,7 @@ static CALI_BPF_INLINE struct fwd calico_tc_skb_accepted(struct __sk_buff *skb,
 	int ct_rc = ct_result_rc(state->ct_result.rc);
 	bool ct_related = ct_result_is_related(state->ct_result.rc);
 	uint32_t seen_mark;
-	size_t csum_off_fixup = 0;
+	size_t l4_csum_off = 0, l3_csum_off;
 
 	CALI_DEBUG("src=%x dst=%x\n", be32_to_host(state->ip_src), be32_to_host(state->ip_dst));
 	CALI_DEBUG("post_nat=%x:%d\n", be32_to_host(state->post_nat_ip_dst), state->post_nat_dport);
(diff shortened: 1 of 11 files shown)


bpf-gpl/tc.c
```diff
@@ -728,16 +730,87 @@ static CALI_BPF_INLINE struct fwd calico_tc_skb_accepted(struct __sk_buff *skb,
...
+
+	__u8 ihl = ip_header->ihl * 4;
+
+	int res = 0;
+	bool encap_needed = false;
+
+	if (state->ip_proto == IPPROTO_ICMP && ct_related) {
+		/* do not fix up embedded L4 checksum for related ICMP */
+	} else {
+		switch (ip_header->protocol) {
```
tomastigera, line 803: `state` carries information about the entire packet, the `state->ip_proto` is still icmp, but we have moved the `ip_header` to point to the inner header. All the editing it done where point to, so we should use that info even in the case of regular NAT

bpf-gpl/tc.c
```diff
@@ -699,23 +710,14 @@ static CALI_BPF_INLINE struct fwd calico_tc_skb_accepted(struct __sk_buff *skb,
...
-	size_t l4_csum_off = 0, l3_csum_off;
-	int res = 0;
-	bool encap_needed = false;
-	uint32_t fib_flags = 0;
-
 	/* XXX we cannot pass the related ICMP after NATing back yet, so we need
 	 * to act here, we know we are forwarding.
 	 */
+
+	/* XXX XXX XXX */
```
fasaxc, line 717: ?

bpf-gpl/tc.c
```diff
@@ -728,16 +730,87 @@ static CALI_BPF_INLINE struct fwd calico_tc_skb_accepted(struct __sk_buff *skb,
...
+		if (ip_header->protocol == IPPROTO_ICMP) {
+			struct icmphdr *icmp;
+			bool outer_ip_snat;
+
+			/* if we do SNAT ... */
+			outer_ip_snat = ct_rc == CALI_CT_ESTABLISHED_SNAT;
+			/* ... there is a return path to the tunnel ... */
+			outer_ip_snat = outer_ip_snat && state->ct_result.tun_ret_ip;
+			/* ... and should do encap and it is not DSR or it is
+			 * it is leaving host and either DSR from WEP or
```
fasaxc, line 744: typo "it is it is"

bpf-gpl/tc.c
```diff
@@ -873,6 +946,38 @@ static CALI_BPF_INLINE struct fwd calico_tc_skb_accepted(struct __sk_buff *skb,
...
+				state->ip_dst = state->ct_result.tun_ret_ip;
+				seen_mark = CALI_SKB_MARK_BYPASS_FWD_SRC_FIXUP;
+				goto nat_encap;
+			} else if (CALI_F_TO_HEP) {
+				/* Special case for ICMP error being returned by the host with the
+				 * backing workload into the tunnel back to the original host. It is
+				 * ICMP related and there is a return tunnel path. We need to change
+				 * both the source and destination at once.
+				 *
+				 * XXX the packet was routed to the original client as if it was XXX
```
fasaxc, line 970: Think I need you to walk me through this case

bpf-gpl/tc.c
```diff
@@ -728,16 +730,87 @@ static CALI_BPF_INLINE struct fwd calico_tc_skb_accepted(struct __sk_buff *skb,
...
+				if (res) {
+					reason = CALI_REASON_CSUM_FAIL;
+					goto deny;
+				}
+				CALI_DEBUG("ICMP related: outer IP SNAT to %x\n",
+						be32_to_host(state->ct_result.nat_ip));
+			}
+
+			if (!icmp_skb_get_hdr(skb, &icmp)) {
+				CALI_DEBUG("Ooops, we already passed one such a check!!!\n");
```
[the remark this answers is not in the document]
tomastigera: perhaps, but the function above also logs and this is more like this never happens, I'd panic if I could but did not feel like I should let this go completely unhandled.

bpf-gpl/tc.c
tomastigera: thanks for reminding me of this. Fixed the comment. I decided that it is easier to handle it in here than to let the host to handle it. At least for now.

PR #2251, opened 2020-04-08
BPF: tunnel MTU configurable via VXLAN MTU

tomastigera:
## Description

MTU used by NP tunnel is configurable, or rather tight to, vxlan mtu

Patching values into a precompiled binary is isolateded into a bpf.Binary object

UT tests can use the patching to reduced the number of precompiled binaries, eventually avoiding most of the combinations.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```


dataplane/linux/bpf_ep_mgr.go
```diff
@@ -500,8 +500,11 @@ func (m *bpfEndpointManager) ensureQdisc(ifaceName string) {
 func (m *bpfEndpointManager) attachWorkloadProgram(endpoint *proto.WorkloadEndpoint, polDirection PolDirection) error {
 	ap := m.calculateTCAttachPoint(tc.EpTypeWorkload, polDirection, endpoint.Name)
 	// Host side of the veth is always configured as 169.254.1.1.
-	hostIP := net.IPv4(169, 254, 1, 1)
-	err := tc.AttachProgram(ap, hostIP)
+	ap.IP = net.IPv4(169, 254, 1, 1)
+	// vxlanMTU is the MTU of workload devices, we need to use another 50 less
```
fasaxc, line 504: I think this comment is a bit misleading.  The VXLANMTU configuration parameter is the MTU that felix sets on the VXLAN device that it creates for VXLAN IP pools.  That's its original purpose.

So, I think we need to carefully explain where the `-50` come from here.  I think it's something like:

* VXLAN MTU should be the eth0 MTU -50, in order to allow space for VXLAN.
* We also expect that to be the MTU used on veths.
* We do encap on the veths, and there's a bogus kernel MTU check in the BPF helper for resizing the packet, so we have to reduce the apparent MTU by another 50 bytes.

fasaxc: Maybe the variable should be renamed ICMP_TOO_BIG_MTU, or something?

bpf-gpl/bpf.h
```diff
@@ -179,4 +179,20 @@ static CALI_BPF_INLINE void ip_dec_ttl(struct iphdr *ip)
...
+{												\
+	__u32 ret;										\
+	asm("%0 = " #pattern ";" : "=r"(ret) /* output */ : /* no inputs */ : /* no clobber */);\
+	return ret;										\
+}
+
+#define CALI_CONFIGURABLE(name)	cali_configurable_##name()
+
+CALI_CONFIGURABLE_DEFINE(host_ip, 0x54534f48) /* be 0x54534f48 = ASCII(HOST) */
+CALI_CONFIGURABLE_DEFINE(tunnel_mtu, 0x55544d54) /* be 0x55544d54 = ASCII(HOST) */
```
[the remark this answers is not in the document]
tomastigera: indeed :-)

dataplane/linux/bpf_ep_mgr.go
tomastigera: applied the comment kept the var name

PR #2256, opened 2020-04-09
Update routetable to accept deltas

robbrockbank:
## Description
This enhances the routable module to provide:
-  An interface which accepts delta route programming
-  Exact interface name matches
-  A "None interface" for non-link rules along with handling for throw, blackhole and prohibit route types.
-  Support for table index (rather than just the main table)

## Todos
- [ ] Integration tests (delete as appropriate) Not needed
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/dataplane/linux/int_dataplane.go
+++ b/dataplane/linux/int_dataplane.go
@@ -375,8 +375,8 @@ func NewIntDataplaneDriver(config Config) *InternalDataplane {
 	dp.ipSets = append(dp.ipSets, ipSetsV4)
 
 	if config.RulesConfig.VXLANEnabled {
-		routeTableVXLAN := routetable.New([]string{"vxlan.calico"}, 4, true, config.NetlinkTimeout, config.DeviceRouteSourceAddress,
-			config.DeviceRouteProtocol, true)
+		routeTableVXLAN := routetable.New([]string{"^vxlan.calico$"}, 4, true, config.NetlinkTimeout,
+			config.DeviceRouteSourceAddress, config.DeviceRouteProtocol, true, 0)
 
 		vxlanManager := newVXLANManager(
 			ipSetsV4,
@@ -572,8 +572,12 @@ func NewIntDataplaneDriver(config Config) *InternalDataplane {
 		}
 	}
 
-	routeTableV4 := routetable.New(config.RulesConfig.WorkloadIfacePrefixes, 4, false, config.NetlinkTimeout,
-		config.DeviceRouteSourceAddress, config.DeviceRouteProtocol, config.RemoveExternalRoutes)
+	interfaceRegexes := make([]string, len(config.RulesConfig.WorkloadIfacePrefixes))
+	for i, r := range config.RulesConfig.WorkloadIfacePrefixes {
+		interfaceRegexes[i] = "^" + r + ".*"
+	}
+	routeTableV4 := routetable.New(interfaceRegexes, 4, false, config.NetlinkTimeout,
+		config.DeviceRouteSourceAddress, config.DeviceRouteProtocol, config.RemoveExternalRoutes, 0)
 
 	epManager := newEndpointManager(
 		rawTableV4,
@@ -639,7 +643,9 @@ func NewIntDataplaneDriver(config Config) *InternalDataplane {
 		dp.iptablesMangleTables = append(dp.iptablesMangleTables, mangleTableV6)
 		dp.iptablesFilterTables = append(dp.iptablesFilterTables, filterTableV6)
 
-		routeTableV6 := routetable.New(config.RulesConfig.WorkloadIfacePrefixes, 6, false, config.NetlinkTimeout, config.DeviceRouteSourceAddress, config.DeviceRouteProtocol, config.RemoveExternalRoutes)
+		routeTableV6 := routetable.New(
+			interfaceRegexes, 6, false, config.NetlinkTimeout,
+			config.DeviceRouteSourceAddress, config.DeviceRouteProtocol, config.RemoveExternalRoutes, 0)
 
 		if !config.BPFEnabled {
 			dp.RegisterManager(newIPSetsManager(ipSetsV6, config.MaxIPSetSize, callbacks))
--- a/routetable/route_table.go
+++ b/routetable/route_table.go
@@ -17,6 +17,7 @@ package routetable
 import (
 	"errors"
 	"net"
+	"reflect"
 	"regexp"
 	"strings"
 	"syscall"
@@ -29,6 +30,8 @@ import (
 	"github.com/projectcalico/felix/conntrack"
 	"github.com/projectcalico/felix/ifacemonitor"
 	"github.com/projectcalico/felix/ip"
+	netlinkshim "github.com/projectcalico/felix/netlink"
+	timeshim "github.com/projectcalico/felix/time"
 	"github.com/projectcalico/libcalico-go/lib/set"
 )
 
@@ -62,11 +65,25 @@ func init() {
 	prometheus.MustRegister(listIfaceTime, perIfaceSyncTime)
 }
 
+const (
+	// Use this for targets with no outbound interface.
+	InterfaceNone = "*NoOIF*"
+)
+
 type TargetType string
 
 const (
 	TargetTypeVXLAN   TargetType = "vxlan"
 	TargetTypeNoEncap TargetType = "noencap"
+
+	// The following target types should be used with InterfaceNone.
+	TargetTypeBlackhole TargetType = "blackhole"
+	TargetTypeProhibit  TargetType = "prohibit"
+	TargetTypeThrow     TargetType = "throw"
+)
+
+const (
+	maxApplyRetries = 2
 )
 
 type L2Target struct {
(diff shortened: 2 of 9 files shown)


routetable/route_table.go
```diff
@@ -631,6 +852,64 @@ func (r *RouteTable) syncRoutesForLink(ifaceName string) error {
...
+		// In case this IP is being re-used, wait for any previous conntrack entry
+		// to be cleaned up.  (No-op if there are no pending deletes.)
+		r.waitForPendingConntrackDeletion(cidr.Addr())
+		if err := nl.RouteDel(&route); err != nil {
+			logCxt.WithError(err).Warn("Failed to delete route")
+			updatesFailed = true
+		}
+	}
+
+	// Delete routes followed by creates.
```
fasaxc, line 881: Out of date comment.

routetable/route_table.go
```diff
@@ -426,88 +533,203 @@ func (r *RouteTable) ensureL2Dataplane(linkAttrs *netlink.LinkAttrs, target L2Ta
...
+		route.Scope = netlink.SCOPE_UNIVERSE
+		route.SetFlag(syscall.RTNH_F_ONLINK)
+	}
+
+	return route
+}
+
+// fullSyncRoutesForLink performs a full resync of the routes by first listing current routes and correlating against
+// the expected set.
+func (r *RouteTable) fullSyncRoutesForLink(
```
fasaxc, line 700: Could the "full resync" case and the "delta updates" case be combined to avoid having two fairly complex codepaths to maintain?

Elsewhere, I've used this pattern to force both cases through the same logic:

* Track our complete, "desired state" of the dataplane
* Track which parts of the dataplane we think are dirty
* When making a delta change, update desired state and mark everything touched as dirty
* When doing a resync, throw away the dirty set and recalculate it from scratch; the resync doesn't actually modify the dataplane, only our picture of what's dirty
* Then (whether or not we did a resync) to apply updates, iterate over the dirty set and queue up an update to bring the dataplane into sync with the desired state.  If the resync found everything was already in sync, the dirty set ends up empty and we do nothing at this step.

Example: https://github.com/projectcalico/felix/blob/master/dataplane/linux/bpf_route_mgr.go#L124

fasaxc:
Should we make the API to specify the names directly take a regex pattern?  Then the caller can control whether it's a prefix or not?  Might be cleaner than yet-another-bool?

routetable/route_table.go
```diff
@@ -535,16 +770,22 @@ func (r *RouteTable) syncRoutesForLink(ifaceName string) error {
...
-			routeProblems = append(routeProblems, "incorrect protocol")
+		if dest != ipV6LinkLocalCIDR {
+			if r.deviceRouteProtocol != route.Protocol {
+				routeProblems = append(routeProblems, "incorrect protocol")
+			}
+			if expectedTargetFound && expectedTarget.RouteType() != route.Type {
+				routeProblems = append(routeProblems, "incorrect type")
+			}
 		}
 		if len(routeProblems) == 0 {
```
song-jiang, line 790: @robbrockbank It is great to have the check for `RouteType` here as it could useful for egress ip.
Could we also add an extra check for gateway ip? 
```
	if route.Gw.Equal(expectedTarget.GW.AsNetIP()) {
            routeProblems = append(routeProblems, "mismatch gateway")
	}
```

dataplane/linux/int_dataplane.go
```diff
@@ -374,8 +374,8 @@ func NewIntDataplaneDriver(config Config) *InternalDataplane {
 	dp.ipSets = append(dp.ipSets, ipSetsV4)
 
 	if config.RulesConfig.VXLANEnabled {
-		routeTableVXLAN := routetable.New([]string{"vxlan.calico"}, 4, true, config.NetlinkTimeout, config.DeviceRouteSourceAddress,
-			config.DeviceRouteProtocol, true)
+		routeTableVXLAN := routetable.New([]string{"^vxlan.calico.*"}, 4, true, config.NetlinkTimeout,
```
fasaxc, line 377: FWIW, I think there's only one VXLAN device so no need for a prefix.  @caseydavenport would "just know"

routetable/route_table.go
```diff
@@ -223,22 +297,60 @@ func (r *RouteTable) onIfaceSeen(ifaceName string) {
...
 
+// markIfaceForUpdate marks an interface update is required. This is either a delta update from a route
+// set, update, or remove, or a full resync triggered from start-up processing, QueueResync or a previous failed update.
+func (r *RouteTable) markIfaceForUpdate(ifaceName string, resync bool) {
+	if resync {
+		// This is a full resync so flag as such.
+		r.ifaceNameToUpdateType[ifaceName] = updateTypeFullResync
+	} else if _, ok := r.ifaceNameToUpdateType[ifaceName]; !ok {
+		// This is not a full resync - set the update status if not already set (because we don't want to "downgrade"
+		// a full-resync to a delta update).
```
fasaxc, line 308: Very clear now, thanks!

routetable/route_table.go
```diff
@@ -426,104 +538,230 @@ func (r *RouteTable) ensureL2Dataplane(linkAttrs *netlink.LinkAttrs, target L2Ta
...
-	leaveDirty := false
-
 	// If this is a modify or delete, grab a copy of the existing targets so we can clean up
 	// conntrack entries even if the routes have been removed.  We'll remove any still-required
 	// CIDRs from this set below.  We don't apply the grace period to this calculation because
 	// it only removes routes that the datamodel previously said were there and then were
-	// removed.  In that case, we know we're up to date.
-	oldCIDRs := set.New()
-	if updatedTargets, ok := r.pendingIfaceNameToTargets[ifaceName]; ok {
+	// removed.  In that case, we know we're up to date.n
```
fasaxc, line 553: Typo: `n` at the end

routetable/route_table.go
```diff
@@ -631,6 +861,64 @@ func (r *RouteTable) syncRoutesForLink(ifaceName string) error {
...
+	}
+
+	// Delete routes first.
+	updatesFailed := false
+	for cidr, target := range deletedCIDRsToTarget {
+		route := r.createL3Route(linkAttrs, target)
+
+		// In case this IP is being re-used, wait for any previous conntrack entry
+		// to be cleaned up.  (No-op if there are no pending deletes.)
+		r.waitForPendingConntrackDeletion(cidr.Addr())
```
fasaxc, line 883: I don't think you should be waiting here.  The conntrack deletion is done async because it's slow and we don't want to block deletion itself.  Usually, IPs are reused in FIFO fashion so we take advantage of that by kicking off cleanup in the background on deletion but only waiting for it to finish when we add a new route with the same CIDR.

Also, conntrack cleanup should be done after removing a route so that no new packets can flow and create conntrack entries in parallel.

routetable/route_table.go
```diff
@@ -304,226 +448,345 @@ func (r *RouteTable) Apply() error {
...
+
+	// Any deleted route that is not being replaced by a route with the same CIDR should have the corresponding
+	// conntrack entry removed.
+	defer func() {
+		cidrsToTarget := r.ifaceNameToTargets[ifaceName]
+		for _, route := range routesToDelete {
+			if cidr := ip.CIDRFromIPNet(route.Dst); cidr == nil {
+				// No parseable CIDR destination in route.
+			} else if _, ok := cidrsToTarget[cidr]; !ok {
+				// Route is deleted and CIDR should not be routable anymore - remove conntrack entries.
```
fasaxc, line 578: Not sure if this condition is right.  The point of deleting conntrack entries is in case an IP gets reused:
* workload A has 10.0.0.1
* felix gets disconnected form datastore/whatever for 30s
* workload A is removed
* workload B gets created with same ip 10.0.0.1
* felix wakes up, goes through route reconciliation: we _want_ to remove conntrack entries for workload A's use of 10.0.0.1 so that workload B can't re-use them

routetable/route_table.go
```diff
@@ -535,100 +798,73 @@ func (r *RouteTable) syncRoutesForLink(ifaceName string) error {
...
 
-		routeExpected := expectedCIDRs.Contains(dest)
+		expectedTarget, expectedTargetFound := expectedTargets[dest]
+		routeExpected := expectedTargetFound || (r.ipVersion == 6 && dest == ipV6LinkLocalCIDR)
 		var routeProblems []string
 		if !routeExpected {
 			routeProblems = append(routeProblems, "unexpected route")
 		}
 		if !r.deviceRouteSourceAddress.Equal(route.Src) {
+			//TODO(rlb): What about ipv6 link local?  Should we remove, if so, how does it get added back?
```
fasaxc, line 808: We've always let the kernel handle the link-locals before, maybe we should just ignore those completely?  The source will be our own link local, I'd have thought.

routetable/route_table.go
robbrockbank: @fasaxc :  Doesn't this check just say "only kick off contract deletion if the CIDR is no longer valid for the interface" - which I think is ok no?  In the case of IP reuse the CIDR will not be valid for the interface anymore.

fasaxc: Agreed, I'd misunderstood

routetable/route_table.go
fasaxc: As discussed, I now think this is source IP selection tweaking and I think it should be inside the following conditional.  It probably works in the mainline case because  `deviceRouteSourceAddress` is empty and so is `Src` for the IPv6 route (but I haven't verified that).

PR #2258, opened 2020-04-13
Configurable source for routing information

caseydavenport:
## Description

This PR introduces a config option to allow configuration of different
sources of routing information in Felix's L3 route resolver. 

For now, it supports two modes:
- calicoIPAM: current behavior, drive routing decisions off Calico IPAM.
- workloadIPs: new option, driven based off WEP IP addresses.

The `workloadIPs` option allows Felix to be used for routing even if not
using Calico IPAM (felix is currently used for routing of VXLAN routes
and when using the eBPF tech-preview dataplane.

A few things left:
- [x] Expose as an option on FelixConfiguration API in libcalico
- [x] Decide if we need to also generate routes for
node.spec.BGP.XTunnelAddress IPs.
- [x] Test using eBPF dataplane on live cluster. 

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
Felix can now calculate routes without dependency on Calico IPAM.
```

### DIFF
--- a/calc/l3_route_resolver.go
+++ b/calc/l3_route_resolver.go
@@ -107,7 +107,7 @@ func (c *L3RouteResolver) RegisterWith(allUpdDispatcher, localDispatcher *dispat
 	// Depending on if we're using workload endpoints for routing information, we may
 	// need all WEPs, or only local WEPs.
 	logrus.WithField("routeSource", c.routeSource).Info("Registering for L3 route updates")
-	if c.routeSource == "workloadIPs" {
+	if c.routeSource == "WorkloadIPs" {
 		// Driven off of workload IP addressess. Register for all WEP udpates.
 		allUpdDispatcher.Register(model.WorkloadEndpointKey{}, c.OnWorkloadUpdate)
 	} else {
@@ -726,6 +726,9 @@ func (r *RouteTrie) RemoveWEP(cidr ip.V4CIDR, nodename string) {
 		if ri.WEP.RefCount[nodename] < 0 {
 			logrus.WithField("cidr", cidr).Panic("BUG: Asked to decref a workload past 0.")
 		}
+		if len(ri.WEP) == 0 {
+			ri.WEP = nil
+		}
 	})
 }
 
@@ -817,9 +820,11 @@ func (r RouteInfo) IsValidRoute() bool {
 // explicitly copy them so that they are not shared between the copies.
 func (r RouteInfo) Copy() RouteInfo {
 	cp := r
-	cp.WEP.RefCount = map[string]int{}
-	for n, c := range r.WEP.RefCount {
-		cp.WEP.RefCount[n] = c
+	if len(r.WEP) != 0 {
+		cp.WEP.RefCount = map[string]int{}
+		for n, c := range r.WEP.RefCount {
+			cp.WEP.RefCount[n] = c
+		}
 	}
 	return cp
 }
--- a/calc/l3_route_resolver.go
+++ b/calc/l3_route_resolver.go
@@ -64,6 +64,7 @@ type L3RouteResolver struct {
 	allPools               map[string]model.IPPool
 	workloadIDToCIDRs      map[model.WorkloadEndpointKey][]cnet.IPNet
 	useNodeResourceUpdates bool
+	routeSource            string
 }
 
 type l3rrNodeInfo struct {
@@ -75,7 +76,7 @@ func (i l3rrNodeInfo) AddrAsCIDR() ip.V4CIDR {
 	return i.Addr.AsCIDR().(ip.V4CIDR)
 }
 
-func NewL3RouteResolver(hostname string, callbacks PipelineCallbacks, useNodeResourceUpdates bool) *L3RouteResolver {
+func NewL3RouteResolver(hostname string, callbacks PipelineCallbacks, useNodeResourceUpdates bool, routeSource string) *L3RouteResolver {
 	logrus.Info("Creating L3 route resolver")
 	return &L3RouteResolver{
 		myNodeName: hostname,
@@ -88,6 +89,7 @@ func NewL3RouteResolver(hostname string, callbacks PipelineCallbacks, useNodeRes
 		allPools:               map[string]model.IPPool{},
 		workloadIDToCIDRs:      map[model.WorkloadEndpointKey][]cnet.IPNet{},
 		useNodeResourceUpdates: useNodeResourceUpdates,
+		routeSource:            routeSource,
 	}
 }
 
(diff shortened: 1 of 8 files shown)


calc/l3_route_resolver.go
```diff
@@ -763,11 +789,12 @@ type RouteInfo struct {
...
-	LocalWEP struct {
-		// Count of local WEPs that have this CIDR.  Normally, this will be 0 or 1 but Felix has to be tolerant
-		// to bad data (two local WEPs with the same CIDR) so we do ref counting.
-		RefCount int
+	// WEP contains information extracted from the workload endpoints.
+	WEP struct {
+		// Count of WEPs that have this CIDR.  Normally, this will be 0 or 1 but Felix has to be tolerant
+		// to bad data (two WEPs with the same CIDR) so we do ref counting.
+		// The RefCount is tracked per-node, to properly handle remote weps.
+		RefCount map[string]int
```
caseydavenport, line 797: I made this a map since previously the RefCount could only apply to local WEPs and so the node name was implicitly known.

We now need to track the node name for each WEP, and we don't want to blat over it if we end up in a transient scenario where two CIDRs are owned by different WEPs on different nodes, and thus end up with a route pointing at the wrong node. This seems to let us maintain both nodenames, and will eventually resolve itself down to a single entry.

LMK if you have better ideas on how to do this!

calc/l3_route_resolver.go
```diff
@@ -525,11 +536,20 @@ func (c *L3RouteResolver) flush() {
 					rt.Type = proto.RouteType_REMOTE_HOST
 				}
 			}
-			if ri.LocalWEP.RefCount > 0 {
-				// We have a local WEP with this IP.
-				rt.DstNodeName = c.myNodeName
-				rt.Type = proto.RouteType_LOCAL_WORKLOAD
-				rt.LocalWorkload = true
+
+			for nodename, _ := range ri.WEP.RefCount {
```
fasaxc, line 540: Map iteration is not deterministic (it's randomised each time).  I think it's really important to keep the calc graph deterministic for a few reasons:

- It really helps when reasoning about it (e.g. when diagnosing an issue hit in the field)
- It's part of the contract that other parts of felix expect (I don't want to have to revisit other parts of felix to check for invalidated assumptions if we allow a bit of non-determinism).
- It's expected and asserted by the calc graph FV tests (which you should use to test this new feature).   If you add a test in that framework that has two overlapping WEPs, it should detect the non-determinism

One way to quickly get determinism here is to use a tie-break rule such as "highest nodename wins".

calc/l3_route_resolver.go
fasaxc: I think you're right that we need to solve that problem.  I might be tempted to optimise for occupancy rather than efficiency.  For example, you could store the same information as a `[]struct{Node string, RefCount int}`; in the mainline case where there's exactly 1, it'd probably be faster than a map anyway and you could sort the slice on nodename on update so that lookup is O(1) (pick the 0th entry)

calc/l3_route_resolver.go
```diff
@@ -783,7 +810,18 @@ func (r RouteInfo) IsValidRoute() bool {
...
 		r.Pool.NATOutgoing ||
-		r.LocalWEP.RefCount > 0
+		len(r.WEP.RefCount) > 0
+}
+
+// Copy returns a copy of the RouteInfo. Since some fields are pointers, we need to
+// explicitly copy them so that they are not shared between the copies.
+func (r RouteInfo) Copy() RouteInfo {
+	cp := r
+	cp.WEP.RefCount = map[string]int{}
```
fasaxc, line 820: For occupancy's sake, I think I'd handle the nil map case explicitly.  I.e. if we have 0 refcounts to store, set the map to nil and explicitly check/initialise it if we have something to store.

If you switch to using a slice, likewise; I'd set it to nil whenever it's length 0.  That also makes the Equals() method work in all cases; right now, it'll gte tripped up by the difference between a nil map and an empty map.

config/config_params.go
```diff
@@ -135,6 +135,11 @@ type Config struct {
 	TyphaReadTimeout    time.Duration `config:"seconds;30;local"`
 	TyphaWriteTimeout   time.Duration `config:"seconds;10;local"`
 
+	// Configure where Felix gets its routing information.
+	// - workloadIPs: use workload endpoints to construct routes.
+	// - calicoIPAM: use IPAM data to contruct routes.
+	RouteSource string `config:"oneof(workloadIPs,calicoIPAM);calicoIPAM"`
```
fasaxc, line 141: Maybe we should align new params with libcalico-go naming?  I.e `WorkloadIPs` (c.f. `Allow|Deny` in policies)?

calc/l3_route_resolver.go
fasaxc: Same goes for the zero check.  Don't want a 0-length slice to prevent cleanup

calc/l3_route_resolver.go
```diff
@@ -688,17 +709,54 @@ func (r *RouteTrie) RemoveHost(cidr ip.V4CIDR, nodeName string) {
...
-func (r *RouteTrie) AddLocalWEP(cidr ip.V4CIDR) {
+func (r *RouteTrie) AddWEP(cidr ip.V4CIDR, nodename string) {
 	r.updateCIDR(cidr, func(ri *RouteInfo) {
-		ri.LocalWEP.RefCount++
+		if ri.WEP.RefCount == nil {
+			ri.WEP.RefCount = map[string]int{}
+		}
+		ri.WEP.RefCount[nodename]++
+
+		// Groom the nodename slice.
```
caseydavenport, line 719: I would like to switch to using a list to store the WEPs, like you suggested. This was quicker and gets the same result IIUC, and I'd like to progress with testing, so I went with this for now.

Should be relatively seamless to switch the implementation to a `[]struct{}` rather than a map once I've got some more tests written.

calc/l3_route_resolver.go
```diff
@@ -688,17 +709,54 @@ func (r *RouteTrie) RemoveHost(cidr ip.V4CIDR, nodeName string) {
...
+				found = true
+				break
+			}
+		}
+		if !found {
+			log.Fatalf("Unable to find element %s in %s", e, s)
+		}
+
+		// Remove it.
+		s = append(s[:i], s[i+1:]...)
```
fasaxc, line 745: Linter spotted that this side effect is not published.  When you do 
```
s = append(s[:i], s[i+1:]...)
```
then append will re-use the same backing array but the change in length of the slice is local to the new `s` that is returned, the input `s` will still have the old length.

calc/l3_route_resolver.go
```diff
@@ -63,7 +63,9 @@ type L3RouteResolver struct {
 	blockToRoutes          map[string]set.Set
 	allPools               map[string]model.IPPool
 	workloadIDToCIDRs      map[model.WorkloadEndpointKey][]cnet.IPNet
+	nodeToCIDRs            map[string]set.Set
```
caseydavenport, line 66: UT spotted that we weren't marking routes dirty when they targeted at Node and that node's IP changed.

e.g., the case where a WEP appears in the syncer before the corresponding node does, so we don't know the node's IP.

I added a new map to track the CIDRs for each node so that when the node IP changes we can mark those CIDRs dirty.

fasaxc: That's meant to be handled by the call to `c.markAllNodeRoutesDirty(nodeName)`, but looks like that method is implemented with a loop over (all) blocks right now.

Adding a node to CIDR cache would make it more efficient but it'd need to be refcounted or something since you might have duplicate CIDRs on a node (at least transiently).

calc/l3_route_resolver.go
```diff
@@ -783,7 +870,17 @@ func (r RouteInfo) IsValidRoute() bool {
...
-		r.LocalWEP.RefCount > 0
+		len(r.WEPs) > 0
+}
+
+// Copy returns a copy of the RouteInfo. Since some fields are pointers, we need to
+// explicitly copy them so that they are not shared between the copies.
+func (r RouteInfo) Copy() RouteInfo {
+	cp := r
+	if len(r.WEPs) != 0 {
+		cp.WEPs = append(cp.WEPs, r.WEPs...)
```
fasaxc, line 881: `cp` will have WEPs initialised to the same slice so this will result in `cp.WEPs` being a doubled copy of `r.WEPs` (and it may share backing storage debending on the capacity of `r.WEPs`.)  I think you want:
```
cp.WEPs = make([]WEP, len(r.WEPs))
copy(cp.WEPs, r.WEPs)
```

caseydavenport:
GitHub seems to gotten into a funky state on this PR due to the intermittent instability they were having this morning - new commits aren't showing up, and it won't let me re-open this PR :D 

I've opened a new one here: https://github.com/projectcalico/felix/pull/2269

PR #2260, opened 2020-04-16
Fix VXLAN race condition

caseydavenport:
## Description

Fixes a race in VXLAN resolver where we don't get cross subnet routes.

Already fixed in master due to a refactoring of the route resolution code by @fasaxc .

This is a small patch to earlier releases to fix the same issue. 

https://github.com/projectcalico/calico/issues/3231

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
Fix race condition in VXLAN CrossSubnet code, issue https://github.com/projectcalico/calico/issues/3231
```

### DIFF
--- a/bpf/ut/bpf_prog_test.go
+++ b/bpf/ut/bpf_prog_test.go
@@ -60,9 +60,8 @@ func init() {
 
 // Constants that are shared with the UT binaries that we build.
 const (
-	natTunnelMTU       = uint16(700)
-	ethernetHeaderSize = 14
-	testVxlanPort      = uint16(5665)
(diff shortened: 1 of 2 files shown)


bpf/ut/bpf_prog_test.go
```diff
@@ -60,9 +60,8 @@ func init() {
 
 // Constants that are shared with the UT binaries that we build.
 const (
-	natTunnelMTU       = uint16(700)
-	ethernetHeaderSize = 14
```
caseydavenport, line 64: Static checks complained about this const being unused.

/shrug

PR #2261, opened 2020-04-16
Disable unprivileged BPF mode by default.

fasaxc:
## Description

Set the `kernel.unprivileged_bpf_disabled` sysctl by default.  This locks unprivileged users out of the BPF syscall.  Include a config param to disable the behaviour.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
In BPF mode, Felix now sets the kernel.unprivileged_bpf_disabled sysctl by default to restrict access to the BPF syscall.  This behaviour is controlled by the BPFDisableUnprivileged configuration parameter.
```


fasaxc:
> I notice you like log messages ending with a full stop now; why is that, and should it be promulgated to others?

I hadn't given any thought to it to be honest!

calc/profile_decoder.go
```diff
@@ -34,7 +34,7 @@ type ProfileDecoder struct {
 }
 
 func NewProfileDecoder(callbacks passthruCallbacks) *ProfileDecoder {
-	return &ProfileDecoder{callbacks: callbacks, converter: conversion.Converter{}}
+	return &ProfileDecoder{callbacks: callbacks, converter: conversion.NewConverter()}
```
fasaxc, line 37: Required by the libcalico-go changes

PR #2263, opened 2020-04-17
In BPF mode, add bpf to cluster type

sridhartigera:
The changes adds the bpf string to the clustertype, based on whether BPF is enabled or not.
## Todos
UT logs attached
[bpf-365_ut.log](https://github.com/projectcalico/felix/files/4492096/bpf-365_ut.log)

## Release Note

```release-note
None required
```

### DIFF
--- a/usagerep/usagerep.go
+++ b/usagerep/usagerep.go
@@ -97,7 +97,8 @@ func (u *UsageReporter) PeriodicallyReportUsage(ctx context.Context) {
 
 	doReport := func() {
 		alpEnabled := (config["PolicySyncPathPrefix"] != "")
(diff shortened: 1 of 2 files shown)


usagerep/usagerep.go
```diff
@@ -97,6 +97,9 @@ func (u *UsageReporter) PeriodicallyReportUsage(ctx context.Context) {
 
 	doReport := func() {
 		alpEnabled := (config["PolicySyncPathPrefix"] != "")
+		if config["BPFEnabled"] == "true" {
+			config["ClusterType"] = config["ClusterType"] + ",bpf"
```
fasaxc, line 101: I don't think you should change the incoming map here since this code doesn't own it (and it may be going to multiple places).  Please can you define a new variable to modify locally.

PR #2268, opened 2020-04-21
Implement routing table config and manager

neiljerram:
## Description
Together with https://github.com/projectcalico/libcalico-go/pull/1226, this PR establishes a common config model and allocation framework for Felix's use of additional Linux routing tables.  The current intention is that:

- our Wireguard code, which needs just one additional routing table, will allocate that at start of day by calling `RouteTableManager.GrabIndex()`

- other functions that need routing tables can allocate those by calling `RouteTableManager.GrabIndex()` or `RouteTableManager.GrabAllRemainingIndices()`.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/routetable/usage_manager.go
+++ b/routetable/usage_manager.go
@@ -1,0 +1,48 @@
+// Copyright (c) 2020 Tigera, Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package routetable
+
+import (
+	"errors"
+
+	"github.com/golang-collections/collections/stack"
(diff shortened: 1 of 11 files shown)


routetable/usage_manager.go
```diff
@@ -0,0 +1,59 @@
...
+
+func (r *RouteTableManager) GrabAllRemainingIndices() set.Set {
+	remainingIndices := set.New()
+	idx, err := r.GrabIndex()
+	for err == nil {
+		remainingIndices.Add(idx)
+		idx, err = r.GrabIndex()
+	}
+	return remainingIndices
+}
```
song-jiang, line 59: I am thinking it is better to promote this package to upper level and to be more generic. Similar to `markbits`. Maybe rename it to `indexmanager` ?

routetable/usage_manager.go
```diff
@@ -0,0 +1,59 @@
...
+
+import (
+	"errors"
+
+	"github.com/golang-collections/collections/stack"
+	v3 "github.com/projectcalico/libcalico-go/lib/apis/v3"
+	"github.com/projectcalico/libcalico-go/lib/set"
+)
+
+type RouteTableManager struct {
```
fasaxc: Instead of `Manager`, how about `Allocator`; Manager is always a bit woolly (wish I had a more specific name for the endpoint manager et al)

neiljerram: Yes, have just pushed a change to make this more generic, named IndexAllocator.

PR #2270, opened 2020-04-21
RPF check

tomastigera:
## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
In BPF mode, Calico now does an RPF check on every packet (not just the first packet in a flow).  This prevents workloads on the same host from injecting packets into another workload's flow (and various other attacks).
```


Makefile
```diff
@@ -238,9 +238,9 @@ ifeq ($(ARCH),amd64)
 	docker tag $(BUILD_IMAGE):latest-$(ARCH) $(BUILD_IMAGE):latest
 endif
 
-image-test: image fv/Dockerfile.test.amd64
+image-test: image fv/Dockerfile.test.amd64 bin/pktgen
```
fasaxc, line 241: Think you need test-connection and test-workload here too

bpf-gpl/conntrack.h
```diff
@@ -72,8 +72,12 @@ struct calico_ct_leg {
 	__u32 whitelisted:1;
 
 	__u32 opener:1;
+
+	__u32 ifindex; /* where the packet entered the system from */
 };
 
+#define CT_INVALID_IFINDEX	((__u32) -1)
```
fasaxc, line 79: From what I can tell, 0 is not a valid ifIndex (but all other uint values are) so why not use 0 for the invalid value?

fv/bpf_test.go
```diff
@@ -637,6 +637,157 @@ func describeBPFTests(opts ...bpfTestOpt) bool {
...
+
+						cc.ExpectSome(w[1][0], w[0][0])
+						cc.ExpectSome(w[1][1], w[0][0])
+						cc.CheckConnectivity()
+					})
+
+					/* XXX temporary will be removed and replaced by enfocing it
+					 * XXX from felix in BPF
+					 */
+					By("setting strict RPF check", func() {
```
fasaxc, line 666: Need to move this to felix.

fv/bpf_test.go
```diff
@@ -637,6 +637,157 @@ func describeBPFTests(opts ...bpfTestOpt) bool {
...
+						felixes[1].Exec("sysctl", "-w", "net.ipv4.conf.all.rp_filter=1")
+						felixes[1].Exec("sysctl", "-w", "net.ipv4.conf.default.rp_filter=1")
+					})
+
+					By("testing that packet sent by another workload is dropped", func() {
+						tcpdump := w[0][0].AttachTCPDump()
+						tcpdump.SetLogEnabled(true)
+						matcher := fmt.Sprintf("IP %s\\.30444 > %s\\.30444: UDP", w[1][0].IP, w[0][0].IP)
```
fasaxc, line 674: Best to use backtick quotes for regexps to avoid double-escaping.

Could use `regep.QuoteMeta()`

bpf-gpl/conntrack.h
fasaxc: https://elixir.bootlin.com/linux/latest/source/include/net/flow.h#L18

tomastigera: though that 0 is valid, will change it, good catch

fv/bpf_test.go
tomastigera: Wouldn't QuoteMate also quote `.*` and similar if I changed the regexp? Will probably stick with manual escaping in some form

fasaxc: Fine not to use QuoteMeta; I still think backtick quotes are better for regexps:
```
`IP %s\.30444 > %s\.30444: UDP`
```
if more readable than
```
"IP %s\\.30444 > %s\\.30444: UDP"
```
(and the difference only grows with the complexity of the regexp.)

PR #2282, opened 2020-04-28
changes to support ICMP type and code check in policy in bpf mode

sridhartigera:
yet to do the version check and add traffic cases in bpf/ut


bpf-gpl/jump.h
```diff
@@ -23,14 +23,25 @@
...
+};
 struct cali_tc_state {
 	__be32 ip_src;
 	__be32 ip_dst;
 	__be32 post_nat_ip_dst;
 	__be32 nat_tun_src;
 	__s32 pol_rc;
 	__u16 sport;
-	__u16 dport;
+	struct cali_tc_state_dst dst;
```
fasaxc, line 44: How about using an anonymous union for this; I think it'd be clearer:
```
...
__u16 sport;
union {
    __u16 dport;
    struct {
        __u8 icmp_type;
        __u8 icmp_code;
    };
};
__u16 post_nat_dport;
...
```

Then you can access it as `state->icmp_type`, `state->dport` etc without needing to name the union.

bpf/polprog/pol_prog_builder.go
```diff
@@ -302,6 +304,29 @@ func (p *Builder) writeRule(rule *proto.Rule, passLabel string) {
 		p.writePortsMatch(true, legDest, rule.NotDstPorts, rule.NotDstNamedPortIpSetIds)
 	}
 
+	if rule.IpVersion == 4 {
```
fasaxc, line 307: Don't think you need this check here.  When we start handling ICMPv6 we'll use the same location in the state struct so the match logic will still be correct,.

bpf/polprog/pol_prog_builder.go
```diff
@@ -86,6 +86,8 @@ var (
 	_                            = stateOffDstPort
 	stateOffPostNATDstPort int16 = 24
 	stateOffIPProto        int16 = 26
+	stateOffICMPType       int16 = 22
```
fasaxc, line 89: Please move these up to line 86 so the numbers are in order.

bpf-gpl/tc.c
```diff
@@ -106,7 +106,7 @@ int calico_tc_norm_pol_tail(struct __sk_buff *skb)
 	}
 
 	state->pol_rc = execute_policy_norm(skb, state->ip_proto, state->ip_src,
-					    state->ip_dst, state->sport, state->dport);
+					    state->ip_dst, state->sport, state->dst.u.port);
```
tomastigera, line 109: this is probably gonna get revisited and we will just pass the state, otherwise would perhaps make some sense to reuse both ports rather than a single 16-bit value for 2 8-bit ones. But I had the impression that you want to use the other for some additional id, right?

bpf/polprog/pol_prog_builder.go
sridhartigera: For IPv6,  can we still use Rule_IcmpTypeCode?

bpf-gpl/tc.c
sridhartigera: Using ID in place of src port was for conntrack, if we are to do it like how linux does

bpf/polprog/pol_prog_builder.go
```diff
@@ -302,6 +309,27 @@ func (p *Builder) writeRule(rule *proto.Rule, passLabel string) {
...
+		switch icmp := rule.NotIcmp.(type) {
+		case *proto.Rule_NotIcmpTypeCode:
+			p.writeICMPTypeMatch(true, uint8(icmp.NotIcmpTypeCode.Type))
+			p.writeICMPCodeMatch(true, uint8(icmp.NotIcmpTypeCode.Code))
+		case *proto.Rule_NotIcmpType:
+			p.writeICMPTypeMatch(true, uint8(icmp.NotIcmpType))
+		}
+	}
+
 	// TODO ICMP
```
fasaxc, line 333: Can remove this TODO now :-)

bpf-gpl/tc.c
```diff
@@ -611,6 +611,12 @@ static CALI_BPF_INLINE int calico_tc(struct __sk_buff *skb)
 			}
 		}
 	}
+	if (state.ip_proto == IPPROTO_ICMP)
+	{
+		state.icmp_type = icmp_header->type;
```
fasaxc, line 616: Any reason to leave this initialisation so late?  We initialise state->sport/dport all the way in the switch at line 465.  I'd be inclined to do the initialisation in that switch for symmetry.

bpf/polprog/pol_prog_builder.go
```diff
@@ -205,6 +208,30 @@ func (p *Builder) setUpIPSetKey(ipsetID uint64, keyOffset, ipOffset, portOffset
 	p.b.StoreStack32(R1, keyOffset+ipsKeyID+4)
 }
 
+func validateNet (rule *proto.Rule) (*proto.Rule) {
```
fasaxc, line 211: This formatting looks wrong, you should do `make install-git-hooks` to add the pre-commit check hooks and run `make fix` to format the code before check-in.

fasaxc: I think you should 

* Move this function over to the rules package so it can be used from both places; there's still a lot of duplication between this and `ProtoRuleToIptablesRules`.
* Move the check of rule.IpVersion into the start of this function.  (i.e. return nil if the rule doesn't match the IP version that we're rendering for.)

Also, I think the naming could be improved.  How about `FilterRuleToIPVersion(rule, ipVersion)`?

bpf-gpl/tc.c
sridhartigera: We do conntrack lookup based on state->sport, dport. Since we are not clear with conntrack for icmp, i have added it before we jump into policy

bpf/polprog/pol_prog_builder.go
sridhartigera: ipversion check for policy includes v6 as well right? For BPF, we dont want to go ahead with policy writing if its v6. I feel ipversion check inside this function may not be possible until we support v6 in bpf mode.

sridhartigera: Can have additional flags and check if its BPF mode or not. your comments?

fasaxc: Here's what I'd do: pass in the IP version that you're rendering for:
```
func FilterRuleToIPVersion(ipVersion int, rule *proto.Rule) *proto.Rule {
    if rule.IpVersion != 0 && rule.IpVersion != ipVersion {
        log.Debugf ("Skipping rule it is for a different IP version.")
        return nil
    }

    ...
}
```

Then, in the BPF code, pass in the static `ipVersion=4`.  In the iptables-mode code, pass in the `ipVersion` variable that contains the version that we're rendering for.

bpf-gpl/tc.c
fasaxc: Ah, worth a comment then so someone doesn't make the mistake I made.  I'd probably say something like
```
// icmp_type and icmp_code share storage with the ports; now we've used
// the ports set to 0 to do the conntrack lookup, we can set the ICMP fields 
// for policy.
```

bpf/ut/pol_prog_test.go
```diff
@@ -846,9 +854,55 @@ var polProgramTests = []polProgramTest{
...
+		PolicyName: "allow icmp packet with type not equal to 8 and code not equal to 3",
+		Policy: [][][]*proto.Rule{{{{
+			Action:  "Allow",
+			NotIcmp: &proto.Rule_NotIcmpTypeCode{&proto.IcmpTypeAndCode{Type: 8, Code: 3}},
+		}}}},
+		AllowedPackets: []packet{
+			icmpPkt_with_type_code("10.0.0.1", "10.0.0.2", 10, 0)},
+		DroppedPackets: []packet{
+			icmpPkt_with_type_code("10.0.0.1", "10.0.0.2", 8, 3),
+			icmpPkt_with_type_code("10.0.0.1", "10.0.0.2", 8, 4),
```
fasaxc, line 903: I think this packet should be allowed; I think the match should be `!((type == 8) && (code == 3))` so `code=4` should make the packet accepted

rules/policy.go
```diff
@@ -78,23 +77,20 @@ func filterNets(mixedCIDRs []string, ipVersion uint8) (filtered []string, filter
...
-	//
-	// That handles the mainline case well, where the IP version is missing but the rule is
-	// otherwise consistent since we'll render the rule only for the matching version.
-	//
-	// It also handles rules like "allow from 10.0.0.1,feed::beef" in an intuitive way.  Only
-	// rules of the form "allow from 10.0.0.1,feed::beef to 10.0.0.2" will get filtered out,
-	// and only for IPv6, where there's no obvious meaning to the rule.
-	ruleCopy := *pRule
+func FilterRuleToIPVersion(ipVersion uint8, pRule *proto.Rule) *proto.Rule {
+	ruleCopy := pRule
```
fasaxc, line 81: Since `pRule` is a pointer to a rule, this isn't making a copy of the rule, it's making a copy of the pointer.  That means that the changes below will modify the input rule, which will cause problems.

Couple of ways round that:

- Change this to `ruleCopy = *pRule` and change the return to be `return &ruleCopy` (this is 100% safe in go, unlike in C)
- Change the argument to be `pRule proto.Rule` and the result to be `proto.Rule`; then the copy semantics will be explicit in the signature.

rules/policy.go
```diff
@@ -111,7 +107,28 @@ func (r *DefaultRuleRenderer) ProtoRuleToIptablesRules(pRule *proto.Rule, ipVers
 	if filteredAll {
 		return nil
 	}
+	return ruleCopy
+}
+func (r *DefaultRuleRenderer) ProtoRuleToIptablesRules(pRule *proto.Rule, ipVersion uint8) []iptables.Rule {
```
fasaxc, line 112: Nit: should be a blank line between functions.

rules/policy.go
```diff
@@ -535,6 +550,8 @@ func appendProtocolMatch(match iptables.MatchCriteria, protocol *proto.Protocol,
 	return match
 }
 
+var SkipRule = errors.New("Rule skipped")
+
 func (r *DefaultRuleRenderer) CalculateRuleMatch(pRule *proto.Rule, ipVersion uint8) (iptables.MatchCriteria, error) {
```
fasaxc, line 555: Looks like this function no longer returns an error in any situation so why not remove the error argument and the error check above?

rules/policy.go
```diff
@@ -535,6 +550,8 @@ func appendProtocolMatch(match iptables.MatchCriteria, protocol *proto.Protocol,
 	return match
 }
 
+var SkipRule = errors.New("Rule skipped")
```
fasaxc, line 553: Any reason to move this?   If not, please put it back to reduce the probability of conflicts with later merges/cherry-picks.

bpf-gpl/tc.c
```diff
@@ -611,6 +611,15 @@ static CALI_BPF_INLINE int calico_tc(struct __sk_buff *skb)
 			}
 		}
 	}
+	// icmp_type and icmp_code share storage with the ports; now we've used
+	// the ports set to 0 to do the conntrack lookup, we can set the ICMP fields
+	// for policy.
+	if (state.ip_proto == IPPROTO_ICMP)
+	{
```
tomastigera, line 618: `{` should be at the end of the if line

rules/policy.go
```diff
@@ -111,7 +106,29 @@ func (r *DefaultRuleRenderer) ProtoRuleToIptablesRules(pRule *proto.Rule, ipVers
 	if filteredAll {
 		return nil
 	}
+	return &ruleCopy
+}
+
+func (r *DefaultRuleRenderer) ProtoRuleToIptablesRules(pRule *proto.Rule, ipVersion uint8) []iptables.Rule {
+	// Filter the CIDRs to the IP version that we're rendering.  In general, we should have an
```
tomastigera, line 113: nit: In golang, for the purpose of documentation, it is common (linter may - like in my vi - warn about that) that exported functions are prepended with a comment that starts `// <fucn name>` and explains what the function does. It can be then rendered by godoc.

```
// MyCoolFunction is cool
func MyCoolFunction () {
    ...
}
```

Seems like this comment falls in that category.

The same holds for exported types, variables, etc.

fasaxc: Agree, code would be improved by moving the comment to the new function.

Here, all the reader needs to know is that a rule may not match the IP version that we're rendering for and hence we may filter it out.  Don't even need a comment for that, the code is very clear.

However, the algorithm is worth explaining in the new function since it's a bit subtle.

rules/policy.go
```diff
@@ -78,7 +76,10 @@ func filterNets(mixedCIDRs []string, ipVersion uint8) (filtered []string, filter
 	return
 }
 
-func (r *DefaultRuleRenderer) ProtoRuleToIptablesRules(pRule *proto.Rule, ipVersion uint8) []iptables.Rule {
+//FilterRuleToIPVersion: If the rule applies to the give IP version, returns a copy of the rule
```
fasaxc, line 79: Typo: `s/give/given/`

Also, please put a space after the `//`

PR #2286, opened 2020-04-30
Create a wireguard rule for each local source CIDR

robbrockbank:
## Description
This PR adds a rule per local workload when routing to the wireguard table.

This ensures host to pod traffic will not be encrypted.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```


dataplane/linux/wireguard_mgr.go
```diff
@@ -60,33 +66,57 @@ func (m *wireguardManager) OnUpdate(protoBufMsg interface{}) {
...
+		switch msg.Type {
+		case proto.RouteType_LOCAL_WORKLOAD:
+			if cidr := ip.MustParseCIDROrIP(msg.Dst); cidr == nil {
+				log.Errorf("error parsing RouteUpdate CIDR: %s", msg.Dst)
+			} else if cidr.Version() == 4 {
+				log.Debugf("Update for IPv4 CIDR: %s", cidr)
+				m.localCIDRs.Add(cidr)
+				m.wireguardRouteTable.LocalWorkloadCIDRAdd(cidr)
+			}
+		case proto.RouteType_REMOTE_WORKLOAD:
```
fasaxc, line 78: A CIDR can change from a RouteType_LOCAL_WORKLOAD to something else and in that case you'll get an update rather than a delete followed by an add.  That means you also need to do deletion processing for route updates where the type is no RouteType_LOCAL_WORKLOAD.

fasaxc: Ooh, that also applies to RouteType_REMOTE_WORKLOAD; you might get a CIDR that changes from being RouteType_REMOTE_WORKLOAD to something else too.

dataplane/linux/wireguard_mgr.go
```diff
@@ -60,33 +66,57 @@ func (m *wireguardManager) OnUpdate(protoBufMsg interface{}) {
...
-			m.wireguardRouteTable.EndpointAllowedCIDRAdd(msg.DstNodeName, cidr)
 		}
 	case *proto.RouteRemove:
 		log.WithField("msg", msg).Debug("RouteRemove update")
 		cidr := ip.MustParseCIDROrIP(msg.Dst)
-		if cidr != nil {
+		if cidr == nil {
+			log.Errorf("error parsing RouteRemove CIDR: %s", msg.Dst)
+		} else if cidr.Version() != 4 {
+			log.Debugf("Route removal for non-IPv4 CIDR: %s", cidr)
```
fasaxc, line 94: Aren't you just ignoring this case on the add?  Probably don't need to do the remove here?  (Also the route resolver is IPv4 only right now so you won't ever see an IPv6 CIDR)

routerule/rule_lib.go
```diff
@@ -104,7 +112,8 @@ func RulesMatchSrcFWMark(r, p *Rule) bool {
 		(r.nlRule.Invert == p.nlRule.Invert) &&
 		(r.nlRule.Mark == p.nlRule.Mark) &&
 		(r.nlRule.Mask == p.nlRule.Mask) &&
-		(r.nlRule.Src.String() == p.nlRule.Src.String())
+		((r.nlRule.Src != nil && p.nlRule.Src != nil && r.nlRule.Src.String() == p.nlRule.Src.String()) ||
```
fasaxc, line 115: If Src is a net.IP then it has an Equal() method; any reason not to use that?  (It handles corner cases like an IPv4 being stored in an IPv6-length slice)

wireguard/wireguard.go
```diff
@@ -338,6 +368,26 @@ func (w *Wireguard) EndpointAllowedCIDRRemove(cidr ip.CIDR) {
 	w.setPeerUpdate(name, update)
 }
 
+func (w *Wireguard) LocalWorkloadCIDRAdd(cidr ip.CIDR) {
+	w.logCxt.Debugf("LocalWorkloadCIDRAdd: cidr=%v", cidr)
+	if cidr.Prefix() == ipPrefixLen {
```
fasaxc, line 373: Worth a comment to say why this splitting /32 vs non-/32 makes sense?

wireguard/wireguard.go
```diff
@@ -1256,92 +1306,86 @@ func (w *Wireguard) ensureLinkAddressV4(netlinkClient netlinkshim.Netlink) error
...
-				return err
-			}
-		}
-	}
-
-	return nil
+// createRouteRule creates a routing rule to route a local source CIDR to the wireguard table (if wireguard firewall
+// mark is not set).
+func (w *Wireguard) createRouteRule(cidr ip.CIDR) *routerule.Rule {
+	rule := routerule.NewRule(ipVersion, w.config.RoutingRulePriority).
```
fasaxc, line 1370: Is it OK to have multiple rules with the same priority?

wireguard/wireguard_test.go
```diff
@@ -295,17 +271,17 @@ var _ = Describe("Enable wireguard", func() {
...
 
 				wg.QueueResync()
 				err = wg.Apply()
 				Expect(err).ToNot(HaveOccurred())
 
-				Expect(wgDataplane.AddedRules).To(HaveLen(0))
-				Expect(wgDataplane.DeletedRules).To(HaveLen(1))
-				Expect(wgDataplane.DeletedRules[0]).To(Equal(*incorrectRule))
+				Expect(rrDataplane.AddedRules).To(HaveLen(0))
+				Expect(rrDataplane.DeletedRules).To(HaveLen(1))
```
fasaxc, line 283: You can combine these asserts into one (with better diags, maybe?):
```
Expect(rrDataplane.DeletedRules).To(ConsistOf(*incorrectRule))
```

dataplane/linux/wireguard_mgr.go
robbrockbank: ah yup - didn't mean to do that

routerule/rule_lib.go
robbrockbank: These are IPNet which do not have an Equal method.  I added a helper method to ip to compare ipNet pointers.

wireguard/wireguard.go
robbrockbank: Yup

wireguard/wireguard.go
```diff
@@ -242,119 +272,201 @@ func (w *Wireguard) EndpointUpdate(name string, ipv4Addr ip.Addr) {
...
+	if cidr.Prefix() == ipPrefixLen {
+		w.localIPs.Add(cidr.Addr())
+	} else {
+		w.localCIDRs.Add(cidr)
+	}
+	// Only flag the CIDRs for update if it not wholly covered by the already filtered local CIDRs.
+	contained := false
+	w.localCIDRsFiltered.Iter(func(item interface{}) error {
+		filtered := item.(ip.CIDR).ToIPNet()
+		if filtered.Contains(cidr.ToIPNet().IP) {
```
fasaxc, line 388: By this test, I think `10.0.0.0/32` would contain `10.0.0.0/16`, which is wrong.  You need to check the length too.

robbrockbank: heh.  oh yes, good spot!

dataplane/linux/wireguard_mgr.go
```diff
@@ -60,33 +60,53 @@ func (m *wireguardManager) OnUpdate(protoBufMsg interface{}) {
 		m.wireguardRouteTable.EndpointRemove(msg.Hostname)
 	case *proto.RouteUpdate:
 		log.WithField("msg", msg).Debug("RouteUpdate update")
-		if msg.Type != proto.RouteType_REMOTE_WORKLOAD {
-			log.Debug("RouteUpdate is not a peer workload update, ignoring")
+		if msg.Type == proto.RouteType_CIDR_INFO {
```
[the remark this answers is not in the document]
robbrockbank: I think I misunderstood the comment in the protobuf code - I assumed CIDR_INFO was informational only and not directly related to setting up routes

PR #2287, opened 2020-04-30
Add a kernel version check for BPF mode.

fasaxc:
## Description

Disable BPF mode on kernel < v5.3.0.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
Felix now checks the kernel version before enabling BPF mode.  v5.3.0 is required.
```

### DIFF
--- a/bpf/bpf.go
+++ b/bpf/bpf.go
@@ -94,6 +94,9 @@ var (
 	// v4Dot20Dot0 is the first kernel version that has all the
 	// required features we use for sidecar acceleration
 	v4Dot20Dot0 = versionparse.MustParseVersion("4.20.0")
+	// v5Dot3Dot0 is the first kernel version that has all the
+	// required features we use for BPF dataplane mode
+	v5Dot3Dot0 = versionparse.MustParseVersion("5.3.0")
 )
 
 func (m XDPMode) String() string {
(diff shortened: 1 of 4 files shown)


bpf/bpf.go
```diff
@@ -2188,6 +2191,19 @@ func SupportsSockmap() error {
 	return nil
 }
 
+func SupportsBPFDataplane() error {
```
neiljerram, line 2194: I guess felix-private will need a corresponding  func in bpf_windows.go.

daemon/daemon.go
```diff
@@ -335,6 +332,17 @@ configRetry:
...
+	if configParams.BPFEnabled {
+		// Check for BPF dataplane support before we do anything that relies on the flag being set one way or another.
+		if err := bpf.SupportsBPFDataplane(); err != nil {
+			log.Error("BPF dataplane mode enabled but not supported by the kernel.  Disabling BPF mode.")
+			_, err := configParams.OverrideParam("BPFEnabled", "false")
+			if err != nil {
+				log.WithError(err).Panic("Bug: failed to override config parameter")
+			}
+		}
+	}
```
neiljerram, line 344: So IIUC the override is only computed when someone tries to enable BPF?  I imagined instead that the overrides would be computed independently (and unconditionally) at start of day.  What is your thinking on the preferred architecture here?

fasaxc: I wondered about that, but it made it easier to only log the error in the "enabled but not supported case" if I did it this way.  Also felix restarts if the flag changes so we'll go through this code again if the flag is enabled while felix is running.

If I unconditionally set the override, felix wouldn't restart, but you'd also get no message saying that the feature was disabled so I could go either way.

PR #2289, opened 2020-04-30
WireGuard FV

realgaurav:
## Description

WireGuard FV.

## Todos
[x] All tests passing

## Release Note
```release-note
None required
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -134,6 +134,8 @@ clean:
 	find . -name "*.pyc" -type f -delete
 	$(DOCKER_GO_BUILD) make -C bpf-apache clean
 	$(DOCKER_GO_BUILD) make -C bpf-gpl clean
+	-docker rmi $(BUILD_IMAGE)-wgtool:latest-amd64
+	-docker rmi $(BUILD_IMAGE)-wgtool:latest
 
 ###############################################################################
 # Automated pin updates
@@ -236,10 +238,16 @@ ifeq ($(ARCH),amd64)
 	docker tag $(BUILD_IMAGE):latest-$(ARCH) $(BUILD_IMAGE):latest
 endif
 
-image-test: image fv/Dockerfile.test.amd64 bin/pktgen bin/test-workload bin/test-connection
+image-test: image fv/Dockerfile.test.amd64 bin/pktgen bin/test-workload bin/test-connection image-wgtool
 	docker build -t $(BUILD_IMAGE)-test:latest-$(ARCH) --build-arg QEMU_IMAGE=$(CALICO_BUILD) --file ./fv/Dockerfile.test.$(ARCH) bin;
 ifeq ($(ARCH),amd64)
 	docker tag $(BUILD_IMAGE)-test:latest-$(ARCH) $(BUILD_IMAGE)-test:latest
+endif
+
+image-wgtool: fv/Dockerfile.wgtool.amd64
+	docker build -t $(BUILD_IMAGE)-wgtool:latest-$(ARCH) --file ./fv/Dockerfile.wgtool.$(ARCH) fv;
+ifeq ($(ARCH),amd64)
+	docker tag $(BUILD_IMAGE)-wgtool:latest-$(ARCH) $(BUILD_IMAGE)-wgtool:latest
 endif
 
 ###############################################################################
@@ -398,6 +406,20 @@ fv fv/latency.log: $(REMOTE_DEPS) image-test bin/iptables-locker bin/test-worklo
 
 fv-bpf:
 	$(MAKE) fv FELIX_FV_ENABLE_BPF=true
+
+KO_DIR := "/lib/modules/$(shell uname -r)/"
+WIREGUARD_KO_PATH := $(shell find $(KO_DIR) -name "wireguard.ko")
+fv-wireguard:
+ifndef FORCE_WIREGUARD_FV
+	@if test -z $(WIREGUARD_KO_PATH); then \
+		echo "WireGuard not available."; \
+		exit 1; \
+	else \
+		$(MAKE) fv FELIX_FV_WIREGUARD_AVAILABLE=true GINKGO_FOCUS="WireGuard-Supported"; \
+	fi
+else
+	$(MAKE) fv FELIX_FV_WIREGUARD_AVAILABLE=true GINKGO_FOCUS="WireGuard-Supported"
+endif
 
 ###############################################################################
 # K8SFV Tests
--- a/fv/wireguard_test.go
+++ b/fv/wireguard_test.go
@@ -1,0 +1,482 @@
+// Copyright (c) 2020 Tigera, Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// +build fvtests
+
+package fv_test
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"regexp"
+	"strings"
+	"time"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	"github.com/projectcalico/felix/fv/connectivity"
+	"github.com/projectcalico/felix/fv/infrastructure"
+	"github.com/projectcalico/felix/fv/tcpdump"
+	"github.com/projectcalico/felix/fv/workload"
+	"github.com/projectcalico/libcalico-go/lib/apiconfig"
+	api "github.com/projectcalico/libcalico-go/lib/apis/v3"
+	"github.com/projectcalico/libcalico-go/lib/clientv3"
+	"github.com/projectcalico/libcalico-go/lib/ipam"
+	"github.com/projectcalico/libcalico-go/lib/net"
+	"github.com/projectcalico/libcalico-go/lib/options"
+)
+
+const (
+	wireguardInterfaceNameDefault       = "wireguard.cali"
+	wireguardMTUDefault                 = 1420
+	wireguardRoutingRulePriorityDefault = "99"
+	wireguardListeningPortDefault       = "51820"
+
+	fakeWireguardPubKey = "jlkVyQYooZYzI2wFfNhSZez5eWh44yfq1wKVjLvSXgY="
+)
+
+var _ = infrastructure.DatastoreDescribe("WireGuard-Supported", []apiconfig.DatastoreType{apiconfig.EtcdV3, apiconfig.Kubernetes}, func(getInfra infrastructure.InfraFactory) {
+	const nodeCount = 2
+
+	var (
+		infra        infrastructure.DatastoreInfra
+		felixes      []*infrastructure.Felix
+		client       clientv3.Interface
+		tcpdumps     []*tcpdump.TCPDump
+		wls          [nodeCount]*workload.Workload // simulated host workloads
+		cc           *connectivity.Checker
+		routeEntries [nodeCount]string
+		ruleCIDRs    [nodeCount]string
+	)
+
+	BeforeEach(func() {
+		// Run these tests only when the Host has Wireguard kernel module installed.
+		if os.Getenv("FELIX_FV_WIREGUARD_AVAILABLE") != "true" {
+			Skip("Skipping Wireguard supported tests.")
+		}
+
+		infra = getInfra()
+		felixes, client = infrastructure.StartNNodeTopology(nodeCount, wireguardTopologyOptions(), infra)
+
+		// To allow all ingress and egress, in absence of any Policy.
+		infra.AddDefaultAllow()
+
+		for i := range wls {
+			wlIP := fmt.Sprintf("10.65.%d.2", i)
+			wlName := fmt.Sprintf("wl%d", i)
+
+			err := client.IPAM().AssignIP(context.Background(), ipam.AssignIPArgs{
+				IP:       net.MustParseIP(wlIP),
+				HandleID: &wlName,
+				Attrs: map[string]string{
+					ipam.AttributeNode: felixes[i].Hostname,
+				},
+				Hostname: felixes[i].Hostname,
+			})
+			Expect(err).NotTo(HaveOccurred())
+
+			wls[i] = workload.Run(felixes[i], wlName, "default", wlIP, "8055", "tcp")
+			wls[i].ConfigureInDatastore(infra)
+
+			// Prepare substring to match in rule.
+			ruleCIDRs[i] = fmt.Sprintf("10.65.%d.0/26", i)
+			// Prepare route entry.
+			routeEntries[i] = fmt.Sprintf("10.65.%d.0/26 dev %s scope link", i, wireguardInterfaceNameDefault)
+
+			felixes[i].TriggerDelayedStart()
+		}
+		// Swap route entry to match between workloads.
+		routeEntries[0], routeEntries[1] = routeEntries[1], routeEntries[0]
+
+		cc = &connectivity.Checker{}
+	})
+
+	AfterEach(func() {
+		if CurrentGinkgoTestDescription().Failed {
+			for _, felix := range felixes {
+				felix.Exec("ip", "link", "show")
+				felix.Exec("ip", "rule", "list")
+				felix.Exec("ip", "route", "show", "table", "all")
+				felix.Exec("ip", "route", "show", "cached")
+				felix.Exec("wg")
+			}
+		}
+
+		for _, wl := range wls {
+			wl.Stop()
+		}
+		for _, tcpdump := range tcpdumps {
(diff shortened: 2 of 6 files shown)


fv/wireguard_test.go
```diff
@@ -0,0 +1,444 @@
...
+
+			for i := range felixes {
+				Eventually(tcpdumpMatchCount(i, "numInWorkloadPackets")).Should(BeNumerically(">", 0))
+				Eventually(tcpdumpMatchCount(i, "numOutWorkloadPackets")).Should(BeNumerically(">", 0))
+				Eventually(tcpdumpMatchCount(i, "numTunnelOutboundPackets")).Should(BeNumerically(">", 0))
+				Eventually(tcpdumpMatchCount(i, "numTunnelInboundPackets")).Should(BeNumerically(">", 0))
+			}
+		})
+
+		It("the Wireguard rule-entry/route-table should exist", func() {
```
fasaxc, line 194: Usually, I wouldn't check this level of detail (better to check the rules in UT and then just verify the behaviour here).  However, combined with the negated check when wg is disabled, maybe it's a good cross check.

fv/wireguard_test.go
```diff
@@ -0,0 +1,444 @@
...
+		It("the Wireguard device shouldn't exist", func() {
+			for _, felix := range felixes {
+				Eventually(func() string {
+					out, _ := felix.ExecOutput("ip", "-d", "link", "show", wireguardInterfaceNameDefault)
+					return out
+				}, "10s", "100ms").Should(BeEmpty())
+			}
+		})
+
+		It("the Wireguard rule-entry/route-table shouldn't list wireguard rule", func() {
```
fasaxc, line 324: I guess this is an example where an FV check is a good idea since it's hard to prove a negative otherwise.

fasaxc: Suggest you make a common assertion function between the positive "rule exists" and the negative "rule doesn't exist" case.  Then you can do `Eventually(wireguardRuleExists).Should(BeTrue(), "wireguard rule was missing")` and  `Eventually(wireguardRuleExists).Should(BeFalse(), "wireguard rule was unexpectedly present")`

The presence of the positive version bolsters the negated check: if the code changes so the positive version fails to detect the rule, the FV will fail.

fv/wireguard_test.go
```diff
@@ -0,0 +1,444 @@
...
+				Eventually(func() string {
+					node, err := client.Nodes().Get(context.Background(), felix.Hostname, options.GetOptions{})
+					Expect(err).NotTo(HaveOccurred())
+					wgPubKey := node.Status.WireguardPublicKey
+					return wgPubKey
+				}, "2s", "100ms").Should(BeEmpty())
+			}
+		})
+
+		It("workload connectivity remains but uses un-encrypted tunnel", func() {
```
fasaxc, line 349: You don't actually check the connectivity?

Makefile
```diff
@@ -399,6 +399,18 @@ fv fv/latency.log: $(REMOTE_DEPS) image-test bin/iptables-locker bin/test-worklo
 fv-bpf:
 	$(MAKE) fv FELIX_FV_ENABLE_BPF=true
 
+KO_DIR := "/lib/modules/$(shell uname -r)/"
+WIREGUARD_KO_PATH := $(shell find $(KO_DIR) -name "wireguard.ko")
+fv-wireguard:
```
fasaxc, line 404: When running in CI, I recommend having a force-enable flag that disables the auto-detection, then make sure it's run on a platform that supports it.  We don't want the FV to silently fail in CI, even if it's failing because the autodetection failed.

fv/wireguard_test.go
```diff
@@ -0,0 +1,444 @@
...
+
+		infra = getInfra()
+		felixes, client = infrastructure.StartNNodeTopology(nodeCount, wireguardTopologyOptions(), infra)
+
+		// To allow all ingress and egress, in absence of any Policy.
+		infra.AddDefaultAllow()
+
+		for _, felix := range felixes {
+			// Copy wg binary (from ROOTDIR/bin).
+			felix.Container.EnsureBinary("wg")
```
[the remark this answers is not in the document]
realgaurav: I do have local changes for test-container. I was not sure between choosing an "available" `wg` or installing one.
Thanks for the suggestion. I will back this out and propose installing `wg` in test container.

fv/wireguard_test.go
```diff
@@ -0,0 +1,442 @@
...
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"regexp"
+	"strings"
+	"time"
+
+	"github.com/projectcalico/felix/fv/connectivity"
```
robbrockbank, line 27: nit:  import in wrong place.

fv/wireguard_test.go
```diff
@@ -0,0 +1,442 @@
...
+			return out
+		}, "10s", "100ms").Should(BeEmpty())
+	})
+})
+
+func wireguardTopologyOptions() infrastructure.TopologyOptions {
+	topologyOptions := infrastructure.DefaultTopologyOptions()
+
+	// For attaching tcpdump.
+	topologyOptions.DelayFelixStart = true
```
robbrockbank, line 414: We also need to delay start to ensure the calico-node resource is created before felix is started.

fv/wireguard_test.go
```diff
@@ -0,0 +1,455 @@
...
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+	felixConfig, err := client.FelixConfigurations().Get(ctx, "default", options.GetOptions{})
+	Expect(err).NotTo(HaveOccurred())
+	felixConfig.Spec.WireguardEnabled = &enabled
+	felixConfig, err = client.FelixConfigurations().Update(ctx, felixConfig, options.SetOptions{})
+	Expect(err).NotTo(HaveOccurred())
+}
+
+func wireguardRuleRouteEntryExists(felix *infrastructure.Felix, exp *regexp.Regexp, entry string) bool {
```
robbrockbank, line 438: nit: RuleRouteEntry .... everywhere else you use RoutingRule - this makes it sound like something different.

realgaurav: We are checking both RoutingRule and RouteEntry here, updated to `wireguardRoutingRuleRouteEntryExists`.

robbrockbank: This is a very odd interface.  You have a regex hardcoded into the method (which is fine for the test code), but then you pass in another regex to extract the priority.

What is exp?  Why does it need to be specified?

Also, there are now potentially multiple rules that will be installed.  This might be more useful if, given a "priority" and "table", it returns a list of from CIDRs.... or something like that.  Something that at least allows you to check multiple rules.  You can use it to check if rules have been deleted by checking for no CIDRs.  That kind of thing.

fv/wireguard_test.go
```diff
@@ -0,0 +1,455 @@
...
+	rule, err := felix.ExecOutput("ip", "rule", "show", "pref", wireguardRoutingRulePriorityDefault)
+	Expect(err).NotTo(HaveOccurred())
+	rule = strings.TrimSpace(rule)
+	if rule == "" {
+		return false
+	}
+	Expect(rule).To(MatchRegexp(`\d+:\s+from [\d\.]+/\d+ fwmark 0/0x\d+ lookup \d+`))
+
+	// Check route-table entry for Wireguard device.
+	priority := exp.FindString(rule)
```
robbrockbank, line 449: I think this is supposed to be table?

realgaurav: yeah, table-id. good catch.

fv/wireguard_test.go
```diff
@@ -0,0 +1,455 @@
...
+	defer cancel()
+	felixConfig, err := client.FelixConfigurations().Get(ctx, "default", options.GetOptions{})
+	Expect(err).NotTo(HaveOccurred())
+	felixConfig.Spec.WireguardEnabled = &enabled
+	felixConfig, err = client.FelixConfigurations().Update(ctx, felixConfig, options.SetOptions{})
+	Expect(err).NotTo(HaveOccurred())
+}
+
+func wireguardRoutingRuleRouteEntryExists(felix *infrastructure.Felix, exp *regexp.Regexp, entry string) bool {
+	// Get WireGuard route-table index.
```
robbrockbank, line 439: This is a very odd interface. You have a regex hardcoded into the method (which is fine for the test code), but then you pass in another regex to extract the priority.

What is exp? Why does it need to be specified?

Also, there are now potentially multiple rules that will be installed. This might be more useful if, given a "priority" and "table", it returns a list of from CIDRs.... or something like that. Something that at least allows you to check multiple rules. You can use it to check if rules have been deleted by checking for no CIDRs. That kind of thing.

robbrockbank: Also, the rules and routes have multiple entries and contain different data.  Wouldn't it be better to have separate validation functions for each rather than trying to combine them?

fv/wireguard_test.go
```diff
@@ -0,0 +1,455 @@
...
+
+		It("workload connectivity remains but uses un-encrypted tunnel", func() {
+			cc.ExpectSome(w[0], w[1])
+			cc.ExpectSome(w[1], w[0])
+			cc.CheckConnectivity()
+
+			for _, felix := range felixes {
+				Eventually(func() string {
+					// No tunnel implies un-encrypted communication.
+					out, _ := felix.ExecOutput("wg")
```
robbrockbank, line 349: Need to check for no error?

fv/wireguard_test.go
```diff
@@ -0,0 +1,455 @@
...
+				Eventually(func() string {
+					out, _ := felix.ExecOutput("ip", "-d", "link", "show", ifaceName)
+					return out
+				}, "10s", "100ms").Should(ContainSubstring(fmt.Sprintf("mtu %d", mtu)))
+			}
+
+			// Expect the settings to be changed on the device.
+			for _, felix := range felixes {
+				Eventually(func() string {
+					out, _ := felix.ExecOutput("ip", "-d", "link", "show", wireguardInterfaceNameDefault)
```
robbrockbank, line 213: For situations where you check for zero value, you also need to check for no error.

fv/wireguard_test.go
```diff
@@ -0,0 +1,455 @@
...
+			By("verifying peer tunnel transmit details are complimentary")
+			var sent, rcvd [nodeCount]string
+			for i, felix := range felixes {
+				out, _ := felix.ExecOutput("wg")
+				matches := xferRegExp.FindStringSubmatch(out)
+				Expect(matches).NotTo(BeNil())
+				rcvd[i] = matches[1]
+				sent[i] = matches[2]
+			}
+			Expect(rcvd[0]).To(Equal(sent[1]))
```
robbrockbank, line 300: this is a good idea.  Do we need to check that the wireguard interfaces are configured with the correct peers first though - otherwise isn't there a window where one peer could be configured, but not the other ... so you might get some asymmetric data.

fv/wireguard_test.go
```diff
@@ -0,0 +1,455 @@
...
+			for _, felix := range felixes {
+				Eventually(func() string {
+					out, _ := felix.ExecOutput("ip", "link", "show", wireguardInterfaceNameDefault)
+					return out
+				}, "10s", "100ms").Should(BeEmpty())
+			}
+		})
+
+		It("the Wireguard rule-entry/route-table shouldn't exist", func() {
+			for i, felix := range felixes {
```
robbrockbank, line 325: I mention this below, but I think rule and route checks are different things - much cleaner if you split them out.

fv/wireguard_test.go
```diff
@@ -0,0 +1,455 @@
...
+				exists := wireguardRoutingRuleRouteEntryExists(felix, routingRuleRegExp, routeEntries[i])
+				Eventually(exists).Should(BeFalse(), "wireguard rule unexpectedly present")
```
robbrockbank, line 327: or is it the route?  who can say?  :-)

fv/wireguard_test.go
realgaurav: Probably not. Transmit data in `wg` output applies to each peer section. In a 2 node cluster, if the regex matches it is implied that that Wireguard tunnel is established with the only other pair in this case.

fv/wireguard_test.go
realgaurav: Result from routing-rule check (i.e. table-id) feeds into the route-table entry check. For reliably checking route-entry I am using rule to feed the tableId. To avoid repeating the tableId extraction, I combined the tests in 1.

robbrockbank: But we know that table ID upfront don't we?  It's always the first ID to be selected from the configured range - so we don't need to extract it to then extract the rules..... or am I missing something here?

realgaurav: table-id allocation is a separate code not related to WireGuard. You are right that we will see its value to be 1, at this time. But if a new table-id allocation is ever added in our code-path or if the range changes from 1-250 to some other values, this test will fail for no direct reason. I am trying to avoid that assumption by extracting the actual table-id.

I think FVs are pretty descriptive in pointing to the exact line of failure in a spec. So between the rule and route-entry failure, it should be easy to point to right failure.

fv/wireguard_test.go
```diff
@@ -0,0 +1,455 @@
...
+
+func wireguardRoutingRuleRouteEntryExists(felix *infrastructure.Felix, exp *regexp.Regexp, entry string) bool {
+	// Get WireGuard route-table index.
+	rule, err := felix.ExecOutput("ip", "rule", "show", "pref", wireguardRoutingRulePriorityDefault)
+	Expect(err).NotTo(HaveOccurred())
+	rule = strings.TrimSpace(rule)
+	if rule == "" {
+		return false
+	}
+	Expect(rule).To(MatchRegexp(`\d+:\s+from [\d\.]+/\d+ fwmark 0/0x\d+ lookup \d+`))
```
robbrockbank, line 446: This is probably ok for the first iteration, but I think this check is ultimately too restrictive.  Won't this break if there are multiple matching rules?

We can probably also validate the CIDR and fwmark in these rules are as expected. 

Fine for the first pass though.  We can refine the checks after these go in.

fv/wireguard_test.go
robbrockbank: I'm fine with this for now - but we should update to check the table ID.

I explicitly agreed with Neil that Wireguard would "get in first" with the table index allocation. It is important for the table index to be deterministic to claim that we won't drop traffic unnecessarily.  Having an FV that checked that and stated it explicitly in the event of a failure would probably be a good thing.

fv/wireguard_test.go
```diff
@@ -0,0 +1,455 @@
...
+		if CurrentGinkgoTestDescription().Failed {
+			infra.DumpErrorData()
+		}
+		infra.Stop()
+	})
+
+	It("no Wireguard device exists", func() {
+		Eventually(func() string {
+			out, _ := felixes[0].ExecOutput("ip", "link", "show", wireguardInterfaceNameDefault)
+			return out
```
robbrockbank, line 393: You are comparing a Zero value, so need to check for error.  In general it's safer to just always check the error value as well.

realgaurav: yeah, I added checking error for all `ExecOutput`, testing now.

fv/wireguard_test.go
realgaurav: For the FV setup, I don't see why there will be multiple rules, unless I am missing something.

>We can probably also validate the CIDR and fwmark in these rules are as expected.

Good point. I can look into extending the regex.

Makefile
bcreane: @realgaurav - if this issue is resolved, please mark it so. I'd like for @robbrockbank to help prioritize getting this PR merged.

fv/wireguard_test.go
bcreane: @realgaurav - same here. Resolved?

fv/wireguard_test.go
bcreane: @realgaurav - resolved?

fv/wireguard_test.go
bcreane: @realgaurav - resolved?

realgaurav: yup, it has been addressed in the latest update. I didn't mark any issue resolved because @robbrockbank asked me not to.

bcreane: okay. So back to you @robbrockbank - sounds like all issued are resolved, but we're not able to track whether issues are resolved, by design.

fv/wireguard_test.go
```diff
@@ -0,0 +1,482 @@
...
+			for _, felix := range felixes {
+				Eventually(func() string {
+					node, err := client.Nodes().Get(context.Background(), felix.Hostname, options.GetOptions{})
+					Expect(err).NotTo(HaveOccurred())
+					return node.Status.WireguardPublicKey
+				}, "5s", "100ms").Should(BeEmpty())
+			}
+		})
+
+		It("workload connectivity remains but uses un-encrypted tunnel", func() {
```
robbrockbank, line 350: nit:  tunnel -> connection

fv/wireguard_test.go
robbrockbank: I guess this ok for now, but you should determine whether "probably not" is "actually not", otherwise I'm worried this might be a future source of flakes.  Should be easy enough to test - delete the peer config on the remote end and see if we still see the sent packets increasing, but no packets received.   If we do then we *may* want to 
tighten up the initial testing before we start sending packets, or work with delta packet counts.  Not sure.

PR #2297, opened 2020-05-05
Add local throw routes to wireguard table

robbrockbank:
## Description
Previous PR added local CIDRs as rules so that only traffic from local pods would be routed to the wireguard device.

I realized that we also need to include these rules in the wireguard route table to ensure we don't route over wireguard for local routing.

Ordinarily this wouldn't happen, but when IPs are borrowed it's possible that without the local routes we'll route to remote host for the node CIDR.

PR contains a couple things:
-  I use the processing that calculates the local non-overlapping CIDRs for both calculating the rules and routes.
-  To simplify the route calculation, I am storing the local routes in the "peerData" and "peerUpdateData" .... so I renamed these (back) to be nodeData and nodeUpdateData.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/wireguard/wireguard.go
+++ b/wireguard/wireguard.go
@@ -63,7 +63,7 @@ type noOpConnTrack struct{}
 
 func (*noOpConnTrack) RemoveConntrackFlows(ipVersion uint8, ipAddr net.IP) {}
 
-type peerData struct {
+type nodeData struct {
 	ipv4EndpointAddr      ip.Addr
 	publicKey             wgtypes.Key
 	cidrs                 set.Set
@@ -71,13 +71,13 @@ type peerData struct {
 	routingToWireguard    bool
 }
 
-func newPeerData() *peerData {
-	return &peerData{
+func newNodeData() *nodeData {
+	return &nodeData{
 		cidrs: set.New(),
 	}
 }
 
-func (n *peerData) allowedCidrsForWireguard() []net.IPNet {
+func (n *nodeData) allowedCidrsForWireguard() []net.IPNet {
 	cidrs := make([]net.IPNet, 0, n.cidrs.Len())
 	n.cidrs.Iter(func(item interface{}) error {
 		cidrs = append(cidrs, item.(ip.CIDR).ToIPNet())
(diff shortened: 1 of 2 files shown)


wireguard/wireguard.go
```diff
@@ -10,6 +10,7 @@
 // distributed under the License is distributed on an "AS IS" BASIS,
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
+// See the License for the specific language governing permissions and
```
bcreane, line 13: dup'ed

wireguard/wireguard.go
```diff
@@ -368,9 +371,12 @@ func (w *Wireguard) routeRemove(name string, cidr ip.CIDR) {
...
 
-// Add a local workload CIDR. These CIDRs are used for the source-matched wireguard routes to limit wireguard encryption
-// to traffic to/from local workloads. The workload CIDRs may overlap, in which case the updateAndApplyRouteRules
-// method will determine the minimal set of non-overlapping CIDRs.
+// Add a local workload CIDR. These CIDRs are used for:
+// - the source-matched wireguard routing rules to limit wireguard encryption to traffic to/from local workloads.
+// - add throw routes to the wireguard route table to ensure we throw to the main table for local routing.
+//
+// Note that the workload CIDRs may overlap. This method determines the minimal overlapping set of routes and then
+// programs that as a .
```
fasaxc, line 379: Tails off

wireguard/wireguard.go
```diff
@@ -382,23 +388,30 @@ func (w *Wireguard) localWorkloadCIDRAdd(cidr ip.CIDR) {
...
-		return nil
-	})
-	if !contained {
-		w.localCIDRsUpdated = true
 	}
 }
 
-// Remove a local workload CIDR. These CIDRs are used for the source-matched wireguard routes to limit wireguard
-// encryption to traffic to/from local workloads.
+// Add a local workload CIDR. These CIDRs are used for:
```
fasaxc, line 410: Pasta

PR #2314, opened 2020-05-09
BPF: multi nic rpf fix

tomastigera:
## Description

    bpf: NP pkts after decap need to skip rpf check
    
    When nodes have multiple NICs, one for the private/cluster network and
    one for reaching the outside world, a packet after being decaped from
    the tunnel would be dropped by a strict RPF check because it looks it
    arrived from the private network iface but the best path is straight to
    the public net.

    bpf: do not set BPF_FIB_LOOKUP_OUTPUT after encap
    
    Its is not correct to set it makes the packet to go out the same iface.
    It breaks if forwarding though another iface and FIB is successful.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```


bpf-gpl/nat.h
```diff
@@ -154,18 +154,29 @@ static CALI_BPF_INLINE struct calico_nat_dest* calico_v4_nat_lookup2(__be32 ip_s
...
+				return NULL;
+			}
+
+			if (!cali_rt_is_host(rt)) {
+				CALI_DEBUG("NAT: route dest not a host\n");
+				return NULL;
+			}
+		} else {
+			/* we got here because the original node that forwarded
+			 * it through the tunnel though it is a nodeport, we can
```
fasaxc, line 173: Typo: though

bpf-gpl/nat.h
```diff
@@ -154,18 +154,29 @@ static CALI_BPF_INLINE struct calico_nat_dest* calico_v4_nat_lookup2(__be32 ip_s
...
+			 * use the wildcard nodeport entry.
+			 *
+			 * If the nodes have multiple IPs/NICs, RT entries would
+			 * not know the other IPs of other nodes.
+			 */
+			CALI_DEBUG("NAT: skip route lookup from tunnel, it is a nodeport\n");
```
fasaxc, line 179: I think this misses a case:

- external node connects to new service cluster IP via node 1
- node 1 tunnels to node 2
- node 2 hasn't heard about cluster IP yet, gets NAT miss, treats it as nodeport when it shouldn't

bpf-gpl/bpf.h
```diff
@@ -163,6 +163,7 @@ enum calico_skb_mark {
 	CALI_SKB_MARK_BYPASS                 = CALI_SKB_MARK_SEEN | 0x10000,
 	CALI_SKB_MARK_BYPASS_FWD             = CALI_SKB_MARK_SEEN | 0x30000,
 	CALI_SKB_MARK_BYPASS_FWD_SRC_FIXUP   = CALI_SKB_MARK_SEEN | 0x50000,
+	CALI_SKB_MARK_SKIP_RPF               = CALI_SKB_MARK_SEEN | 0x40000,
```
fasaxc, line 166: The 4-bit overlaps with the 5 above, doe we always check explicitly for the 5, or do we treat the bypass and src fixup bits separately?

bpf-gpl/nat.h
tomastigera: That is true. Unless we know that the other node has 2 IPs we either cannot handle that case or we need to live with the possible incorrectness.

I can rephrase the now relaxed condition to do exactly what it used to do plus if there is an RT miss because in the multi NIC scenario because we do not know that the other host has also another IP, we can then treat it as nodeport. This would still make a possible false assumption about a nodeport.

bpf-gpl/bpf.h
tomastigera: Except pure CALI_SKB_MARK_SEEN (which is a prefix), we always test the whole value. We do not have more bits unless we claim more :-/

bpf-gpl/nat.h
tomastigera: I tried to adjust the condition, but not sure if it is any better. It should be subject of a runtime configuration - eventually.

fv/bpf_test.go
```diff
@@ -1353,6 +1353,22 @@ func describeBPFTests(opts ...bpfTestOpt) bool {
 									Expect(err).NotTo(HaveOccurred())
 								})
 
+								By("setting up routes to .20 net on dest node to trigger RPF check", func() {
+									// set up a dumme interface just for the routing purpose
```
fasaxc, line 1357: typo: dumme

PR #2315, opened 2020-05-12
Drop packets with IP options 

sridhartigera:
If the packet is from or to workload endpoint and if it has IP options, drop the packet. This includes additional UT to verify the above mentioned behavior.


bpf-gpl/tc.c
```diff
@@ -455,6 +455,13 @@ static CALI_BPF_INLINE int calico_tc(struct __sk_buff *skb)
 		}
 	}
 
+	// Drop packets from/to workload Endpoints with IP options
+	if (CALI_F_TO_WEP || CALI_F_FROM_WEP) {
+		if (ip_header->ihl > 5) {
+			CALI_DEBUG("Drop packets with IP options\n");
+			return TC_ACT_SHOT;
```
tomastigera, line 462:
```suggestion
			goto deny;
```

Just to send it through the common path. We should also set `fwd.reason = ` the reasons are in `reasons.h` and we should add a new reason if none applies (most likely)

bpf-gpl/tc.c
```diff
@@ -455,6 +455,13 @@ static CALI_BPF_INLINE int calico_tc(struct __sk_buff *skb)
 		}
 	}
 
+	// Drop packets from/to workload Endpoints with IP options
+	if (CALI_F_TO_WEP || CALI_F_FROM_WEP) {
```
tomastigera, line 459: I think we should do this drop on all TC attach points. We discussed that we should not drop packet that are destined for the host itself, but this would allow the HEP program to drop a packet that is for a workload.

I think we would need to check, after we knew the ihl>5, whether it is for the host IP using routing lookup and `goto allow` as that would skip any further parsing of the packet. Unfortunately, that would give a blind cheque to any packet that is for the host and carries IP option and thus it might be used to bypass any policy in the future, but since we do not have any HEP policy yet, it does not matter much and we will address it once we have the policies and once we decide to handle options.

bpf/ut/ip_options_test.go
```diff
@@ -0,0 +1,44 @@
...
+func TestIPOptionsFromToWEP(t *testing.T) {
+	RegisterTestingT(t)
+
+	iphdr := *ipv4Default
+	iphdr.IHL = 6
+
+	_, _, _, _, pktBytes, err := testPacket(nil, &iphdr, nil, nil)
+	Expect(err).NotTo(HaveOccurred())
+
+	runBpfTest(t, "calico_from_workload_ep", rulesDefaultAllow, func(bpfrun bpfProgRunFn) {
```
tomastigera, line 32: I think we should run the test for all from/to wep/hep combinations and if my comment above on the HEP is correct, we should also have a test that would not drop a packet to HEP with the host ip and should not create any CT entry.

bpf-gpl/tc.c
fasaxc: This comment needs to be addresed

bpf-gpl/tc.c
```diff
@@ -680,7 +686,10 @@ int calico_tc_skb_accepted_entrypoint(struct __sk_buff *skb)
 		goto deny;
 	}
 	ip_header = skb_iphdr(skb);
-
+	if (ip_header->ihl > 5) {
```
fasaxc, line 689: Is this one needed?  This is in the epilogue program so the packet should have already been dropped above if it had options.

bpf-gpl/tc.c
```diff
@@ -455,6 +455,12 @@ static CALI_BPF_INLINE int calico_tc(struct __sk_buff *skb)
 		}
 	}
 
+	// Drop packets with IP options
+	if (ip_header->ihl > 5) {
```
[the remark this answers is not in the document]
fasaxc: As @tomastigera said, it'd make sense to do a routing lookup here and allow packets:

- in from_hep if the dest IP is a local host IP
- in to_hep

That would stop us from blocking traffic to/from the host itself if the host uses options.

bpf-gpl/tc.c
sridhartigera: yes. This is not needed

bpf-gpl/tc.c
sridhartigera: I am not sure if we will know whether a packet is for the host IP before NAT. @fasaxc @tomastigera please confirm.

bpf-gpl/tc.c
```diff
@@ -455,6 +455,25 @@ static CALI_BPF_INLINE int calico_tc(struct __sk_buff *skb)
...
+		fwd.reason = CALI_REASON_IP_MALFORMED;
+		CALI_DEBUG("Drop malformed IP packets\n");
+		goto deny;
+	}
+	else if (ip_header->ihl > 5) {
+		/* Drop packets with IP options from/to WEP.
+		 * Also drop packets with IP options if the dest IP is not host IP
+		 */
+		if (CALI_F_FROM_WEP || CALI_F_TO_WEP ||
+			!cali_rt_flags_local_host(cali_rt_lookup_flags(ip_header->daddr))) {
```
fasaxc, line 469: I think you should only check the dest IP in the FROM_HEP program.  TO_HEP is egress from the host so it'd be the source IP that'd be relevant there.

No need to check the source IP in TO_HEP since either:
* it's from the host so we should allow
* it's from a workload, but then we'd have dropped it already in the FROM_WEP hook.

bpf-gpl/tc.c
```diff
@@ -455,6 +455,25 @@ static CALI_BPF_INLINE int calico_tc(struct __sk_buff *skb)
...
+		fwd.reason = CALI_REASON_IP_MALFORMED;
+		CALI_DEBUG("Drop malformed IP packets\n");
+		goto deny;
+	}
+	else if (ip_header->ihl > 5) {
+		/* Drop packets with IP options from/to WEP.
+		 * Also drop packets with IP options if the dest IP is not host IP
+		 */
+		if (CALI_F_FROM_WEP || CALI_F_TO_WEP ||
+			(CALI_F_FROM_HEP && !cali_rt_flags_local_host(cali_rt_lookup_flags(ip_header->daddr)))) {
```
tomastigera, line 469: I have introduced `rt_addr_is_local_host()` for another usecase in https://github.com/projectcalico/felix/pull/2326

PR #2316, opened 2020-05-12
Reduce cases where we flap BPF programs, improve diags

fasaxc:
## Description

* Check for changes before re-applying BPF progam.
* Improve diags in attach.go.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
In BPF mode, reduce the number of updates that trigger BPF programs to be replaced.
```

### DIFF
--- a/bpf/maps.go
+++ b/bpf/maps.go
@@ -176,7 +176,7 @@ func (b *PinnedMap) Iter(f MapIter) error {
 	args := cmd[1:]
 
 	printCommand(prog, args...)
-	output, err := exec.Command(prog, args...).CombinedOutput()
+	output, err := exec.Command(prog, args...).Output()
 	if err != nil {
 		return errors.Errorf("failed to dump in map (%s): %s\n%s", b.versionedFilename(), err, output)
 	}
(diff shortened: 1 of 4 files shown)


fasaxc:
> Do I get it right that the flapping is reduced by if we manage to renew dhcp lease within the deferred work interval, we would not see it as a dirty iface?

Yes, exactly, if the IP remove and re-add happens quickly enough then we suppress the remove.

bpf/maps.go
```diff
@@ -176,7 +176,7 @@ func (b *PinnedMap) Iter(f MapIter) error {
 	args := cmd[1:]
 
 	printCommand(prog, args...)
-	output, err := exec.Command(prog, args...).CombinedOutput()
+	output, err := exec.Command(prog, args...).Output()
```
fasaxc, line 179: Hit a flake here where I think there may have been some output to stderr that got mixed in with the output from Stdout.  Hence switching to `Output()`, which does also capture stderr as `err.Stderr`

tomastigera:
> > Do I get it right that the flapping is reduced by if we manage to renew dhcp lease within the deferred work interval, we would not see it as a dirty iface?
> 
> ~Yes, exactly, if the IP remove and re-add happens quickly enough then we suppress the remove.~
> 
> #2311 makes it so we suppress the IP deletion event if it is followed quickly by an add. This PR 

Now it makes more sense :-)

PR #2317, opened 2020-05-13
BPF: VXLAN tunnel source/dest policing

tomastigera:
## Description

    bpf: check source of tunnel pkts on both nodes
    
    Since we also record the next hop node when we first receive the
    nodeport packet, we can also change if we receive the packets from the
    same node back.

    bpf: no RP lookup for conntracked nodeport tunnel
    
    We remember the destination node when we first create the conntrack
    entry and we use the recorded destination node when we do consequent
    conntrack lookups.

    bpf rename calico_ct_result->tun_ret_ip to tun_ip

    bpf: rename ct_ctx->nat_tun_src to tun_ip
    
    the same for state

    bpf: store tunnel dest in conntrack

    bpf: check source of tunnel pkts on dest node
    
    The source node IP gets recorded when we first approve it and when we
    first create a conntrack record. If the source node IP changed, we drop
    such packets unless we are able to confirm that the node IP changed but
    the node is the same. We cannot do that atm.

    bpf: do not accept tunnel from non-node src IP
    
    Prevent accepting decaped packet if sent from a src IP that is not a
    known node IP. We fail to create the necessary conntrack entry and that
    drop the packet.
    
    We defer the check to conntrack creation as if done earlier, we would
    need to do a routing lookup for every packet.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/bpf-gpl/conntrack.h
+++ b/bpf-gpl/conntrack.h
@@ -58,8 +58,11 @@ enum cali_ct_type {
 					 */
 };
 
-#define CALI_CT_FLAG_NAT_OUT	0x01
-#define CALI_CT_FLAG_DSR_FWD	0x02 /* marks entry into the tunnel on the fwd node when dsr */
+#define CALI_CT_FLAG_NAT_OUT	(1 << 0)
+#define CALI_CT_FLAG_DSR_FWD	(1 << 1) /* marks entry into the tunnel on the fwd node when dsr */
+#define CALI_CT_FLAG_NP_FWD	(1 << 2) /* marks entry into the tunnel on the fewd node */
+
+#define ct_result_np_node(res)		((res).flags & CALI_CT_FLAG_NP_FWD)
 
 struct calico_ct_leg {
 	__u32 seqno;
@@ -215,10 +218,10 @@ create:
 	ct_value.flags = ctx->flags;
 	CALI_DEBUG("CT-ALL tracking entry flags 0x%x\n", ct_value.flags);
 
-	if (type == CALI_CT_TYPE_NAT_REV && ctx->nat_tun_src) {
+	if (type == CALI_CT_TYPE_NAT_REV && !(ctx->flags & CALI_CT_FLAG_NP_FWD) && ctx->nat_tun_src) {
 		struct cali_rt *rt = cali_rt_lookup(ctx->nat_tun_src);
 		if (!rt || !cali_rt_is_host(rt)) {
-			CALI_DEBUG("CT-ALL nat tunnel src not a host %x\n", be32_to_host(ctx->nat_tun_src));
+			CALI_DEBUG("CT-ALL nat tunnel IP not a host %x\n", be32_to_host(ctx->nat_tun_src));
 			err = -1;
 			goto out;
 		}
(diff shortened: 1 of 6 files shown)


bpf-gpl/conntrack.h
```diff
@@ -58,8 +58,11 @@ enum cali_ct_type {
 					 */
 };
 
-#define CALI_CT_FLAG_NAT_OUT	0x01
-#define CALI_CT_FLAG_DSR_FWD	0x02 /* marks entry into the tunnel on the fwd node when dsr */
+#define CALI_CT_FLAG_NAT_OUT	(1 << 0)
+#define CALI_CT_FLAG_DSR_FWD	(1 << 1) /* marks entry into the tunnel on the fwd node when dsr */
+#define CALI_CT_FLAG_NP_FWD	(1 << 2) /* marks entry into the tunnel on the fewd node */
```
fasaxc, line 63: Typo: fewd

bpf-gpl/tc.c
```diff
@@ -896,7 +896,10 @@ static CALI_BPF_INLINE struct fwd calico_tc_skb_accepted(struct __sk_buff *skb,
 		// If we get here, we've passed policy.
 
 		if (nat_dest == NULL) {
-			conntrack_create(&ct_nat_ctx, CT_CREATE_NORMAL);
+			if (conntrack_create(&ct_nat_ctx, CT_CREATE_NORMAL)) {
+				CALI_DEBUG("Creating normal conntrack failed\n");
```
fasaxc, line 906: Change to deny here; is that right if policy has accepted and we're creating a normal entry (so can't hit the new source filtering logic)?

bpf-gpl/conntrack.h
```diff
@@ -127,7 +127,7 @@ struct ct_ctx {
 	__u16 dport;
 	__u16 orig_dport;
 	struct tcphdr *tcp;
-	__be32 nat_tun_src;
+	__be32 tun_ip;
```
fasaxc, line 130: Please add comment explaining which types of NAT entry have this set and what it is set to in each case.

bpf-gpl/tc.c
tomastigera: Hmmm yeah we might let this through, subsequent packets may be luckier and create an entry. If not, it is just a small perf issue in this direction. However, what if policy does not allow untracked packets back? It is as bad as dropping now, isn't it? I can revert this for sure.

tomastigera: I will keep it this way for now

bpf-gpl/tc.c
```diff
@@ -922,31 +931,53 @@ static CALI_BPF_INLINE struct fwd calico_tc_skb_accepted(struct __sk_buff *skb,
...
+					reason = CALI_REASON_RT_UNKNOWN;
+					goto deny;
+				}
+				CALI_DEBUG("rt found for 0x%x local %d\n",
+						be32_to_host(state->post_nat_ip_dst), !!cali_rt_is_local(rt));
+
+				encap_needed = !cali_rt_is_local(rt);
+			}
+
+			if (encap_needed) {
```
fasaxc, line 957: Duplicate condition?  Could these two blocks be combined?

bpf-gpl/tc.c
```diff
@@ -922,31 +931,53 @@ static CALI_BPF_INLINE struct fwd calico_tc_skb_accepted(struct __sk_buff *skb,
...
 		 */
 		if (ct_rc == CALI_CT_NEW) {
-			if (CALI_F_DSR && CALI_F_FROM_HEP &&
-					encap_needed && state->nat_tun_src == 0) {
-				ct_nat_ctx.flags |= CALI_CT_FLAG_DSR_FWD;
+			struct cali_rt * rt;
+			int nat_type = CT_CREATE_NAT;
+
+			if (encap_needed) {
+				rt = cali_rt_lookup(state->post_nat_ip_dst);
```
fasaxc, line 946: Think these blocks need comments explaining the "why".

bpf-gpl/tc.c
tomastigera: The first block may change the condition for the second, but we can merge them, There was previously more code in-between. Done.

PR #2318, opened 2020-05-13
Examine the local routing table instead of link addresses

fasaxc:
## Description

This makes Felix aware of local IPs that are only programmed into the local routing table but not actually programmed as an address on an interface.  This is an unusual corner case but there's at least one cloud (GCP) that programs additional local IPs into the local routing table.  In GCP's case this happens when you point a load balancer at the node.

## Todos
- [x] Release note

## Release Note

```release-note
Felix now derives its list of interface addresses from the local routing table instead of the interface address list.  This picks up IPs that have been added only to the local routing table, such as load balancer external IPs in GCE.
```

### DIFF
--- a/ifacemonitor/update_filter.go
+++ b/ifacemonitor/update_filter.go
@@ -20,8 +20,10 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/davecgh/go-spew/spew"
 	"github.com/sirupsen/logrus"
 	"github.com/vishvananda/netlink"
+	"golang.org/x/sys/unix"
 
 	timeshim "github.com/projectcalico/felix/time"
 )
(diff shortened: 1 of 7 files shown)


ifacemonitor/update_filter.go
```diff
@@ -98,19 +100,29 @@ mainLoop:
 					ReadyAt: u.Time.Now().Add(delay),
 					Update:  linkUpd,
 				})
-		case addrUpd := <-addrInC:
-			idx := addrUpd.LinkIndex
+		case routeUpd := <-routeInC:
+			logrus.WithField("route", spew.Sdump(routeUpd)).Debug("Route update")
```
neiljerram, line 104: Curious what Sdump output looks like.  Does it cross multiple lines, and if so does that matter and look OK in the log output?

fasaxc: Didn't intend to leave that in, it'll calculate the string every time even if debug is disabled.

PR #2319, opened 2020-05-15
Avoid programming every policy to every iptables table.

fasaxc:
## Description

Reduces the amount of time spent in iptables by a factor of 3 when updating policies since, in practice each policy can only ever be in one table at a time.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
In iptables mode, Felix now programs policy only to those iptables tables where it is needed.  This can give a 3x reduction in dataplane apply times with lots of policies in use.
```


fasaxc:
Tried a different approach which came out cleaner: #2323

PR #2321, opened 2020-05-18
Changes to use syscall version for deleting BPF maps

sridhartigera:
## Description
Deleting BPF map was done using bpftool instead of the syscall version. This is changed to use the syscall version. This is expected to be faster compared to the current method. 
1) To get or update BPF maps bpf_attr union is filled with fd, key and the value. bpf_attr_setup_map_elem is used to fill the attribute union.
2) In case of deletion, only fd and key is needed. Hence bpf_attr_setup_map_elem cannot be used. bpf_attr_setup_map_elem_for_delete is defined to fill only the key and the fd of the union. 
3) Unit testing done and pass logs attached.
## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```


bpf/maps.go
```diff
@@ -204,33 +203,11 @@ func (b *PinnedMap) Get(k []byte) ([]byte, error) {
...
-				logrus.WithField("k", k).Debug("Item didn't exist.")
-				return os.ErrNotExist
-			}
-		}
-		logrus.WithField("out", string(out)).Error("Failed to run bpftool")
+	if b.perCPU {
+		logrus.Panic("Per-CPU operations not implemented")
 	}
-	return err
+	return DeleteMapEntry(b.fd, k, b.ValueSize)
```
fasaxc, line 210: The code that calls PinnedMap.Delete is expecting a non-exists error if the KV is missing and some other error if not.  I think this code should translate the errno errors from DeleteMapEntry to distinct errors for missing key and missing map.

You either need to translate errors here or you need to adjust the calling code.

Note: the error behaviour in the calling code isn't well covered by tests since it only affects how we log.

PR #2323, opened 2020-05-18
Only program iptables chains that are referenced.

fasaxc:
## Description

This avoids programming policy chains that are not in use in the particular table. Should significantly reduce iptables use and contention when updating large policies.

When Felix was first designed we were able to run itpables-restore in parallel to mitigate the slowness of programming policies multiple times.  However, the ability to run in parallel was removed once iptables-restore adopted the xtables lock with no way to disable it.

## Todos
- [x] Release note

## Release Note

```release-note
Felix now only programs policies into the iptables tables where they are in use.  This significantly reduces iptables use and contention when updating policies on a system with lots of iptables rules.
```

### DIFF
--- a/iptables/actions.go
+++ b/iptables/actions.go
@@ -21,7 +21,7 @@ type Action interface {
 }
 
 type Referrer interface {
-	ReferredToChain() string
+	ReferencedChain() string
 }
 
 type GotoAction struct {
@@ -37,11 +37,11 @@ func (g GotoAction) String() string {
 	return "Goto->" + g.Target
 }
 
-func (g GotoAction) ReferredToChain() string {
+func (g GotoAction) ReferencedChain() string {
 	return g.Target
 }
 
-var _ Referrer = (*GotoAction)(nil)
+var _ Referrer = GotoAction{}
 
 type JumpAction struct {
 	Target   string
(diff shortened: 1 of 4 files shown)


iptables/actions.go
```diff
@@ -20,6 +20,10 @@ type Action interface {
 	ToFragment(features *Features) string
 }
 
+type Referrer interface {
+	ReferredToChain() string
```
neiljerram, line 24: "ReferencedChain" ?  "To" suggests "To" vs "From" to me, as that occurs a lot in our iptables handling code.

iptables/actions.go
```diff
@@ -33,6 +37,12 @@ func (g GotoAction) String() string {
 	return "Goto->" + g.Target
 }
 
+func (g GotoAction) ReferredToChain() string {
+	return g.Target
+}
+
+var _ Referrer = (*GotoAction)(nil)
```
neiljerram, line 44: I see CI is passing, so I'm curious why that's OK.  I'd expect an error like "*GotoAction does not satisfy Referrer interface, but GotoAction does".

iptables/table.go
```diff
@@ -196,7 +196,12 @@ type Table struct {
 	// "--match foo --jump DROP" (i.e. omitting the action and chain name, which are calculated
 	// as needed).
 	chainNameToChain map[string]*Chain
-	dirtyChains      set.Set
+	// chainRefCounts counts the number of chains that refer to a given chain.  Transitive
+	// reachabilty isn't tracked but testing whether a chain is referenced does allow us to
```
neiljerram, line 200: sp: reachabilty

iptables/actions.go
fasaxc: I believe non-pointer receiver methods can be called on pointers but not the other way around.  That said, I've changed this to use the non-pointer type since that's what I actually require.

PR #2327, opened 2020-05-20
Git pre-commit check for license, copyright in .c and .h files

sridhartigera:
## Description
1) New script to check for GPL, apache license and copyright in .c and .h files under bpf-gpl and bpf-apache directories
2) Fixed the license and copyright statements in files under bpf-gpl and bpf-apache directory.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```


bpf-apache/bpf.h
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2019 Tigera, Inc. All rights reserved.
+// Copyright (c) 2020 Tigera, Inc. All rights reserved.
```
fasaxc, line 1: Copyright _year_ should only be checked for modified files.  You're only supposed to update it when the file is changed.

git-hooks/pre-commit-in-container
```diff
@@ -250,6 +250,6 @@ if $go_apache_check_failed || $go_copyright_check_failed || \
    $junit_hook_check_failed || $stale_file_in_skip_files; then
   exit 1
 fi
-
+exec ./git-hooks/pre-commit-gpl-apache
 # Remove the trap handler.
 trap "" EXIT
```
fasaxc, line 255: This line is now not hit because `exec` doesn't return.

Feels like a bit of a code smell to have one script with lots of low-level function call another similar script at the end.  I think you should either add the check to this file, even though it's getting a bit big, or, refactor so there's a short script in pre-commit-in-container that invokes other scripts to do well-defined pieces of work.

PR #2329, opened 2020-05-20
Disable accept_ra in workloads

lmm:
## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/dataplane/linux/endpoint_mgr.go
+++ b/dataplane/linux/endpoint_mgr.go
@@ -193,7 +193,6 @@ type endpointManager struct {
 	OnEndpointStatusUpdate EndpointStatusUpdateCallback
 	callbacks              endpointManagerCallbacks
 	bpfEnabled             bool
-	openStackActive        bool
 }
 
 type EndpointStatusUpdateCallback func(ipVersion uint8, id interface{}, status string)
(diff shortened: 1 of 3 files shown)


neiljerram:
@lmm @caseydavenport If accept_ra is disabled, doesn't that mean that a workload won't get an IPv6 default route ?  I'm pretty sure that RAs are critical for this purpose for OpenStack VMs; is the situation different for Pod workloads?

dataplane/linux/endpoint_mgr.go
```diff
@@ -1047,6 +1047,12 @@ func (m *endpointManager) configureInterface(name string) error {
 			"Skipping configuration of interface because it is oper down.")
 		return nil
 	}
+
+	err := m.writeProcSys(fmt.Sprintf("/proc/sys/net/ipv6/conf/%s/accept_ra", name), "0")
```
neiljerram, line 1051: I'm concerned that this will break IPv6 for an OpenStack VM. If I'm correctly understanding that this will prevent RAs getting through to the workload, I believe that means that a VM won't set up an IPv6 default route.

PR #2337, opened 2020-05-22
Fix automatic disabling of BPF mode on older kernels

fasaxc:
## Description

Previously, when processing env vars, we'd do a lookup for, say 'bpfenabled', when the map contained 'BPFEnabled'.  This would result in preferring the env var over an internal override.

Also, fix that we had two BPF supported checks in different places and combine them into one.

## Todos
- [x] Integration tests (delete as appropriate) Not needed
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
Fix config inheritance so that the BPF kernel version check takes precedence  over environment variables.
```

### DIFF
--- a/config/config_params_test.go
+++ b/config/config_params_test.go
@@ -155,9 +155,27 @@ var _ = Describe("Config override empty", func() {
 			Expect(cp.BPFEnabled).To(BeFalse())
 
 			By("Ignoring a lower-priority config update")
-			changed, err = cp.UpdateFrom(map[string]string{"BPFEnabled": "true"}, EnvironmentVariable)
+			// Env vars get converted to lower-case before calling UpdateFrom.
+			changed, err = cp.UpdateFrom(map[string]string{"bpfenabled": "true"}, EnvironmentVariable)
 			Expect(changed).To(BeFalse())
 			Expect(err).NotTo(HaveOccurred())
+			Expect(cp.BPFEnabled).To(BeFalse())
+		})
+	})
+
+	Describe("with env var set", func() {
+		BeforeEach(func() {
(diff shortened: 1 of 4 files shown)


daemon/daemon.go
```diff
@@ -359,11 +359,6 @@ configRetry:
 
 	// Start up the dataplane driver.  This may be the internal go-based driver or an external
 	// one.
-	if configParams.BPFEnabled && !bpf.SyscallSupport() {
```
fasaxc, line 362: This was an out-of-date dupe of the check on line 335

config/config_params_test.go
```diff
@@ -155,12 +155,30 @@ var _ = Describe("Config override empty", func() {
 			Expect(cp.BPFEnabled).To(BeFalse())
 
 			By("Ignoring a lower-priority config update")
-			changed, err = cp.UpdateFrom(map[string]string{"BPFEnabled": "true"}, EnvironmentVariable)
+			// Env vars get converted to lower-case before calling UpdateFrom.
+			changed, err = cp.UpdateFrom(map[string]string{"bpfenabled": "true"}, EnvironmentVariable)
```
neiljerram, line 159: I see other test places that are still passing a mixed case name here. Should they be updated too?

(dataplane/linux/int_dataplane_test.go, daemon/daemon_test.go, and other places in this file)

neiljerram: OK, but those other cases that I noted are also using `EnvironmentVariable`.

fasaxc: Ah, I see what you mean; would be more realistic but I think the bases are covered now so would rather avoid another respin on this PR.

PR #2347, opened 2020-05-25
bpf: HOST_IP represents the same node ip on all host ifaces

tomastigera:
Instead of having the iface ip hardwired in the attached tc program, we
hardwire the single node IP that is used to pass packets within the
cluster (the NP vxlan tunnel) and to respond with ICMP errors.

This allows, for instance, for scenarious with multiple NICs that share
the same IP, but the IP is not assinged to any of them, but is
represented by local routes.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```
In BPF mode, Calico now uses the node IP for generated encap packets.
```

### DIFF
--- a/fv/bpf_test.go
+++ b/fv/bpf_test.go
@@ -1351,6 +1351,12 @@ func describeBPFTests(opts ...bpfTestOpt) bool {
 									// Add a route to felix[1] to be able to reach the nodeport
 									_, err = eth20.RunCmd("ip", "route", "add", felixes[1].IP+"/32", "via", "10.0.0.20")
 									Expect(err).NotTo(HaveOccurred())
+									// This multi-NIC scenario works only if the kernel's RPF check
+									// is not strict so we need to override it for the test and must
+									// be set properly when product is deployed. We reply on
+									// iptables to do require check for us.
+									felixes[1].Exec("sysctl", "-w", "net.ipv4.conf.eth0.rp_filter=2")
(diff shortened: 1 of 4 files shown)


fv/bpf_test.go
```diff
@@ -1351,6 +1351,12 @@ func describeBPFTests(opts ...bpfTestOpt) bool {
 									// Add a route to felix[1] to be able to reach the nodeport
 									_, err = eth20.RunCmd("ip", "route", "add", felixes[1].IP+"/32", "via", "10.0.0.20")
 									Expect(err).NotTo(HaveOccurred())
+									// This multi-NIC scenario works only if the kernel's RPF check
```
fasaxc, line 1354: Can you explain to me what goes wrong here?  Can the test be adjusted to set up working routing instead?

tomastigera: When we receive a NP packet on an iface the encaped packet pkt will have src IP == node IP which is local but from a different iface - the one through which it will be most likely forwarded. Seems like strict rpf does not like it even when accept local is enabled.

I think setting up local routes would fix it. It would be the same as the dual ToR (?) in GCE.

tomastigera: For the sake of this one test, imo it does not matter how we fix the problem as in production, things will highly depend on how the networking is configured. So I am thinking about leaving it this way for now.

fasaxc:
CI was a Wireguard flake that I think is fixed on master.

PR #2348, opened 2020-05-26
Address WireGuard FV flakes

realgaurav:
## Description

Address WireGuard FV flakes:
* adjust **wireguard disabled** tests
* adjust **wireguard config change** test
* increase timeout

## Release Note
```release-note
None required
```


fasaxc:
@realgaurav Merged, but any ideas why it's taking so long for felix to restart, for example?  Normally 5s is long enough since datastore changes get signalled in <100ms and felix has a 2 second timer before it restarts.

PR #2351, opened 2020-05-26
Update of libcalico-go to use new CRDs

caseydavenport:
## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -63,4 +63,5 @@ htmlcov/
 
 # CRDs used in the FV tests, pulled from libcalico.
 fv/infrastructure/crds.yaml
+fv/infrastructure/crds
 Makefile.common*
--- a/fv/infrastructure/infra_k8s.go
+++ b/fv/infrastructure/infra_k8s.go
@@ -256,13 +256,13 @@ func setupK8sDatastoreInfra() (*K8sDatastoreInfra, error) {
 
 	log.Info("Started controller manager.")
 
-	// Copy CRD registration manifest into the API server container, and apply it.
-	err := kds.k8sApiContainer.CopyFileIntoContainer("infrastructure/crds.yaml", "/crds.yaml")
+	// Copy CRD registration manifests into the API server container, and apply it.
+	err := kds.k8sApiContainer.CopyFileIntoContainer("infrastructure/crds", "/crds")
 	if err != nil {
 		TearDownK8sInfra(kds)
 		return nil, err
 	}
-	err = kds.k8sApiContainer.ExecMayFail("kubectl", "apply", "-f", "/crds.yaml")
+	err = kds.k8sApiContainer.ExecMayFail("kubectl", "apply", "-f", "/crds/")
 	if err != nil {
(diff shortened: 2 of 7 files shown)


.gitignore
```diff
@@ -63,4 +63,5 @@ htmlcov/
 
 # CRDs used in the FV tests, pulled from libcalico.
 fv/infrastructure/crds.yaml
```
neiljerram, line 65: Should this entry for crds.yaml be removed now?

fv/infrastructure/infra_k8s.go
```diff
@@ -256,13 +256,13 @@ func setupK8sDatastoreInfra() (*K8sDatastoreInfra, error) {
...
-	// Copy CRD registration manifest into the API server container, and apply it.
-	err := kds.k8sApiContainer.CopyFileIntoContainer("infrastructure/crds.yaml", "/crds.yaml")
+	// Copy CRD registration manifests into the API server container, and apply it.
+	err := kds.k8sApiContainer.CopyFileIntoContainer("infrastructure/crds", "/crds")
 	if err != nil {
 		TearDownK8sInfra(kds)
 		return nil, err
 	}
-	err = kds.k8sApiContainer.ExecMayFail("kubectl", "apply", "-f", "/crds.yaml")
+	err = kds.k8sApiContainer.ExecMayFail("kubectl", "apply", "-f", "/crds/")
```
neiljerram, line 265: https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#apply indicates we'd need `-k` for applying a directory, instead of `-f` ?

k8sfv/run-test
```diff
@@ -175,12 +175,12 @@ if [ ! -e ${CRD_PATH} ]; then
...
 
 # Copy CRD manifest into running k8sfs-apiserver
-while ! docker cp ${CRD_PATH} ${prefix}-apiserver:/crds.yaml; do
+while ! docker cp ${CRD_PATH} ${prefix}-apiserver:/crds; do
     sleep 2
 done
 
 # Apply CRD manifest to pre-create resource definitions
-while ! docker exec ${prefix}-apiserver kubectl apply -f /crds.yaml; do
+while ! docker exec ${prefix}-apiserver kubectl apply -f /crds; do
```
neiljerram, line 183: Again, maybe need `-k` here.

.gitignore
caseydavenport: I left it because you might switch between new / old versions of the repo that use the old file.

caseydavenport:
@neiljerram thanks for the review! I think I've addressed the comments.

caseydavenport:
@neiljerram it turns out `-k` is not what we want here, since it's specifically for kustomize files

> error: unable to find one of 'kustomization.yaml', 'kustomization.yml' or 'Kustomization' in directory '/crds'
3183

I think `-f` is what we want, even for directories.

PR #2353, opened 2020-05-27
Adding support for Load balancer SourceIPRange 

sridhartigera:
LBSourceIPRange is specified to allow access to the cluster only from certain source IP range
In order achieve this, packets are to be dropped in the NAT lookup stage if its from an unauthorised source. Hence NAT_FE map is used.
1) NAT_FE is changed from hash to LPM_TRIE to match source IP ranges
2) Prefix len and source IP is added as the key to NAT FE map.
3) By default, the prefixlen for NAT lookup is 56 with <dest IP, dest port and protocol> and src IP specified as 0.
4) Whenever there is a LBSourceIPRange config specified, the source IP CIDR is added to the NAT FE table and prefix len adjusted.
5) A drop entry is created for the <dest IP, dest port and protocol> match with 0.0.0.0/0. 
6) Packets which hits a valid NAT entry passes through to the next stage, whereas packets hitting the drop entry is dropped in the NAT_FE_LOOKUP stage.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```


bpf/proxy/syncer.go
```diff
@@ -664,31 +665,70 @@ func getSvcNATKey(svc k8sp.ServicePort) (nat.FrontendKey, error) {
...
+	log.Debugf("loadbalancer %v", loadBalancerSourceRanges)
+	proto, err := protoV1ToInt(svc.Protocol())
 	if err != nil {
-		return err
+		return keys, err
+	}
+	key := nat.NewNATKey(ipaddr, uint16(port), proto)
+	keys = append(keys, key)
+	for _, src := range loadBalancerSourceRanges {
+		key := nat.NewNATKeySrc(ipaddr, uint16(port), proto, ip.MustParseCIDROrIP(src).(ip.V4CIDR))
```
fasaxc, line 684: This will panic if the IP is an IPv6 range.  I think we should completely ignore IPv6 services instead.

bpf/proxy/syncer.go
```diff
@@ -664,31 +665,70 @@ func getSvcNATKey(svc k8sp.ServicePort) (nat.FrontendKey, error) {
...
+		copy(affkey[:], key[4:12])
+	} else {
+		keys, err := getSvcNATKeyLBSrcRange(svc)
+		if err != nil {
+			return err
+		}
+		for index, key := range keys {
+			if index == 0 {
+				err = s.writeSvcNatKey(key, nat.NewNATValue(math.MaxUint32, uint32(0), uint32(local), affinityTimeo))
+				copy(affkey[:], key[4:12])
```
fasaxc, line 720: This is too low level to have outside the nat package. Suggest adding a method to Key that returns the AffKey.

bpf/proxy/syncer.go
```diff
@@ -664,31 +665,70 @@ func getSvcNATKey(svc k8sp.ServicePort) (nat.FrontendKey, error) {
...
+			} else {
+				err = s.writeSvcNatKey(key, val)
+			}
+			if err != nil {
+				return errors.Errorf("bpfSvcs.Update: %s", err)
+			}
+		}
 	}
```
fasaxc, line 728: How are updates handled?  If the set of allowed IPs changes or is removed completely?

sridhartigera: Have UTs for addition, updation, deletion of source IP ranges.

bpf-gpl/nat.h
```diff
@@ -118,9 +133,17 @@ static CALI_BPF_INLINE struct calico_nat_dest* calico_v4_nat_lookup2(__be32 ip_s
 								     __be32 ip_dst,
 								     __u8 ip_proto,
 								     __u16 dport,
-								     bool from_tun)
+								     bool from_tun,
+								     __u8 *drop)
 {
 	struct calico_nat_v4_key nat_key = {
+		.prefixlen = 88,
```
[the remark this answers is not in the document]
sridhartigera: This has to be (sizeof(struct calico_nat_v4_key) - sizeof(uint32_t) - sizeof(nat_key.pad))*8
I feel this as well has to be kept in sync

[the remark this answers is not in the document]
tomastigera: I was thinking the same as Shaun, took me a while to see where the 88 comes from. I'd define a marco, something like `NAT_KEY_WITH_SRC_MATCH`, whatever makes sense so that the code is readable right away.

bpf-gpl/nat.h
```diff
@@ -188,7 +211,10 @@ static CALI_BPF_INLINE struct calico_nat_dest* calico_v4_nat_lookup2(__be32 ip_s
 		}
 		CALI_DEBUG("NAT: nodeport hit\n");
 	}
-
+	if (nat_lv1_val->id == 0xffffffff && nat_lv1_val->count == 0) {
+		*drop = 1;
```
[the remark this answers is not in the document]
tomastigera: And preferably, `0xffffffff` defined as something like `REJECTED_SRC_IP_ID` so that it is easier to read the code and match it with a similar constant in golang.

Since we use cgo already, perhaps we should move such constants in some header file that we can include in go.

bpf/proxy/syncer.go
```diff
@@ -664,31 +665,73 @@ func getSvcNATKey(svc k8sp.ServicePort) (nat.FrontendKey, error) {
...
+		}
+		copy(affkey[:], key.Affinitykey())
+	} else {
+		keys, err := getSvcNATKeyLBSrcRange(svc)
+		if err != nil {
+			return err
+		}
+		for index, key := range keys {
+			if index == 0 {
+				err = s.writeSvcNatKey(key, nat.NewNATValue(math.MaxUint32, uint32(0), uint32(local), affinityTimeo))
```
tomastigera, line 722: nit: The other values are also irrelevant.

```suggestion
				err = s.writeSvcNatKey(key, nat.NewNATValue(math.MaxUint32, uint32(0), uint32(0), 0)
```

bpf-gpl/nat.h
```diff
@@ -45,25 +45,43 @@
 #define VXLAN_ENCAP_SIZE	(sizeof(struct ethhdr) + sizeof(struct iphdr) + \
 				sizeof(struct udphdr) + sizeof(struct vxlanhdr))
 
-// Map: NAT level one.  Dest IP and port -> ID and num backends.
+/* Prefix len = (dst_addr + port + protocol + src_addr) in bits. */
+#define NAT_PREFIX_LEN_WITH_SRC_MATCH 88
```
tomastigera, line 49: could be defined using the sizeof as Shaun suggested to give the value a meaning? It still going to need some maintenance, but at least the calculation of the value is explained by the code.

bpf/proxy/syncer.go
```diff
@@ -17,6 +17,7 @@ package proxy
 import (
 	"context"
 	"fmt"
+	"math"
```
fasaxc, line 20: I took a look at the surrounding code and I'm wondering if startupSync() is now out of date since it doesn't consider the new field and in general it seems to assume a 1-1 correspondence between services and map entries.  Would be worth adding a test that begins with some out-of-date source IPs in the dataplane and checks they get cleaned up.

bpf/proxy/syncer.go
fasaxc: Indeed, looks like the cases I was worried about are handled by the surrounding code.  It must convert an update into a delete then an add?

tomastigera: After all new services are added, we delete the ones that no longer exist, however, the LB takes a different path. When the source IPs range changes, the service itself changes and we always first delete a service in this case.

The LB handling is done in a slightly different way then the other services, which may or may not be OK - normally, the ClusterIP service is the "primary service" and then other services are treated as "derived". This implementation diverts from that pattern, but given the different purpose of the clusterIP (a black hole) and derived services (the actual NAT) in this case it may be ok.

Note that the "derived" lpm entries never make it in the `newSvcMap/prevSvcMap`

One thing I am not sure will work well is if we have an `ExternalIP` assigned. The iptables kube-proxy also does the filtering on the IP. I think the way we have it know, we would bypass the filtering as the ExternalIP would be just derived from the ClusterIP service. We should also cover this in a unittest.

Regarding recovery after crash, we need to cover that in a UT (another reason why not to reuse an existing service - tuning it into an LB - in the test). Since the keys will have the the src IP prefix part, we will know that it is an LB source IP range match entry, but it does not seem we handle that case and we need to make sure that we do not delete these entries because we cannot match them with the new set of services. And we must make sure we delete stale if the LB does not exist anymore.

bpf/proxy/syncer.go
```diff
@@ -271,6 +273,16 @@ func (s *Syncer) cleanupDerived(id uint32) error {
...
+			keys, err := getSvcNATKeyLBSrcRange(si.svc)
+			if err != nil {
+				return err
+			}
+			for _, key = range keys {
+				log.Debugf("bpf map deleting derived %s:%s", key, nat.NewNATValue(id, 0, 0, 0))
+				if err := s.bpfSvcs.Delete(key[:]); err != nil {
+					return errors.Errorf("bpfSvcs.Delete: %s", err)
+				}
+			}
```
tomastigera, line 285: I think this is executed for every occurrence of the ID in the prevMap, that is for the ClusterIP and ExternalIP as well - the LB is unlikely to have any other entries. You could do it just for the ExternalIP. No a big deal, but if `writeLBSrcRangeSvcNATKey` inserted those src range entries in the `newSvcMap` they would be in the `prevSvcMap` here.

bpf/proxy/syncer.go
```diff
@@ -418,6 +430,12 @@ func (s *Syncer) applyDerived(sname k8sp.ServicePortName, t svcType, sinfo k8sp.
 		if err := s.writeSvc(sinfo, svc.id, count, local); err != nil {
 			return err
 		}
+		if svcTypeExternalIP == t {
+			err := s.writeLBSrcRangeSvcNATKey(sinfo, svc.id, count, local)
+			if err != nil {
+				log.Debugf("Failed to write LB source range NAT keys")
+			}
+		}
```
tomastigera, line 438: You could promote this block into the caller (`apply()`) add a type `svcTypeExternalIPSrc` or something and create keys into the map that also include the src prefix...

sridhartigera: As we need to write keys with src ip range for every external IP, I feel the current approach would be better. Let me know your thoughts

bpf/proxy/syncer.go
sridhartigera: I am not sure if we can distinguish whether it is external IP or cluster IP here

bpf/proxy/syncer.go
```diff
@@ -776,7 +859,22 @@ func (s *Syncer) matchBpfSvc(bsvc nat.FrontendKey, svcs k8sp.ServiceMap) *svcKey
 			}
 			continue
 		}
+		matchLBSrcIp := func() uint8 {
+			for _, srcip := range info.LoadBalancerSourceRanges() {
+				cidr := ip.MustParseCIDROrIP(srcip).(ip.V4CIDR)
+				addr := cidr.Addr().AsNetIP().To4().String()
+				prefix := uint32(cidr.Prefix() - nat.ZeroCIDRPrefixLen)
+				if addr == bsvc.SrcAddr().String() || prefix == bsvc.PrefixLen() {
```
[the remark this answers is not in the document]
fasaxc: You should beef up the tests to the point that they catch this issue and then fix it.

bpf/proxy/syncer.go
```diff
@@ -17,6 +17,8 @@ package proxy
 import (
 	"context"
 	"fmt"
+
+	//"math"
```
fasaxc, line 21: Please remove commented out code

bpf/proxy/lb_src_range_test.go
```diff
@@ -0,0 +1,228 @@
...
+			defer func() {
+				log("svcs = %+v\n", svcs)
+				log("eps = %+v\n", eps)
+			}()
+
+			step()
+		}
+	}
+
+	saddr1 := ip.MustParseCIDROrIP("35.0.1.2/24").(ip.V4CIDR)
```
fasaxc, line 79: Thanks for pulling these up to the top, lot easier to follow now.

bpf/proxy/syncer.go
```diff
@@ -776,6 +859,29 @@ func (s *Syncer) matchBpfSvc(bsvc nat.FrontendKey, svcs k8sp.ServiceMap) *svcKey
 			}
 			continue
 		}
+		matchLBSrcIp := func() bool {
+			// Allow further comparison if CIDR is 0. It is treated as an entry without src addr
+			if bsvc.SrcCIDR() == nat.ZeroCIDR {
```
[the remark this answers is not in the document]
sridhartigera: If the source IP matches with any of the LB src range, i just let it move to the further comparison of cluster IP or external IP. I am treating src CIDR as something similar to a protocol match.
This is my logic
1) If the entry has a zeroCIDR, then treat it as an entry without src IP and let it match the clusterIP or external IP.
2) If the entry has a Non zero source CIDR, and if the service does not have LB source range, then it is to be treated as a stale entry. so the match function returns false, and i would just continue to the next entry.
3) If the entry has a Non-zero source CIDR and it matches with any of the LB source range, it has to be further compared with cluster Ip or external IP. This can be optimized to compare only the external IP.

fv/bpf_test.go
```diff
@@ -831,6 +831,203 @@ func describeBPFTests(opts ...bpfTestOpt) bool {
...
+						felixes[1].Exec("ip", "route", "add", "local", "10.1.2.3", "dev", "eth0")
+						felixes[0].Exec("ip", "route", "add", "local", "10.1.2.3", "dev", "eth0")
+						cc.ExpectSome(w[1][0], TargetIP(ip[0]), port)
+						cc.ExpectSome(w[1][1], TargetIP(ip[0]), port)
+						cc.ExpectNone(w[0][1], TargetIP(ip[0]), port)
+						cc.CheckConnectivity()
+					})
+				})
+
+				Context("Test LB-service with external Client, LB src ranges", func() {
```
fasaxc, line 941: This test and the one below act as a pair to verify that the external client is only allowed if its IP is in the range.  IMO, it's worth engineering the two tests to share code so that all the set-up is identical apart from the source IP and the expectation of whether the connection is allowed.  One way to do that is to make a table-driven test where each entry in the table is a pair or source IP range and expected outcome.  Another would be to move all the common set-up to a shared before each so that the BeforeEach for each test is only tweaking the source IPs.

Otherwise, a bug in the copying of the set-up here could make the test pass/fail for the wrong reason.

bpf-gpl/nat.h
```diff
@@ -202,7 +230,12 @@ static CALI_BPF_INLINE struct calico_nat_dest* calico_v4_nat_lookup2(__be32 ip_s
 		goto skip_affinity;
 	}
 
-	affkey.nat_key =  nat_key;
+        struct calico_nat_v4 nat_data = {
```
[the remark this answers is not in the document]
fasaxc: Still needs addressing

fv/bpf_test.go
```diff
@@ -831,6 +833,140 @@ func describeBPFTests(opts ...bpfTestOpt) bool {
...
+						cc.ExpectNone(w[0][1], TargetIP(ip[0]), port)
+						cc.CheckConnectivity()
+					})
+				})
+
+				Describe("Test load balancer service with external Client,src ranges", func() {
+					var testSvc *v1.Service
+					tgtPort := 8055
+					externalIP := []string{extIP}
+					srcIPRange := []string{"10.65.1.3/24"}
```
fasaxc, line 920: Tests may run in any order so you should define srcIPRange like this:
```
var srcIPRange []string
```
and then always set it in the `BeforeEach()` immediately below.

In general, if you ever modify a variable inside a Describe block then you should always set it back to its default value in the BeforeEach of the Describe.

fv/bpf_test.go
```diff
@@ -831,6 +833,140 @@ func describeBPFTests(opts ...bpfTestOpt) bool {
...
+							testOpts.protocol, externalIP, srcIPRange)
+						// when we point Load Balancer to a node in GCE it adds local routes to the external IP on the hosts.
+						// Similarity add local routes for externalIP on felixes[0], felixes[1]
+						felixes[1].Exec("ip", "route", "add", "local", extIP, "dev", "eth0")
+						felixes[0].Exec("ip", "route", "add", "local", extIP, "dev", "eth0")
+						ip = testSvc.Spec.ExternalIPs
+						port = uint16(testSvc.Spec.Ports[0].Port)
+					})
+
+					Context("Test LB-service with external IP and workloads", func() {
```
fasaxc, line 855: Nit: no need for an extra Context if you don't have an extra BeforeEach.

fv/bpf_test.go
```diff
@@ -831,6 +833,140 @@ func describeBPFTests(opts ...bpfTestOpt) bool {
...
+						It("should have connectivity from workloads[1][0],[1][1] and [0][1] via external IP to workload 0", func() {
+							cc.ExpectSome(w[1][0], TargetIP(ip[0]), port)
+							cc.ExpectSome(w[1][1], TargetIP(ip[0]), port)
+							cc.ExpectSome(w[0][1], TargetIP(ip[0]), port)
+							cc.CheckConnectivity()
+						})
+					})
+
+					Context("Test LB-service with external IP and external Client", func() {
+						BeforeEach(func() {
```
fasaxc, line 866: Think I'd set up the external client in the BeforeEach() above, just to save one extra nesting.  In fact, you could combine the workload test above with `cc.ExpectSome(externalClient, TargetIP(ip[0]), port)`.  FVs run so slowly, it's sometimes better to combine them instead of having super-precise tests.

PR #2354, opened 2020-05-28
change calicoRouterIP to a constant

latteczy:
## Description
change calicoRouterIP to a constant


fasaxc:
Unfortunately that change is not valid, IPs are ultimately slices, which can't be const.

PR #2362, opened 2020-06-03
Adding more WireGuard FVs

realgaurav:
Adding more WireGuard FVs:
* WireGuard traffic test with GNP
* borrowed-ip handling in WireGuard route-table.
* nodes with mixed WireGuard enabled/disabled settings.

## Release Note
```release-note
None required
```


fv/wireguard_test.go
```diff
@@ -351,6 +329,26 @@ var _ = infrastructure.DatastoreDescribe("WireGuard-Supported", []apiconfig.Data
 			Expect(rcvd[1]).NotTo(BeEmpty())
 			Expect(rcvd[1]).To(Equal(sent[0]))
 		})
+
+		It("between pod to pod should be encrypted using wg tunnel with egress policy applied", func() {
```
robbrockbank, line 333: How do you know the policy is doing what you'd expect?

I think you should also have a similar test where the policy is set to "Deny" and you check the packet counts drop to 0.  Perhaps have it as part of the same test, so you modify the existing policy and then retry.

realgaurav: I had both tests initially but removed the `deny` test thinking `allow` test will be sufficient to invoke the policy code-path. Added both `allow` and `deny` tests now.

fv/wireguard_test.go
```diff
@@ -348,6 +335,52 @@ var _ = infrastructure.DatastoreDescribe("WireGuard-Supported", []apiconfig.Data
...
+					return tcpdumps[i].MatchCount("numInTunnelPackets")
+				}, "5s", "100ms").Should(BeNumerically("==", 0))
+				Consistently(func() int {
+					return tcpdumps[i].MatchCount("numOutTunnelPackets")
+				}, "5s", "100ms").Should(BeNumerically("==", 0))
+			}
+
+			cc.ResetExpectations()
+
+			policy.Name = "f01-egress-allow"
```
robbrockbank, line 364: I think this relies on the policy ordering being alphanumeric - which we don't actually expose anywhere.

I would either:
-  Modify the original policy, or
-  Set the Order number so this is definitely lower order than the first.

fv/wireguard_test.go
```diff
@@ -348,6 +335,52 @@ var _ = infrastructure.DatastoreDescribe("WireGuard-Supported", []apiconfig.Data
...
+			_, err = client.GlobalNetworkPolicies().Create(utils.Ctx, policy, utils.NoOptions)
+			Expect(err).NotTo(HaveOccurred())
+
+			cc.ExpectSome(wls[0], wls[1])
+			cc.ExpectSome(wls[1], wls[0])
+			cc.CheckConnectivity()
+
+			By("verifying tunnelled packet count is non-zero")
+			for i := range felixes {
+				Eventually(func() int {
```
robbrockbank, line 376: For this to be useful it needs to check counts before and after you run connectivity.

IIUC, the first connectivity could result in tunnel packets while the policy is being installed.

robbrockbank: I've added a ResetCounts to the tcpdump - maybe useful once I've checked in my changes

fv/wireguard_test.go
```diff
@@ -348,6 +335,52 @@ var _ = infrastructure.DatastoreDescribe("WireGuard-Supported", []apiconfig.Data
...
+					return tcpdumps[i].MatchCount("numInTunnelPackets")
+				}, "10s", "100ms").Should(BeNumerically(">", 0))
+				Eventually(func() int {
+					return tcpdumps[i].MatchCount("numOutTunnelPackets")
+				}, "10s", "100ms").Should(BeNumerically(">", 0))
+			}
```
robbrockbank, line 382: I think 10s is gonna be too long for all of these tcp dump checks.  From a quick glance I think wireguard are the only tests that don't use the defaults (1s) for checking the tcp dump stats.  Do you think these need to be longer than the other tests?

realgaurav: There are a few tests using `tcpdump` in available examples. Original motivation to bump from `5s` to `10s` was flakes seen earlier, which may have been addressed by recent changes. In terms of impact though, I think polling interval will have actual impact for `Eventually` than the total time which is actually from `10ms` to `100ms` which isn't huge IMO. We can revisit this if needed.

fv/wireguard_test.go
realgaurav: added order

fv/wireguard_test.go
```diff
@@ -348,6 +335,52 @@ var _ = infrastructure.DatastoreDescribe("WireGuard-Supported", []apiconfig.Data
...
+			policy.Spec.Egress = []api.Rule{{Action: api.Deny}}
+			policy.Spec.Selector = fmt.Sprintf("name in { '%s', '%s'}", wls[0].Name, wls[1].Name)
+			_, err := client.GlobalNetworkPolicies().Create(utils.Ctx, policy, utils.NoOptions)
+			Expect(err).NotTo(HaveOccurred())
+
+			cc.ExpectNone(wls[0], wls[1])
+			cc.ExpectNone(wls[1], wls[0])
+			cc.CheckConnectivity()
+
+			By("verifying tunnelled packet count is non-zero")
```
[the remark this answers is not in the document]
realgaurav: done

fv/wireguard_test.go
realgaurav: If I am understanding this comment correctly, we need to make sure that the policy is installed correctly before testing the connectivity. Will it be a good/clearer approach -- if we wait a few seconds and then read the policy and compare it with the policy created? If it matches then test the connectivity?

realgaurav: I updated the PR with suggested point above.

PR #2363, opened 2020-06-04
Make TC program application graceful

fasaxc:
## Description

- [x] Never delete the qdisc.
- [x] Add new program then remove the old.
- [x] Support policy update without replacing program.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
In BPF mode, Felix now applies policy updates without reapplying the BPF programs; this gives a performance boost and closes a window where traffic was not policed.
```

### DIFF
--- a/bpf/tc/attach.go
+++ b/bpf/tc/attach.go
@@ -65,13 +65,10 @@ func (e ErrAttachFailed) Error() string {
 }
 
 var ErrDeviceNotFound = errors.New("device not found")
-var	prefHandleRe = regexp.MustCompile(`pref ([^ ]+) .* handle ([^ ]+)`)
+var prefHandleRe = regexp.MustCompile(`pref ([^ ]+) .* handle ([^ ]+)`)
 
 // AttachProgram attaches a BPF program from a file to the TC attach point
 func (ap AttachPoint) AttachProgram() error {
-	// FIXME we use this lock so that two copies of tc running in parallel don't re-use the same jump map.
-	// This can happen if tc incorrectly decides the two programs are identical (when in fact they differ by attach
-	// point).
 	logCxt := log.WithField("attachPoint", ap)
 
 	tempDir, err := ioutil.TempDir("", "calico-tc")
@@ -92,6 +89,9 @@ func (ap AttachPoint) AttachProgram() error {
 		return err
 	}
 
+	// FIXME we use this lock so that two copies of tc running in parallel don't re-use the same jump map.
+	// This can happen if tc incorrectly decides the two programs are identical (when in fact they differ by attach
+	// point).
 	logCxt.Debug("AttachProgram waiting for lock...")
 	tcLock.Lock()
 	defer tcLock.Unlock()
--- a/bpf/tc/attach.go
+++ b/bpf/tc/attach.go
@@ -168,6 +168,7 @@ func (ap AttachPoint) AttachProgram() error {
 	}
 
 	// Success: clean up the old programs.
+	var progErrs []error
 	for _, p := range progsToClean {
 		log.WithField("prog", p).Debug("Cleaning up old calico program")
 		tcCmd := exec.Command("tc", "filter", "del", "dev", ap.Iface, string(ap.Hook), "pref", p.pref, "handle", p.handle, "bpf")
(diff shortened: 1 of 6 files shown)


bpf/tc/attach.go
```diff
@@ -104,13 +91,60 @@ func (ap AttachPoint) AttachProgram() error {
...
+		pref   string
+		handle string
+	}
+	var progsToClean []progToCleanUp
+	for _, line := range strings.Split(string(out), "\n") {
+		if !strings.Contains(line, "calico") {
+			continue
+		}
+		// find the pref and the handle
+		prefHandleRe := regexp.MustCompile(`pref ([^ ]+) .* handle ([^ ]+)`)
```
tomastigera, line 121: This one is a constant pattern, can be compiled just once in init time

bpf/tc/attach.go
```diff
@@ -130,7 +164,19 @@ func (ap AttachPoint) AttachProgram() error {
...
+
+	// Success: clean up the old programs.
+	for _, p := range progsToClean {
+		log.WithField("prog", p).Debug("Cleaning up old calico program")
+		tcCmd := exec.Command("tc", "filter", "del", "dev", ap.Iface, string(ap.Hook), "pref", p.pref, "handle", p.handle, "bpf")
+		err := tcCmd.Run()
+		if err != nil {
+			// TODO: filter error to avoid spam if interface is gone.
+			log.WithError(err).WithField("prog", p).Warn("Failed to clean up old calico program.")
+			return errors.WithMessage(err, "failed to clean up old program")
```
neiljerram, line 178: Would it be better to try to clean up the whole of progsToClean, before returning an error?

fv/bpf_test.go
```diff
@@ -1416,7 +1416,7 @@ func describeBPFTests(opts ...bpfTestOpt) bool {
 								if testOpts.connTimeEnabled {
 									Skip("FIXME externalClient also does conntime balancing")
 								}
-
+								externalClient.EnsureBinary("test-connection")
```
neiljerram, line 1419: Is this a related change?

dataplane/linux/bpf_ep_mgr.go
```diff
@@ -496,44 +496,48 @@ func (m *bpfEndpointManager) applyPolicy(wlID proto.WorkloadEndpointID) error {
...
+			return err
 		}
-	}()
 
+		jumpMapFD, err := FindJumpMap(ap)
+		if err != nil {
+			return errors.Wrap(err, "failed to look up jump map")
+		}
+		if iface.jumpMapFD == nil {
+			iface.jumpMapFD = map[PolDirection]bpf.MapFD{}
```
neiljerram, line 531: Can we simplify by doing this in the iface constructor?  (So that the map always exists.)

dataplane/linux/bpf_ep_mgr.go
```diff
@@ -496,44 +496,48 @@ func (m *bpfEndpointManager) applyPolicy(wlID proto.WorkloadEndpointID) error {
...
+		}
+		iface.jumpMapFD[polDirection] = jumpMapFD
+		m.ifaces[endpoint.Name] = iface
+	}
+
+	return m.updatePolicyProgram(iface.jumpMapFD[polDirection], rules)
+}
+
+func (m *bpfEndpointManager) updatePolicyProgram(jumpMapFD bpf.MapFD, rules [][][]*proto.Rule) error {
```
neiljerram, line 540: Is there any change needed here because of the jump map fd being held open?  E.g. if it was like a normal file, it would point to a particular offset in the file, and that might need resetting?

bpf/tc/attach.go
fasaxc: Yes, good catch.

dataplane/linux/bpf_ep_mgr.go
fasaxc: No, I don't think BPF map FDs are stateful.

dataplane/linux/bpf_ep_mgr.go
fasaxc: Tried to do this but found a problem: we're currently using the zero value returned by `ifaces[<unknown key>]` in anger.  If I change to using a constructor, I'll have to do 
```
if iface, ok := ifaces[key]; !ok { 
   // create the entry
}
```
everywhere, or wrap that in a method.  Not sure that ends up any cleaner in the end?

fv/bpf_test.go
fasaxc: No, looks like an accidental commit.

dataplane/linux/bpf_ep_mgr.go
neiljerram: OK, thanks for looking at it.

PR #2374, opened 2020-06-10
bpf: Map.Iter() uses bpf.GetMapNextKey syscall

tomastigera:
Iteration is delete-safe, that is, deleting the current key does not
break iterating over the map, no key will be skipped and no key is
visited more then once.

The above holds as long as there is no other concurrent user.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/bpf/bpf_syscall.go
+++ b/bpf/bpf_syscall.go
@@ -69,6 +69,15 @@ import (
 //    attr->flags = flags;
 // }
 //
+// // bpf_attr_setup_map_get_next_key sets up the bpf_attr union for use with BPF_MAP_GET_NEXT_KEY
+// // A C function makes this easier because unions aren't easy to access from Go.
+// void bpf_attr_setup_map_get_next_key(union bpf_attr *attr, __u32 map_fd, void *key, void *next_key, __u64 flags) {
+//    attr->map_fd = map_fd;
+//    attr->key = (__u64)(unsigned long)key;
+//    attr->next_key = (__u64)(unsigned long)next_key;
+//    attr->flags = flags;
+// }
+//
 // // bpf_attr_setup_map_elem_for_delete sets up the bpf_attr union for use with BPF_MAP_DELETE_ELEM
 // // A C function makes this easier because unions aren't easy to access from Go.
 // void bpf_attr_setup_map_elem_for_delete(union bpf_attr *attr, __u32 map_fd, void *pointer_to_key) {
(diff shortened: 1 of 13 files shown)


bpf/bpf_syscall.go
```diff
@@ -367,3 +376,46 @@ func DeleteMapEntry(mapFD MapFD, k []byte, valueSize int) error {
 	}
 	return nil
 }
+
+// GetMapNextKey returns the next key for the given key if the current key exists.
+// Otherwise it returns the first key. Order is implemention / map type
+// dependent.
+//
+// Start iterating by passing a nil key.
+func GetMapNextKey(mapFD MapFD, k []byte, keySize int) ([]byte, error) {
```
fasaxc, line 385: Might be worth having a stateful iterator object for this so we can avoid allocating and copying for every iteration?

tomastigera: I will defer this into another PR. We can do the same for lookup and we can make the `PinnedMap.Iter()` allocate the memory once and we may also use it for deletes etc. So might be a bigger standalone change.

bpf/conntrack/cleanup.go
```diff
@@ -79,7 +79,8 @@ func (l *LivenessScanner) Scan() {
 			// Look up the reverse entry, where we do the book-keeping.
 			revEntryBytes, err := l.ctMap.Get(ctVal.ReverseNATKey().AsBytes())
 			if err != nil && bpf.IsNotExists(err) {
-				// Forward entry exists but no reverse entry (and the grace period has expired).
+				// Forward entry exists but no reverse entry. We might have come across the reverse
```
fasaxc, line 82: Think you need to check the creation timestamp here to make sure that the entry wasn't just created (or we might be racing with the BPF program creating the pair of keys).

tomastigera: Imo not an issue - not sure if by desing or by accident ;-) but the original code did not deal with that - afaict in contrary to the comment - either. We always create the fwd entry _after_ we create the rev entry in BPF. So if we hit a fwd which does not have rev entry we must have deleted it or there was some weird inconsistency - in either case it seems correct to remove the orphaned fwd entry.

Otoh it is a bit brittle since it depends on how BPF code creates the entries and if that changes, this will break. I will at least update the comment.

However, Since we can/should not delete two keys in one iteration, I was pondering collecting all the rev keys that we asses as expired and defer deleting them at the end of the clean up. This way we would never remove rev key before fwd key.

PR #2376, opened 2020-06-10
Update pins / fix prometheus client

Brian-McM:
## Description
This commit does the following:
- updates and tidys the pins
- uses the NewSummary method in the new libcalico-go version to create prometheus summaries, as the defaults were removed in the new version of the prometheus client

## Todos
- [x] Release note

## Release Note
```release-note
Update and tidy pins
Use libcalico-go NewSummary method for creating prometheus summaries (which has the Objective defaults)
```


doublek:
I tried to build using master and I get the same changes in the PR for protobuf.

proto/felixbackend.pb.go
```diff
@@ -68,21 +68,16 @@
 */
 package proto
 
-import (
```
fasaxc, line 71: Please run `make fix` then this diff will go away.  No sense in having everyone else get a diff after they run `make fix`

Brian-McM: done

PR #2378, opened 2020-06-11
Include wireguard tunnel IP address in wireguard routing

robbrockbank:
## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```


fv/wireguard_test.go
```diff
@@ -348,6 +347,73 @@ var _ = infrastructure.DatastoreDescribe("WireGuard-Supported", []apiconfig.Data
...
+						hep.Spec.InterfaceName = "*"
+					} else {
+						hep.Spec.InterfaceName = "eth0"
+					}
+					_, err := client.HostEndpoints().Create(utils.Ctx, hep, options.SetOptions{})
+					Expect(err).NotTo(HaveOccurred())
+				}
+
+				By("Checking there is connectivity between the workloads")
+				cc.ExpectSome(wls[0], wls[1])
```
fasaxc, line 412: Could do with some traffic that gets blocked here too, otherwise the check might sneak in before the policy is applied.

rules/rule_defs.go
```diff
@@ -57,9 +57,10 @@ const (
 	IPSetIDNATOutgoingAllPools  = "all-ipam-pools"
 	IPSetIDNATOutgoingMasqPools = "masq-ipam-pools"
 
-	IPSetIDAllHostNets        = "all-hosts-net"
-	IPSetIDAllVXLANSourceNets = "all-vxlan-net"
-	IPSetIDThisHostIPs        = "this-host"
+	IPSetIDAllHostNets          = "all-hosts-net"
+	IPSetIDAllWireguardHostNets = "all-wg-hosts-net"
```
fasaxc, line 61: Don't think you used this in the end?

robbrockbank: Ah no - I will remove.  Cruft from my cherry-pick from the other PR

fv/wireguard_test.go
robbrockbank: That's a good idea.  Will do!

dataplane/driver.go
```diff
@@ -129,6 +129,17 @@ func StartDataplaneDriver(configParams *config.Config,
...
+		// If wireguard is enabled, update the failsafe inbound ports to inculde the wireguard port.
+		failsafeInboundHostPorts := configParams.FailsafeInboundHostPorts
+		if configParams.WireguardEnabled {
+			failsafeInboundHostPorts = make([]config.ProtoPort, len(configParams.FailsafeInboundHostPorts)+1)
+			copy(failsafeInboundHostPorts, configParams.FailsafeInboundHostPorts)
+			failsafeInboundHostPorts[len(configParams.FailsafeInboundHostPorts)] = config.ProtoPort{
+				Port:     uint16(configParams.WireguardListeningPort),
+				Protocol: "udp",
+			}
+		}
```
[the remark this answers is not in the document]
robbrockbank: yup.  that fixed the test failure :-)

fv/wireguard_test.go
robbrockbank: I've added a Consistently check for now - not ideal, but I think passable

fv/wireguard_test.go
```diff
@@ -253,6 +255,81 @@ var _ = infrastructure.DatastoreDescribe("WireGuard-Supported", []apiconfig.Data
...
+				} else if err := waitForPackets(tcpdumps[i], timeout, "numWorkload1to0Packets", 0); err != nil {
+					return err
+				}
+			}
+
+			By("Checking the packet stats from wg")
+			for i := range felixes {
+				rcvd, sent := getWgStatistics(felixes[i])
+				Expect(rcvd).NotTo(BeEmpty())
+				Expect(sent).NotTo(BeEmpty())
```
realgaurav, line 328: @robbrockbank - why are we removing sent/rcvd comparison check?

robbrockbank: At the moment I'm trying to reduce test flakiness.  It's possible I've been a little overzealous, but at the moment I haven't convinced myself yet that the sent and received packets should always align pm either side - in particular during establishment of the routing on both sides.

I think I'd rather err on the side of caution and just check that "yes we have some traffic" and that we are hitting the correct ports on the host interface.

PR #2380, opened 2020-06-11
Reject connections to services with no backends

sridhartigera:
As of now, connections to services without backends are allowed. This is changed to reject the connection with ICMP port unreachable

## Todos
- [ ] Documentation
- [ ] Backport
- [x] Release note

## Release Note

```release-note
In BPF mode, connections to Kubernetes services with no backends are now rejected with ICMP port unreachable.

### CODE REVIEW

File: bpf-gpl/tc.c [hunk 1]
```diff
@@ -698,6 +716,14 @@ static CALI_BPF_INLINE int calico_tc(struct __sk_buff *skb)
 	CALI_DEBUG("Tail call to policy program failed: DROP\n");
 	return TC_ACT_SHOT;
 
+icmp_allow:
+	if (update_state_map(&state)) {
+		goto deny;
+	}
+	bpf_tail_call (skb, &cali_jump, 2);
+	/* should not reach here */
+	goto deny;
```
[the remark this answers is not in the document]
tomastigera: I would suggest to define some good constants for the tail call indexes like `JUMP_IDX_ICMP_REPLY` or maybe even something like `#define jump_icmp_send_reply(skb) bpf_tail_call (skb, &cali_jump, 2)` ...

bpf-gpl/tc.c
```diff
@@ -1272,6 +1253,82 @@ static CALI_BPF_INLINE struct fwd calico_tc_skb_accepted(struct __sk_buff *skb,
...
+						uint8_t type, uint8_t code)
+{
+	uint32_t fib_flags = 0;
+	__be32 un = 0;
+	int rc = TC_ACT_UNSPEC;
+	enum calico_reason reason = CALI_REASON_UNKNOWN;
+	uint32_t seen_mark;
+
+	if (code == ICMP_FRAG_NEEDED)
+	{
```
tomastigera, line 1266:
```suggestion
	if (code == ICMP_FRAG_NEEDED) {
```

bpf-gpl/tc.c
```diff
@@ -1272,6 +1253,82 @@ static CALI_BPF_INLINE struct fwd calico_tc_skb_accepted(struct __sk_buff *skb,
 	}
 }
 
+static CALI_BPF_INLINE struct fwd calico_tc_skb_icmp_accepted(struct __sk_buff *skb, struct iphdr *ip,
```
tomastigera, line 1256: `_accepted` ? 

As much as I favour breaking code into small brief functions, since it is the core of `calico_tc_skb_send_icmp_replies` itself, I would perhaps put all the code straight there.

bpf-gpl/tc.c
```diff
@@ -1272,6 +1253,82 @@ static CALI_BPF_INLINE struct fwd calico_tc_skb_accepted(struct __sk_buff *skb,
...
+static CALI_BPF_INLINE struct fwd calico_tc_skb_icmp_accepted(struct __sk_buff *skb, struct iphdr *ip, 
+						uint8_t type, uint8_t code)
+{
+	uint32_t fib_flags = 0;
+	__be32 un = 0;
+	int rc = TC_ACT_UNSPEC;
+	enum calico_reason reason = CALI_REASON_UNKNOWN;
+	uint32_t seen_mark;
+
+	if (code == ICMP_FRAG_NEEDED)
```
tomastigera, line 1265: we may just remove the other callers of `icmp_v4_reply` as they are not needed anymore, right?

bpf-gpl/tc.c
```diff
@@ -1272,6 +1253,82 @@ static CALI_BPF_INLINE struct fwd calico_tc_skb_accepted(struct __sk_buff *skb,
...
+	{
+		struct {
+			__be16  unused;
+			__be16  mtu;
+		} frag = {
+			.mtu = host_to_be16(TUNNEL_MTU),
+		};
+
+		un = *(__be32 *)&frag;
```
tomastigera, line 1274: we may perhaps reuse one of the 32bit fields like `pol_rc` or `tun_ip`, which are not going to be used downstream and pass the `un` from where the icmp response type is decide rather than having a special case here.

bpf/ut/icmp_port_unreachable_test.go
```diff
@@ -0,0 +1,117 @@
...
+
+		pktR := gopacket.NewPacket(res.dataOut, layers.LayerTypeEthernet, gopacket.Default)
+		fmt.Printf("pktR = %+v\n", pktR)
+
+		checkICMPPortUnreachable(pktR, ipv4)
+	})
+
+	err = natMap.Update(
+		natkey,
+		nat.NewNATValue(0, 0, 0, 0).AsBytes(),
```
tomastigera, line 83: nit: perhaps a comment how this case and the case above differ in what they test.

bpf-gpl/tc.c
```diff
@@ -1183,64 +1208,21 @@ static CALI_BPF_INLINE struct fwd calico_tc_skb_accepted(struct __sk_buff *skb,
...
-
-	ip_header = skb_iphdr(skb);
-	/* we silently drop the packet if things go wrong */
-
-	/* XXX we should check if it is broadcast or multicast and not respond */
-
-	/* do not respond to IP fragments except the first */
-	if (ip_frag_no(ip_header)) {
-		goto deny;
-	}
```
tomastigera, line 1200: I think this is still important and should not be removed.

bpf-gpl/icmp.h
```diff
@@ -185,12 +163,24 @@ static CALI_BPF_INLINE int icmp_v4_too_big(struct __sk_buff *skb)
 
 	CALI_DEBUG("Sending ICMP too big mtu=%d\n", be16_to_host(frag.mtu));
 
-	return icmp_v4_reply(skb, ICMP_DEST_UNREACH, ICMP_FRAG_NEEDED, *(__be32 *)&frag);
+	if (skb_too_short(skb)) {
+		CALI_DEBUG("ICMP v4 reply: too short before making room\n");
+		return -1;
+	}
```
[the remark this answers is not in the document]
sridhartigera: Moving it to the caller does not make the verifier happy.

fv/bpf_test.go
```diff
@@ -908,6 +908,59 @@ func describeBPFTests(opts ...bpfTestOpt) bool {
...
+								Source: api.EntityRule{
+									Nets: []string{
+										externalClient.IP + "/32",
+									},
+								},
+							},
+						}
+						pol = updatePolicy(pol)
+					})
+					It("should have connectivity from workloads[1][0],[1][1] via external IP to workload 0", func() {
```
fasaxc, line 944: Our of date spec name.

fv/bpf_test.go
```diff
@@ -908,6 +908,59 @@ func describeBPFTests(opts ...bpfTestOpt) bool {
...
+
+						tcpdump := externalClient.AttachTCPDump("any")
+						tcpdump.SetLogEnabled(true)
+						matcher := fmt.Sprintf("ICMP %s udp port %d unreachable", extIP, port)
+						tcpdump.AddMatcher("ICMP", regexp.MustCompile(matcher))
```
fasaxc, line 949: I still have a slight doubt that we're not testing the end-user behaviour here; what if tcpdump is outputting `ICMP %s udp port %d unreachable bad checksum` or something?

fv/bpf_test.go
```diff
@@ -908,6 +908,59 @@ func describeBPFTests(opts ...bpfTestOpt) bool {
...
+						tcpdump.Start()
+						defer tcpdump.Stop()
+
+						// send a packet from the correct workload to create a conntrack entry
+						externalClient.Exec("/pktgen", externalClient.IP, extIP, "udp",
```
fasaxc, line 954: This test is always using UDP but the test will be defined for the whole matrix of tests so we'll end up with the exact same test running multiple times.

fv/bpf_test.go
```diff
@@ -915,19 +915,20 @@ func describeBPFTests(opts ...bpfTestOpt) bool {
 					srcIPRange := []string{}
 					testSvcName := "test-lb-service-extip"
 					var port uint16
+					var ip []string
 
 					BeforeEach(func() {
-						if testOpts.connTimeEnabled {
+						if testOpts.connTimeEnabled || testOpts.udpUnConnected {
 							Skip("FIXME externalClient also does conntime balancing")
```
fasaxc, line 922: Nit: skip text is wrong for new condition

fv/connectivity/conncheck.go
```diff
@@ -195,6 +199,9 @@ func (c *Checker) ExpectedConnectivityPretty() []string {
 				result[i] += fmt.Sprintf(" (maxLoss: %.1f%%)", exp.ExpectedPacketLoss.MaxPercent)
 			}
 		}
+		if len(exp.Error) > 0 {
+			result[i] += exp.Error
```
fasaxc, line 203: Probably want a space before the error or the text will run together

fv/test-connection/test-connection.go
```diff
@@ -308,6 +308,7 @@ func tryConnect(remoteIPAddr, remotePort, sourceIPAddr, sourcePort, protocol str
 	tc, err := NewTestConn(remoteIPAddr, remotePort, sourceIPAddr, sourcePort, protocol,
 		time.Duration(seconds)*time.Second, sendLen, recvLen)
 	if err != nil {
+		os.Stdout.WriteString(err.Error())
```
fasaxc, line 311: Would prefer if we outputted a RESULT= line like we do on success in this case; have been trying to funnel all results through that mechanism.

fv/test-connection/test-connection.go
```diff
@@ -386,7 +387,20 @@ func (tc *testConn) tryLoopFile(loopFile string) error {
...
-
+func (tc *testConn) sendErrorResp(err error) {
+	var resp connectivity.Response
+	mtuPair := connectivity.MTUPair{}
+	resp.Error = err.Error()
+	res := connectivity.Result{
+		LastResponse: resp,
+		Stats: connectivity.Stats{
+			RequestsSent:      1,
+			ResponsesReceived: 1,
```
fasaxc, line 398: This should be 0, I think.

fv/test-connection/test-connection.go
```diff
@@ -386,7 +387,20 @@ func (tc *testConn) tryLoopFile(loopFile string) error {
 	res.PrintToStdout()
 	return nil
 }
-
+func (tc *testConn) sendErrorResp(err error) {
+	var resp connectivity.Response
+	mtuPair := connectivity.MTUPair{}
```
fasaxc, line 392: Can probably leave out MTU pair here and in the struct

fv/test-connection/test-connection.go
```diff
@@ -386,7 +387,20 @@ func (tc *testConn) tryLoopFile(loopFile string) error {
 	res.PrintToStdout()
 	return nil
 }
-
+func (tc *testConn) sendErrorResp(err error) {
```
fasaxc, line 390: Nit: please put one line space between functions.

fv/connectivity/conncheck.go
```diff
@@ -460,6 +475,11 @@ func (e Expectation) Matches(response *Result, checkSNAT bool) bool {
 
 	} else {
 		if response != nil {
```
fasaxc, line 477: I think the response can now equal nil, but we want the check to fail if e.Error is set but the response is nil.

fv/bpf_test.go
```diff
@@ -908,6 +908,46 @@ func describeBPFTests(opts ...bpfTestOpt) bool {
...
+					var testSvc *v1.Service
+					tgtPort := 8055
+					externalIP := []string{extIP}
+					srcIPRange := []string{}
+					testSvcName := "test-lb-service-extip"
+					var port uint16
+					var ip []string
+
+					BeforeEach(func() {
+						if testOpts.connTimeEnabled || testOpts.udpUnConnected {
```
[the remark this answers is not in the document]
sridhartigera: In case of UDP unconnected, its responsibility of the application to handle ICMP port unreachable, whereas in case of TCP and UDP-connected, the kernel responds with a "connection Refused". For UDP-unconnected, "test-connection" proceeds with send and does not get any data back and timesout. Hence UDP-unconnected is skipped.

fv/bpf_test.go
```diff
@@ -908,6 +908,46 @@ func describeBPFTests(opts ...bpfTestOpt) bool {
 					})
 				})
 
+				Context("Test load balancer service with no backend", func() {
```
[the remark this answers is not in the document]
sridhartigera: No particular reason to pick LB. Thought could just use the existing infra. Can also use a normal service

fv/bpf_test.go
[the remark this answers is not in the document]
fasaxc: I think that's OK; the main point of the test is to verify that we're generating the ICMP message correctly. The fact that the kernel accepts it for both UDP and TCP seems pretty conclusive.

fasaxc:
@tomastigera correct, we can't do anything at connect time but we'll do the NAT lookup at egress too IIRC so we'll catch it there.

fv/bpf_test.go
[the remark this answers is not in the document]
tomastigera: OK, me surprised that works differently for the two versions of UDP, might be the libs, if it works, it works :)

PR #2381, opened 2020-06-11
Add AWS source-destination-check handling

realgaurav:
## Description

Add AWS source-destination-check handling.

I have manually tested the implementation on EKS cluster to validate the core functionality. In process of adding UTs to increase test coverage.

## Todos
- [ ] Documentation
- [x] Release note

## Release Note
```release-note
Implementation to handle setting source-destination-check for AWS EC2 instances.
```

### DIFF
--- a/aws/ec2.go
+++ b/aws/ec2.go
@@ -1,0 +1,192 @@
+package aws
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/aws/ec2metadata"
+	awssession "github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	log "github.com/sirupsen/logrus"
+)
+
+const (
+	defaultTimeOut = 20
+	defaultRetries = 3
+)
+
+func getEC2InstanceID(ctx context.Context, svc *ec2metadata.EC2Metadata) (string, error) {
+	idDoc, err := svc.GetInstanceIdentityDocumentWithContext(ctx)
+	if err != nil {
+		return "", err
+	}
+	log.Infof("ec2-instance-id: %s", idDoc.InstanceID)
+	return idDoc.InstanceID, nil
+}
+
+func getEC2Region(ctx context.Context, svc *ec2metadata.EC2Metadata) (string, error) {
+	region, err := svc.RegionWithContext(ctx)
+	if err != nil {
+		return "", err
+	}
+	log.Infof("region: %s", region)
+	return region, nil
+}
+
+func convertError(err error) string {
+	if awsErr, ok := err.(awserr.Error); ok {
+		return fmt.Sprintf("%s: %s", awsErr.Code(), awsErr.Message())
+	}
+
+	return fmt.Sprintf("%v", err.Error())
+}
+
+func retriable(err error) bool {
+	if awsErr, ok := err.(awserr.Error); ok {
+		switch awsErr.Code() {
+		case "InternalError":
+				return true
+		case "InternalFailure":
+				return true
+		case "RequestLimitExceeded":
+				return true
+		case "ServiceUnavailable":
+				return true
+		case "Unavailable":
+				return true
+		}
+	}
+
+	return false
+}
+
+func getEC2NetworkInterfaceId(ctx context.Context, svc *ec2.EC2, instanceId string) (networkInstanceId string, err error) {
+	input := &ec2.DescribeInstancesInput{
+		InstanceIds: []*string{
+			aws.String(instanceId),
+		},
+		Filters: []*ec2.Filter{
+			{
+				Name: aws.String("network-interface.attachment.device-index"),
+				Values: []*string{
+					aws.String("0"), // Only device-id-0
+				},
+			},
+		},
+	}
+
+	i := 0
+retry:
+	out, err := svc.DescribeInstancesWithContext(ctx, input)
+	if err != nil {
+		if retriable(err) && i < defaultRetries {
+			log.WithField("instance-id", instanceId).Info("retrying getting network-interface-id")
+			i++
+			goto retry
+		}
+		return "", err
+	}
+
+	if len(out.Reservations) <= 0 {
+		return "", fmt.Errorf("no network-interface-id found for EC2 instance %s", instanceId)
+	}
+
+	var interfaceId string
+	for _, instance := range out.Reservations[0].Instances {
+		if len(instance.NetworkInterfaces) == 0 {
+			return "", fmt.Errorf("no network-interface-id found for EC2 instance %s", instanceId)
+		}
+		for _, networkInterface := range instance.NetworkInterfaces {
+			if *(networkInterface.Attachment.DeviceIndex) == 0 {
+				interfaceId = *(networkInterface.NetworkInterfaceId)
+				if interfaceId != "" {
+					log.Infof("instance-id: %s, network-interface-id: %s", instanceId, interfaceId)
+					return interfaceId, nil
+				}
+			}
+			log.Infof("instance-id: %s, network-interface-id: %s", instanceId, interfaceId)
+		}
+		if interfaceId == "" {
+			return "", fmt.Errorf("no network-interface-id found for EC2 instance %s", instanceId)
+		}
+	}
+	return interfaceId, nil
+}
+
+func setEC2SourceDestinationCheck(ctx context.Context, svc *ec2.EC2, ec2NetId string, checkVal bool) error {
+	input := &ec2.ModifyNetworkInterfaceAttributeInput{
+		NetworkInterfaceId: aws.String(ec2NetId),
+		SourceDestCheck: &ec2.AttributeBooleanValue{
+			Value: aws.Bool(checkVal),
+		},
+	}
+
+	i := 0
+retry:
+	_, err := svc.ModifyNetworkInterfaceAttributeWithContext(ctx, input)
+	if err != nil {
+		if retriable(err) && i < defaultRetries {
+			log.WithField("net-instance-id", ec2NetId).Infof("retrying setting source-destination-check")
+			i++
+			goto retry
+		}
+		return err
+	}
+
+	log.Infof("set source-destination-check to %v on network-interface-id: %s", checkVal, ec2NetId)
+	return nil
+}
+
+func checkSourceDestinationValueIsDisable(check string) bool {
+	return check == "disable"
+}
+
+func UpdateSrcDstCheck(check string) error {
+	ctx, cancel := context.WithTimeout(context.Background(), defaultTimeOut*time.Second)
+	defer cancel()
+
+	awsSession, err := awssession.NewSession()
+	if err != nil {
+		return fmt.Errorf("error creating AWS session: %v", err)
+	}
+
+	ec2MetadataSvc := ec2metadata.New(awsSession)
+	if ec2MetadataSvc == nil {
+		return fmt.Errorf("error connecting to EC2 Metadata service: %v", err)
+	}
+
+	if !ec2MetadataSvc.AvailableWithContext(ctx) {
+		return fmt.Errorf("EC2 metadata service is unavailable")
+	}
+
+	ec2Region, err := getEC2Region(ctx, ec2MetadataSvc)
+	if err != nil {
+		return fmt.Errorf("error getting ec2 region: %s", convertError(err))
+	}
+
+	ec2Id, err := getEC2InstanceID(ctx, ec2MetadataSvc)
+	if err != nil {
+		return fmt.Errorf("error getting ec2 instance-id: %s", convertError(err))
+	}
+
+	ec2Svc := ec2.New(awsSession, aws.NewConfig().WithRegion(ec2Region))
+	if ec2Svc == nil {
+		return fmt.Errorf("error connecting to EC2 service")
+	}
+
+	ec2NetId, err := getEC2NetworkInterfaceId(ctx, ec2Svc, ec2Id)
+	if err != nil {
+		return fmt.Errorf("error getting ec2 network-interface-id: %s", convertError(err))
+	}
+
+	checkVal := !checkSourceDestinationValueIsDisable(check)
+	err = setEC2SourceDestinationCheck(ctx, ec2Svc, ec2NetId, checkVal)
+	if err != nil {
+		return fmt.Errorf("error setting src-dst-check for network-interface-id: %s", convertError(err))
+	}
+
+	return nil
+}
--- a/aws/ec2.go
+++ b/aws/ec2.go
@@ -1,3 +1,17 @@
+// Copyright (c) 2020 Tigera, Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 package aws
 
 import (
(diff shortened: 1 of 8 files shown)


aws/ec2.go
```diff
@@ -0,0 +1,225 @@
...
+		case "Unavailable":
+			return true
+		}
+	}
+
+	return false
+}
+
+func checkSourceDestinationValueIsDisable(check string) bool {
+	return check == "disable"
```
mgleung, line 49: nit: we could assign these strings to a constant so that if they change at any point, we'll just need to change the constant. Since the `disable` value is only really passed in for the tests, this isn't really necessary at this point.

aws/ec2.go
```diff
@@ -0,0 +1,225 @@
...
+		return "", fmt.Errorf("no network-interface-id found for EC2 instance %s", c.ec2InstanceId)
+	}
+
+	var interfaceId string
+	for _, instance := range out.Reservations[0].Instances {
+		if len(instance.NetworkInterfaces) == 0 {
+			return "", fmt.Errorf("no network-interface-id found for EC2 instance %s", c.ec2InstanceId)
+		}
+		for _, networkInterface := range instance.NetworkInterfaces {
+			if *(networkInterface.Attachment.DeviceIndex) == 0 {
```
mgleung, line 189: Is there a specific reason we want to return just the first network index we find? Would we ever want to return a different one?

aws/ec2.go
```diff
@@ -0,0 +1,225 @@
...
+			return true
+		case "Unavailable":
+			return true
+		}
+	}
+
+	return false
+}
+
+func checkSourceDestinationValueIsDisable(check string) bool {
```
caseydavenport, line 48: Do we really need a function just to do a string comparison? Why not just do the comparison inline?

aws/ec2.go
```diff
@@ -0,0 +1,225 @@
...
+
+	log.Infof("Successfully set source-destination-check to %t on network-interface-id: %s", checkVal, ec2NetId)
+	return nil
+}
+
+type ec2MetadataClient struct {
+	ec2MetadataSvc *ec2metadata.EC2Metadata
+}
+
+func NewEc2MetadataClient(awsSession *awssession.Session) (*ec2MetadataClient, error) {
```
caseydavenport, line 104: nit: Ec2 should be EC2 throughout.

config/config_params.go
```diff
@@ -211,6 +211,8 @@ type Config struct {
 	IpInIpMtu        int    `config:"int;1440;non-zero"`
 	IpInIpTunnelAddr net.IP `config:"ipv4;"`
 
+	AWSSrcDstCheck   string `config:"oneof(donothing,enable,disable);donothing;non-zero"`
```
caseydavenport, line 214: Shouldn't these be `DoNothing`, `Enable` and `Disable`? Should match the libcalico-go API.

aws/ec2.go
caseydavenport: Yep. We already have constants for these in the libcalico-go API, so we should be using those throughout IIUC.

aws/ec2_suite_test.go
```diff
@@ -0,0 +1,40 @@
...
+	"github.com/onsi/ginkgo/reporters"
+	//"github.com/sirupsen/logrus"
+
+	//"github.com/projectcalico/libcalico-go/lib/logutils"
+	"github.com/projectcalico/libcalico-go/lib/testutils"
+)
+
+func init() {
+	testutils.HookLogrusForGinkgo()
+	//logrus.AddHook(&logutils.ContextHook{})
```
caseydavenport, line 32: Commented code - should just remove.

daemon/daemon.go
```diff
@@ -361,6 +363,20 @@ configRetry:
 		simulateDataRace()
 	}
 
+	// Set source-destination-check on AWS EC2 instance.
```
caseydavenport, line 366: IIUC, we'll want to do this in a component hooked into the calculation graph rather than part of daemon.go.

@fasaxc does that match your understanding? Feels like we at least want it to be driven off of config updates post validation filter. And, potentially, handle these changes without triggering a felix restart. Feels like it doesn't need to go into the data plane though, since it's agnostic to which data plane is being used. WDYT?

aws/ec2.go
```diff
@@ -0,0 +1,225 @@
...
+func UpdateSrcDstCheck(check string) error {
+	ctx, cancel := context.WithTimeout(context.Background(), defaultTimeOut*time.Second)
+	defer cancel()
+
+	awsSession, err := awssession.NewSession()
+	if err != nil {
+		return fmt.Errorf("error creating AWS session: %v", err)
+	}
+
+    cli, err := NewEc2MetadataClient(awsSession)
```
caseydavenport, line 61: Indentation is off - probably need to run `goimports` across the code.

aws/ec2.go
```diff
@@ -0,0 +1,225 @@
...
+	}
+
+	i := 0
+retry:
+	out, err := c.ec2Svc.DescribeInstancesWithContext(ctx, input)
+	if err != nil {
+		if retriable(err) && i < defaultRetries {
+			log.WithField("instance-id", c.ec2InstanceId).Debug("retrying getting network-interface-id")
+			i++
+			goto retry
```
caseydavenport, line 174: Can we do this without a `goto` statement?

aws/ec2.go
```diff
@@ -0,0 +1,225 @@
...
+	}
+
+	i := 0
+retry:
+	_, err := c.ec2Svc.ModifyNetworkInterfaceAttributeWithContext(ctx, input)
+	if err != nil {
+		if retriable(err) && i < defaultRetries {
+			log.WithField("net-instance-id", ec2NetId).Debug("retrying setting source-destination-check")
+			i++
+			goto retry
```
caseydavenport, line 220: Same here - let's ditch the `goto` and instead use a loop. That's not a pattern we typically use in Calico.

aws/ec2.go
```diff
@@ -0,0 +1,225 @@
...
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/aws/ec2metadata"
+	awssession "github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	log "github.com/sirupsen/logrus"
+)
+
+const (
+	defaultTimeOut = 20
```
caseydavenport, line 17: nit: Timeout instead of TimeOut

caseydavenport: Also, we never change these values, whereas `default` implies they can be overridden.

aws/ec2.go
```diff
@@ -0,0 +1,225 @@
...
+	defaultRetries = 3
+)
+
+func convertError(err error) string {
+	if awsErr, ok := err.(awserr.Error); ok {
+		return fmt.Sprintf("%s: %s", awsErr.Code(), awsErr.Message())
```
caseydavenport, line 23: What does an AWS error normally stringify as?

aws/ec2.go
caseydavenport, line 104: Also, I think this is just an internally used function, so we should probably make it private.

aws/ec2.go
caseydavenport: I thought we passed in the device index to the query, so shouldn't it only ever return us a single network interface with device index 0? Do we need this loop at all?

daemon/daemon.go
realgaurav: This is currently invoked after the configuration/validation and the outcome of src-dst-check operation doesn't cause  Felix to restart.

aws/ec2.go
realgaurav: For UT.

aws/ec2.go
realgaurav: Yeah, it doesn't seem to work. Documentation allows returning using the filter but the response contains all the interfaces. Looping is to make sure` device-index = 0` is modified.

config/config_params.go
realgaurav: Most of the current configuration uses small-caps. Per our conversation, will match it to `libcalico-go`.

aws/ec2.go
caseydavenport: I don't think it's worth it. We don't need a UT for the `==` operator.

realgaurav: It's more about checking the valid values. Please have a look at the testing section and the idea that by default the input to set-src-check-dst will always be true unless set to false by `disable`.

aws/ec2.go
realgaurav: IIUC `Code`, `Message` and `OrigError`.

config/config_params.go
```diff
@@ -211,6 +211,8 @@ type Config struct {
 	IpInIpMtu        int    `config:"int;1440;non-zero"`
 	IpInIpTunnelAddr net.IP `config:"ipv4;"`
 
+	AWSSrcDstCheck string `config:"oneof(Donothing,Enable,Disable);DoNothing;non-zero"`
```
caseydavenport, line 214: Donothing -> DoNothing

aws/ec2.go
```diff
@@ -0,0 +1,226 @@
...
+
+	return false
+}
+
+func checkSourceDestinationValueIsDisable(check string) bool {
+	return check == apiv3.AWSSrcDstCheckOptionDisable
+}
+
+func UpdateSrcDstCheck(check string) error {
+	ctx, cancel := context.WithTimeout(context.Background(), timeout*time.Second)
```
fasaxc, line 57: Any reason not to make the constant `30 * time.Second`  Then the unit of the constant is clear.

aws/ec2.go
```diff
@@ -0,0 +1,226 @@
...
+
+type ec2Client struct {
+	EC2Svc        ec2iface.EC2API
+	ec2InstanceId string
+}
+
+func newEC2Client(ctx context.Context) (*ec2Client, error) {
+	awsSession, err := awssession.NewSession()
+	if err != nil {
+		return nil, fmt.Errorf("error creating AWS session: %v", err)
```
fasaxc, line 113: Use `%w` with `Errorf()`; it returns a wrapped error so the original error can be accessed if needed.

aws/ec2.go
```diff
@@ -0,0 +1,226 @@
...
+		if len(instance.NetworkInterfaces) == 0 {
+			return "", fmt.Errorf("no network-interface-id found for EC2 instance %s", c.ec2InstanceId)
+		}
+		// We are only modifying network interface with device-id-0 to update
+		// instance source-destination-check.
+		// An instance can have multiple interfaces and the API response can be
+		// out-of-order interface list. We compare the device-id in the
+		// response to make sure the right device is updated.
+		for _, networkInterface := range instance.NetworkInterfaces {
+			if *(networkInterface.Attachment.DeviceIndex) == deviceIndexZero {
```
fasaxc, line 184: I'd be inclined to put a nil check here; don't want felix to crash if the AWS API returns something unexpected.

daemon/daemon.go
fasaxc: Will this code work unmodified on Windows?  If not, it'll break when merged over to enterprise.  I'd be inclined to move it to the dataplane driver since it is part of setting up the dataplane (and that will prevent it from breaking on Windows).  If it's needed by multiple dataplane types then they can call through to a utility function that does this work.   (But it's probably not needed for all dataplane types that we may ever support.)

Felix normally restarts when its config changes; I'm not sure I see a reason to do the extra work for this config parameter to handle it without restart.  The main issue is that needs a lot more testing.

Should we back off to an occasional retry instead of trying to filter retriable/non-retriable errors?  I'm thinking a non-retriable error could be something like a permissions error; but the user could fix a permissions error at any time.  OTOH, if you have a large cluster with thousands of nodes hitting the API, that could cause throttling.

aws/ec2.go
fasaxc: Agree with Casey, I think it'd be clearer to check inline.

aws/ec2.go
```diff
@@ -0,0 +1,226 @@
...
+	if err != nil {
+		return err
+	}
+
+	ec2NetId, err := ec2Cli.getEC2NetworkInterfaceId(ctx)
+	if err != nil {
+		return fmt.Errorf("error getting ec2 network-interface-id: %s", convertError(err))
+	}
+
+	checkVal := !checkSourceDestinationValueIsDisable(check)
```
fasaxc, line 70: This line is confusing because it has ended up a double negative `not(is _dis_abled)`.  Why not just write it as `checkVal := check == apiv3.AWSSrcDstCheckOptionEnabled`

Given this is dealing with booleans and awkward naming from  AWS, I'd probably use more verbose variable name here:
```
s/check/caliCheckAction/
s/checkVal/checkEnabled/
```

aws/ec2.go
fasaxc: So, so we really need this formatting function?  Would we get something reasonable with `fmt.Errorf("... %w", err)`?  Using `%w` has the effect that the original error is wrapped and preserved so the caller can inspect it if that ever becomes needed.

dataplane/driver.go
```diff
@@ -279,6 +281,20 @@ func StartDataplaneDriver(configParams *config.Config,
...
+		const healthName = "felix-startup"
+		// Set source-destination-check on AWS EC2 instance.
+		if configParams.AWSSrcDstCheck != string(apiv3.AWSSrcDstCheckOptionDoNothing) {
+			go func(check, healthName string, healthAgg *health.HealthAggregator) {
+				log.Infof("Setting AWS EC2 source-destination-check to %s", check)
+				err := aws.UpdateSrcDstCheck(check)
+				if err != nil {
+					log.WithField("src-dst-check", check).Errorf("Failed to set source-destination-check: %v", err)
+					// set not-ready.
+					healthAggregator.Report(healthName, &health.HealthReport{Live: true, Ready: false})
```
fasaxc, line 293: Does it make sense to report not-ready if we're not going to retry?

Also, probably makes sense to give this its own health check name; that'll make it easier to debug if it fails.

I think I suggested adding a slow retry here since even errors like permissions could be fixed by the user.  WDYT?

aws/ec2.go
realgaurav: yeah, I think this is needed. Based on [the documentation](https://docs.aws.amazon.com/sdk-for-go/v1/developer-guide/handling-errors.html), `awserr` string-ifies Code, Message and OrigError when available.

To understand this better, I wrote a stub --
```
...
    interfaceID := "abcdef"
    srcDstCheckValue := true
    networkInterfaceAttribIn := &ec2.ModifyNetworkInterfaceAttributeInput{
        NetworkInterfaceId: aws.String(interfaceID),
        SourceDestCheck: &ec2.AttributeBooleanValue{
            Value: aws.Bool(srcDstCheckValue),
        },
    }

    networkInterfaceAttribOut, err := ec2Svc.ModifyNetworkInterfaceAttribute(networkInterfaceAttribIn)
    if err != nil {
        if aerr, ok := err.(awserr.Error); ok {
            fmt.Printf("err: %v\n", err)
            fmt.Printf("awserr: %s: %s\n", aerr.Code(), aerr.Message())
            return
...
```

here is the output for the stub --
```
err: InvalidNetworkInterfaceId.Malformed: Invalid id: "abcdef" (expecting "eni-...")
        status code: 400, request id: 57e804e8-0655-42bc-b26f-2d955eb7566f
awserr: InvalidNetworkInterfaceId.Malformed: Invalid id: "abcdef" (expecting "eni-...")
```

Notice the `OrigErr()` part of the output in the second line above. `OrigErr()` is not always set. But when set, shows _finer_ details which may not really useful to address the issue. When `OrigErr()` is not set (e.g. when AWS_REGION is not set), the output is identical for both Printf() in the stub. `convertError()` makes sure that the output always appears in the same format.

> Using %w has the effect that the original error is wrapped and preserved so the caller can inspect it if that ever becomes needed.

Regarding using `%w`, I am missing the point here. In general it's a good idea. However in this case, the caller is fixed. Had this code been part of a common API with several possible consumers, I do see the usefulness of error wrapping there. IMO, if this code were to ever change for such a use-case, we can adjust this code accordingly then.

For the current proposed change though, I don't see any disadvantage with the current format. Happy to be corrected if I missed any use-case or scenario.

dataplane/driver.go
realgaurav: Thanks for taking a look Shaun.

> Also, probably makes sense to give this its own health check name; that'll make it easier to debug if it fails.

:+1: 

> I think I suggested adding a slow retry here since even errors like permissions could be fixed by the user. WDYT?
>
>Should we back off to an occasional retry instead of trying to filter retriable/non-retriable errors? I'm thinking a non-retriable error could be something like a permissions error; but the user could fix a permissions error at any time. OTOH, if you have a large cluster with thousands of nodes hitting the API, that could cause throttling.

Actually, this is exactly the concern I had in mind. Should we keep trying indefinitely? A periodic fixed/random try from a node is OK but in case of a large cluster this could be an issue, at a minimum get throttled. This is the reason I am explicitly identifying the _retriable errors_ and capping the errors to a maximum of 3 retries. If we think permission issue is a _retriable error_, I can add that as such. But trying infinitely for any error possible IMO leaves this change more susceptible to being throttled. Idea here is that by marking the status as not-ready, we will let the Admin address the error by looking at the logs. Once the error is addressed, admin can restart Felix to retry setting src-dst-check value.

Please let me know what you think.

> Does it make sense to report not-ready if we're not going to retry?

I am reading this as - if it makes sense to set the status as not-ready if we are going to try in fixed/random interval? - I think this is useful. We can possibly set the status to not-ready for an explicit configuration (AWSSrcDstCheck=Enable). By setting the status we want Admin to take a note of the failure.

fasaxc: SGTM

PR #2386, opened 2020-06-15
Make race detector optional.  Enable by default.

fasaxc:
## Description

Add a new build option to enable/disable the race detector:

```
make fv FV_RACE_DETECTOR_ENABLED=true/false
```

Enabled by default.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -94,6 +94,7 @@ FV_NUM_BATCHES?=1
 # (with FV_NUM_BATCHES=1) to check that it's not a flake.
 FV_BATCHES_TO_RUN?=$(shell seq $(FV_NUM_BATCHES))
 FV_SLOW_SPEC_THRESH=90
+FV_RACE_DETECTOR_ENABLED?=true
 
 # Linker flags for building Felix.
 #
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -94,6 +94,7 @@ FV_NUM_BATCHES?=1
 # (with FV_NUM_BATCHES=1) to check that it's not a flake.
 FV_BATCHES_TO_RUN?=$(shell seq $(FV_NUM_BATCHES))
 FV_SLOW_SPEC_THRESH=90
+FV_RACE_DETECTOR_ENABLED?=true
```
Brian-McM, line 97: Can we default this to false for the moment? The fv's are likely to fail right now, aren't they?

Brian-McM: Realised this is fine for OS and we'll disable it for enterprise

Makefile
```diff
@@ -246,8 +247,14 @@ ifeq ($(ARCH),amd64)
 	docker tag $(BUILD_IMAGE):latest-$(ARCH) $(BUILD_IMAGE):latest
 endif
 
-image-test: image fv/Dockerfile.test.amd64 bin/pktgen bin/test-workload bin/test-connection bin/calico-felix-race-amd64 image-wgtool
-	docker build -t $(BUILD_IMAGE)-test:latest-$(ARCH) --build-arg QEMU_IMAGE=$(CALICO_BUILD) --file ./fv/Dockerfile.test.$(ARCH) bin;
+ifeq ($(FV_RACE_DETECTOR_ENABLED),true)
+FV_BINARY=calico-felix-race-amd64
```
realgaurav, line 251: This seem to be forcing `ARCH=amd64` but the `docker build` command below consider $(ARCH).
Is this expected behavior?

Brian-McM: @fasaxc, can you follow up with this later please? Since this isn't a regression we're merging this through so that I can get unblocked.

fasaxc: Yeah, FVs are amd64 only (previously that was just hard-coded in the dockerfile).

PR #2389, opened 2020-06-16
Add BPF dataplane support for Wireguard

neiljerram:
We need a new BPF program variant for Wireguard interfaces, because a
Wireguard device sees an SKB with the IP header at offset 0, in both
directions.  (aka It's an L3 device.)

For now we prevent direct redirect from the Wireguard ingress problem,
because that would need adding back an Ethernet header.  There will be
follow-on work to try to do that.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/dataplane/linux/bpf_ep_mgr.go
+++ b/dataplane/linux/bpf_ep_mgr.go
@@ -614,6 +614,8 @@ func (m *bpfEndpointManager) attachDataIfaceProgram(ifaceName string, polDirecti
 	epType := tc.EpTypeHost
 	if ifaceName == "tunl0" {
 		epType = tc.EpTypeTunnel
+	} else if ifaceName == "wireguard.cali" {
+		epType = tc.EpTypeWireguard
 	}
 	ap := m.calculateTCAttachPoint(epType, polDirection, ifaceName)
 	ap.HostIP = m.hostIP
--- a/bpf-gpl/bpf.h
+++ b/bpf-gpl/bpf.h
@@ -106,11 +106,12 @@ struct bpf_map_def_extended {
 };
 
 /* These constants must be kept in sync with the calculate-flags script. */
-#define CALI_TC_HOST_EP	(1<<0)
-#define CALI_TC_INGRESS	(1<<1)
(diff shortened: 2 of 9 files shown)


.semaphore/semaphore.yml
```diff
@@ -89,7 +89,7 @@ blocks:
       - ./.semaphore/on-test-vm make --directory=${REPO_NAME} fv-wireguard FORCE_WIREGUARD_FV="1"
     - name: FV on newer kernel
       commands:
-      - ./.semaphore/on-test-vm make --directory=${REPO_NAME} fv-bpf GINKGO_FOCUS=BPF-SAFE FV_NUM_BATCHES=${SEMAPHORE_JOB_COUNT} FV_BATCHES_TO_RUN="${SEMAPHORE_JOB_INDEX}"
+      - ./.semaphore/on-test-vm make --directory=${REPO_NAME} fv-bpf GINKGO_FOCUS=BPF-SAFE FV_NUM_BATCHES=${SEMAPHORE_JOB_COUNT} FV_BATCHES_TO_RUN="${SEMAPHORE_JOB_INDEX}" FELIX_FV_WIREGUARD_AVAILABLE=true
```
fasaxc, line 92: This flag overrides the kernel version check so that the test blows up if wireguard can't be enabled but it doesn't actually enable it!

dataplane/linux/bpf_ep_mgr.go
```diff
@@ -614,6 +614,8 @@ func (m *bpfEndpointManager) attachDataIfaceProgram(ifaceName string, polDirecti
 	epType := tc.EpTypeHost
 	if ifaceName == "tunl0" {
 		epType = tc.EpTypeTunnel
+	} else if ifaceName == "wireguard.cali" {
```
tomastigera, line 617: nit: a switch perhaps?

neiljerram: If you don't mind I'll leave this for now (to avoid another CI cycle), but will keep it in mind for further work.

PR #2393, opened 2020-06-17
fv: ExpectConnectivity is a universal expectation 

tomastigera:
## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/fv/connectivity/conncheck.go
+++ b/fv/connectivity/conncheck.go
@@ -40,8 +40,9 @@ import (
 // actual state of the connectivity between the given workloads.  It is expected to be used like so:
 //
 //     var cc = &connectivity.Checker{}
-//     cc.ExpectNone(w[2], w[0], 1234)
-//     cc.ExpectSome(w[1], w[0], 5678)
+//     cc.Expect(None, w[2], w[0], 1234)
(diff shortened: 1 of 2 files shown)


fv/connectivity/conncheck.go
```diff
@@ -56,24 +56,25 @@ type Checker struct {
...
+	c.expect(false, from, to, ExpectWithPorts(explicitPort...))
 }
 
 // ExpectConnectivity asserts existing connectivity between a ConnectionSource
 // and ConnectionTarget with details configurable with ExpectationOption(s).
 // This is a super set of ExpectSome()
-func (c *Checker) ExpectConnectivity(from ConnectionSource, to ConnectionTarget,
-	ports []uint16, opts ...ExpectationOption) {
-	c.expect(true, from, to, ports, opts...)
+func (c *Checker) ExpectConnectivity(connectivity bool,
```
fasaxc, line 74: How about dropping `Connectivity` since it's implicit given this is a method of connectivity checker?

I'm not so keen on a bare true/false argument, how about `cc.Expect(conncheck.Blocked|Allowed, ...)`  (but then the conncheck prefix is a bit verbose)

PR #2394, opened 2020-06-17
bpf: Reduced allocations in map related syscalls

tomastigera:
We use the slice buffer without allocating additional space.
## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/bpf/bpf_syscall.go
+++ b/bpf/bpf_syscall.go
@@ -30,6 +30,9 @@ import (
 // #include <linux/bpf.h>
 // #include <stdlib.h>
 // #include <string.h>
+// #include <errno.h>
+// #include <unistd.h>
+// #include <sys/syscall.h>
 //
 // union bpf_attr *bpf_attr_alloc() {
 //    union bpf_attr *attr = malloc(sizeof(union bpf_attr));
@@ -133,6 +136,18 @@ import (
 // __u32 bpf_attr_prog_run_duration(union bpf_attr *attr) {
 //    return attr->test.duration;
 // }
+//
+// int bpf_map_call(int cmd, __u32 map_fd, void *pointer_to_key, void *pointer_to_value, __u64 flags) {
+//    union bpf_attr attr = {};
+//
+//    attr.map_fd = map_fd;
+//    attr.key = (__u64)(unsigned long)pointer_to_key;
+//    attr.value = (__u64)(unsigned long)pointer_to_value;
+//    attr.flags = flags;
+//
+//    return syscall(SYS_bpf, cmd, &attr, sizeof(attr)) == 0 ? 0 : errno;
+// }
+//
 import "C"
 
 func SyscallSupport() bool {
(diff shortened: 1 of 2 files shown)


tomastigera:
The failed test is an unrelated wireguard flake

bpf/bpf_syscall.go
```diff
@@ -390,32 +379,76 @@ func GetMapNextKey(mapFD MapFD, k []byte, keySize int) ([]byte, error) {
...
+	if vSize > 0 {
+		x.cV = C.malloc(C.size_t(vSize))
+	}
+
+	return x
+}
+
+func (x bpfSyscallMapMemCtx) Free() {
+	C.free(x.cK)
+	C.free(x.cV)
```
fasaxc, line 409: Is freeing a NULL pointer OK?

bpf/bpf_syscall.go
```diff
@@ -390,32 +379,76 @@ func GetMapNextKey(mapFD MapFD, k []byte, keySize int) ([]byte, error) {
...
-		defer C.free(cK)
+	if len(k) >= 1 {
+		cK = unsafe.Pointer(&k[0])
 	}
-	cV := C.malloc(C.size_t(keySize)) // value has the size of a key - it is the key!
-	defer C.free(cV)
 
-	C.bpf_attr_setup_map_get_next_key(bpfAttr, C.uint(mapFD), cK, cV, 0)
+	C.bpf_attr_clear(x.attr)
+	C.bpf_attr_setup_map_get_next_key(x.attr, C.uint(mapFD), cK, x.cK, 0)
```
fasaxc, line 425: Assuming k is allocated as a Go slice, I don't think it's valid to embed a pointer to k[0] in the attr union.  It breaks one of the pointer rules in the [cgo docs](https://golang.org/cmd/cgo/#hdr-Passing_pointers):

> C code may not keep a copy of a Go pointer after the call returns.

and

> Go code may not store a Go pointer in C memory. C code may store Go pointers in C memory, subject to the rule above: it must stop storing the Go pointer when the C function returns.

tomastigera: Hmmm my understanding was that this does not fall in that cathegory as C technically does not keep that pointer and go still has a live pointer, but you are likely right that letting one function to fill it in and another function to make the syscall may violate it. I will unify it so that it does not split in 2 calls.

bpf/bpf_syscall.go
```diff
@@ -390,32 +379,76 @@ func GetMapNextKey(mapFD MapFD, k []byte, keySize int) ([]byte, error) {
...
+	x := allocBpfSyscallMapMemCtx(keySize, 0)
+	defer x.Free()
 
+	return x.GetMapNextKey(mapFD, k, keySize)
+}
+
+type bpfSyscallMapMemCtx struct {
+	cK   unsafe.Pointer
+	cV   unsafe.Pointer
+	attr *[72]byte
```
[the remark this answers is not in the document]
tomastigera: no need for this now

PR #2395, opened 2020-06-18
bpf/proxy flaky UTs fixes

tomastigera:
## Description

It is perhaps a little too complicated, but I felt like changing completely to `Eventually` and turning all the `Expect` in `checkstate` into a single true/false function less readable and more difficult to maintain.

If a test fails, we also get which of the assertions was wrong - the first that failed in the last iteration.

    bpf/proxy: StaleUDPSvcs not used anymore
    
    Syncer calculates this on its own when necessary

    bpf/proxy: checkState checks "eventually"
    
    We need to use moc syncer's Expect in checkState if we want to wait for
    the expectations to be OK eventually.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/bpf/proxy/proxy.go
+++ b/bpf/proxy/proxy.go
@@ -19,7 +19,6 @@ package proxy
 
 import (
 	"net"
-	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


tomastigera:
> (Went digging because I knew you can change the Gomega fail handler; didn't realise they'd have a built-in way to temporarily swap it out!)

This is great! I stopped where I found the exchange of the global handler. This will make it all much simpler! Thanks!

PR #2398, opened 2020-06-24
New unit and FV tests to test service type changes

sridhartigera:
UT and FV tests to verify changes in service type

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/bpf/proxy/service_type_change_test.go
+++ b/bpf/proxy/service_type_change_test.go
@@ -343,7 +343,8 @@ func setSvcTypeToClusterIP(testSvc *v1.Service, k8s *fake.Clientset) {
 	testSvc.Spec.LoadBalancerSourceRanges = []string{}
 	testSvc.Spec.Type = v1.ServiceTypeClusterIP
 	testSvc.Spec.Ports[0].NodePort = 0
-	k8s.CoreV1().Services(v1.NamespaceDefault).Update(testSvc)
+	_, err := k8s.CoreV1().Services(v1.NamespaceDefault).Update(testSvc)
+	Expect(err).NotTo(HaveOccurred())
 }
 
 func setSvcTypeToExternalIP(testSvc *v1.Service, extIP []string, k8s *fake.Clientset) {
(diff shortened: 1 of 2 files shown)


bpf/proxy/service_type_change_test.go
```diff
@@ -0,0 +1,362 @@
...
+				if len(front.m) == 1 && ret1 {
+					return true
+				}
+				return false
+			}).Should(BeTrue())
+		})
+		// cluster IP -> external IP
+		By("Update the service type from ClusterIP to ExternalIP", func() {
+			setSvcTypeToExternalIP(testSvc, []string{extIPstr}, k8s)
+			Eventually(func() bool {
```
tomastigera, line 117: we could wrap the testing function in something like 

```
func succeeds(conds func()) bool {
	front.Lock()
	defer front.Unlock()
        return len(interceptGomegaFailures(conds)) > 0
}       
```

and then we could use the gomega matchers to express the test like 

```
Eventually(succeeds(func() {
    Expect(front.m).To(HaveLen(2))
    Expect(front.m).To(HaveKey(key_clusterIP))
    Expect(front.m).To(HaveKey(key_extIP))
}))

Which might be a bit more readable. 

Or `ret1` & co. may become `key_clusterIP_exists` so that the final condition is clear.

File: bpf/proxy/service_type_change_test.go [hunk 2]
```diff
@@ -0,0 +1,362 @@
...
+	initIP := net.IPv4(1, 1, 1, 1)
+
+	front := newMockNATMap()
+	back := newMockNATBackendMap()
+	aff := newMockAffinityMap()
+	ct := mock.NewMockMap(conntrack.MapParams)
+	p, _ := proxy.StartKubeProxy(k8s, "test-node", front, back, aff, ct, proxy.WithImmediateSync())
+	p.OnHostIPsUpdate([]net.IP{initIP})
+
+	key_clusterIP := nat.NewNATKey(clusterIP, port, proxy.ProtoV1ToIntPanic(proto))
```
tomastigera, line 92: The 'C'ish naming convention is unGOish and the only true way ;-) is `camelCase`

```suggestion
	keyClusterIP := nat.NewNATKey(clusterIP, port, proxy.ProtoV1ToIntPanic(proto))
```

bpf/proxy/service_type_change_test.go
```diff
@@ -0,0 +1,362 @@
...
+				_, ret3 := front.m[key_extIP]
+				if ret1 && ret2 && !ret3 {
+					return true
+				}
+				return false
+			}).Should(BeTrue())
+		})
+
+		// NodePort -> LoadBalancer
+		By("Update the service type from NodePort to LoadBalancer", func() {
```
tomastigera, line 260: Since we already do all this sanity checking, perhaps LB with src ip range too?

fv/bpf_test.go
```diff
@@ -1005,6 +1005,291 @@ func describeBPFTests(opts ...bpfTestOpt) bool {
 					})
 				})
 
+				Context("Test Service type transitions", func() {
```
tomastigera, line 1008: I wonder what is the delta in runtime now, would it be worth to limit it to say TCP only?

PR #2404, opened 2020-06-26
Add FV tests for WireGuard host to workload

neiljerram:
## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/fv/wireguard_test.go
+++ b/fv/wireguard_test.go
@@ -830,7 +830,7 @@ var _ = infrastructure.DatastoreDescribe("WireGuard-Supported 3 node cluster", [
 			cc.CheckConnectivity()
 
 			By("verifying packets between felix-0 and felix-1 is encrypted")
-			for i := range []int{0, 1} {
+			for _, i := range []int{0, 1} {
 				Eventually(func() int {
 					return tcpdumps[i].MatchCount("numTunnelPackets01")
 				}, "10s", "100ms").Should(BeNumerically(">", 0))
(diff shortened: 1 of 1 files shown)


fv/wireguard_test.go
```diff
@@ -820,6 +824,37 @@ var _ = infrastructure.DatastoreDescribe("WireGuard-Supported 3 node cluster", [
 				}, "10s", "100ms").Should(BeNumerically(">", 0))
 			}
 		})
+
+		It("WireGuard should be used for host to workload connections on WireGuard enabled nodes", func() {
+			cc.ExpectSome(felixes[0], wls[1])
```
realgaurav, line 829: `cc.ExpectSome(felixes[1], wls[0])`?

fv/wireguard_test.go
```diff
@@ -820,6 +824,37 @@ var _ = infrastructure.DatastoreDescribe("WireGuard-Supported 3 node cluster", [
...
+			}
+		})
+
+		It("WireGuard should not be used for host to workload connections when WireGuard disabled on either node", func() {
+			cc.ExpectSome(felixes[0], wls[2])
+			cc.ExpectSome(felixes[2], wls[0])
+			cc.CheckConnectivity()
+
+			By("verifying packets between felix-0 and felix-2 are not encrypted")
+			for i := range []int{0, 2} {
```
realgaurav, line 849: Should this be?

`for _, i := range []int{0, 2} {`

fv/wireguard_test.go
neiljerram: No; as that should be exactly symmetric, I thought the encrypted packets testing would be more precise if it was done with just one attempted connection.

fv/wireguard_test.go
neiljerram: Thanks, good catch!

PR #2409, opened 2020-06-30
Adding support to detect RHEL with kernel version 4.18.0

sridhartigera:
## Description
Changes to detect RHEL with kernel version 4.18.0. To check if the kernel supports BPF, a new mapping between the linux distribution name and the kernel version supported is added. 
1) Read the distribution name and kernel version from /proc/version
2) Check if the read version is same as the expected one in the mapping table.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
In BPF mode, Felix now detects BPF support on Red Hat kernels with backports as well as generic kernels.
```

### DIFF
--- a/bpf/bpf.go
+++ b/bpf/bpf.go
@@ -92,11 +92,8 @@ var (
 	// required features we use for XDP filtering
 	v4Dot16Dot0 = versionparse.MustParseVersion("4.16.0")
 	// v4Dot18Dot0 is the kernel version in RHEL that has all the
-	// required features for BPF dataplane
+	// required features for BPF dataplane, sidecar acceleration
 	v4Dot18Dot0 = versionparse.MustParseVersion("4.18.0")
-	// v4Dot20Dot0 is the first kernel version that has all the
-	// required features we use for sidecar acceleration
-	v4Dot20Dot0 = versionparse.MustParseVersion("4.20.0")
 	// v5Dot3Dot0 is the first kernel version that has all the
 	// required features we use for BPF dataplane mode
 	v5Dot3Dot0 = versionparse.MustParseVersion("5.3.0")
--- a/bpf/bpf.go
+++ b/bpf/bpf.go
@@ -37,7 +37,6 @@ import (
 	"strings"
 	"syscall"
 
-	version "github.com/hashicorp/go-version"
 	log "github.com/sirupsen/logrus"
 	"golang.org/x/sys/unix"
 
@@ -102,7 +101,7 @@ var (
 	v5Dot3Dot0 = versionparse.MustParseVersion("5.3.0")
 )
 
-var distToVersionMap = map[string]*version.Version{
+var distToVersionMap = map[string]*versionparse.Version{
 	"ubuntu":  v5Dot3Dot0,
 	"rhel":    v4Dot18Dot0,
 	"default": v5Dot3Dot0,
@@ -2169,7 +2168,7 @@ func (b *BPFLib) RemoveSockmapEndpointsMap() error {
 	return os.Remove(mapPath)
 }
 
-func isAtLeastKernel(v *version.Version) error {
+func isAtLeastKernel(v *versionparse.Version) error {
 	versionReader, err := versionparse.GetKernelVersionReader()
 	if err != nil {
 		return fmt.Errorf("failed to get kernel version reader: %v", err)
@@ -2180,7 +2179,7 @@ func isAtLeastKernel(v *version.Version) error {
 		return fmt.Errorf("failed to get kernel version: %v", err)
 	}
 
-	if versionparse.Compare(kernelVersion, v) < 0 {
+	if kernelVersion.Compare(v) < 0 {
 		return fmt.Errorf("kernel is too old (have: %v but want at least: %v)", kernelVersion, v)
 	}
 
@@ -2200,7 +2199,7 @@ func SupportsSockmap() error {
 	return nil
 }
 
-func GetMinKernelVersionForDistro(distName string) *version.Version {
+func GetMinKernelVersionForDistro(distName string) *versionparse.Version {
 	return distToVersionMap[distName]
 }
 
--- a/bpf/bpf.go
+++ b/bpf/bpf.go
@@ -93,7 +93,7 @@ var (
 	v4Dot16Dot0 = versionparse.MustParseVersion("4.16.0")
 	// v4Dot18Dot0 is the kernel version in RHEL that has all the
 	// required features for BPF dataplane, sidecar acceleration
-	v4Dot18Dot0 = versionparse.MustParseVersion("4.18.0")
+	v4Dot18Dot0 = versionparse.MustParseVersion("4.18.0-193")
 	// v5Dot3Dot0 is the first kernel version that has all the
 	// required features we use for BPF dataplane mode
 	v5Dot3Dot0 = versionparse.MustParseVersion("5.3.0")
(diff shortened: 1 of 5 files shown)


bpf/bpf.go
```diff
@@ -2191,8 +2197,13 @@ func SupportsSockmap() error {
 	return nil
 }
 
+func GetExpectedVersionFromMap(distName string) *version.Version {
```
fasaxc, line 2200: Name of the function could be improved.  Right now it's too generic: "expected for what?" "which map?"  How about `MinKernelVersionForDistro()`

bpf/bpf.go
```diff
@@ -91,14 +91,20 @@ var (
 	// v4Dot16Dot0 is the first kernel version that has all the
 	// required features we use for XDP filtering
 	v4Dot16Dot0 = versionparse.MustParseVersion("4.16.0")
-	// v4Dot20Dot0 is the first kernel version that has all the
-	// required features we use for sidecar acceleration
-	v4Dot20Dot0 = versionparse.MustParseVersion("4.20.0")
+	// v4Dot18Dot0 is the kernel version in RHEL that has all the
+	// required features for BPF dataplane, sidecar acceleration
+	v4Dot18Dot0 = versionparse.MustParseVersion("4.18.0-193")
```
fasaxc, line 96: Unfortunately, I think the hashicorp version library we're using interprets  `x.y.z-1234` as coming before  `x.y.z`, which is not what we want.   (That's because it uses semver, which is not the scheme that redhat uses: https://play.golang.org/p/k_g9cOONGV5)

bpf/bpf_test.go
```diff
@@ -761,3 +762,46 @@ func TestIPv6NotSupported(t *testing.T) {
 		t.Fatalf("creating an IPv6 blacklist should have failed")
 	}
 }
+
+func TestVersionParse(t *testing.T) {
+	t.Log("Test version parsing")
+	ubuntuVersionStr := "Linux version 5.3.0-39-generic (buildd@lcy01-amd64-016) (gcc version 9.3.0 (Ubuntu 9.3.0-10ubuntu2)) #43-Ubuntu SMP Fri Jun 19 10:28:31 UTC 2020"
+	rhelVersionStr := "Linux version 4.18.0-193.el8.x86_64 (mockbuild@x86-vm-08.build.eng.bos.redhat.com) (gcc version 8.3.1 20191121 (Red Hat 8.3.1-5) (GCC)) #1 SMP Fri Mar 27 14:35:58 UTC 2020"
+	fedVersionStr := "Linux version 5.3.0-193.el8.x86_64 (mockbuild@x86-vm-08.build.eng.bos.redhat.com) (gcc version 8.3.1 20191121 (Fedora 8.3.1-5) (GCC)) #1 SMP Fri Mar 27 14:35:58 UTC 2020"
```
fasaxc, line 770: Should we try to treat Fedora and Centos as RH distros?  Do they use the same kernels?

sridhartigera: Centos is same as RHEL : 
Linux version 4.18.0-193.6.3.el8_2.x86_64 (mockbuild@kbuilder.bsys.centos.org) (gcc version 8.3.1 20191121 (Red Hat 8.3.1-5) (GCC)) #1 SMP Wed Jun 10 11:09:32 UTC 2020

sridhartigera: Fedora:
Linux version 2.6.8-1.523 (user@foo.redhat.com) (gcc version 3.4.1 20040714 \  (Red Hat Enterprise Linux 3.4.1-7)) #1 Mon Aug 16 13:27:03 EDT 2004

As both centos and fedora has "Red Hat" in the show version output, it will be checked for kernel version 4.18.0-193

versionparse/lib.go
```diff
@@ -20,47 +20,158 @@ import (
...
+		return -1
+	}
+	if vlen > olen {
+		return 1
+	}
+	return 0
+}
+
+func convertVersionToIntSlice(s string) ([]int, error) {
+	splitStrDash := strings.Split(s, "-")
```
fasaxc, line 87: Think this logic would work out simpler if you made a simple regex to split the on either `.` or `-`:
```
var splitRe = regexp.MustParse(`[\.-]`)
func convertVersionToIntSlice(s string) ([]int, error) {
    parts := splitRe.Split(s)
```

versionparse/lib.go
```diff
@@ -20,47 +20,158 @@ import (
...
+	"strings"
 
-	version "github.com/hashicorp/go-version"
 	log "github.com/sirupsen/logrus"
 )
 
 var (
-	kernelVersionRegexp = regexp.MustCompile(`Linux version (\d+\.\d+\.\d+)`)
+	kernelVersionRegexp     = regexp.MustCompile(`Linux version (\d+\.\d+\.\d+)`)
+	kernelVersionRHELRegexp = regexp.MustCompile(`Linux version (\d+\.\d+\.\d-\d+)`)
```
fasaxc, line 31: Are we sure that RH always append a `-%d+` on their versions?  What if they ship a vanilla kernel as the first one?  Think I'd make the regex generic so it can match either type:
```
regexp.MustCompile(`Linux version (\d+\.\d+\.\d+(?:-\d+)?)`)
```

bpf/bpf_test.go
```diff
@@ -761,3 +762,75 @@ func TestIPv6NotSupported(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("Parsing kernel version failed")
+	}
+	expVer = GetMinKernelVersionForDistro(distName)
+	if parsedVer.Compare(expVer) != 0 {
+		t.Fatalf("Parsed version not same as expected version")
+	}
+
+	// Tests to verify the compare function in versionparse
+	ver1 := versionparse.MustParseVersion("4.18.0-193")
```
fasaxc, line 811: Consider making a table-driven test for this kind of testing.  Go makes that quite easy to do (since tests are "just go") it's a good technique to have in your arsenal:
```
type comparisonTest struct {a,b string, expected int}
tests:= []comparisonTest{
  {a: "4.18.0-194", b: "4.18.0-192", expected: -1},
  ...
}
for _, test := range tests {
   t.Log("Comparing ", a, " to ", b)
   // parse a
   // parse b
   // compare them
}
```

bpf/bpf_test.go
```diff
@@ -761,3 +762,75 @@ func TestIPv6NotSupported(t *testing.T) {
...
+	ver2 := versionparse.MustParseVersion("4.18.0-194")
+	if ver1.Compare(ver2) != -1 {
+		t.Fatalf("Comparing kernel versions 4.18.0-193 and 4.18.0-194 failed")
```
fasaxc, line 814: Nothing wrong with this but I typically use the Gomega assertions even in `go test` style tests.  They're concise and they tend to give good diags: `Expect(ver1.Compare(ver2)).To(Equal(-1))`

To use them, you have to call `RegisterTestingT(t)` at the top of the function.

versionparse/lib.go
```diff
@@ -20,47 +20,132 @@ import (
...
+		return -1
+	}
+	if vlen > olen {
+		return 1
+	}
+	return 0
+}
+
+func convertVersionToIntSlice(s string) ([]int, error) {
+	var splitRe = regexp.MustCompile(`[\.-]`)
```
fasaxc, line 86: Suggest moving the regex creation out of the function so it only gets compiled once.

PR #2410, opened 2020-06-30
Allow tc to run in parallel again.

fasaxc:
## Description

TC based object-pinned map names on the hash of the object file but we really want one map per attached instance of the program.  

- Add a UUID to each binary to avoid hash clashes.
- Use RWMutex to allow attach to run in parallel.
- Remove map repinning logic.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
In BPF mode, Felix now attaches programs in parallel for improved performance.
```

### DIFF
--- a/bpf/binary.go
+++ b/bpf/binary.go
@@ -18,9 +18,11 @@ package bpf
 
 import (
 	"bytes"
+	"crypto/rand"
 	"encoding/binary"
 	"io/ioutil"
 	"net"
+	"os"
 
 	"github.com/pkg/errors"
 )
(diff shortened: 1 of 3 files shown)


bpf/binary.go
```diff
@@ -44,7 +46,34 @@ func BinaryFromFile(ifile string) (*Binary, error) {
...
-	return ioutil.WriteFile(ofile, b.raw, 0600)
+	err := ioutil.WriteFile(ofile, b.raw, 0600)
+	if err != nil {
+		return err
+	}
+
+	// Append a UUID to the file.  We want each attachment point to get its own jump map
+	// but tc names jump maps by hash of the binary, which means they can clash if we load
+	// the same binary onto multiple interfaces.
+	f, err := os.OpenFile(ofile, os.O_APPEND|os.O_WRONLY, 0600)
```
tomastigera, line 57: nice! I guess it does not matter what rubbish follows the ELF, or does it?

fasaxc: Yes, tc doesn't seem to care about extra data at the end.  I think it's pretty common to stuff data at the end of an executable and the ELF spec seems to allow it (e.g. an auto-extracting archive).

PR #2417, opened 2020-07-07
Support for both appending and inserting chains in iptables dataplane

caseydavenport:
## Description

Goal of this PR is to enable both insertion and appending of iptables
chains (previously only one or the other was supported), to enable using 
that functionality (in a follow on PR) to make sure kube-proxy rules are hit even when policy
accepts the traffic by appending the accept rule after the kube-proxy
rules.

### DIFF
--- a/dataplane/linux/int_dataplane.go
+++ b/dataplane/linux/int_dataplane.go
@@ -927,13 +927,13 @@ func (d *InternalDataplane) setUpIptablesBPF() {
 				Comment: []string{"To workload, BPF will handle."},
 			})
 		}
-		t.SetRuleInsertions("INPUT", inputRules)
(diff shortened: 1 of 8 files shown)


neiljerram:
@song-jiang Per our discussion yesterday, I understood that you have already done this, so I wonder what this new PR is adding?

fasaxc:
@neiljerram this PR is a cherry-pick of Song's work

fasaxc:
I think you should bank this PR as a simple cherry-pick then follow up with the PR to fix services with no endpoints.

caseydavenport:
Moved the fix for service rejection to a separate PR: https://github.com/projectcalico/felix/pull/2424

PR #2418, opened 2020-07-08
Do FIB lookup on to-host path.

fasaxc:
## Description

Looks like a bug that we were doing it on the opposite path.
## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/bpf-gpl/tc.c
+++ b/bpf-gpl/tc.c
@@ -805,7 +805,7 @@ static CALI_BPF_INLINE struct fwd calico_tc_skb_accepted(struct __sk_buff *skb,
 		// to trigger that and leave the fib lookup disabled.
 		seen_mark = CALI_SKB_MARK_NAT_OUT;
 	} else {
-		// Non-SNAT case, see if we can do the FIB lookup.  Note: tried to pass in
+		// Non-SNAT case, allow FIB lookup only if RPF check passed.  Note: tried to pass in
 		// the calculated value from calico_tc but hit verifier issues so recalculate it
 		// here.
 		if (CALI_F_TO_HOST && !ct_result_rpf_failed(state->ct_result.rc)) {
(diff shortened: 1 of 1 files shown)


bpf-gpl/tc.c
```diff
@@ -586,9 +586,9 @@ static CALI_BPF_INLINE int calico_tc(struct __sk_buff *skb)
 	}
 
 	/* We are possibly past (D)NAT, but that is ok, we need to let the IP
-	 * stack do the RPF check on the source, dest is not importatnt.
+	 * stack do the RPF check on the source, dest is not important.
 	 */
-	if (CALI_F_TO_HOST && ct_result_rpf_failed(state.ct_result.rc)) {
```
tomastigera, line 591: The tiny benefit of the `CALI_F_TO_HOST &&` is that it is compiled out for other programs. Probably makes little difference in runtime and the code is simpler without it.

bpf-gpl/tc.c
```diff
@@ -801,12 +801,14 @@ static CALI_BPF_INLINE struct fwd calico_tc_skb_accepted(struct __sk_buff *skb,
...
 	if (CALI_F_FROM_WEP && (state->flags & CALI_ST_NAT_OUTGOING)) {
+		// We are going to SNAT this traffic, using iptables SNAT so set the mark
+		// to trigger that and leave the fib lookup disabled.
 		seen_mark = CALI_SKB_MARK_NAT_OUT;
 	} else {
-		/* XXX we do it here again because doing it in one place only
-		 * XXX in calico_tc() irritates the verifier :'(
-		 */
-		if (!CALI_F_TO_HOST || !ct_result_rpf_failed(state->ct_result.rc)) {
+		// Non-SNAT case, see if we can do the FIB lookup.  Note: tried to pass in
```
tomastigera, line 808:
```suggestion
		// Non-SNAT case, allow FIB lookup only if RPF check passed.  Note: tried to pass in
```

bpf-gpl/tc.c
fasaxc: I'm hoping the compiler is smart enough to spot that fib is already set to false at the top so this conditional is a no-op.

fasaxc:
Think @tomastigera was happy with this and CI fail was a semaphore funny (on an unrelated code path).

PR #2423, opened 2020-07-13
Drop packets to unknown WEPs

fasaxc:
## Description

* Add an iptables chain that accepts traffic to known workloads.
* Refactor the mark bits to use fewer bytes and to give a clean bit flag for "bypass".
* Rearrange the base iptables rules to:

  * Check the bypass bit first and immediately accept.
  * Check for missing marks (i.e. missing programs).
  * Jump to the new chain.

## Todos
- [x] Ensure only packets that are approved at both ends get bypass (might already be true)
- [x] Release note

## Release Note

```release-note
In BPF mode, traffic to unknown workload interfaces is now blocked (as long as Felix was running long enough to insert its policing rules).
```

### DIFF
--- a/dataplane/linux/bpf_ep_mgr.go
+++ b/dataplane/linux/bpf_ep_mgr.go
@@ -382,7 +382,7 @@ func (m *bpfEndpointManager) CompleteDeferredWork() error {
 	m.ensureStarted()
 
 	m.applyProgramsToDirtyDataInterfaces()
-	m.applyProgramsToDirtyWorkloadEndpoints()
+	m.updateWEPsInDataplane()
 
 	if m.happyWEPsDirty {
 		chains := m.ruleRenderer.WorkloadInterfaceAllowChains(m.happyWEPs)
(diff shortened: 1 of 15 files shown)


dataplane/linux/bpf_ep_mgr.go
```diff
@@ -485,22 +506,40 @@ func (m *bpfEndpointManager) applyProgramsToDirtyWorkloadEndpoints() {
 		err := errs[wlID]
 		if err == nil {
 			log.WithField("id", wlID).Info("Applied policy to workload")
+			if m.allWEPs[wlID] != nil {
+				if _, ok := m.happyWEPs[wlID]; !ok {
```
tomastigera, line 510: nit: A bit confused by the two types of testing existence. Thought for a while that `nil` might be a valid happy value ...

dataplane/linux/int_dataplane.go
```diff
@@ -887,48 +889,84 @@ func (d *InternalDataplane) doStaticDataplaneConfig() {
...
-			fwdRules = append(fwdRules, iptables.Rule{
-				Match:   iptables.Match().OutInterface(prefix + "+"),
-				Action:  iptables.AcceptAction{},
-				Comment: []string{"To workload, BPF will handle."},
-			})
+
+		if t.IPVersion == 6 {
+			for _, prefix := range d.config.RulesConfig.WorkloadIfacePrefixes {
+				// In BPF mode, we don't support IPv6 yet.  Drop it.
+				fwdRules = append(fwdRules, iptables.Rule{
```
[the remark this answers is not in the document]
fasaxc: This is a backstop in case the bpf program isn't installed.

dataplane/linux/bpf_ep_mgr.go
fasaxc: Agreed, not ideal.  I'm about to pout up another PR so will address under that one.

PR #2424, opened 2020-07-13
Move policy acceptance to end of the filter table forward chain

caseydavenport:
## Description

Extension to https://github.com/projectcalico/felix/pull/2417

Will fix: https://github.com/projectcalico/calico/issues/3548

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
Connections to services without endpoints are now properly rejected in iptables dataplane mode
```


romulus-ai:
May it is somehow possible to instead of simply adding those rule at the End of the Forward Chain, to insert those Rule right before the first REJECT or DROP rule found in the FORWARD Chain? This would increase compatibility with i.e. firewalld and other tools.

PR #2426, opened 2020-07-14
Speed up FV container tear down.

fasaxc:
## Description

When we're tearing down the rig, there's no need to do it gracefully.  

Gracefully shutting down takes a couple of seconds so use SIGKILL and do k8s infra teardown in parallel so avoid that delay.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/fv/containers/containers.go
+++ b/fv/containers/containers.go
@@ -165,9 +165,11 @@ func (c *Container) Signal(sig os.Signal) {
 }
 
 type RunOpts struct {
-	AutoRemove    bool
-	WithStdinPipe bool
-	SameNamespace *Container
+	AutoRemove      bool
+	WithStdinPipe   bool
+	SameNamespace   *Container
+	StopTimeoutSecs int
+	StopSignal      string
 }
 
 func NextContainerIndex() int {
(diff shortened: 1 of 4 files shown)


fv/containers/containers.go
```diff
@@ -191,7 +193,11 @@ func RunWithFixedName(name string, opts RunOpts, args ...string) (c *Container)
 
 	// Prep command to run the container.
 	log.WithField("container", c).Info("About to run container")
-	runArgs := []string{"run", "--name", c.Name}
+	runArgs := []string{"run", "--name", c.Name, "--stop-timeout", fmt.Sprint(opts.StopTimeoutSecs)}
+
+	if opts.StopSignal != "" {
+		runArgs = append(runArgs, "--stop-signal", opts.StopSignal)
+	}
```
neiljerram, line 200: Are there any cases where it isn't right to use SIGKILL and 0 timeout?

fasaxc: I think some of the other repos import this code and use it in other ways so I was scared of changing the default.

fasaxc:
Have rerun a few times, I think it's at least as fast and it removes all the "goodbye cruel world" log spam from etcd and the k8s API server.

PR #2427, opened 2020-07-16
Rename shims to avoid name clashes

fasaxc:
## Description

Was finding the code using the time shim very confusing because we end up with multiple imports called Time; and two completely different types named `time.Time`.

The netlink shim uses the same pattern so I renamed that too.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```


timeshim/time.go
```diff
@@ -12,22 +12,24 @@
...
-type Time interface {
+type Interface interface {
 	Now() time.Time
 	Since(t time.Time) time.Duration
 	Until(t time.Time) time.Duration
 	After(t time.Duration) <-chan time.Time
 }
 
-func NewRealTime() Time {
-	return &realTime{}
```
song-jiang, line 30: Just wondering is this an error to return a pointer as `Time` interface where the `realTime` structure's method only has value receivers. Am I correct that `time.NewRealTime().Now()` will return compile error and `(*time.NewRealTime()).Now()` will not.

fasaxc: No, it worked fine before; you can call a non-pointer receiver on a pointer but it has copy semantics so side effects to the struct would be lost.  (This struct is 0-size so no issue here either way.)

PR #2428, opened 2020-07-16
Rate limit jump map cleanup

fasaxc:
## Description

* Add a generic rate-limited runner with UT.
* Do the cleanup from an instance of the rate-limited runner.
* Add FV to verify that cleanup happens as expected.

## Todos
- [x] Release note

## Release Note

```release-note
In BPF mode, Felix now rate-limits stale BPF map cleanup in order to save CPU.
```

### DIFF
--- a/dataplane/linux/bpf_ep_mgr.go
+++ b/dataplane/linux/bpf_ep_mgr.go
@@ -42,6 +42,8 @@ import (
 	"github.com/projectcalico/libcalico-go/lib/set"
 )
 
+const jumpMapCleanupInterval = 10 * time.Second
+
 type epIface struct {
 	ifacemonitor.State
 	jumpMapFDs map[PolDirection]bpf.MapFD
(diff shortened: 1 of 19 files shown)


dataplane/linux/bpf_ep_mgr.go
```diff
@@ -356,6 +360,9 @@ func (m *bpfEndpointManager) markProfileUsersDirty(id proto.ProfileID) {
 }
 
 func (m *bpfEndpointManager) CompleteDeferredWork() error {
+	// Do one-off initialisation.
+	m.startupOnce.Do(m.Start)
```
tomastigera, line 364: Why not to do this in `func (m *bpfEndpointManager) Start()` ?

dataplane/linux/bpf_ep_mgr.go
```diff
@@ -111,6 +111,10 @@ func newBPFEndpointManager(
 		dsrEnabled:          dsrEnabled,
 		ipSetMap:            ipSetMap,
 		stateMap:            stateMap,
+		mapCleanupRunner: ratelimited.NewRunner(10*time.Second, func(ctx context.Context) {
```
tomastigera, line 114: I wonder if the timeout should be configurable. Or at least defined somewhere at the top as a constant?

fasaxc: Moved to a constant, don't think it's worth the extra work for a config flag.

PR #2433, opened 2020-07-22
Add config param to control the interface monitor's resync interval.

fasaxc:
## Description

Noticed the interface monitor's resync was using 500ms+ of CPU in a sale run trace and it had a fixed interval of 10s.

Add a config param to control the interval and increase the default to 90s.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
Previously, Felix had a fixed 10s timer on which it resynced its list of local interfaces with the dataplane.  To reduce CPU usage, the timer has been increased to 90s by default and a config parameter (InterfaceRefreshInterval) added to control it.
```


config/config_params.go
```diff
@@ -164,6 +164,7 @@ type Config struct {
 
 	IptablesBackend                    string            `config:"oneof(legacy,nft,auto);legacy"`
 	RouteRefreshInterval               time.Duration     `config:"seconds;90"`
+	InterfaceRefreshInterval           time.Duration     `config:"seconds;90"`
```
song-jiang, line 167: Shall we set default to 10 seconds as before?

fasaxc: I think 10 seconds was too aggressive, it was using more CPU than expected so I decided to increase it.

PR #2436, opened 2020-07-29
Handling LoadBalancerIPStrings as a derived service

sridhartigera:
In case of LoadBalancers, the public IP is present in the LoadBalancerIPStrings in the service info. This is handled as a derived service and proper NAT entries are programmed for the LB public IP
1) This fixes the GCP loadbalancer issue
2) In case of aws, we get Loadbalancer domain names and not public IP. Hence we dont see a NAT entry for the resolved domain name in FE BPF map. The loadbalancer in turn does a DNAT of the <public ip:80> to <node_ip:nodeport>. Hence we will only nodeport NAT entries for aws. 

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/bpf/proxy/lb_src_range_test.go
+++ b/bpf/proxy/lb_src_range_test.go
@@ -35,7 +35,7 @@ func init() {
 	logrus.SetLevel(logrus.DebugLevel)
 }
 
-func testfn(option func(ips []string) proxy.K8sServicePortOption, arg []string) {
+func testfn(makeIPs func(ips []string) proxy.K8sServicePortOption) {
 	svcs := newMockNATMap()
 	eps := newMockNATBackendMap()
 	aff := newMockAffinityMap()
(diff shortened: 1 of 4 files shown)


bpf/proxy/lb_src_range_test.go
```diff
@@ -57,7 +57,7 @@ var _ = Describe("BPF Load Balancer source range", func() {
 				net.IPv4(10, 0, 0, 2),
 				2222,
 				v1.ProtocolTCP,
-				proxy.K8sSvcWithExternalIPs([]string{"35.0.0.2"}),
+				proxy.K8sSvcWithLoadBalancerIPs([]string{"35.0.0.2"}),
```
tomastigera, line 60: I wonder if we should execute this test 2x, once for ExternalIPs and once for LBIPs

sridhartigera: yes. Will add one more test for external IP

tomastigera: We may just turn this set of tests into a function and run it 2x with 2 differnt options :thinking:

bpf/proxy/lb_src_range_test.go
```diff
@@ -36,193 +36,212 @@ func init() {
...
+				err := s.Apply(state)
+				Expect(err).NotTo(HaveOccurred())
+
+				Expect(svcs.m).To(HaveLen(0))
+				Expect(eps.m).To(HaveLen(0))
+			}))
+
+		})
+	}
+	testfn([]string{}, []string{"35.0.0.2"})
```
tomastigera, line 245: Since the two test cases are pretty much mutually exclusive, you could just pass the one option.

PR #2445, opened 2020-08-07
Fix race between interface deletion and cleanup.

fasaxc:
## Description

- Make sure we don't retry indefinitely if an interface is removed while we're trying to clean it up.
- Improve logging in such cases by wrapping `tc` and generating a clean error that we can match on.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```


dataplane/linux/bpf_ep_mgr.go
```diff
@@ -661,6 +671,18 @@ func (m *bpfEndpointManager) applyPolicy(ifaceName string) error {
 
 	wep := m.allWEPs[*endpointID]
 
+	// Attach the qdisc first; it is shared between the directions.
+	err := tc.EnsureQdisc(ifaceName)
```
tomastigera, line 675: I wonder if we just want to call `HasQdisc(ifaceName)`. What if we manage to install qdisc? Wouldn't is create another kind of race? It means that the other programs are  not there :thinking:

fasaxc: Not sure I follow!  This code is before we try to add the programs below.  All I'm trying to do here is move the EnsureQdisc call from the per-direction goroutines to this one so that it happens exactly once rather than the two directions racing to add the qdisc and triggering a retry of adding the programs.

PR #2447, opened 2020-08-10
Reduce log spam from missing interfaces.

fasaxc:
## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
Reduce log spam when an interface is removed from the dataplane.
```

### DIFF
--- a/dataplane/linux/endpoint_mgr.go
+++ b/dataplane/linux/endpoint_mgr.go
@@ -139,6 +139,7 @@ type endpointManager struct {
 	ruleRenderer rules.RuleRenderer
 	routeTable   routeTable
 	writeProcSys procSysWriter
+	osStat       func(path string) (os.FileInfo, error)
 	epMarkMapper rules.EndpointMarkMapper
 
 	// Pending updates, cleared in CompleteDeferredWork as the data is copied to the activeXYZ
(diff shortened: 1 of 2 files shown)


dataplane/linux/endpoint_mgr.go
```diff
@@ -682,7 +686,12 @@ func (m *endpointManager) resolveWorkloadEndpoints() {
 		ifaceName := item.(string)
 		err := m.configureInterface(ifaceName)
 		if err != nil {
-			log.WithError(err).Warn("Failed to configure interface, will retry")
+			if exists, err := m.interfaceExistsInProcSys(ifaceName); err == nil && !exists {
+				// Suppress log spam if interface has been removed.
+				log.WithError(err).Debug("Failed to configure interface and it seems to be gone")
```
song-jiang, line 691: Should the Iter function return set.RemoveItem here? I don't see how log spam is suppressed.

fasaxc: If it doesn't exist then we log at debug instead of warning so that prevents the spam.

We return nil to flag to the rest of the code that the interface is not configured (resulting in its status being changed to down).  Since we don't remove the flag, we'll then retry in case the interface comes back (or, we'll get the WEP removed message).

PR #2448, opened 2020-08-10
Fix slow retries in route_table and finesse logging.

fasaxc:
## Description

The route table used to retry each interface before looping to
the next interface.

Swap the loop order so that we try all interfaces once before
retrying. This addresses failures caused by routes being moved
from one interface to another (can't add the route to interface
A until it is removed from interface B).

Finesse the logging to avoid warnings around the above, expected,
case.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
Fix route_table.go slow retries (and reduce log spam) when a route is moved from one interface to another.
```


fasaxc:
CI failure was a known flake.

PR #2449, opened 2020-08-11
Fix that profile refs weren't cleaned up when removing endpoint.

fasaxc:
# Description

Caused a panic if the profile was subsequently updated.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/dataplane/linux/bpf_ep_mgr.go
+++ b/dataplane/linux/bpf_ep_mgr.go
@@ -247,77 +247,11 @@ func (m *bpfEndpointManager) onWorkloadEndpointUpdate(msg *proto.WorkloadEndpoin
 	log.WithField("wep", msg.Endpoint).Debug("Workload endpoint update")
 	wlID := *msg.Id
 	oldWEP := m.allWEPs[wlID]
+	m.removeWEPFromIndexes(wlID, oldWEP)
+
 	wl := msg.Endpoint
-	if oldWEP != nil {
-		for _, t := range oldWEP.Tiers {
-			for _, pol := range t.IngressPolicies {
-				polSet := m.policiesToWorkloads[proto.PolicyID{
-					Tier: t.Name,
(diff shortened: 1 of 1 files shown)


dataplane/linux/bpf_ep_mgr.go
```diff
@@ -329,29 +263,9 @@ func (m *bpfEndpointManager) onWorkloadEnpdointRemove(msg *proto.WorkloadEndpoin
...
-			polSet := m.policiesToWorkloads[proto.PolicyID{
-				Tier: t.Name,
-				Name: pol,
-			}]
-			if polSet == nil {
-				continue
-			}
-			polSet.Discard(wlID)
-		}
-	}
```
fasaxc, line 353: Bug was that this code didn't match the code above: it was missing the cleanup of profiles.  To fix, I made the code common and called the same code from both places.

dataplane/linux/bpf_ep_mgr.go
```diff
@@ -967,3 +881,87 @@ func (m *bpfEndpointManager) isWorkloadIface(iface string) bool {
...
+		m.addPolicyToWEPMappings(t.IngressPolicies, wlID)
+		m.addPolicyToWEPMappings(t.EgressPolicies, wlID)
+	}
+	m.addProfileToWEPMappings(wl.ProfileIds, wlID)
+}
+
+func (m *bpfEndpointManager) addPolicyToWEPMappings(polNames []string, wlID proto.WorkloadEndpointID) {
+	for _, pol := range polNames {
+		polID := proto.PolicyID{
+			Tier: "default",
```
neiljerram, line 896: I believe we've lost the real Tier name here.  Is that OK?  (Feels like a gotcha for when this is eventually merged into a multi-tier system.)

PR #2453, opened 2020-08-12
bpf/conntrack: fix assumed ordering in NewStaleNATScanner

tomastigera:
Legs A-B do not need to have the same ordering in fwd key and the rev key.

Port is also significant as there may be services with multiple ports.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/bpf/conntrack/cleanup.go
+++ b/bpf/conntrack/cleanup.go
@@ -54,7 +54,7 @@ type ScanVerdict int
 
 const (
 	// ScanVerdictOK means entry is fine and should remain
-	ScanVerdictOK = iota
+	ScanVerdictOK ScanVerdict = iota
 	// ScanVerdictDelete means entry should be deleted
 	ScanVerdictDelete
 )
@@ -272,7 +272,7 @@ type NATChecker func(frontIP net.IP, frontPort uint16, backIP net.IP, backPort u
 func NewStaleNATScanner(frontendHasBackend NATChecker) EntryScanner {
 	debug := log.GetLevel() >= log.DebugLevel
 
-	return func(k Key, v Value, get EntryGet) ScanVerdict {
+	return func(k Key, v Value, _ EntryGet) ScanVerdict {
 		switch v.Type() {
 		case TypeNormal:
 			// skip non-NAT entry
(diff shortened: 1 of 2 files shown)


bpf/conntrack/cleanup.go
```diff
@@ -318,18 +322,29 @@ func NewStaleNATScanner(frontendHasBackend NATChecker) EntryScanner {
...
 				epIP = revB
-				epPort = revKey.PortB()
+				epPort = revBport
 				svcIP = kB
-				svcPort = k.PortB()
-			} else if kA.Equal(revB) {
+				svcPort = kBport
+			} else if kB.Equal(revA) && kBport == revAport {
+				epIP = revB
+				epPort = revAport
```
fasaxc, line 332: `revBport`?

bpf/conntrack/cleanup.go
```diff
@@ -318,18 +322,29 @@ func NewStaleNATScanner(frontendHasBackend NATChecker) EntryScanner {
...
+			} else if kA.Equal(revB) && kAport == revBport {
 				epIP = revA
-				epPort = revKey.PortA()
+				epPort = revAport
 				svcIP = kA
-				svcPort = k.PortA()
+				svcPort = kAport
+			} else if kB.Equal(revB) && kBport == revBport {
+				epIP = revA
+				epPort = revBport
```
fasaxc, line 342: revAport to go with the `epIP`?

bpf/conntrack/cleanup.go
```diff
@@ -318,18 +322,29 @@ func NewStaleNATScanner(frontendHasBackend NATChecker) EntryScanner {
...
+			} else if kB.Equal(revA) && kBport == revAport {
+				epIP = revB
+				epPort = revAport
+				svcIP = kA
+				svcPort = kAport
+			} else if kA.Equal(revB) && kAport == revBport {
 				epIP = revA
-				epPort = revKey.PortA()
+				epPort = revAport
 				svcIP = kA
```
fasaxc, line 338: think this should be `kB`  if `kA == revB` then `kA` is the source pod IP si it's not the service IP

bpf/conntrack/cleanup.go
```diff
@@ -318,18 +322,29 @@ func NewStaleNATScanner(frontendHasBackend NATChecker) EntryScanner {
...
-				svcPort = k.PortA()
+				svcPort = kAport
+			} else if kB.Equal(revB) && kBport == revBport {
+				epIP = revA
+				epPort = revBport
+				svcIP = kB
```
fasaxc, line 343: Think this should be `kA` since `kB.Equal(revB)` means that `kB` is the source pod IP.

tomastigera:
The test now checks whether what de selected as front/back ends is what we expect

PR #2454, opened 2020-08-13
bpf/proxy: Conntrack cleaner runs immediately after sync up

tomastigera:
Immediatally after the state of proxy is synced up and the fresh state
is applied, start the conntrack cleaner. It makes its first sweep
immediately. It needs to be after the first sync up to avoid breaking
connections that may look like towards stale backends as the internal
state was not updated yet.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/bpf/proxy/kube-proxy.go
+++ b/bpf/proxy/kube-proxy.go
@@ -76,6 +76,13 @@ func StartKubeProxy(k8s kubernetes.Interface, hostname string,
 	}
 
 	go func() {
+		// Before we start, scan for all finished / timed out connections to
+		// free up the conntrack table asap as it may take time to sync up the
+		// proxy and kick off the first full cleaner scan.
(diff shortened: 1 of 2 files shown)


bpf/proxy/kube-proxy.go
```diff
@@ -76,6 +76,13 @@ func StartKubeProxy(k8s kubernetes.Interface, hostname string,
 	}
 
 	go func() {
+		// Before we start, scan for all finished / timed out connections to
+		// free up the conntrack table asap as it may take time to sync up the
+		// proxy and kick off the first full cleaner scan.
+		lc := conntrack.NewLivenessScanner(kp.conntrackTimeouts, kp.dsrEnabled)
```
fasaxc, line 82: Since it's a one-off, worth putting an info log before and after.

PR #2461, opened 2020-08-20
BPF: If packet is too short, try to pull more data

fasaxc:
## Description

We use direct packet access to read and write the packet; for that to be effective we need to make sure that the SKB has enough data in its head segment.

If we detect that the packet is too short, issue a bpf_skb_data_pull.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
In BPF mode, fix that packets could be dropped if the UDP/TCP header didn't fit in the SKB's head buffer.
```


neiljerram (approved):
LGTM

PR #2480, opened 2020-08-26
Only enable drop rules if appropriate encap is enabled

doublek:
## Todos
- [x] Unit tests (full coverage)
- [x] Integration tests Not needed
- [ ] Backport
- [x] Release note

## Release Note

```release-note
Only drop encapsulation traffic from workloads if corresponding encapsulation is enabled
```

PR #2481, opened 2020-08-28
Add specific deny rules for IPIP and VXLAN traffic originating from workloads.

doublek:
## Description

Rather than blanket deny all VXLAN and IPIP traffic from workloads, tweak the existing rule so that we only drop if the packet's destination IP address is a node with a tunnel endpoint. This will provide some additional flexibility where if a workload does want to run VXLAN/IPIP traffic, we still let the workload do this, but prevent spoofed packets to our nodes.

Related PR with an alternate fix: #2480 

## Todos
- [x] Integration tests (delete as appropriate) Not needed
- [ ] Backport
- [x] Release note

## Release Note

```release-note
Only drop specific encapsulation traffic originating from workloads if corresponding encapsulation is enabled and if packet is destined to a host.
```

### DIFF
--- a/dataplane/linux/endpoint_mgr_test.go
+++ b/dataplane/linux/endpoint_mgr_test.go
@@ -183,8 +183,8 @@ func chainsForIfaces(ipVersion uint8, ifaceMetadata []string,
 		dropEncapRules = []iptables.Rule{
 			{
 				Match: iptables.Match().ProtocolNum(ProtoUDP).
-					DestIPSet(ipSetVXLANSourceHosts).
-					DestPorts(uint16(VXLANPort)),
+					DestPorts(uint16(VXLANPort)).
+					DestIPSet(ipSetVXLANSourceHosts),
 				Action:  iptables.DropAction{},
 				Comment: []string{"Drop VXLAN encapped packets originating in pods destined to the cluster nodes"},
 			},
--- a/dataplane/linux/endpoint_mgr_test.go
+++ b/dataplane/linux/endpoint_mgr_test.go
@@ -654,7 +654,6 @@ func endpointManagerTests(ipVersion uint8) func() {
 				IptablesMarkNonCaliEndpoint: 0x0100,
 				KubeIPVSSupportEnabled:      true,
 				WorkloadIfacePrefixes:       []string{"cali", "tap"},
-				VXLANEnabled:                true,
 				VXLANPort:                   4789,
 				VXLANVNI:                    4096,
 			}
--- a/dataplane/linux/endpoint_mgr_test.go
+++ b/dataplane/linux/endpoint_mgr_test.go
@@ -172,13 +172,13 @@ func chainsForIfaces(ifaceMetadata []string,
 				DestIPSet(ipSetVXLANSourceHosts).
 				DestPorts(uint16(VXLANPort)),
 			Action:  iptables.DropAction{},
-			Comment: []string{"Drop VXLAN encapped packets originating in pods"},
+			Comment: []string{"Drop VXLAN encapped packets originating in pods destined to the cluster nodes"},
 		},
 		{
 			Match: iptables.Match().ProtocolNum(ProtoIPIP).
 				DestIPSet(ipSetAllHosts),
 			Action:  iptables.DropAction{},
-			Comment: []string{"Drop IPinIP encapped packets originating in pods"},
+			Comment: []string{"Drop IPinIP encapped packets originating in pods destined to the cluster nodes"},
 		},
 	}
 
--- a/dataplane/linux/endpoint_mgr_test.go
+++ b/dataplane/linux/endpoint_mgr_test.go
@@ -117,25 +117,25 @@ var fromHostDispatchEmpty = []*iptables.Chain{
 	},
 }
 
-func hostChainsForIfaces(ifaceMetadata []string, epMarkMapper rules.EndpointMarkMapper) []*iptables.Chain {
-	return append(chainsForIfaces(ifaceMetadata, epMarkMapper, true, "normal"),
-		chainsForIfaces(ifaceMetadata, epMarkMapper, true, "applyOnForward")...,
+func hostChainsForIfaces(ipVersion uint8, ifaceMetadata []string, epMarkMapper rules.EndpointMarkMapper) []*iptables.Chain {
+	return append(chainsForIfaces(ipVersion, ifaceMetadata, epMarkMapper, true, "normal"),
+		chainsForIfaces(ipVersion, ifaceMetadata, epMarkMapper, true, "applyOnForward")...,
 	)
 }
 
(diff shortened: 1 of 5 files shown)


dataplane/linux/endpoint_mgr_test.go
```diff
@@ -164,15 +164,19 @@ func chainsForIfaces(ifaceMetadata []string,
...
 	epmarkFromPrefix := outPrefix[:6]
+	ipSetVXLANSourceHosts := "cali40all-vxlan-net"
+	ipSetAllHosts := "cali40all-hosts-net"
 	dropEncapRules := []iptables.Rule{
 		{
 			Match: iptables.Match().ProtocolNum(ProtoUDP).
+				DestIPSet(ipSetVXLANSourceHosts).
 				DestPorts(uint16(VXLANPort)),
 			Action:  iptables.DropAction{},
 			Comment: []string{"Drop VXLAN encapped packets originating in pods"},
```
bcreane, line 175: in hindsight, this comment was inaccurate, but should be correct now.

dataplane/linux/endpoint_mgr_test.go
```diff
@@ -164,15 +164,19 @@ func chainsForIfaces(ifaceMetadata []string,
 	epMarkFromName := "cali-from-endpoint-mark"
 	epMarkSetOnePrefix := "cali-sm-"
 	epmarkFromPrefix := outPrefix[:6]
+	ipSetVXLANSourceHosts := "cali40all-vxlan-net"
+	ipSetAllHosts := "cali40all-hosts-net"
 	dropEncapRules := []iptables.Rule{
 		{
 			Match: iptables.Match().ProtocolNum(ProtoUDP).
+				DestIPSet(ipSetVXLANSourceHosts).
```
bcreane, line 172: I'm not completely following `DestIPSet` combined with `SourceHosts`. There's not a `SrcIPSet`?

robbrockbank: I'd assumed the ipSetVXLANSourceHosts was the set of hosts that support VXLAN - since I'm not sure it makes sense to think of source and destination separately for that.

doublek: Yeah, I don't think it makes sense to think of it as source/dest. Maybe I can update the variable name here. 

The ipset that felix maintains has the variable name `IPSetIDAllVXLANSourceNets` and is actually a list of IP addresses of hosts (VXLAN parent device's IP address). I think that it was named as source nets because until now its only use was as a source match ipset.

dataplane/linux/endpoint_mgr_test.go
robbrockbank: the comment was accurate.  these rules are added to the cali interface chains so they explicitly cover packets originating in pods.  The changes is that we are now allowing vxlan packets unless they are going to the cluster VXLAN host IPs for this cluster. 

So the comment should now be something like:  "Drop VXLAN encapped packets originating in pods destined to the cluster nodes".

doublek: Updated with tweaks for ip versions as well.

rules/endpoints.go
```diff
@@ -308,17 +322,38 @@ func (r *DefaultRuleRenderer) endpointIptablesChain(
...
 	})
 
-	if dropEncap {
+	if dropEncap && r.Config.VXLANEnabled {
+		// If VXLANEnabled, block the endpoint from sending VXLAN traffic with destination IP
+		// address matching any of the VXLAN hosts.
+		match := Match().ProtocolNum(ProtoUDP).DestPorts(uint16(r.Config.VXLANPort))
+		comment := "Drop VXLAN encapped packets originating in pods"
+		if ipVersion == 4 {
+			// The VXLAN blocking rules use a IPv4 specific ipset, so in the
```
doublek, line 331: s/use/uses.

rules/endpoints.go
```diff
@@ -308,17 +322,38 @@ func (r *DefaultRuleRenderer) endpointIptablesChain(
...
+			Comment: []string{comment},
 		})
+	}
+	if dropEncap && r.Config.IPIPEnabled {
+		// If IPIP is enabled, block the endpoint from sending IPIP traffic with destination IP
+		// address matching any of our hosts.
+		match := Match().ProtocolNum(ProtoIPIP)
+		comment := "Drop IPinIP encapped packets originating in pods"
+		if ipVersion == 4 {
+			// The IPIP blocking rules use a IPv4 specific ipset, so in the
```
doublek, line 348: s/use/uses

rules/endpoints.go
```diff
@@ -308,17 +322,41 @@ func (r *DefaultRuleRenderer) endpointIptablesChain(
 		},
 	})
 
-	if dropEncap {
+	if dropEncap && r.Config.VXLANEnabled {
```
doublek, line 325: I could possibly remove the `r.Config.VXLANEnabled ` if we just want behavior similar to before where we added the drop rules unconditionally leaving the tweaks to the actual the destination endpoint.

robbrockbank: Was wondering about this myself.  Is it possible to have a single cluster with mixed VXLAN/no VXLAN or IPIP/no IPIP usage?   If so then an individual node may have encapEnabled set to false but you still need to police the encap IPSet....

Feels like it might be safer.  In the normal case, if encapsulated is disabled then the IPSet will be empty anyways.

caseydavenport: Is there a reason we only want this with VXLAN enabled? I thought the rule tweaks were so that we could always have this rule in place and still allow the necessary traffic?

caseydavenport: I guess it's to make this work for IPv6 as well, where we don't have the IP set to match on?

caseydavenport: yeah, my vote would be to always program the rules, but just make them more specific so they only match on traffic from wep to a known Calico node.

For IPv4 vs IPv6, do we need separate rule sets actually? We don't support VXLAN or IPIP for v6 anyway.

doublek: Yup, will update to always program the rule.

doublek: I was looking into why the FVs started to fail after removing the check to only program the rules if the VXLAN config is enabled and it turns out that Felix only creates and maintains the corresponding host ipset if IPIP or VXLAN is enabled. 

I think a few options are:

Option 1: 
Keep the logic for only enabling the encap spoofing deny rules, which will prevent us from dropping legitimate encap traffic on hosts where encap is disabled. If encap is enabled and traffic is ipv4 then the ipset based rule will refine the traffic we drop based on destination IP check.

Option 2:
Create an empty ipset even if the encap method is disabled. While this will fix the FV's, this may not prevent the bug of incorrectly blocking legitimate encap traffic. For the fix to actually work, we may have to start maintaining a list of calico hosts and add this to the rule instead? The current fix uses specific encap related ipsets, and these may not be the same as a Calico host?

doublek:
Closing in favor of #2484

PR #2484, opened 2020-09-01
Add knobs to disable adding encap drop rules from workloads

doublek:
## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```


rules/endpoints.go
```diff
@@ -308,13 +308,15 @@ func (r *DefaultRuleRenderer) endpointIptablesChain(
 		},
 	})
 
-	if dropEncap {
+	if dropEncap && r.Config.DropVXLANPacketsFromWorkloads {
```
fasaxc, line 311: Maybe the parameter should be consulted higher up the stack so that we only set dropEncap if we're rendering a workload egress chain and the flag is set?

fasaxc (changes requested):
One comment about the placement of the condition, otherwise looks good to me.

rules/endpoints.go
doublek: The reason why I did it this way was to add individual knobs to each type of encap. I could perhaps modify the type or add an additional parameter to `endpointIptablesChain` to distinguish IPIP versus VXLAN. WDYT?

config/config_params.go
```diff
@@ -213,6 +213,11 @@ type Config struct {
 	IpInIpMtu        int    `config:"int;1440;non-zero"`
 	IpInIpTunnelAddr net.IP `config:"ipv4;"`
 
+	// Knobs provided to explicitly control whether we add rules to drop encap traffic
+	// from workloads. We always add them unless explicitly disabled.
+	DropVXLANPacketsFromWorkloads bool `config:"bool;true"`
```
caseydavenport, line 218: Are most of our other fields default-zero-value?

Would `AllowVXLANPacketsFromWorkloads` be a better formulation of this? I think it would make it easier for golang users of the API (default value matches the type zero-value)

doublek: Updated.

rules/endpoints.go
doublek: I've moved this up and tried to clean up a bit based on Casey's feedback. Let me know if this works.

fasaxc (approved):
LGTM.  Agree that AllowXXX is better :+1:

PR #2485, opened 2020-09-02
bpf: use state from map, free up stack

tomastigera:
We are running out of space on stack, especially when DEBUG is enabled.
Therefore this change frees up 60 bytes of stack space (limit is 512)
move the jump state away from the stack and getting it from the map
where we are going to update it anyway if we needed to go through
policy.

This adds a lookup in CPU local array (fast) for each packet, removes
updates for packets that go through policy.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/bpf-gpl/list-ut-objs
+++ b/bpf-gpl/list-ut-objs
@@ -60,3 +60,5 @@ for ep_type in $ep_types; do
     done
   done
 done
+
+echo "bin/test_from_hep_fib_no_log_skb0x0.o"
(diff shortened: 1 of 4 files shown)


fasaxc (approved):
LGTM but has a conflict.  I think you could make the benchmark into a "proper" go benchmark by using `B.ReportMetric()` https://golang.org/pkg/testing/#B.ReportMetric

PR #2489, opened 2020-09-03
In BPF mode, check that the BPF mark bits are assigned.

fasaxc:
## Description

Since we can't change the mark bits used by the BPF programs, at least make sure they are assigned to felix in the config:

- Pre-claim the BPF mark bits so they can't be used for other bits.
- In BPF mode, do not claim bits that are not used.
- Ignore missing bits in BPF mode.

## Todos
- [ ] Documentation
- [x] Release note

## Release Note

```release-note
In BPF mode, Felix now checks that the mark bits used by the BPF programs are assigned to Felix.  Previously, it would use them without checking, possibly creating conflicts.
```

### DIFF
--- a/dataplane/driver.go
+++ b/dataplane/driver.go
@@ -127,7 +127,7 @@ func StartDataplaneDriver(configParams *config.Config,
 
 		// Mark bits for endpoint mark. Currently Felix takes the rest bits from mask available for use.
 		markEndpointMark, allocated := markBitsManager.NextBlockBitsMark(markBitsManager.AvailableMarkBitCount())
-		if kubeIPVSSupportEnabled && allocated == 0 {
+		if kubeIPVSSupportEnabled {
 			if allocated == 0 {
 				log.WithFields(log.Fields{
 					"Name":     "felix-iptables",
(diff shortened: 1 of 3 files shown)


dataplane/driver.go
```diff
@@ -61,53 +57,86 @@ func StartDataplaneDriver(configParams *config.Config,
...
 
-		// Mark bits for end point mark. Currently felix takes the rest bits from mask available for use.
+		// Mark bits for endpoint mark. Currently Felix takes the rest bits from mask available for use.
 		markEndpointMark, allocated := markBitsManager.NextBlockBitsMark(markBitsManager.AvailableMarkBitCount())
 		if kubeIPVSSupportEnabled && allocated == 0 {
-			log.WithFields(log.Fields{
-				"Name":     "felix-iptables",
-				"MarkMask": configParams.IptablesMarkMask,
-			}).Panic("Not enough mark bits available for endpoint mark.")
+			if allocated == 0 {
```
neiljerram, line 131: We already tested `allocated == 0` in the previous line.

neiljerram: Looks like you meant to remove `&& allocated == 0` from the preceding line.

neiljerram (changes requested):
One catch (I believe).

neiljerram (approved):

PR #2491, opened 2020-09-04
bpf: harden accept_local using iptables rpfilter

tomastigera:
We need accept_local for traffic that local IP when we turn the packet
around from ingress to egress side. The primary reason is ICMP reponses,
however, it is also used the vxlan NodePort tunnel.

The NodePort tunnel could avoid that, however, that would require
relaxing the RPF check when an overlay is used for pod-pod networking as
we decap the packets on external phys interfaces, but the best route is
through the tunnel interface.

It is easier to harden accept_local then to relax rpf for selected
packet in iptables.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/dataplane/linux/int_dataplane.go
+++ b/dataplane/linux/int_dataplane.go
@@ -981,13 +981,22 @@ func (d *InternalDataplane) setUpIptablesBPF() {
 	}
 
 	for _, t := range d.iptablesRawTables {
+		// Do not RPF check what is marked as to be skipped by RPF check.
 		rpfRules := []iptables.Rule{{
 			Match:  iptables.Match().MarkMatchesWithMask(tc.MarkSeenBypassSkipRPF, tc.MarkSeenBypassSkipRPFMask),
 			Action: iptables.ReturnAction{},
 		}}
 
+		// For anything we approved for forward, permit accept_local as it is
+		// traffic encapped for NodePort, ICMP replies etc. - stuff we trust.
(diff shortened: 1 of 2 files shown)


dataplane/linux/int_dataplane.go
```diff
@@ -981,13 +981,22 @@ func (d *InternalDataplane) setUpIptablesBPF() {
...
+		// Do not RPF check what is marked as to be skipped by RPF check.
 		rpfRules := []iptables.Rule{{
 			Match:  iptables.Match().MarkMatchesWithMask(tc.MarkSeenBypassSkipRPF, tc.MarkSeenBypassSkipRPFMask),
 			Action: iptables.ReturnAction{},
 		}}
 
+		// For anything we approved for forward, permit accept_local as it is
+		// traffic encapped for NodePort, ICMP replies etc. - stuff we trust.
+		rpfRules = append(rpfRules, iptables.Rule{
+			Match:  iptables.Match().MarkMatchesWithMask(tc.MarkSeenBypassForward, tc.MarksMask).RPFCheckPassed(true),
```
fasaxc, line 993: What about MarkSeenBypassForwardSourceFixup, should that match too?

tomastigera: I believe not. That one is specifically to fix that we do not have the local IP set as source yet.

fasaxc (approved):

PR #2495, opened 2020-09-11
Fix that broadcast routes weren't filtered out.

fasaxc:
## Description

The condition in the route filter was wrong, it used a bitwise and, instead of a simple `==` to look for the route type.  Since `RTN_BROADCAST == 3` and `RTN_LOCAL == 2`, the condition passed for both.

## Todos
- [ ] Backport
- [x] Release note

## Release Note

```release-note
Fix that broadcast routes weren't filtered out of felix's list of local IPs.  In BPF mode, this caused dataplane route flaps.
```


neiljerram (approved):
LGTM

song-jiang (commented):
LGTM

PR #2499, opened 2020-09-18
Address/Investigate a WG fv flake

realgaurav:
## Description

Address/investigate a WG fv flake, by:
* increasing connectivity-test timeout.
* adding a callback to check the connection if the connectivity-test fails. This will help isolate if there are any issue in connectivity-test or just pod connectivity in general.

This is a speculative fix and in case of issue will provide debugging information using failed callback tests.

This being a flake, ran it several times through CI to see if the error can be reproduced. I wasn't able to repro the flake in 10 CI runs.

## Release Note
```release-note
None required
```

### DIFF
--- a/fv/wireguard_test.go
+++ b/fv/wireguard_test.go
@@ -24,6 +24,8 @@ import (
 	"strings"
 	"time"
 
+	log "github.com/sirupsen/logrus"
+
 	"github.com/projectcalico/libcalico-go/lib/numorstring"
 
 	. "github.com/onsi/ginkgo"
@@ -452,24 +454,6 @@ var _ = infrastructure.DatastoreDescribe("_BPF-SAFE_ WireGuard-Supported", []api
 			return fmt.Errorf("policy not applied")
 		}
 
-		AfterEach(func() {
-			fmt.Printf("Check if the workload connectivity is indeed bad")
-
-			fmt.Printf("Test wls0 to wls1 connectivity")
-			err, errstr := wls[0].SendPacketsTo(wls[1].IP, 5, 100)
-			if err != nil {
-				fmt.Printf("Connection from wls0 to wls1 failed: %s\n", errstr)
-			}
-			Expect(err).NotTo(HaveOccurred())
-
-			fmt.Printf("Test wls1 to wls0 connectivity")
-			err, errstr = wls[1].SendPacketsTo(wls[0].IP, 5, 100)
-			if err != nil {
-				fmt.Printf("Connection from wls1 to wls0 failed: %s\n", errstr)
-			}
-			Expect(err).NotTo(HaveOccurred())
-		})
-
 		It("between pod to pod should be encrypted using wg tunnel with egress policies applied", func() {
 			policy := api.NewGlobalNetworkPolicy()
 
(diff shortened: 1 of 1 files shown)


fv/wireguard_test.go
```diff
@@ -484,6 +484,7 @@ var _ = infrastructure.DatastoreDescribe("_BPF-SAFE_ WireGuard-Supported", []api
 
 			cc.ResetExpectations()
 
+			policy = api.NewGlobalNetworkPolicy()
```
neiljerram, line 487: I agree that it's good practice to use a new GNP object here.  But do you know exactly what the problem was as the code was before, when it was reusing the same object?

fv/wireguard_test.go
```diff
@@ -496,6 +497,17 @@ var _ = infrastructure.DatastoreDescribe("_BPF-SAFE_ WireGuard-Supported", []api
...
+			err, errstr := wls[0].SendPacketsTo(wls[1].IP, 5, 100)
+			if err != nil {
+				fmt.Printf("Connection from wls0 to wls1 failed: %s\n", errstr)
+			}
+			Expect(err).NotTo(HaveOccurred())
+			err, errstr = wls[1].SendPacketsTo(wls[0].IP, 5, 100)
+			if err != nil {
+				fmt.Printf("Connection from wls1 to wls0 failed: %s\n", errstr)
+			}
+			Expect(err).NotTo(HaveOccurred())
```
neiljerram, line 509: These additions look like might have been just debugging.  Do we still need them?  If so, can you add a comment to explain why?

neiljerram (changes requested):
Couple of queries here.

fv/wireguard_test.go
realgaurav: Thanks for taking a look @neiljerram. I agree this _should_ not be needed.

With the current proposed changes, I am trying to repro the flake we see with respect to workload connectivity when `Apply` policy is applied. In my experiment, without these changes, I was able to repro the flake in 1 out of 4 runs. With these changes however, I haven't been able to repro the flake with 10 successful runs so far. I am not sure atm what's "fixing" the flake. I have tried `5 seconds sleep` before `CheckConnectivity()` but that didn't address the flake ruling out the timing issue.

Can you think of any check to do before `CheckConnectivity()` other than reading the policy for verifying that the policy is applied, which is already being done?

fv/wireguard_test.go
realgaurav: The intent was to compare if this block of test passes (verifying workload connectivity)  while `CheckConnectivity()` fails thereby inferring that there may be an issue with `CheckConnectivity()`. However, I haven't been able to repro the issue with this proposed change block present.

realgaurav: This is NA anymore, marking resolved.

fv/wireguard_test.go
neiljerram: So I think you're saying that this empirically reduces the occurrence of the flake, but you don't yet know why.  Is that right?

Have you tried dumping and comparing the detailed `*policy` that is applied, with and without the addition of line 489 ?

neiljerram: As regards sleeping before `CheckConnectivity()`,

- I think we should always avoid hardcoded sleeping, and instead use something like Eventually or an explicit retry loop

- it shouldn't be needed, because `CheckConnectivity()` already does the right kind of retrying, I think.

neiljerram (approved):
Thanks @realgaurav , would just like the narrative to be clear that this is a partly speculative fix + debugging.

PR #2511, opened 2020-09-30
Calculate MTU based on host interfaces

caseydavenport:
## Description

Detect MTU!

## Todos
- [x] Implement for Linux dataplane
- [x] Release note

In a subsequent PR:
- [ ] Reconcile existing veth MTUs
- [ ] Documentation

Other:
- [x] libcalico-go changes for interface regex configuration
- [x] Corresponding changes for CNI plugin

## Release Note

```release-note
Automatic MTU detection
```

### DIFF
--- a/dataplane/linux/int_dataplane.go
+++ b/dataplane/linux/int_dataplane.go
@@ -192,6 +192,9 @@ type Config struct {
 	KubeClientSet *kubernetes.Clientset
 
 	FeatureDetectOverrides map[string]string
+
+	// Populated with the smallest host MTU based on auto-detection.
+	hostMTU int
 }
 
 // InternalDataplane implements an in-process Felix dataplane driver based on iptables
@@ -294,9 +297,24 @@ func NewIntDataplaneDriver(config Config) *InternalDataplane {
 		config.RulesConfig.IptablesMarkEndpoint,
 		config.RulesConfig.IptablesMarkNonCaliEndpoint)
 
-	// Determine values to use for MTU and default any unspecified fields.
-	if err := configureMTU(&config); err != nil {
-		log.WithError(err).Fatal("Unable to configure MTU, shutting down")
+	// Auto-detect host MTU.
+	if mtu, err := findHostMTU(); err != nil {
+		log.WithError(err).Fatal("Unable to detect host MTU, shutting down")
+	} else {
+		// We found the host's MTU. Default any MTU configurations that have not been set.
+		config.hostMTU = mtu
+		if config.RulesConfig.IPIPEnabled && config.IPIPMTU == 0 {
+			config.IPIPMTU = mtu - 20
+		}
+		if config.RulesConfig.VXLANEnabled && config.VXLANMTU == 0 {
+			config.VXLANMTU = mtu - 50
+		}
+		if config.Wireguard.Enabled && config.Wireguard.MTU == 0 {
+			config.Wireguard.MTU = mtu - 60
+		}
+	}
+	if err := writeMTUFile(config); err != nil {
+		log.Fatal("Failed to write MTU file")
 	}
 
 	dp := &InternalDataplane{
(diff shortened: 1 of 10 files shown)


caseydavenport:
One of the main questions I have is where should this go in Felix - current PR has it in the data plane driver, since that seems the logical place to know about how to calculate the MTU (e.g., different for windows, etc).

It does mean a change in behavior the config provided over the dataplane interface. Previously, `0` was an invalid value and would never be sent. Now, `0` means "auto-detect a good value for this".

fasaxc:
Definitely the dataplane driver.  I try to avoid loops from the dataplane back into the calc graph since that'd make the whole thing prone to oscillation and hard to test.

Should the code go in libcalico-go so we can call it from the CNI plugin too if the mtu file isn't present?

Have you thought about adjusting veth MTUs on the fly?  I think that'd be a good thing to do since it'd allow for hte MTU to be fixed non-disruptively.  I'm assuming that the netlink lib somehow lets us look up the netns of the veth peer and then we can enter the netns and set the other MTU with a couple more netlink calls.

dataplane/linux/int_dataplane.go
```diff
@@ -109,6 +113,17 @@ func init() {
 	prometheus.MustRegister(summaryIfaceBatchSize)
 	prometheus.MustRegister(summaryAddrBatchSize)
 	processStartTime = time.Now()
+
+	// Initialize a regex for matching interfaces we want to ignore for automatic MTU detection.
+	ignores := []string{
+		"docker.*", "cbr.*", "dummy.*",
```
fasaxc, line 119: This will want to be configurable.

dataplane/linux/int_dataplane.go
```diff
@@ -866,6 +886,75 @@ func (d *InternalDataplane) RecvMessage() (interface{}, error) {
 	return <-d.fromDataplane, nil
 }
 
+// configureMTU auto-detects the MTU to use for IPIP, VXLAN, and Wireguard interfaces
+// if not already set, and writes the smallest MTU to disk for use by other components.
+func configureMTU(config *Config) error {
```
fasaxc, line 891: Recommend splitting this into a detection function and a function to write out the data.  Then, you can periodically call the detection function in a loop and, if the MTU changes, trigger felix to restart so that we apply the new MTU.

caseydavenport:
> Have you thought about adjusting veth MTUs on the fly? 

I have thought about this a little bit. My plan was to bank the static approach first and then revisit adding MTU adjustment for existing veths as a follow-on. 

> Should the code go in libcalico-go so we can call it from the CNI plugin too if the mtu file isn't present?

I did originally consider this, but I think decided that this should go into Felix. Especially once we have the on-the-fly adjustment we won't need this in the CNI plugin. 

Plus, it would mean adding dataplane-specific logic to libcalico

caseydavenport:
Corresponding CNI PR here: https://github.com/projectcalico/cni-plugin/pull/962

caseydavenport:
libcalico-go PR here: https://github.com/projectcalico/libcalico-go/pull/1331

dataplane/linux/int_dataplane.go
```diff
@@ -279,6 +287,32 @@ func NewIntDataplaneDriver(config Config) *InternalDataplane {
 		config.RulesConfig.IptablesMarkEndpoint,
 		config.RulesConfig.IptablesMarkNonCaliEndpoint)
 
+	// Auto-detect host MTU.
+	if mtu, err := findHostMTU(config.MTUIfacePattern); err != nil {
+		log.WithError(err).Fatal("Unable to detect host MTU, shutting down")
```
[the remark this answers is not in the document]
fasaxc: I guess we only hit this if we can't list interfaces, which is pretty bad.  OK, seems fine to exit.

dataplane/linux/int_dataplane.go
```diff
@@ -741,6 +775,82 @@ func NewIntDataplaneDriver(config Config) *InternalDataplane {
...
+	}
+	return smallest, nil
+}
+
+// writeMTUFile writes the smallest MTU among enabled encapsulation types to disk
+// for use by other components (e.g., CNI plugin).
+func writeMTUFile(config Config) error {
+	// Make sure directory exists.
+	if err := os.MkdirAll("/var/lib/calico", os.ModePerm); err != nil {
+		return err
```
fasaxc, line 809: Worth a `fmt.Errorf("failed to create /var/lib/calico directory: %w", err)`?

fasaxc (changes requested):
Looking good overall; one thought on how to make the restart more smooth and a couple of minor suggestions.

caseydavenport:
@fasaxc I think this is ready for another look!

fasaxc (approved):

NicolasT:
This seems to set the system MTU to the lowest MTU across all interfaces found on the machine (whose name matches a particular pattern). Would it make sense for this to be the MTU of the interface that to which the (potentially auto-detected) IP of the node is attached?

Use-case: servers have two interfaces, one for a control-plane network, one for workload-plane traffic. Control-plane is plain 10GigE, MTU 1500. Workload-plane is, say 40GigE, jumbo-frames MTU 9000. IP auto-detection is set to match the workload-plane interface/CIDR/... so we likely want the discovered MTU to be 9000, not 1500...

caseydavenport:
@NicolasT Yes, I think that's a valid scenario and we might be able to improve upon that. 

I do think you can account for that in the current scenario by using the MTUInterfacePattern configuration option so that Calico will ignore the control-plane interfaces when performing the calculation.

I did consider using the auto-detected interface, but I don't think that's necessarily the right option either (not all installs have such an interface, either because they aren't using Calico for host->host networking, or because they are using multiple data interfaces for pod network traffic)

PR #2520, opened 2020-10-06
Enable UDP bypass, tighten RPF check accordingly

fasaxc:
## Description

* Enable the bypass mark pattern for UDP flows.
* Tighten up the RPF logic so that the initial response packet goes through RPF.  This is more important for UDP because UDP may be unidirectional so there's an arbitrarily large window where another interface could slip in a "response" packet.
* Add a TCP spoofing test.  This works by adding a secondary interface to the workload and moving routes between interfaces so that we can move an active TCP connection from one interface to the other.

## Todos
- [x] Release note

## Release Note

```release-note
In BPF mode, skip second conntrack lookup when flow is established.  Tighten up the RPF check to ensure this is safe.
```

### DIFF
--- a/bpf-gpl/conntrack.h
+++ b/bpf-gpl/conntrack.h
@@ -713,7 +713,7 @@ static CALI_BPF_INLINE struct calico_ct_result calico_ct_v4_lookup(struct ct_ctx
 		}
 		CALI_CT_VERB("B: whitelisted %d.\n", v->b_to_a.whitelisted);
 
-		if (tcp_header && v->a_to_b.whitelisted && v->b_to_a.whitelisted) {
+		if (v->a_to_b.whitelisted && v->b_to_a.whitelisted) {
 			result.rc = CALI_CT_ESTABLISHED_BYPASS;
 		} else {
 			result.rc = CALI_CT_ESTABLISHED;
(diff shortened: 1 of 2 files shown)


bpf-gpl/conntrack.h
```diff
@@ -708,7 +708,7 @@ static CALI_BPF_INLINE struct calico_ct_result calico_ct_v4_lookup(struct ct_ctx
 		}
 		CALI_CT_VERB("B: whitelisted %d.\n", v->b_to_a.whitelisted);
 
-		if (tcp_header && v->a_to_b.whitelisted && v->b_to_a.whitelisted) {
+		if ((tcp_header || proto_orig == IPPROTO_UDP) && v->a_to_b.whitelisted && v->b_to_a.whitelisted) {
```
tomastigera, line 711: If we were to fix it for UDP, perhaps we should do the same for all protocols.

```suggestion
		if (v->a_to_b.whitelisted && v->b_to_a.whitelisted) {
```

tomastigera (commented):
See https://github.com/projectcalico/felix/pull/2523 for the fv fix

tomastigera (approved):
Cool,LGTM!

PR #2529, opened 2020-10-13
Fix L2 dataplane reconciliation

caseydavenport:
## Description

This PR changes the logic so that we attempt to remove both the FDB entry and the ARP entry for a given IP / MAC pairing, assuming the MAC address isn't in the expected set, and expect that one of them may fail with "does not exist", which we will no longer consider an error.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```


song-jiang (approved):
LGTM

PR #2532, opened 2020-10-15
Service loop prevention

neiljerram:
## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```


proto/felixbackend.proto
```diff
@@ -548,3 +551,7 @@ message WireguardEndpointRemove {
   // The name of the wireguard host.
   string hostname = 1;
 }
+
+message GlobalBGPConfigUpdate {
+  repeated string service_cluster_ips = 1;
```
fasaxc, line 556: `service_cluster_cidrs`?

neiljerram: (I had that in my mind too, but forgot to action it.)

fasaxc (changes requested):
Code looks sound, just one nit.  Of course, this won't cover eBPF mode, maybe we need to follow up with a fix for that.  Unfortunately, that'd probably need a different approach, adding the CIDR to the L3 route resolver.

fv/service_loop_prevention_test.go
```diff
@@ -116,13 +119,77 @@ var _ = infrastructure.DatastoreDescribe("service loop prevention; with 2 nodes"
...
+			// Return the number of packets observed to/from 10.96.0.19.
+			return tcpdumpF.MatchCount("serviceIPPackets")
+		}
+
+		if expectLoop {
+			// Tcpdump should see more than 2 packets, because of looping.  Note: 2
+			// packets would be Felix receiving the ping and then forwarding it out
+			// again.  I want to check here that it's also looped around again by the
+			// gateway, resulting in MORE THAN 2 packets.
+			Expect(tryPing()).To(BeNumerically(">", 2))
```
fasaxc, line 178: `MatchCount` is async (it depends on reading from the stdout pipe of tcpdump) so it should always be used with an `Eventually` (or a retry loop).

It's usually quite fast but I don't think it's guaranteed to work here.

fasaxc (changes requested):
Mostly looking good, just a small problem with the way you're using tcpdump

fasaxc (approved):

neiljerram:
Docs and release note are at https://github.com/projectcalico/calico/pull/4104

PR #2535, opened 2020-10-16
Fix wireguard resync processing: correctly flag already handled nodes

robbrockbank:
## Description
Fix wireguard re-sync processing.

There were a couple of issues:
-  Nodes that were not deleted were the being treated as new and undergoing a full update (actually they might actually have two updates).  The tracking of whether a node was handled via delta was incorrect.
-  The tracking did not handle CIDRs missing from the node allowed IPs.

This PR fixes both of these.

The existing tests should still cover this - it's just before we ended up doing a full node update so the data plane values were still correct. 

I have, however, beefed up the tests to check the actual contents of the updates to check:
-  We don't have multiple entries for the same peer in a single update
-  The correct update/replace flag is set

Hopeful this will fix https://github.com/projectcalico/calico/issues/4091

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
Fix Wireguard resync processing to properly track deltas
```

### DIFF
--- a/netlinkshim/mocknetlink/netlink.go
+++ b/netlinkshim/mocknetlink/netlink.go
@@ -224,6 +224,7 @@ type MockNetlinkDataplane struct {
 	NumRuleAddCalls        int
 	NumRuleDelCalls        int
 	WireguardConfigUpdated bool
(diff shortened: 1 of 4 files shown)


caseydavenport (approved):
Seems legit.

PR #2538, opened 2020-10-19
bpf: pick a random backend after affinity timeo expires

tomastigera:
When session affinity time out expires, we should make a new pick of the
backend.

Currently, we were using a stable hash of the client's IP which would
result in the same pick if the set of backends did not change.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
In BPF mode, choose the backend randomly for service VIPs with service affinity enabled.  This ensures a more random distribution and good rebalancing (as expected by the k8s conformance tests).
```

### DIFF
--- a/fv/bpf_test.go
+++ b/fv/bpf_test.go
@@ -1898,22 +1898,51 @@ func describeBPFTests(opts ...bpfTestOpt) bool {
 							"Service endpoints didn't get created? Is controller-manager happy?")
 					})
 
-					It("should have connectivity from a workload to a service with multiple backends", func() {
-						ip := testSvc.Spec.ClusterIP
-						port := uint16(testSvc.Spec.Ports[0].Port)
+					// FIXME we can only do the test with regular NAT as
+					// cgroup shares one random affinity map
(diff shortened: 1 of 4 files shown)


fv/bpf_test.go
```diff
@@ -1899,6 +1899,12 @@ func describeBPFTests(opts ...bpfTestOpt) bool {
 					})
 
 					It("should have connectivity from a workload to a service with multiple backends", func() {
+						// FIXME we can only do the test with regular NAT as
+						// cgroup shares one random affinity map
+						if testOpts.connTimeEnabled {
+							return
```
fasaxc, line 1905: Please can you wrap the `It` in this `if { ... }` instead; skipping a test in this way means that we do about 30s of set-up (creating containers and workloads) before the skip so it's very inefficient.

fasaxc (approved):
Thanks! Approved with one tweak I think you should make.

fv/bpf_test.go
tomastigera: Ohhh sure, needs to be outside :+1:

PR #2546, opened 2020-10-21
Export multipath route to kernel

misi:
Allow ECMP egress gw propagation to kernel.
Actual behavior that bird sees the received multipath routes (e.g. default routes 0.0.0.0/0) 
but exports only one path to the kernel not all paths.

## Description
See issue that I opened here: https://github.com/projectcalico/calico/issues/4114
(Please move this issue to this repository)


neiljerram (approved):
LGTM, thanks!

PR #2549, opened 2020-10-29
bpf: nodeport return with multiple ifaces

tomastigera:
In environments like EKS, where the egress route is determined by the
source IP as well, the return VXLAN packet for the nodeport tunnel may
end up at a wrong interface (and route) because the packet has the
source IP of the backing pod - to avoid being dropped by RPF check etc.

This fix uses the knowledge of the iface where the nodeport ingress
packets came from (recorded for the purpose of RPF check) and send the
packet out straight to the same interface.

Until we get bpf_redirect_neigh() in our supported kernels (5.10+) we
need to maintain an ARP lookup cache that allows us to patch in the
right MAC addresses so that the packet, once forwarded by the host
interface, can reach the next hop.

We use the tunnel IP where we are sending the egress packet to and the
ifindex of the device where it came from look up the MAC addresses we should
patch in.

The lookup map is an LRU hash with 10k entries. That allows 10k other
nodes to forward their nodeports to a node.

## Todos
- [ ] Documentation
- [ ] Backport
- [x] Release note

## Release Note

```release-note
In eBPF mode, return nodeport traffic from the backing pod to the same interface that it arrived at.  This prevents incorrect routing in environments like EKS, which use source-based routing.
```

### DIFF
--- a/bpf-gpl/arp.h
+++ b/bpf-gpl/arp.h
@@ -1,0 +1,33 @@
+// Project Calico BPF dataplane programs.
+// Copyright (c) 2020 Tigera, Inc. All rights reserved.
+//
+// This program is free software; you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation; either version 2 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License along
+// with this program; if not, write to the Free Software Foundation, Inc.,
+// 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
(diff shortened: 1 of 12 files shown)


dataplane/linux/bpf_arp_mgr.go
```diff
@@ -67,16 +105,31 @@ func (m *bpfARPManager) CompleteDeferredWork() error {
...
 			log.WithField("IP", n.IP).Info("Ignoring non-IPv4 neighbour.")
 			continue
 		}
 
+		var ipv4key [4]byte
+		copy(ipv4key[:], ipv4)
+
+		if _, ok := m.remoteNodes[ip.V4Addr(ipv4key)]; !ok {
+			// Throw away anything that is not pointing to a node as that is what we need
+			// and we do not want to overfill the table in a large cluster. If we through
```
seh, line 119: s/through/throw/

fasaxc (changes requested):
Seems like a pragmatic fix

fasaxc (approved):
Realised all my comments were in the file you deleted so I guess this is approved :-)

caseydavenport:
Reminder that this will need cherry-picking if you'd like this to make it to v3.17 :) 

@tomastigera

PR #2550, opened 2020-10-30
Expand MTU interface regex

fasaxc:
## Description

Tests were failing on some systems because Felix couldn't detect MTU.  Turns out there are more possible prefixes for data interfaces.  This PR expands the regex to cover other cases.  The prefixes were found by reviewing hte systemd source code where they are generated: https://github.com/systemd/systemd/blob/master/src/udev/udev-builtin-net_id.c#L801

Beefed up the BPF regex equivalently.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
Broaden the data interface detection regex to cover wifi and other devices.
```

### DIFF
--- a/config/config_params.go
+++ b/config/config_params.go
@@ -117,7 +117,7 @@ type Config struct {
 	BPFEnabled                         bool           `config:"bool;false"`
 	BPFDisableUnprivileged             bool           `config:"bool;true"`
 	BPFLogLevel                        string         `config:"oneof(off,info,debug);off;non-zero"`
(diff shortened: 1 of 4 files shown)


tomastigera (approved):

fasaxc:
Think the regex part of this is sound but the part to return an error instead of 0 missed the case where a manual MTU is specified.

fasaxc:
Split out #2555

PR #2554, opened 2020-11-02
Fix service loop prevention in IPVS mode.

fasaxc:
## Description

Remove service loop prevention rule from OUTPUT chain because
the OUTPUT chain comes _before_ DNAT when IPVS is in use.
This means that we were dropping _all_ service IP traffic.

Unconditionally drop the rule since if traffic does loop back
to us, we'll catch it on the forward path.
## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/rules/static.go
+++ b/rules/static.go
@@ -645,13 +645,6 @@ func (r *DefaultRuleRenderer) filterOutputChain(ipVersion uint8) *Chain {
 		},
 	)
 
-	// Jump to chain for blocking service CIDR loops.
-	rules = append(rules,
-		Rule{
-			Action: JumpAction{Target: ChainCIDRBlock},
-		},
-	)
-
 	return &Chain{
 		Name:  ChainFilterOutput,
 		Rules: rules,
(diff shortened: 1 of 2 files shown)


rules/static.go
```diff
@@ -645,13 +645,6 @@ func (r *DefaultRuleRenderer) filterOutputChain(ipVersion uint8) *Chain {
 		},
 	)
 
-	// Jump to chain for blocking service CIDR loops.
-	rules = append(rules,
-		Rule{
-			Action: JumpAction{Target: ChainCIDRBlock},
-		},
-	)
-
```
song-jiang, line 654: Does that mean we disable service loop prevention for packet generated by local host?

fasaxc: Yes, exactly; but if there is a loop and the packet returns to us; _then_ we'll drop it in the forward chain before we send it back out again.

song-jiang (approved):
LGTM

rules/static.go
neiljerram: Nice reasoning and solution!

neiljerram (approved):

PR #2563, opened 2020-11-03
Handle when unable to detect host MTU

caseydavenport:
## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/dataplane/linux/int_dataplane.go
+++ b/dataplane/linux/int_dataplane.go
@@ -805,7 +805,7 @@ func findHostMTU(matchRegex *regexp.Regexp) (int, error) {
 	if smallest == 0 {
 		// We failed to find a usable interface. Default the MTU of the host
 		// to 1460 - the smallest among common cloud providers.
-		log.Warn("Failed to auto-detect host MTU, assuming 1460")
+		log.Warn("Failed to auto-detect host MTU - no interfaces matched the MTU interface pattern. To use auto-MTU, set mtuIfacePattern to match your host's interfaces")
 		return 1460, nil
 	}
 	return smallest, nil
--- a/dataplane/linux/int_dataplane.go
+++ b/dataplane/linux/int_dataplane.go
@@ -801,6 +801,13 @@ func findHostMTU(matchRegex *regexp.Regexp) (int, error) {
 			smallest = l.Attrs().MTU
 		}
 	}
+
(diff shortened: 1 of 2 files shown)


dataplane/linux/int_dataplane.go
```diff
@@ -801,6 +801,13 @@ func findHostMTU(matchRegex *regexp.Regexp) (int, error) {
 			smallest = l.Attrs().MTU
 		}
 	}
+
+	if smallest == 0 {
+		// We failed to find a usable interface. Default the MTU of the host
+		// to 1460 - the smallest among common cloud providers.
+		log.Warn("Failed to auto-detect host MTU, assuming 1460")
```
fasaxc, line 808: You could improve this log by saying something like "Failed to auto-detect host's MTU, no interfaces matched the MTU data interface pattern; to use auto MTU detection, set <param name> to match your host's interfaces."

fasaxc (changes requested):
Could do with an FV test, should just be a matter of setting the interface pattern to garbage and checking ht MTU file shows up.

caseydavenport:
@fasaxc thanks for the review! I've added a test

dataplane/linux/int_dataplane.go
```diff
@@ -801,6 +801,13 @@ func findHostMTU(matchRegex *regexp.Regexp) (int, error) {
 			smallest = l.Attrs().MTU
 		}
 	}
+
+	if smallest == 0 {
+		// We failed to find a usable interface. Default the MTU of the host
+		// to 1460 - the smallest among common cloud providers.
+		log.Warn("Failed to auto-detect host MTU - no interfaces matched the MTU interface pattern. To use auto-MTU, set to match your host's interfaces")
```
fasaxc, line 808: :facepalm: My comment got mangled, Tried to say `...set <name of parameter> to match...`

fasaxc (changes requested):

dataplane/linux/int_dataplane.go
caseydavenport: lol right, fixed it.

caseydavenport:
@fasaxc I think this is good to go now - I'm going to merge so I can get it into a pin update and into QA, but I'll keep an eye out for any more comments.

PR #2579, opened 2020-11-12
Change default route table range to avoid clash with GKE.

fasaxc:
## Description

GKE uses routing table 1 when intera-node visibility is enabled.  Change our default range to avoid clashing with that (and leave some headroom).

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
Change the default range of routing tables that Calico is allowed to use to avoid clashes with common cloud providers.
```

### DIFF
--- a/config/config_params.go
+++ b/config/config_params.go
@@ -270,7 +270,7 @@ type Config struct {
 	// - calicoIPAM: use IPAM data to contruct routes.
 	RouteSource string `config:"oneof(WorkloadIPs,CalicoIPAM);CalicoIPAM"`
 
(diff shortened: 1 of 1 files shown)


robbrockbank:
My only concern with this is upgrade.

If someone is already using wireguard then we'll have created a wg table w/ index 1 which will never get removed.  Perhaps that's ok though - but we might need to release note it.

realgaurav:
Agree with upgrade concerns here. A possible manifest/operator change for GKE would be more appropriate IMHO.
If we want to use this, we may also want to update the docs to be thorough.

Thanks

fasaxc:
OK, let's take a different path

PR #2582, opened 2020-11-16
Fix application of normal HEP policy to traffic from host to a DNAT'd IP

neiljerram:
## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
Bugfix: HostEndpoint policy was incorrectly enforced on connections from a pod to a service IP, even when that service IP was resolved to another workload on the same host.
```

### DIFF
--- a/fv/hostendpoints_test.go
+++ b/fv/hostendpoints_test.go
@@ -159,6 +159,25 @@ func describeHostEndpointTests(getInfra infrastructure.InfraFactory, allInterfac
 		cc.ExpectSome(w[0], w[1])
 		cc.ExpectSome(w[1], w[0])
 	}
+	expectLocalPodToRemotePodViaServiceTraffic := func() {
+		for i := range felixes {
+			// Allocate a service IP.
+			serviceIP := fmt.Sprintf("10.96.10.%v", i+1)
+
+			// Add a NAT rule for the service IP.
+			felixes[i].Exec(
+				"iptables",
+				"-w", "10", // Retry this for 10 seconds, e.g. if something else is holding the lock
+				"-W", "100000", // How often to probe the lock in microsecs.
+				"-t", "nat", "-A", "PREROUTING",
+				"--destination", serviceIP,
+				"-j", "DNAT", "--to-destination", w[1-i].IP,
+			)
+
+			// Expect to connect from local pod to the service IP.
+			cc.ExpectSome(w[i], connectivity.TargetIP(serviceIP), 8055)
+		}
+	}
 	expectDenyHostToHostTraffic := func() {
 		cc.ExpectNone(felixes[0], hostW[1])
 		cc.ExpectNone(felixes[1], hostW[0])
@@ -222,6 +241,7 @@ func describeHostEndpointTests(getInfra infrastructure.InfraFactory, allInterfac
 			expectHostToOwnPodTraffic()
 			expectHostToOwnPodViaServiceTraffic()
 			expectDenyHostToRemotePodViaServiceTraffic()
+			expectLocalPodToRemotePodViaServiceTraffic()
 			cc.CheckConnectivity()
 		})
 
--- a/fv/hostendpoints_test.go
+++ b/fv/hostendpoints_test.go
@@ -115,6 +115,26 @@ func describeHostEndpointTests(getInfra infrastructure.InfraFactory, allInterfac
 		cc.ExpectSome(felixes[0], w[0])
 		cc.ExpectSome(felixes[1], w[1])
 	}
+	expectHostToOwnPodViaServiceTraffic := func() {
+		// host to own pod always allowed, even via a service IP
+		for i := range felixes {
+			// Allocate a service IP.
+			serviceIP := fmt.Sprintf("10.96.0.%v", i+1)
+
+			// Add a NAT rule for the service IP.
+			felixes[i].Exec(
+				"iptables",
+				"-w", "10", // Retry this for 10 seconds, e.g. if something else is holding the lock
+				"-W", "100000", // How often to probe the lock in microsecs.
+				"-t", "nat", "-A", "OUTPUT",
+				"--destination", serviceIP,
+				"-j", "DNAT", "--to-destination", w[i].IP,
+			)
+
+			// Expect connectivity to the service IP.
+			cc.ExpectSome(felixes[i], connectivity.TargetIP(serviceIP), 8055)
+		}
+	}
 	expectPodToPodTraffic := func() {
 		cc.ExpectSome(w[0], w[1])
 		cc.ExpectSome(w[1], w[0])
@@ -180,6 +200,7 @@ func describeHostEndpointTests(getInfra infrastructure.InfraFactory, allInterfac
 			expectDenyHostToOtherPodTraffic()
 			expectPodToPodTraffic()
 			expectHostToOwnPodTraffic()
+			expectHostToOwnPodViaServiceTraffic()
 			cc.CheckConnectivity()
 		})
 
--- a/fv/hostendpoints_test.go
+++ b/fv/hostendpoints_test.go
@@ -135,6 +135,26 @@ func describeHostEndpointTests(getInfra infrastructure.InfraFactory, allInterfac
 			cc.ExpectSome(felixes[i], connectivity.TargetIP(serviceIP), 8055)
 		}
 	}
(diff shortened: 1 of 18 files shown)


rules/static.go
```diff
@@ -826,6 +844,64 @@ func (r *DefaultRuleRenderer) StaticManglePreroutingChain(ipVersion uint8) *Chai
...
+		)
+	}
+
+	// At this point we know that the packet is not forwarded, so it must be originated by a
+	// host-based process or host-networked pod.
+
+	if ipVersion == 4 {
+		r.addIPIPVXLANEgressAllowRules(&rules)
+	}
+
```
neiljerram, line 877: filterOutput has rules here to catch traffic going to a local workload.  I thought we didn't need that here, but in fact I now suspect we do, so that that traffic doesn't hit policy for the "all-interfaces" host endpoint.  That also means another FV case: Host -> service IP (iptables DNAT) -> local workload with "all-interfaces" host endpoint.

rules/static.go
```diff
@@ -595,11 +603,39 @@ func (r *DefaultRuleRenderer) filterOutputChain(ipVersion uint8) *Chain {
...
+}
+
+func (r *DefaultRuleRenderer) addIPIPVXLANEgressAllowRules(rules *[]Rule) {
+
+	if r.IPIPEnabled {
 		// When IPIP is enabled, auto-allow IPIP traffic to other Calico nodes.  Without this,
 		// it's too easy to make a host policy that blocks IPIP traffic, resulting in very confusing
 		// connectivity problems.
-		rules = append(rules,
+		*rules = append(*rules,
```
fasaxc, line 638: I'd normally fashion a function like this after `append`:
```
appendIPIPVXLANEgressAllowRules(rules []Rule) []Rule
```
Guess this is fine too but I don't often see pointers to slices in Go.

rules/static.go
```diff
@@ -826,6 +844,64 @@ func (r *DefaultRuleRenderer) StaticManglePreroutingChain(ipVersion uint8) *Chai
...
+	if ipVersion == 4 {
+		r.addIPIPVXLANEgressAllowRules(&rules)
+	}
+
+	// Apply host endpoint policy to non-forwarded traffic that has been DNAT'd.  We do this
+	// here, rather than in filter-OUTPUT, because Linux is weird: when a host-originated packet
+	// is DNAT'd (typically in nat-OUTPUT), its destination IP is changed immediately, but Linux
+	// does not recalculate the outgoing interface (OIF) until AFTER the filter-OUTPUT chain.
+	// The OIF has been recalculated by the time we hit THIS chain (mangle-POSTROUTING), so we
+	// can reliably apply host endpoint policy here.
```
fasaxc, line 883: +1 for big explainer comment :-)

fasaxc (commented):
Looking good so far

rules/static.go
neiljerram: Turns out we already have FV for this case, and it was already passing because cali-to-host-endpoint internally includes a check against applying host endpoint policy to traffic going to a local workload.  So we don't need those rules here.

fv/ipip_test.go
```diff
@@ -250,6 +251,106 @@ var _ = infrastructure.DatastoreDescribe("IPIP topology before adding host IPs t
...
+					},
+				}
+				policy.Spec.Selector = fmt.Sprintf("has(host-endpoint)")
+				_, err := client.GlobalNetworkPolicies().Create(utils.Ctx, policy, utils.NoOptions)
+				Expect(err).NotTo(HaveOccurred())
+			})
+
+			// Test each of these cases separately, to avoid the possibility of one case
+			// setting up conntrack state that allows a different case to pass.
+			It("allows host0 to remote host-networked workload", func() {
```
fasaxc, line 301: Might have missed something but I'm not sure this tests the case that was broken before.  I think you need a test that blocks traffic on eth0 but asserts that traffic to a remote pod still gets through (because the eth0 hep doesn't apply to the IPIP/vxlan interface).

rules/static.go
```diff
@@ -636,6 +644,7 @@ func (r *DefaultRuleRenderer) filterOutputChain(ipVersion uint8) *Chain {
 			Action: ClearMarkAction{Mark: r.allCalicoMarkBits()},
 		},
 		Rule{
+			Match:  Match().NotConntrackState("DNAT"),
```
fasaxc, line 647: Worth commenting the extra match

rules/static.go
```diff
@@ -826,6 +837,70 @@ func (r *DefaultRuleRenderer) StaticManglePreroutingChain(ipVersion uint8) *Chai
 	}
 }
 
+func (r *DefaultRuleRenderer) StaticManglePostroutingChain(ipVersion uint8) *Chain {
+	rules := []Rule{}
+
+	// Accept immediately if IptablesMarkAccept is set.  Our filter-FORWARD chain sets this for
+	// any packets that reach the end of that chain.  The principle is that we don't want to
+	// apply normal host endpoint policy to forwarded traffic.
+	rules = append(rules, r.acceptAlreadyAccepted()...)
```
fasaxc, line 846: After a bit of thought, I think the "allow" action for the mangle table should be "RETURN" instead of "ACCEPT", at least by default.  The mangle table is not normally used for policy so its default action is unlikely to be deny (we didn't even have rules in this chain before your work so it can't be deny-by-default).

If the mangle table is in use, it's to tweak packets just before they leave the box so we should just fall through to any rules that the user has.

fasaxc (changes requested):
Generally looking good but I had a couple of thoughts.

fv/ipip_test.go
neiljerram: The test for the previously broken case is in hostendpoints_test.go, line 118 (`expectHostToOwnPodViaServiceTraffic`).

This one is arguably superfluous, and perhaps I should remove it.  It's host to host and - IIUC - doesn't use the IP-IP overlay.  My purpose here (and similarly in vxlan_test.go) was to verify the need (or not) for the IPIP/VXLAN failsafe rules in StaticManglePostrouting, and for that I only really need the "allows host0 to remote Calico-networked workload via service IP" cases.

WDYT?  Should I remove the other cases that I've added?  (They may well duplicate testing that we already have elsewhere.)

fv/hostendpoints_test.go
```diff
@@ -115,10 +115,69 @@ func describeHostEndpointTests(getInfra infrastructure.InfraFactory, allInterfac
...
 		cc.ExpectSome(felixes[1], w[1])
 	}
+	expectHostToOwnPodViaServiceTraffic := func() {
+		// host to own pod always allowed, even via a service IP
+		for i := range felixes {
+			// Allocate a service IP.
+			serviceIP := fmt.Sprintf("10.96.0.%v", i+1)
+
+			// Add a NAT rule for the service IP.
+			felixes[i].Exec(
```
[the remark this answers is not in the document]
neiljerram: Do you feel strongly that way?  I think I prefer the locality as it is now, and that it would be harder to understand with part of this moved to BeforeEach.

rules/static.go
neiljerram: I think this one is trickier than it looks!  We do already have IptablesMangleAllowAction, but

- it defaults to ACCEPT
- it was presumably intended for the pre-existing mangle usage, i.e. pre-DNAT
- to follow through your point above, we maybe need to plumb an allowAction parameter through the whole of the host dispatch and endpoint chains, and review that they work with RETURN.

I'll review the last point in more detail...

neiljerram: Actually the plumbing detail for the last point looks fine.  So it's just a matter of whether the existing IptablesMangleAllowAction would go, or if we need an independent parameter.

fv/hostendpoints_test.go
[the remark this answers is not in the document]
neiljerram: Have introduced felix.ProgramIptablesDNAT to address the duplication here.

fv/ipip_test.go
neiljerram: As agreed, I've removed the superfluous cases now.

rules/static.go
neiljerram: As agreed, I've changed the code to Return; hardcoded as such until a need arises to make it configurable.

fasaxc (approved):
LGTM

PR #2588, opened 2020-11-19
bpf/proxy: fix stale NAT entries cleanup

tomastigera:
When nodeport is forwarded to another node, it does not have the first
node's IP in its tables, therefore it thinks that the conntrack entries
are for a service that is gone. We also need to match against the
wildcard to check if the front end might be a nodeport.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
In eBPF mode, fix that node port conntrack entries could be improperly cleaned up if there was no traffic for some time.  In particular, long-lived connections from AWS load balancers were affected.
```


fasaxc (approved):
Let's follow up with some tests but I'm going to merge this so we can get the fix out there.

PR #2589, opened 2020-11-19
[v3.16] bpf/proxy: fix stale NAT entries cleanup

fasaxc:
## Description

Backport #2588 

When nodeport is forwarded to another node, it does not have the first
node's IP in its tables, therefore it thinks that the conntrack entries
are for a service that is gone. We also need to match against the
wildcard to check if the front end might be a nodeport.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
In eBPF mode, fix that long-lived connections to nodeports could be incorrectly cleaned up resulting in dropped connections, this particularly impacted AWS load balancers.
```

### DIFF
--- a/bpf/proxy/syncer.go
+++ b/bpf/proxy/syncer.go
@@ -1272,7 +1272,13 @@ func (s *Syncer) ConntrackFrontendHasBackend(ip net.IP, port uint16,
 
 	id, ok := s.activeSvcsMap[ipPortProto{ipPort{ip.String(), int(port)}, proto}]
 	if !ok {
-		return false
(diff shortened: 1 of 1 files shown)


seh:
@hakman, I reported this problem afflicting my _kOps_-provisioned clusters in AWS, and confirmed that the fix works. I hope we can include a Calico version with this fix in the next _kOps_ beta version.

Next up is to test whether this resolves projectcalico/calico#4146 as well.

hakman:
@seh kOps 1.19 will be released with Calico v3.15, so only if backported to it.

PR #2601, opened 2020-11-21
NAT stale cleanup tests

tomastigera:
## Description

Add FV tests for #2588 

## Todos
- [ ] Backport

## Release Note

```release-note
None required
```

### DIFF
--- a/fv/bpf_test.go
+++ b/fv/bpf_test.go
@@ -51,6 +51,7 @@ import (
 	"github.com/projectcalico/felix/bpf"
 	"github.com/projectcalico/felix/bpf/conntrack"
 	"github.com/projectcalico/felix/bpf/nat"
+	"github.com/projectcalico/felix/bpf/proxy"
 	. "github.com/projectcalico/felix/fv/connectivity"
 	"github.com/projectcalico/felix/fv/containers"
 	"github.com/projectcalico/felix/fv/infrastructure"
@@ -191,6 +192,8 @@ func describeBPFTests(opts ...bpfTestOpt) bool {
 	for _, o := range opts {
 		o(&testOpts)
 	}
+
+	testIfTCP := testOpts.protocol == "tcp"
 
 	protoExt := ""
 	if testOpts.udpUnConnected {
@@ -2268,6 +2271,49 @@ func describeBPFTests(opts ...bpfTestOpt) bool {
 											Should(BeNumerically(">", 0), "MACs do not match")
 									})
 								}
+							})
+
+							_ = testIfTCP && It("should survive conntrack cleanup sweep", func() {
+								By("checking the connectivity and thus syncing with service creation", func() {
+									cc.ExpectSome(externalClient, TargetIP(felixes[1].IP), npPort)
+									cc.CheckConnectivity()
+								})
+
+								By("monitoring a persistent connection", func() {
+									pc := &PersistentConnection{
+										Runtime:             externalClient,
+										RuntimeName:         externalClient.Name,
+										IP:                  felixes[1].IP,
+										Port:                int(npPort),
+										Protocol:            testOpts.protocol,
+										MonitorConnectivity: true,
+									}
+
+									err := pc.Start()
+									Expect(err).NotTo(HaveOccurred())
+									defer pc.Stop()
+
+									EventuallyWithOffset(1, pc.PongCount, "5s").Should(
+										BeNumerically(">", 0),
+										"Expected to see pong responses on the connection but didn't receive any")
+									log.Info("Pongs received within last 1s")
+
+									// We make sure that at least one iteration of the conntrack
+									// cleanup executes and we periodically monitor the connection if
+									// it is alive by checking that the number of PONGs keeps
+									// increasing.
+									start := time.Now()
+									prevCount := pc.PongCount()
+									for time.Since(start) < 2*proxy.ConntrackCleanerPeriod {
+										time.Sleep(time.Second)
+										newCount := pc.PongCount()
+										Expect(prevCount).Should(
+											BeNumerically("<", newCount),
+											"No new pongs since the last iteration. Connection broken?",
+										)
+										prevCount = newCount
+									}
+								})
 							})
 
 							if !testOpts.dsr {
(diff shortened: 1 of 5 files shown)


fv/bpf_test.go
```diff
@@ -2270,6 +2271,41 @@ func describeBPFTests(opts ...bpfTestOpt) bool {
...
+									Expect(err).NotTo(HaveOccurred())
+									defer pc.Stop()
+
+									EventuallyWithOffset(1, pc.PongCount, "5s").Should(
+										BeNumerically(">", 0),
+										"Expected to see pong responses on the connection but didn't receive any")
+									log.Info("Pongs received within last 1s")
+
+									// We make sure that at least one iteration of the conntrack
+									// cleanup executes and we periodically monitor the connectio if
```
seh, line 2295:
```suggestion
									// cleanup executes, and we periodically monitor the connection if
```

fv/bpf_test.go
```diff
@@ -2270,6 +2271,41 @@ func describeBPFTests(opts ...bpfTestOpt) bool {
...
+									// is is alive be checking that the number of PONGs keeps
```
seh, line 2296:
```suggestion
									// is is alive, confirming that the number of PONGs keeps
```

fv/connectivity/conncheck.go
```diff
@@ -741,3 +743,139 @@ func (cc ConnConfig) GetTestMessageSequence(msg string) (int, error) {
...
+
+	lastPongTime time.Time
+	pongCount    int
+}
+
+func (pc *PermanentConnection) Stop() {
+	Expect(pc.stop()).NotTo(HaveOccurred())
+}
+
+var permConnIdx = 0 // XXX perhaps should be atomic / locked
```
fasaxc, line 777: Should be fine, we only create connections from the main thread.

fv/bpf_test.go
```diff
@@ -2270,6 +2271,48 @@ func describeBPFTests(opts ...bpfTestOpt) bool {
 								}
 							})
 
+							if testOpts.protocol == "tcp" {
```
fasaxc, line 2274: Maybe we should define 
```
ConditionallyIt(testOpts.protocol == "tcp", "should survive conntrack cleanup sweep", func() {
  ...
})
```
To replace `if ... { It(...` with the extra nesting

fv/bpf_test.go
```diff
@@ -2270,6 +2271,48 @@ func describeBPFTests(opts ...bpfTestOpt) bool {
...
+										// We make sure that at least one iteration of the conntrack
+										// cleanup executes and we periodically monitor the connection if
+										// it is alive by checking that the number of PONGs keeps
+										// increasing.
+										start := time.Now()
+										prevCount := pc.PongCount()
+										for time.Since(start) < 2*proxy.ConntrackCleanerPeriod {
+											time.Sleep(time.Second)
+											newCount := pc.PongCount()
+											Expect(prevCount).Should(BeNumerically("<", newCount))
```
fasaxc, line 2309: Nit: if this fails, we'll get `Fail: 6 is not less than 6`, which is a bit lacking for debugging.  `Should` has an optional second arg that takes a message; recommend using that whenever you're comparing random numbers.  For example:
```
Expect(prevCount).Should(BeNumerically("<", newCount), "persistent connection stopped receiving pongs")
```

fv/connectivity/conncheck.go
```diff
@@ -741,3 +743,139 @@ func (cc ConnConfig) GetTestMessageSequence(msg string) (int, error) {
...
 	return strings.HasPrefix(strings.TrimSpace(msg), ConnectionTypeStream)
 }
+
+// Runtime abstracts *containers.Container to avoid import loops
+type Runtime interface {
+	EnsureBinary(name string)
+	ExecMayFail(cmd ...string) error
+}
+
+type PermanentConnection struct {
```
fasaxc, line 753: Nice to have this extracted out, thanks!

Wonder if we should change the name to `PersistentConnection`.  It's a subtle distinction, but I think that's more correct.

fasaxc (changes requested):
Nice, some good cleanup and the test looks good.  Thanks!  Just a few nits.

fv/bpf_test.go
tomastigera: I like that, I will do it!

fv/connectivity/conncheck.go
tomastigera: yeah, will rename it, tbh I was first grepping for persistent ;-)

fv/bpf_test.go
tomastigera: However, it is a bit disappointing that Go is no C and has no preprocessor :-/

It would have to look like 

```
ConditionallyIt(func() bool {return testOpts.protocol == "tcp")}, "should survive conntrack cleanup sweep", func() {
  ...
})
```

it may not be too bad

fv/bpf_test.go
```diff
@@ -192,6 +193,8 @@ func describeBPFTests(opts ...bpfTestOpt) bool {
 		o(&testOpts)
 	}
 
+	TestIfTCP := testOpts.protocol == "tcp"
```
fasaxc, line 196: Nit: non-exported variables should start with lower case.

fv/fv_test.go
```diff
@@ -0,0 +1,27 @@
...
+
+// +build fvtests
+
+package fv_test
+
+import (
+	. "github.com/onsi/ginkgo"
+)
+
+func ConditionIt(cond bool, title string, code func()) {
```
fasaxc, line 23: Worth a comment explaining why you'd want to use this

fv/fv_test.go
```diff
@@ -0,0 +1,27 @@
...
+	if cond {
+		It(title, code)
```
fasaxc, line 25: Can see one downside of this now I see it written out; when a test fails, it'll report this line number instead of the "real one".  That said; I tend to look for the line number of the assert, which should still be correct.

fasaxc (approved):
couple of nits but good to go in once you've addressed

fv/fv_test.go
tomastigera: thank can be fixed :+1:  (iirc)

fasaxc: I couldn't see a variant of `It` with an offset when I looked.  One option would be to do:
```
_ = testIfTCP && It(...)
```
since `It()` returns `true`. Bit of a hack though?

tomastigera: I did not know that It() returns anything until I digged deeper in ginkgo, but it is not too bad. And it makes the lines shorter, so I guess it is a winner! :-)

PR #2422, opened 2020-07-09
Use clang flag -target bpf

netoptimizer:
When compiling BPF code the clang flag `-target bpf` should be used.

The LLVM intermediate representation (IR) tries to stay architecture independent, but in some cases, when not using `-target bpf`, it can generate code (mostly data structure access) that will be rejected by the verifier.

This often happen if the program uses data structures with pointer or `long` / `unsigned long` types that interface with BPF helpers or context data structures.

More details see kernel Documentation:
https://www.kernel.org/doc/html/latest/bpf/bpf_devel_QA.html#q-clang-flag-for-target-bpf

I've only fixed this for the subdir "bpf-apache" as I had other compile issues in "bpf-gpl".
This the initial steps of moving towards using target bpf, the other places should of-cause also be fixed in follow ups.

PR #2608, opened 2020-12-01
Build BPF programs with -target bpf

fasaxc:
## Description

Version of #2422 for the bpf-gpl directory.

In order for clang to compile BPF programs correctly (and with BPF-specific workarounds enabled), compile with `-target=bpf`.  This requires a number of knock-on changes:

- Drop stdlib dependencies apart from a couple of workarounds; it doesn't support BPF.
- Build against the uapi headers.
- Vendor in libbpf so we can use its helper definitions and newer copy of bph.h (go-build is currently on an older version and that's hard to change given that it's based on debian buster).

## Todos

- [x] Release note

## Release Note

```release-note
The eBPF programs are now built with clang flag -target bpf and they use libbpf for helper definitions.  This enables the correct features in clang and improves build stability. Thanks to @netoptimizer for his PR and guidance on how to do this correctly.
```


bpf-gpl/Makefile
```diff
@@ -18,30 +18,31 @@
...
-	--include=$(LINUX_AMD64)/include/generated/uapi/linux/version.h
+	-target bpf \
+	-emit-llvm
+
+# Build against libbpf and its recent copy of the kernel headers.
+# We link against the user API version of the headers because they contain
+# everything we need for now.
+#
+# Note: for headers that aren't in the libbpf include directory, we'll
+# fall back on the system headers, which may be old.
```
neiljerram, line 36: I'm unclear if you are saying this is a possible concern.  Can you extend the comment to clarify?

bpf-gpl/Makefile
```diff
@@ -18,30 +18,31 @@
...
+CFLAGS +=  \
+	-I ./include/libbpf/src/ \
+	-I ./include/libbpf/include/uapi
+
+# Workaround for Debian placing "asm/types.h" in /usr/include/x86_64-linux-gnu
+# We also pick up a couple of other definitions from here, such as the socket
+# type constants (which are architecture dependent for historical reasons).
+TRIPLET := $(shell gcc -dumpmachine)
```
neiljerram, line 44: Is this a new requirement for non-containerised GCC being available on a developer's machine?  (I imagine the Mac users might have trouble with that.)

bpf-gpl/bpf.h
```diff
@@ -18,78 +18,24 @@
...
-#endif
-
-// Due to some late-found issues with pre-5.2.0, requiring v5.2.0+ for now.
-#if LINUX_VERSION_CODE < KERNEL_VERSION(5,2,0)
-#error Attempt to build against too-old kernel headers.
-#endif
-
+// After switching to a build based on libbpf we're now using a mix
+// of headers: newer versions of bpf.h from libbpf and some other
+// older headers from the system so it's not meaninful to pull in
```
neiljerram, line 23: sp: meaningful

bpf-gpl/bpf.h
```diff
@@ -18,78 +18,24 @@
...
+// the system version.h.
+#define KERNEL_VERSION(a,b,c) (((a) << 16) + ((b) << 8) + (c))
+#define LINUX_VERSION_CODE KERNEL_VERSION(5,2,0)
```
neiljerram, line 26: In the places in our code where we test for LINUX_VERSION_CODE, does that actually depend on the running kernel, or on the headers that we build with?

neiljerram (changes requested):
A few points, of which the GCC one might be important.

bpf-gpl/Makefile
fasaxc: This whole makefile is run inside go-build so it shouldn't be a problem.

bpf-gpl/bpf.h
fasaxc: No, this change bypasses those checks.  Perhaps I should just rip out the `#define`s that condition on kernel version; they haven't actually worked for a long time and we're always building against the particular kernel headers in libbpf and the go-build container.

bpf-gpl/Makefile
neiljerram: Ah right, I forgot that this wasn't the main Makefile.

bpf-gpl/bpf.h
neiljerram: Yes, that sounds good.  (I see from https://docs.projectcalico.org/archive/v3.17/maintenance/ebpf/enabling-bpf that we say we require 5.3 or later, or an equivalent backport.)

neiljerram (approved):
LGTM

PR #2612, opened 2020-12-03
Refactor BPF programs: split out headers and introduce context

fasaxc:
## Description

Start to refactor the BPF programs in various directions:

- Introduce a stack allocated context as a source of truth for things like packet pointers.
- Rework all handling of packet pointers to go through the context pointers.
- Establish the packet pointers via inline assembler so that we can ensure they are initialised in a verifier-friendly way.
- Introduce a utility function that establishes the packet pointers and pulls the packet if needed.  This introduces a few packet pulls that were missing before (such as on the VXLAN decap path) and requires new code paths to bail out if those pulls fail.
- Move various chunks of function out of tc.c to library headers.
- Move types into xxx_types.h headers to avoid circular includes.

## Todos
- [ ] Used exit to bail out on one of the new failure paths; may want to handle that more cleanly
- [ ] Had to add a packet length check or two.
- [ ] Maybe in a later increment: want to improve the encapsulation of packet parsing to make host endpoints easier to implement.

## Release Note

```release-note
None required
```


bpf-gpl/conntrack.h
```diff
@@ -22,15 +22,10 @@
 #include "nat.h"
 #include "bpf.h"
 #include "icmp.h"
+#include "types.h"
 
 // Connection tracking.
 
-struct calico_ct_key {
```
fasaxc, line 28: Have moved types from `xxx.h` to `xxx_types.h` so that they can be imported and used in the context structs without circular deps.

bpf-gpl/conntrack.h
```diff
@@ -518,8 +391,19 @@ static CALI_BPF_INLINE void ct_tcp_entry_update(struct tcphdr *tcp_header,
...
+	struct ct_ctx ct_lookup_ctx = {
+		.skb = tc_ctx->skb,
+		.proto	= tc_ctx->state->ip_proto,
+		.src	= tc_ctx->state->ip_src,
+		.sport	= tc_ctx->state->sport,
+		.dst	= tc_ctx->state->ip_dst,
+		.dport	= tc_ctx->state->dport,
+		.tun_ip = tc_ctx->state->tun_ip,
+	};
+	struct ct_ctx *ctx = &ct_lookup_ctx;
```
fasaxc, line 405: Found a bug here; `ctx->tcp` doesn't get populated.

bpf-gpl/tc.c
```diff
@@ -439,164 +119,77 @@ static CALI_BPF_INLINE int calico_tc(struct __sk_buff *skb)
...
-		CALI_DEBUG("Drop malformed IP packets\n");
-		goto deny;
-	} else if (ip_header->ihl > 5) {
-		/* Drop packets with IP options from/to WEP.
-		 * Also drop packets with IP options if the dest IP is not host IP
-		 */
-		if (CALI_F_WEP || (CALI_F_FROM_HEP && !rt_addr_is_local_host(ip_header->daddr))) {
-			fwd.reason = CALI_REASON_IP_OPTIONS;
-			CALI_DEBUG("Drop packets with IP options\n");
+	/* Now we've got as afar as the UDP header, check if this is one of our VXLAN packets, which we
```
neiljerram, line 142: typo: afar

bpf-gpl/tc.c
```diff
@@ -722,44 +299,53 @@ static CALI_BPF_INLINE int calico_tc(struct __sk_buff *skb)
...
 			if (!r || !(r->flags & (CALI_RT_WORKLOAD | CALI_RT_HOST))) {
-				CALI_DEBUG("Outside cluster dest %x\n", bpf_ntohl(state->post_nat_ip_dst));
-				state->flags |= CALI_ST_SKIP_FIB;
+				CALI_DEBUG("Outside cluster dest %x\n", bpf_ntohl(ctx.state->post_nat_ip_dst));
+				ctx.state->flags |= CALI_ST_SKIP_FIB;
 			}
 		}
 	}
+
+	// FIXME Not sure why this is needed
```
neiljerram, line 317: Still a valid comment?

bpf-gpl/tc.c
```diff
@@ -912,15 +514,25 @@ static CALI_BPF_INLINE struct fwd calico_tc_skb_accepted(struct __sk_buff *skb,
...
+			if (skb_refresh_validate_ptrs(ctx, ICMP_SIZE + sizeof(struct iphdr) + 8)) {
+				CALI_DEBUG("Failed to revalidate packet size\n");
 				goto deny;
 			}
 
-			l3_csum_off += sizeof(*ip_header) + sizeof(*icmp);
-			ip_header = (struct iphdr *)(icmp + 1); /* skip to inner ip */
+			/* Skip past the ICMP header and check the inner IP header. */
+			l3_csum_off += sizeof(*ctx->ip_header) + sizeof(struct icmphdr);
+			ctx->ip_header = (struct iphdr *)(ctx->icmp_header + 1); /* skip to inner ip */
```
neiljerram, line 526: Does the context definitely not need the outer IP header any more?

bpf-gpl/ut/nat_encap_test.c
```diff
@@ -21,5 +21,13 @@
...
 {
-	return vxlan_v4_encap(skb, HOST_IP, 0x02020202);
+	struct cali_tc_ctx ctx = {
+		.state = state_get(),
+		.skb = skb,
+		.fwd = {
+			.res = TC_ACT_UNSPEC,
+			.reason = CALI_REASON_UNKNOWN,
+		},
+	};
```
neiljerram, line 31: Check `.state` here, if it's really needed for the next call?

neiljerram (commented):
Just typos and minor queries, I think.

bpf-gpl/tc.c
tomastigera: I believe that this is 1:1 change of what was there originally. From this point on, NAT (iirc) is done on the inner IP header. It might be that with a bigger context and no need to store it on stack, we could have both and make some code simpler. Probably out of the scope of this refactoring.

tomastigera (commented):
LGTM - I only skimmed this, looks like a great improvements and if the tests pass and the verifies swallows it, I believe it does what it used to do :-)

One nit: my slight preference would be `ctx->iphdr` or `ctx->ip` instead of `ctx->[ip|tcp|udp]_header` just to make the text shorter. But that is just a personal preference. Otoh the whole PR makes things easier to read and get oriented.

bpf-gpl/skb.h
```diff
@@ -25,49 +25,53 @@
...
-#define skb_seen(skb) ((skb)->mark & CALI_SKB_MARK_SEEN)
+/* skb_start_ptr is equivalent to (void*)((__u64)skb->data); the read is done
+ * in a way that is acceptable to the verifier and it is done as a volatile read
+ * ensuring that a fresh value is returned and the compiler cannot
+ * reorder/recalculate the value later.
+ */
+static CALI_BPF_INLINE void *skb_start_ptr(struct __sk_buff *skb) {
+	void *ptr;
+	asm volatile (\
+		"%0 = *(u32 *)(%1 + 76)" \
```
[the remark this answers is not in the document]
fasaxc: Found a way to calculate the offset from the struct; you can pass it into the assembler fragment as a constant.  I'm not sure if you can do `"r" (skb->data)` in a more intuitive way; the BPF inline assembler is a bit of a black box with no documentation (so have to guess based on the x86 one).

bpf-gpl/tc.c
fasaxc: Yes @tomastigera is right, it's a port of the old code; I'll add a warning comment for now; I'm not sure how to rework it to avoid the problem.

bpf-gpl/ut/nat_encap_test.c
fasaxc: Doesn't look like it (and the verifier will tell us if we start using it so I'll remove for now).

bpf-gpl/tc.c
neiljerram: Thanks, fair enough.  It sounds fine to leave this until there's some actual need for the outer header.

neiljerram (approved):
LGTM, thanks!

PR #2620, opened 2020-12-07
bpf: encap nodeport return from host networked pods

tomastigera:
When a nodeport is backed by a host-networked pod, we also need to encap
the return traffic and send it back through the vxlan tunnel.  Otherwise
we always act like in DSR mode.

Closing https://github.com/projectcalico/calico/issues/4242
## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
In BPF tunnel mode, fix that connections to node ports that are backed by host-networked pods would be returned as DSR packets instead of being returned over the tunnel.  This caused AWS NLBs to fail unless the chosen host also hosted the backing pod.
```

### DIFF
--- a/bpf-gpl/tc.c
+++ b/bpf-gpl/tc.c
@@ -1234,7 +1234,14 @@ static CALI_BPF_INLINE struct fwd calico_tc_skb_accepted(struct __sk_buff *skb,
 			goto deny;
 		}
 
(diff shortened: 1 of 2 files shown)


fasaxc (approved):
LGTM

fasaxc:
@tomastigera please can you cherry-pick this to release-v3.17 once it's merged?  I think it's worth backporting.

PR #2630, opened 2020-12-09
bpf: fixup dst MAC for host networked nodeport backends

tomastigera:
Because routing happens based on the dest IP (which is at that time the
client's) before we get hands on the packet on HEP, the MAC is likely
different (in the same subnet) than the MAC of the node that forwarded
the nodeport to us. Therefore once we place the packet in VXLAN and thus
change the dest IP, we must fix up the dest MAC based on what we
recorded when we received the inbound packets.

This is a similar case to the MAC fixing for regular pods. The only
difference is that the src MAC is already set correctly if we are on the
right device.

We cannot handle the case (yet) when the routes to the original client
and to the node on the other side of the vxland tunnel use different
NICs. We would need to redirect the packet to the other NIC first.

Fixes https://github.com/projectcalico/calico/issues/4242

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/bpf-gpl/tc.c
+++ b/bpf-gpl/tc.c
@@ -1305,6 +1305,38 @@ icmp_send_reply:
 	goto deny;
 
 nat_encap:
+	/* We are about to encap return trafic that originated on the local host
+	 * namespace - a host networked pod. Routing was based on the dst IP,
+	 * which was the original client's IP at that time, not the node's that
+	 * forwarded it. We need to fix it now.
+	 */
+	if (CALI_F_TO_HEP) {
+		struct arp_value *arpv;
+		struct arp_key arpk = {
+			.ip = state->ip_dst,
+			.ifindex = skb->ifindex,
+		};
+
+		arpv = cali_v4_arp_lookup_elem(&arpk);
+		if (!arpv) {
+			CALI_DEBUG("ARP lookup failed for %x dev %d at HEP\n",
+					bpf_ntohl(state->ip_dst), arpk.ifindex);
+			/* Don't drop it yet, we might get lucky and the MAC is correct */
+		} else {
+			if (skb_shorter(skb, sizeof(struct ethhdr))) {
+				reason = CALI_REASON_SHORT;
+				goto deny;
+			}
+			struct ethhdr *eth_hdr = (void *)(long)skb->data;
+			__builtin_memcpy(&eth_hdr->h_dest, arpv->mac_dst, ETH_ALEN);
+			if (state->ct_result.ifindex_fwd == skb->ifindex) {
+				/* No need to change src MAC, if we are at the right device */
+			} else {
+				/* FIXME we need to redirect to the right device */
+			}
+		}
+	}
+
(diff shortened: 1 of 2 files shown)


bpf/ut/nat_test.go
```diff
@@ -801,6 +805,15 @@ func TestNATNodePort(t *testing.T) {
 			pktR := gopacket.NewPacket(res.dataOut, layers.LayerTypeEthernet, gopacket.Default)
 			fmt.Printf("pktR = %+v\n", pktR)
 
+			ethL := pktR.Layer(layers.LayerTypeEthernet)
+			Expect(ethL).NotTo(BeNil())
+			ethR := ethL.(*layers.Ethernet)
+			Expect(ethR).To(layersMatchFields(&layers.Ethernet{
+				SrcMAC:       macUntouched, // Source is set byt net stack and should not be touched.
```
seh, line 812:
```suggestion
				SrcMAC:       macUntouched, // Source is set by the net stack and should not be touched.
```
Or maybe just:
```suggestion
				SrcMAC:       macUntouched, // Source is set by net stack and should not be touched.
```

bpf/ut/nat_test.go
```diff
@@ -801,6 +805,15 @@ func TestNATNodePort(t *testing.T) {
...
+				DstMAC:       macScr,
```
seh, line 813: Should this be `s/macScr/macSrc/g`? If not, what does "scr" mean here?

bpf-gpl/tc.c
```diff
@@ -1305,6 +1305,38 @@ static CALI_BPF_INLINE struct fwd calico_tc_skb_accepted(struct __sk_buff *skb,
 	goto deny;
 
 nat_encap:
+	/* We are about to encap return trafic that originated on the local host
+	 * namespace - host networked pod. Routing was based on the dst IP of
```
seh, line 1309: Here, when you mention routing, do you mean the routing choice from this machine back to the machine that forwarded us this connection?

Does "dst IP of the original client" mean the destination IP address to which the client (e.g. an NLB listener) connected, or does that mean that from our perspective here our destination address for return traffic is the client's IP address?

bpf/ut/nat_test.go
tomastigera: yup typos,thanks

bpf-gpl/tc.c
tomastigera: > Here, when you mention routing, do you mean the routing choice from this machine back to the machine that forwarded us this connection?

Routing of the packet we are just processing. The packet was routed based on the dst IP, which was the original client's IP at that time. Hence the link layer picked a wrong MAC.

> Does "dst IP of the original client" mean the destination IP address to which the client (e.g. an NLB listener) connected, or does that mean that from our perspective here our destination address for return traffic is the client's IP address?

original client is whoever tried to reach the nodeport. It may be an LB, it may be an external client if the nodeport was exposed to the outter world.

fasaxc (approved):

PR #2632, opened 2020-12-10
BPF conntrack scanner moved outside of kube-proxy

tomastigera:
## Description

It did not make a good sense to have it inside except that it was easier that way. Cleaning up expired connections had nothing to do with kube-proxy and cleaning connections to service that do not exist anymore only need to consult kube-proxy.

Taking it outside allows us to have a single iterator over the conntrack map and to add new entry scanners as the requirements and usecases emerge.

The one downside is that it is harder to sync with kube-proxy now. Therefore we kick off the first sweep asap before we add kube-proxy to clean up the map asap (as we did) but the next scan is not right after the next successful sync up of proxy, but is deffered to the next timer tick. That should not be a major difference given the timer period of 10s.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```


fasaxc (approved):
LGTM, nice cleanup.

Issue #2645, opened 2020-12-18
Support CIDRs in failsafe inbound/outbound configuration


kgtw:
Currently failsafe configurations allow specifying the protocol & port, this functionality should be expanded to include a CIDR for cases when it is more desired to restrict to specific ranges. 

## Expected Behavior

Allow specifying an optional CIDR with the protocol and port. 

```
- name: FELIX_FAILSAFEINBOUNDHOSTPORTS
   value: "172.20.10.0/24:tcp:22,172.20.10.0/24:tcp:5473,172.20.10.0/24:tcp:9098,172.20.10.0/24:tcp:9099,172.20.10.0/24:tcp:50051,172.20.10.0/24:tcp:8080,172.20.10.0/24:tcp:10250"
- name: FELIX_FAILSAFEOUTBOUNDHOSTPORTS
   value: "tcp:53,udp:53,172.20.10.0/24:tcp:5473,172.20.10.0/24:tcp:443,172.20.10.0/24:tcp:9099,172.20.10.0/24:tcp:50051,172.20.10.0/24:tcp:8080,172.20.10.0/24:tcp:80"
```

## Current Behavior

As an example we are running Calico in Kubernetes and configure the failsafe ports via environment variables.

```
- name: FELIX_FAILSAFEINBOUNDHOSTPORTS
   value: "tcp:22,tcp:5473,tcp:9098,tcp:9099,tcp:50051,tcp:8080,tcp:10250"
- name: FELIX_FAILSAFEOUTBOUNDHOSTPORTS
   value: "tcp:53,udp:53,tcp:5473,tcp:443,tcp:9099,tcp:50051,tcp:8080,tcp:80"
```

When these rules are applied, they are applied with a `0/0` CIDR.

## Possible Solution

I've had a quick search and I believe these are the main areas that will need updating.
- Support CIDRs in the `ProtoPort` struct: https://github.com/projectcalico/felix/blob/master/config/config_params.go#L332-L335
- Update parsing to handle CIDRs: https://github.com/projectcalico/felix/blob/master/config/param_types.go#L306-L343
- Update the failsafe in/out chain functions to render static rules with CIDR's from the `ProtoPort` struct: https://github.com/projectcalico/felix/blob/master/rules/static.go#L400-L464

mgleung:
Hi @kgtw thanks for your PR! We always appreciate contributions. 

[...]

PR #2646, opened 2020-12-18
Allow FailsafeInboundHostPorts & FailsafeOutboundHostPorts CIDRs

kgtw:
## Description

Allows FailsafeInboundHostPorts & FailsafeOutboundHostPorts to be configured with an optional CIDR (format: protocol:cidr:port)

Fixes: https://github.com/projectcalico/felix/issues/2645

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
FailsafeInboundHostPorts & FailsafeOutboundHostPorts now support restricting to specific cidrs. New format <protocol>:<net>:<port>
```


caseydavenport:
@kgtw thanks for the PR! and sorry for the delay, I just spotted this one today.

We are actually looking at some options for this exact problem, which may impact this PR a little bit. We'll report back soon.

config/config_params.go
```diff
@@ -242,8 +242,8 @@ type Config struct {
 	PrometheusGoMetricsEnabled      bool   `config:"bool;true"`
 	PrometheusProcessMetricsEnabled bool   `config:"bool;true"`
 
-	FailsafeInboundHostPorts  []ProtoPort `config:"port-list;tcp:22,udp:68,tcp:179,tcp:2379,tcp:2380,tcp:5473,tcp:6443,tcp:6666,tcp:6667;die-on-fail"`
-	FailsafeOutboundHostPorts []ProtoPort `config:"port-list;udp:53,udp:67,tcp:179,tcp:2379,tcp:2380,tcp:5473,tcp:6443,tcp:6666,tcp:6667;die-on-fail"`
+	FailsafeInboundHostPorts  []ProtoPort `config:"port-list;0.0.0.0/0:tcp:22,0.0.0.0/0:udp:68,0.0.0.0/0:tcp:179,0.0.0.0/0:tcp:2379,0.0.0.0/0:tcp:2380,0.0.0.0/0:tcp:5473,0.0.0.0/0:tcp:6443,0.0.0.0/0:tcp:6666,0.0.0.0/0:tcp:6667;die-on-fail"`
```
mgleung, line 245: What do you think about changing the ordering to be `<protocol>:<cidr>:<port>`? I think it matches the most with what we previously had so that it doesn't change too much for our users.

dataplane/driver.go
```diff
@@ -162,19 +162,21 @@ func StartDataplaneDriver(configParams *config.Config,
...
 
-		// If wireguard is enabled, update the failsafe ports to inculde the wireguard port.
+		// If wireguard is enabled, update the failsafe ports to include the wireguard port.
 		failsafeInboundHostPorts := configParams.FailsafeInboundHostPorts
 		failsafeOutboundHostPorts := configParams.FailsafeOutboundHostPorts
 		if configParams.WireguardEnabled {
 			failsafeInboundHostPorts = make([]config.ProtoPort, len(configParams.FailsafeInboundHostPorts)+1)
 			copy(failsafeInboundHostPorts, configParams.FailsafeInboundHostPorts)
 			failsafeInboundHostPorts[len(configParams.FailsafeInboundHostPorts)] = config.ProtoPort{
+				Net: "0.0.0.0/0",
```
mgleung, line 172: This defaulting logic for the wireguard ports I think needs a rework. It looks like if we just add the ports regardless of what was set in the failsafes which worked when we were specifying just ports, but now may be too permissive. I think the logic needs a small rework to check if any of the wireguard ports have been set, and if they have not, then add a rule that allows the wireguard port over all IPs. This should allow for anyone to set their specific wireguard overrides (with the specific node IPs of our cluster) and prevent the wireguard port on all nodes to be open. WDYT?

config/config_params.go
kgtw: We could do, the reason why I did this format was that it made it backwards compatible with the current structure of `<protocol>:<port>`.

See: https://github.com/projectcalico/felix/pull/2646/files#diff-2e8a8c60cee2386f6eb73e92222a9ba3ebb38d34e53edbed3b8fcb6d259c0214L317

If we switch to `<protocol>:<cidr>:<port>` it will be a breaking change. I do however think this structure is more syntactically correct and am happy to change the ordering if this is preferred.

dataplane/driver.go
kgtw: Yes, this makes sense!

config/config_params.go
caseydavenport: I don't think it's a breaking change if we implement it correctly - the field field could just support two formats:

- <protocol>:<port>
- <protocol>:<cidr>:<port>

There's no ambiguity there, because if three segements are given then it is clearly the latter, and if 2 segments are given then it must be the "old" style.

Does that seem right?

kgtw: Ah yes, you're right. Seems I was having a brain-fart moment when replying. Will switch to the new structure.

dataplane/driver.go
kgtw: I've added this check in https://github.com/projectcalico/felix/pull/2646/commits/30a6abbf9dae364754729ea02fbbd204c87a3de0 but I haven't had a chance to test this change. I'm also struggling to find a place where we can test this easily, as it stands that dataplane driver doesn't have a test suite that one can easily utilise.

mgleung: Ah yeah, this one is a little difficult given how we haven't really created a set of dataplane tests for this. Probably the easiest way to quickly verify is to add something like a `Debug` statement or something where the failsafes get applied and then build a new node image with these felix changes. Then you can run the changed node in a cluster quickly, while passing the appropriate configuration, and verify if your debug statement shows the appropriate failsafes when you check the pod logs.

The proper solution is honestly a little more work than I would want for a change as small as this, which would be to create the suite of dataplane tests but even that wouldn't be very easy for this test since there's no easy way to inspect the config that the dataplane is created with without writing a bunch of new methods just to get that information.

kgtw: Hi @mgleung I've tested this within our cluster by using debug logs as you suggested.

Default wireguard config, if a more specific entry for wireguard wasn't found in the failsafe config
```
2021-03-04 14:34:36.501 [DEBUG][52] felix/driver.go 184: Adding permissive FailsafeInboundHostPorts for wireguard
2021-03-04 14:34:36.501 [DEBUG][52] felix/driver.go 207: Adding permissive FailsafeOutboundHostPorts for wireguard

-A cali-failsafe-in -p udp -m comment --comment "cali:sePiEYnWQoE34rxm" -m multiport --dports 51820 -j ACCEPT
-A cali-failsafe-out -p udp -m comment --comment "cali:cKMbiSScdOAMPjj7" -m multiport --dports 51820 -j ACCEPT
```

When a specific entry for wireguard is found in the failsafe config
```
2021-03-04 14:41:30.111 [DEBUG][52] felix/driver.go 176: FailsafeInboundHostPorts is already configured for wireguard net="172.0.0.0/8" port=0xca6c protocol="udp"
2021-03-04 14:41:30.111 [DEBUG][52] felix/driver.go 199: FailsafeOutboundHostPorts is already configured for wireguard net="172.0.0.0/8" port=0xca6c protocol="udp"

-A cali-failsafe-in -s 172.0.0.0/8 -p udp -m comment --comment "cali:KoXgB1rx1y51vKyk" -m multiport --dports 51820 -j ACCEPT
-A cali-failsafe-out -d 172.0.0.0/8 -p udp -m comment --comment "cali:2KMZG0hyUESpYCVf" -m multiport --dports 51820 -j ACCEPT

mgleung (approved):
LGTM

PR #2656, opened 2021-01-16
BPF fs root override

tomastigera:
## Description

    calico-bpf can use alternative bpffs root
    
    For instance, EKS Bottle rocket has the bpffs mounted at
    /.bottlerocket/rootfs/sys/fs/bpf

    bpf: bpf.FSRoot is a configurable bpffs root

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/bpf/maps.go
+++ b/bpf/maps.go
@@ -29,6 +29,8 @@ import (
 
 type IteratorAction string
 
+var FSRoot = "/sys/fs/bpf"
+
 const (
 	IterNone   IteratorAction = ""
 	IterDelete IteratorAction = "delete"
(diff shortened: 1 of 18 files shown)


bpf/maps.go
```diff
@@ -278,7 +282,7 @@ func (b *PinnedMap) Open() error {
 		return err
 	}
 	// FIXME hard-coded dir
```
fasaxc, line 284: You're fixing this fixme

fasaxc (approved):
Generally prefer to avoid globals but feels like a step in the right direction to have one place where we define that prefix

caseydavenport:
@tomastigera is this still a draft? And do we need it for v3.18?

tomastigera:
I am closing this because the only usecase is Bottlerocket and we know the right way how to do it now with the current tool.

PR #2676, opened 2021-02-16
Do not forward incoming VXLAN packets.

fasaxc:
## Description

Disable the FIB when we detect a VXLAN packet to the local host.  The FIB lookup is not helpful and on that code path the FIB input datastructure hasn't be initialised so we end up sending the packet to the default gateway.

Bug was introduced by interaction between #2666 and #2669.  Only seems to show up in E2Es; perhaps the FIB is failing for another reason in the FVs?

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```


neiljerram (approved):

marvin-tigera:

PR #2682, opened 2021-02-17
Ignore log rules in BPF mode.

fasaxc:
## Description

Log rules are not suported but they were causing panics instead of being ignored.

Fixes https://github.com/projectcalico/calico/issues/4389

## Todos
- [ ] Backport
- [x] Release note

## Release Note

```release-note
Fix that, in eBPF mode, a Log rule would result in an error instead of being ignored.  Log rules are not supported but they should be ignored, not cause a failure.
```


neiljerram (approved):
LGTM

marvin-tigera:

PR #2686, opened 2021-02-19
For eBPF data plane, respect externalTrafficPolicy for LB services

caseydavenport:
## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/bpf/proxy/syncer.go
+++ b/bpf/proxy/syncer.go
@@ -567,6 +567,9 @@ func (s *Syncer) applyDerived(
 
 	skey = getSvcKey(sname, getSvcKeyExtra(t, sinfo.ClusterIP().String()))
 	switch t {
+	case svcTypeLoadBalancer:
+		// Handle LB services the same as NodePort type.
+		fallthrough
 	case svcTypeNodePort:
 		if sinfo.OnlyNodeLocalEndpoints() {
 			count = local // use only local eps


tomastigera (approved):
I believe, this should work. Do we have an e2e test for this feature?

caseydavenport:
>  Do we have an e2e test for this feature?

We don't at the moment, it's only tested in the node STs. Do we have FV tests in this repo for the eBPF proxy?

tomastigera:
> We don't at the moment, it's only tested in the node STs. Do we have FV tests in this repo for the eBPF proxy? That's where I'd rather put it, but if needed I can devise an e2e.

This what I think comes closest to it https://github.com/projectcalico/felix/blob/master/bpf/proxy/proxy_test.go

I am not so worried that the NAT table is programmed incorrectly if we get LB IPs, I was more looking for k8s conformance test that would verify that ours and theirs interpretation is the same. Perhaps that comes with upgrading e2es to 1.20?

caseydavenport:
Ah right. Well, I am fairly certain there are upstream tests for externalTrafficPolicy on LB services, just not any specifically targeting our service advertisement features.

salanki:
@caseydavenport: Is it possible that this causes a behavior where `externalTrafficPolicy: Local` LB IPs are not reachable from within the cluster? They used to be in 3.18.

salanki:
I have confirmed that I can reach `externalTrafficPolicy: Local` LoadBalancer services from inside the cluster when this commit is reverted.

PR #2689, opened 2021-02-24
Fix that VXLAN tests only ever tested the WorkloadIP case.

fasaxc:
## Description

The loop variable was being used in a func(), resulting in both instances of the test referring to the same variable.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```


neiljerram (approved):
LGTM

PR #2690, opened 2021-02-26
Service loop prevention for external IPs

neiljerram:
## Description
As it says in service_loop_mgr.go:
```
// The service loop manager maintains an iptables chain in the filter table whose purpose is to
// prevent forwarding to IPs within known service CIDRs.  Traffic that arrives on the node with such
// IPs is supposed to be DNAT'd to an endpoint pod IP:port by kube-proxy iptables or IPVS rules.  If
// that doesn't happen, it means that target service IP:port isn't actually in use; and in that case
// it's better for us to drop it, to avoid a possible routing loop between this node and this node's
// default gateway.  The specific loop-generating scenario is when Calico is configured to advertise
// service CIDRs and IPs over BGP: then the default gateway will have a route back to this node, for
// the service CIDR, and there could be a loop if we allowed non-existent service traffic to be
// forwarded on from here.
```

#2532 previously blocked looping through IPs in service cluster CIDR ranges.  This PR extends that to block looping through External IPs.

## Release Note
```release-note
Prevent looping through a service external IP when traffic is sent to the wrong port for the external IP.
```


caseydavenport (commented):
It LGTM, we'll want to do this for ServiceLoadBalancerIPs as well though

neiljerram:
> It LGTM, we'll want to do this for ServiceLoadBalancerIPs as well though

Could we then merge this one as is, and have someone else use it as a model for ServiceLoadBalancerIPs?  I haven't carefully reviewed the scenario for load balance IPs, and I'd prefer not to take that on now.

caseydavenport (approved):

PR #2696, opened 2021-03-04
CORE-4739 | Install blackhole routes in VXLAN mode

electricjesus:
## Description
- Route table can now handle blackhole routes but does it in a way that is separate to other route types. The main rationale is that the current RouteTable delta tracking heavily relies on CIDR as a key and does not allow dupes.
- Using the new route table handling feature for blackhole routes, vxlan manager generates route target requests based on IPAM blocks. Doesn't make a lot of sense to do the same generation with WorkloadIPs (IMO)

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/dataplane/linux/vxlan_mgr.go
+++ b/dataplane/linux/vxlan_mgr.go
@@ -54,14 +54,15 @@ type vxlanManager struct {
 	sync.Mutex
 
 	// Our dependencies.
-	hostname          string
-	routeTable        routeTable
-	noEncapRouteTable routeTable
+	hostname            string
+	routeTable          routeTable
+	blackholeRouteTable routeTable
+	noEncapRouteTable   routeTable
 
 	// Hold pending updates.
-	routesByDest map[string]*proto.RouteUpdate
-	ipamBlocks   map[string]*proto.RouteUpdate
-	vtepsByNode  map[string]*proto.VXLANTunnelEndpointUpdate
+	routesByDest    map[string]*proto.RouteUpdate
+	localIPAMBlocks map[string]*proto.RouteUpdate
+	vtepsByNode     map[string]*proto.VXLANTunnelEndpointUpdate
 
 	// Holds this node's VTEP information.
 	myVTEP *proto.VXLANTunnelEndpointUpdate
@@ -77,7 +78,6 @@ type vxlanManager struct {
 	ipSetMetadata     ipsets.IPSetMetadata
 	externalNodeCIDRs []string
 	vtepsDirty        bool
-	ipamBlocksDirty   bool
 	nlHandle          netlinkHandle
 	dpConfig          Config
 	noEncapProtocol   int
@@ -85,6 +85,10 @@ type vxlanManager struct {
 	noEncapRTConstruct func(interfacePrefixes []string, ipVersion uint8, vxlan bool, netlinkTimeout time.Duration,
 		deviceRouteSourceAddress net.IP, deviceRouteProtocol int, removeExternalRoutes bool) routeTable
 }
+
+const (
+	defaultVXLANProto = 80
+)
 
 func newVXLANManager(
 	ipsetsDataplane ipsetsDataplane,
@@ -95,9 +99,26 @@ func newVXLANManager(
 ) *vxlanManager {
 	nlHandle, _ := netlink.NewHandle()
 
+	blackHoleProto := defaultVXLANProto
+	if dpConfig.DeviceRouteProtocol != syscall.RTPROT_BOOT {
+		blackHoleProto = dpConfig.DeviceRouteProtocol
+	}
+
+	brt := routetable.New(
+		[]string{routetable.InterfaceNone},
+		4,
+		false,
+		dpConfig.NetlinkTimeout,
+		dpConfig.DeviceRouteSourceAddress,
+		blackHoleProto,
+		false,
+		0,
+		opRecorder,
+	)
+
 	return newVXLANManagerWithShims(
 		ipsetsDataplane,
-		rt,
+		rt, brt,
 		deviceName,
 		dpConfig,
 		nlHandle,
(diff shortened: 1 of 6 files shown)


fasaxc (changes requested):
As dissussed, think you're going to revise this to handle the cleanup case (maybe by adding a proto value to the routes).  I spotted one issue in the route calculation where it'll break once you've got the capability to clean up routes.

dataplane/linux/vxlan_mgr.go
```diff
@@ -228,6 +239,27 @@ func (m *vxlanManager) GetRouteTableSyncers() []routeTableSyncer {
 	return rts
 }
 
+func (m *vxlanManager) vxlanRoutesFromIPAMBlocks() []routetable.Target {
+	rtt := make([]routetable.Target, 0)
+	if m.ipamBlocksDirty {
```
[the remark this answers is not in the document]
electricjesus: You were right, I actually didn't need to add in this additional dirty flag.. used the existing flag instead so it is picked up properly on next CompleteDeferredWork

dataplane/linux/vxlan_mgr.go
```diff
@@ -155,6 +172,13 @@ func (m *vxlanManager) OnUpdate(protoBufMsg interface{}) {
 			m.routesByDest[msg.Dst] = msg
 			m.routesDirty = true
 		}
+
+		// Process IPAM blocks that aren't associated to a single or /32 local workload
+		if msg.Type == proto.RouteType_LOCAL_WORKLOAD && msg.IpPoolType == proto.IPPoolType_VXLAN && !msg.LocalWorkload {
```
fasaxc, line 177: Just thought of a corner case.  If you have a /32 block then it can oscillate between being LocalWorkload=true and LocalWorkload=false.

Fix: add an else block to this that does 
```
else if _, ok := m.ipamBlocks[msg.Dst]; ok {
 logrus.WithField("msg", msg).Info("VXLAN data plane IPAM block changed to something else")
  delete(m.ipamBlocks, msg.Dst)
￼  m.routesDirty = true
}
```

dataplane/linux/vxlan_mgr.go
```diff
@@ -53,12 +53,14 @@ type vxlanManager struct {
...
-	routeTable        routeTable
-	noEncapRouteTable routeTable
+	hostname            string
+	routeTable          routeTable
+	blackholeRouteTable routeTable
+	noEncapRouteTable   routeTable
 
 	// Hold pending updates.
 	routesByDest map[string]*proto.RouteUpdate
+	ipamBlocks   map[string]*proto.RouteUpdate
```
fasaxc, line 63: `localIPAMBlocks` ?  Just to make it clear that we're filtering?

routetable/route_table.go
```diff
@@ -230,7 +231,7 @@ func NewWithShims(
 	conntrack conntrackIface,
 	timeShim timeshim.Interface,
 	deviceRouteSourceAddress net.IP,
-	deviceRouteProtocol int,
+	deviceRouteProtocol, routeProtocol int,
```
fasaxc, line 234: I don't think this new parameter gets used?

Given that we've got a route table for blackhole routes and one for non-blackhole I think each route table only needs one protocol parameter?  (and maybe we should rename the parameter to be more generic rather than the old deviceRouteProtocol?)

fv/vxlan_test.go
```diff
@@ -143,6 +143,26 @@ var _ = infrastructure.DatastoreDescribe("_BPF-SAFE_ VXLAN topology before addin
...
+				It("should have some blackhole routes installed", func() {
+					if routeSource == "WorkloadIPs" {
+						Skip("not applicable for workload ips")
+						return
+					}
+
+					nodes := []string{
+						"blackhole 10.65.0.0/26",
+						"blackhole 10.65.1.0/26",
+						"blackhole 10.65.2.0/26",
```
fasaxc, line 155: Make sure you check the `proto` part of the route

dataplane/linux/vxlan_mgr.go
```diff
@@ -181,10 +218,17 @@ func (m *vxlanManager) OnUpdate(protoBufMsg interface{}) {
 func (m *vxlanManager) deleteRoute(dst string) {
 	_, exists := m.routesByDest[dst]
 	if exists {
+		logrus.Info("deleting route dst ", dst)
```
caseydavenport, line 221: Probably should be a debug level log

dataplane/linux/vxlan_mgr.go
```diff
@@ -228,6 +272,25 @@ func (m *vxlanManager) GetRouteTableSyncers() []routeTableSyncer {
 	return rts
 }
 
+func (m *vxlanManager) blackholeRoutes() []routetable.Target {
+	var rtt []routetable.Target
+	for dst := range m.localIPAMBlocks {
+		cidr, err := ip.CIDRFromString(dst)
+		if err != nil {
+			logrus.WithError(err).Debug(
```
caseydavenport, line 280: Should probably upgrade this to a warning log?

dataplane/linux/vxlan_mgr.go
```diff
@@ -228,6 +272,25 @@ func (m *vxlanManager) GetRouteTableSyncers() []routeTableSyncer {
...
+				"Error processing IPAM block CIDR: ", dst,
+			)
+			continue
+		}
+		rtt = append(rtt, routetable.Target{
+			Type: routetable.TargetTypeBlackhole,
+			CIDR: cidr,
+		})
+	}
+	logrus.Info("calculated blackholes ", rtt)
```
caseydavenport, line 290: And downgrade this to debug

caseydavenport (approved):
Finally got around to looking at this and it looks pretty sensible to me! Just a few log level comments, but otherwise I think it looks good.

Worth checking if @fasaxc has other comments pending before merging though.

fv/vxlan_test.go
```diff
@@ -65,7 +66,7 @@ var _ = infrastructure.DatastoreDescribe("_BPF-SAFE_ VXLAN topology before addin
 					// Install a default profile that allows all ingress and egress, in the absence of any Policy.
 					infra.AddDefaultAllow()
 
-					// Wait until the vxlan device appears.
+					// Wait until the vxlan deShouldvice appears.
```
robbrockbank, line 69: typo

fv/vxlan_test.go
```diff
@@ -133,6 +134,10 @@ var _ = infrastructure.DatastoreDescribe("_BPF-SAFE_ VXLAN topology before addin
 					}
 					for _, felix := range felixes {
 						felix.Stop()
+
+						// see if blackhole routes go away with felix
+						o, _ := felix.ExecOutput("ip", "r", "s", "type", "blackhole")
```
fasaxc, line 139: After calling Stop, any Exec will fail because the container is down so this test isn't useful.  Suggest you just remove it.

fasaxc (changes requested):

fasaxc (approved):

PR #2697, opened 2021-03-04
[v3.16] Disable Semaphore's copy-on-write docker mount.

fasaxc:
## Description

Backport semaphore copy-on-write fix.

The copy-on-write docker mount is intended to provide a cache
of useful images but it doesn't cover any images that we use.

In addition, as we churn through docker containers in the FV tests
the copy-on-write FS grows and grows until it fills the root FS.
It's actually better for us to simly write to the root FS directly
since deleted files on the root FS can actually be reclaimed.
## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -48,6 +48,12 @@ blocks:
   task:
     prologue:
       commands:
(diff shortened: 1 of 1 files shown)


mgleung (approved):
LGTM

PR #2710, opened 2021-03-11
Handle failure of netlink channel.  Reconnect and resync

fasaxc:
## Description

When I added the FilterUpdates mechanism, we lost error checking on the channel read.  Add that back and plumb through a failure reporting callback so that we shut down/restart gracefully instead of panicking.

Handle the failure to read gracefully and reconnect to netlink.  Initially, I reinstated the old behaviour that restarted felix but that made a consistent CI failure.  Let's go the whole hog and do the reconnection in-band...

## Todos
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
Fix that, after a netlink read failure, Felix would tight loop reading from a closed channel.  Restart the event poll in that case.
```

### DIFF
--- a/ifacemonitor/ifacemonitor_test.go
+++ b/ifacemonitor/ifacemonitor_test.go
@@ -46,6 +46,7 @@ type netlinkTest struct {
 	linkUpdates    chan netlink.LinkUpdate
 	routeUpdates   chan netlink.RouteUpdate
 	userSubscribed chan int
+	cancel         chan struct{}
 
 	nextIndex int
 	links     map[string]linkModel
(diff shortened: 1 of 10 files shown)


ifacemonitor/ifacemonitor_test.go
```diff
@@ -405,7 +414,13 @@ var _ = Describe("ifacemonitor", func() {
 				regexp.MustCompile("dummy"),
 			},
 		}
-		im = ifacemonitor.NewWithStubs(config, nl, resyncC)
+		fatalErrC = make(chan struct{})
+		fatalErrCallback := func(err error) {
+			log.WithError(err).Info("Fatal error reported")
+			close(fatalErrC)
+			panic(errFatal)
```
neiljerram, line 421: Why do we signal via channel and also by panicking here?

neiljerram (commented):
An initial comment; still getting my head fully around this.  Also noticed this in the CI:
```
felix/daemon.go 386: Shutting down due to fatal error error=failed to read from netlink: %!w(<nil>)
```
so probably a logging call that needs "f" added.

neiljerram:
Also unusual log formatting here:
```
felix-0-23996-224-felixfv[stderr] time="2021-03-11T12:11:29Z" level=error msg="FilterUpdates: route input channel closed." source="update_filter.go:113"
```

ifacemonitor/ifacemonitor_test.go
fasaxc: The channel signals to the test code.  The panic breaks out of the goroutine (and we handle the panic later)

fasaxc:
> Also unusual log formatting here:

Imported the wrong log package by accident

fasaxc:
> so probably a logging call that needs "f" added.

Fixed

neiljerram (approved):
LGTM, let's see if CI is happy now.

PR #2724, opened 2021-03-22
Set interface IP as source IP for ICMP replies

sridhartigera:
We use HOST_IP for all ICMP replies irrespective of the incoming interface. This results in the ICMP reply getting dropped. Hence we use the Interface IP of the ingress interface for ICMP replies. 

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```


dataplane/linux/bpf_ep_mgr.go
```diff
@@ -1344,3 +1370,23 @@ func FindJumpMap(ap *tc.AttachPoint) (mapFD bpf.MapFD, err error) {
...
 	return 0, errors.New("failed to find TC program")
 }
+
+func (m *bpfEndpointManager) getInterfaceIP(ifaceName string) (net.IP, error) {
+	if ip, ok := m.ifaceToIpMap[ifaceName]; ok {
+		return ip, nil
+	}
+	intf, err := net.InterfaceByName(ifaceName)
+	if err != nil {
+		return net.IPv4zero, err
```
tomastigera, line 1380:
```suggestion
		return nil, err
```

perhaps?

dataplane/linux/bpf_ep_mgr.go
```diff
@@ -1344,3 +1370,23 @@ func FindJumpMap(ap *tc.AttachPoint) (mapFD bpf.MapFD, err error) {
...
+	}
+	addrs, err := intf.Addrs()
+	if err != nil {
+		return net.IPv4zero, err
+	}
+	for _, addr := range addrs {
+		if ipv4Addr := addr.(*net.IPNet).IP.To4(); ipv4Addr != nil {
+			return ipv4Addr, nil
+		}
+	}
```
tomastigera, line 1390: I wonder if we want this part. If we can safely fall back to this, why not to read this address at program load time :thinking: I am not sure. I guess it does not hurt.

tomastigera (commented):
looks good, one nit one question

dataplane/linux/bpf_ep_mgr.go
sridhartigera: ya I agree. We can just mark the interface dirty and return.

dataplane/linux/bpf_ep_mgr.go
```diff
@@ -288,6 +292,22 @@ func (m *bpfEndpointManager) OnUpdate(msg interface{}) {
 	}
 }
 
+func (m *bpfEndpointManager) onInterfaceAddrsUpdate(update *ifaceAddrsUpdate) {
+	if update.Addrs != nil && update.Addrs.Len() > 0 {
```
fasaxc, line 296: What if the len drops to zero?

dataplane/linux/bpf_ep_mgr.go
```diff
@@ -288,6 +292,22 @@ func (m *bpfEndpointManager) OnUpdate(msg interface{}) {
...
+		log.Debugf("Interface %+v received address update %+v", update.Name, update.Addrs)
+		update.Addrs.Iter(func(item interface{}) error {
```
fasaxc, line 298: There could be more than one IP here; we should pick one deterministically.  Maybe put them in a slice and sort the slice and pick the first one?  The algorithm isn't important as long as we always pick the same one.

dataplane/linux/bpf_ep_mgr.go
```diff
@@ -288,6 +292,22 @@ func (m *bpfEndpointManager) OnUpdate(msg interface{}) {
...
+			ip := net.ParseIP(item.(string))
+			if ip.To4() != nil {
+				m.ifaceToIpMap[update.Name] = ip
```
fasaxc, line 301: Can get spurious updates so it is worth avoiding marking the interface dirty if the IP hasn't changed

fasaxc (changes requested):

dataplane/linux/bpf_ep_mgr.go
sridhartigera: Length dropping to zero is equivalent to IP deletion? In that case, make the interfaces dirty?

fasaxc: Yes, it means all the IPs are gone (or the interface itself is gone)

dataplane/linux/bpf_ep_mgr.go
sridhartigera: Can we just pick the first IP?

dataplane/linux/bpf_ep_mgr.go
tomastigera: imo no need to make it dirty, program would not apply until there is an IP. Perhaps would not even get here if no IP was considered as device removed. Will not having an IP trigger removal?

dataplane/linux/bpf_ep_mgr.go
```diff
@@ -1344,3 +1392,23 @@ func FindJumpMap(ap *tc.AttachPoint) (mapFD bpf.MapFD, err error) {
...
+	}
+	intf, err := net.InterfaceByName(ifaceName)
+	if err != nil {
+		return nil, err
+	}
+	addrs, err := intf.Addrs()
+	if err != nil {
+		return nil, err
+	}
+	for _, addr := range addrs {
```
fasaxc, line 1408: Should sort these IPs too

fasaxc (changes requested):
Just one minor issue

dataplane/linux/bpf_ep_mgr.go
```diff
@@ -1344,3 +1392,27 @@ func FindJumpMap(ap *tc.AttachPoint) (mapFD bpf.MapFD, err error) {
...
+	intf, err := net.InterfaceByName(ifaceName)
+	if err != nil {
+		return nil, err
+	}
+	addrs, err := intf.Addrs()
+	if err != nil {
+		return nil, err
+	}
+	sort.Slice(addrs, func(i, j int) bool {
+		return bytes.Compare(addrs[i].(*net.IPNet).IP, addrs[j].(*net.IPNet).IP) < 0
```
fasaxc, line 1409: The cast makes me wonder if there could be a different type of address in some corner case.  Probably best to filter those out, just in case!

fasaxc (approved):
LGTM

PR #2752, opened 2021-04-03
Update IPIP FV connectivity behavior expectation

doublek:
## Description

In projectcalico/libcalico-go#1387 we've changed how we populate IP address information in the `HostIPKey`.  This means that workload connectivity continues to exist despite the removal of BGP. This PR updates the FV to match the new behavior. 
If this isn't the expected behavior then this PR isn't valid and the behavior should be fixed.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```


doublek:
Ah looks like the behavior with etcd is different.

fasaxc:
Think this was addressed by #2755

PR #2755, opened 2021-04-07
Fix for FV failures in l3_route_resolver

sridhartigera:
After enabling the feature to use node's Internal IP/ External IP as fallback HostIP, the IPIP mode FVs were failing when we turn off BGP.
In case of iptables mode, the hostIPs are added to the IPSet even after disabling BGP. As a result, the IPIP traffic is allowed. In case of BPF, this behavior was a bit different. Once we remove the BGP, felix gets RouteRemove Messages for the host and its workload. As a result, the IPIP traffic is dropped. The changes as part of this PR, handles the node addresses and does not send RouteRemove to bpff route manager. Hence the behavior is similar between Iptables and BPF mode. This also makes the behavior with Vxlan the same as IPIP.
## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```


calc/l3_route_resolver.go
```diff
@@ -343,7 +343,17 @@ func (c *L3RouteResolver) OnResourceUpdate(update api.Update) (_ bool) {
 				Addr: ip.FromCalicoIP(*ipv4).(ip.V4Addr),
 				CIDR: ip.CIDRFromCalicoNet(*caliNodeCIDR).(ip.V4CIDR),
 			}
+		} else if len(node.Spec.Addresses) > 0 {
+			ipv4, caliNodeCIDR, err := cnet.ParseCIDROrIP(node.Spec.Addresses[0].Address)
```
fasaxc, line 347: Will the `0` address always be the right one?  I thought the logic in libcalico-go was more complex to find the "right" IP?

Worth commenting here that we're mainly doing this to be consistent with the host metadata IP so that BPF mode and iptables mode behave the same in this corner case.

fv/ipip_test.go
```diff
@@ -323,7 +323,10 @@ var _ = infrastructure.DatastoreDescribe("_BPF-SAFE_ IPIP topology before adding
 	})
 
 	Context("after removing BGP address from nodes", func() {
-		// Simulate having a host send IPIP traffic from an unknown source, should get blocked.
+		// After removing BGP address from nodes, the nodeAddress should be used.
+		// Hence we expect the IPSets to have the HostIPs of Felix[0], Felix[1] in case of IPTables
+		// In BPF mode, the route should not be removed for Felix[0].IP and workload IP
+		// Connectivity between workloads should still work.
```
fasaxc, line 329: The test is no longer verifying the IPIP source policing and I think we need to verify that.  Could we delete the Node entirely, or change its IP in the datamodel to be wrong so that the IP set no longer has the right value in it (for example)?

fasaxc (changes requested):

fv/ipip_test.go
```diff
@@ -327,6 +327,14 @@ var _ = infrastructure.DatastoreDescribe("_BPF-SAFE_ IPIP topology before adding
 		BeforeEach(func() {
 			ctx, cancel := context.WithTimeout(context.Background(), 20*time.Second)
 			defer cancel()
+			if bpfEnabled {
+				infra.RemoveNodeAddresses(felixes[0])
```
fasaxc, line 331: Why this difference?

fasaxc (changes requested):
One query

fv/ipip_test.go
sridhartigera: In case of bpf, we send felixes[0].name in the list options. My understanding is that list nodes will return the node object corresponding the felixes[0] and we disable BGP in felixes[0]. https://github.com/projectcalico/felix/blob/master/fv/ipip_test.go#L331

fasaxc (approved):

PR #2767, opened 2021-04-10
BPF: RPF check in EKS

tomastigera:
## Description

    bpf: EKS workaround iptables rule for BPFExtToServiceMark
    
    EKS tries to restore the 0x80 mark from connection. This removes the
    BPFExtToServiceMark we set on egress from pods if we do not set it on
    the connection in host namespace.
    
    Preferably, EKS would restore the mark only if the mark was set and
    would leave the mark untouched otherwise.

    bpf: BpfExtToServiceConnmark - EKS rpf check
    
    env FELIX_BpfExtToServiceConnmark allows us to set a mark if a packet is
    DNATed at the host interace (therefore is not comming from a pod), which
    allows us to influence routing if the packet would go through the host
    namespace - that means it is not forwarded to another interface
    directly. We set a flag at the connection so that we mark all such
    packets in both directions.
    
    EKS uses a mark to trace nodeport connections to route them to correct
    interface on the way back (the main ENI) in case the backend pod is
    associated with a secondary ENI. Setting the mark allows us to follow
    the same routing and pass RPF check.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
Introduce a new Felix configuration parameter that can be set to mark packets from external hosts to services.  This is useful for working around an incompatibility with the Amazon VPC CNI's source based routing rules.  In EKS, BPFExtToServiceConnmark should be set to 0x80.
```


dataplane/linux/bpf_ep_mgr.go
```diff
@@ -749,6 +752,7 @@ func (m *bpfEndpointManager) attachWorkloadProgram(ifaceName string, endpoint *p
 	//   when we cannot encap the packet - non-GSO & too close to veth MTU
 	ap.TunnelMTU = uint16(m.vxlanMTU - 50)
 	ap.IntfIP = calicoRouterIP
+	ap.ToHostMark = uint32(m.bpfToHostMark)
```
fasaxc, line 755: I think marking all traffic from a workload will be a problem.  For secondary ENI pods, workload traffic that is _not_ responding to a  nodeport _should_ go over the secondary ENI but I think this mark will force it over the primary ENI.

bpf-gpl/tc.c
```diff
@@ -784,6 +784,23 @@ static CALI_BPF_INLINE struct fwd calico_tc_skb_accepted(struct cali_tc_ctx *ctx
 				CALI_DEBUG("rt found for 0x%x local %d\n",
 						bpf_ntohl(state->post_nat_ip_dst), !!cali_rt_is_local(rt));
 
+				if (cali_rt_is_workload(rt) && state->tun_ip == 0) {
```
fasaxc, line 787: Need to check if the rt is to a _local_ workload?

dataplane/linux/int_dataplane.go
```diff
@@ -1263,6 +1265,20 @@ func (d *InternalDataplane) setUpIptablesBPF() {
...
+
+	if d.config.BPFToHostMark != 0 {
+		mark := uint32(d.config.BPFToHostMark)
+		for _, t := range d.iptablesMangleTables {
+			t.InsertOrAppendRules("PREROUTING", []iptables.Rule{{
+				Match: iptables.Match().MarkMatchesWithMask(
+					tc.MarkSeen|mark,
+					tc.MarkSeenMask|mark,
+				),
+				Comment: []string{"Mark connections with ToHostMak"},
```
fasaxc, line 1277: Typo: Mak

config/config_params.go
```diff
@@ -124,6 +124,7 @@ type Config struct {
 	BPFKubeProxyIptablesCleanupEnabled bool           `config:"bool;true"`
 	BPFKubeProxyMinSyncPeriod          time.Duration  `config:"seconds;1"`
 	BPFKubeProxyEndpointSlicesEnabled  bool           `config:"bool;false"`
+	BPFToHostMark                      int            `config:"int;0"`
```
fasaxc, line 127: Should try to make the name capture what this setting does once we pin that down `BPFExtToServiceConnmark` or something?

fasaxc (changes requested):

bpf-gpl/tc.c
tomastigera: yeah, this was taken a bit blindly from the ARP branch, should be for local only. And in fact, we want to do that also on the node where it was forwarded to.

config/config_params.go
tomastigera: Yup. Was originally meant to be a generic ToHostMark ... until it turned out it does not work ;-)

fasaxc (approved):
LGTM.  I think the long-winded name is justified since it's a corner-case parameter that few should need to set.

PR #2770, opened 2021-04-13
Fix handling of overlapping IPs in kube-proxy implementation

fasaxc:
## Description

This fixes the kube-proxy syncer to recover after seeing service VIPs overlap.  Previously, after one of the overlapping services was removed, the dataplane would be left in an inconsistent state because the deletion would trigger cleanup of the single entry in the dataplane,  leaving the external IP out of the BPF maps.

This PR:

- Introduces a new struct `CachingMap`, which maintains a cache of the state of an underlying BPF map and allows it to be resynced against our desired state.
- Simplifies the logic in the syncer, removing its incorrect delta tracking.  Instead, we do a full recalculation each time and blast it into a CachingMap.  The CachingMap then dedupes and avoids writing values that were already present in the dataplane.

The new implementation is slower at high scale since it does extra recalculation; we may want to add some variant of delta tracking back in.

## Todos
- [x] Release note

## Release Note

```release-note
In BPF mode: Fix that changing the type of a service or having multiple services with overlapping external IPs would result in incorrect load balancing, even after the overlap was resolved.
```


bpf/cachingmap/caching_map_test.go
```diff
@@ -0,0 +1,369 @@
...
+	Expect(err).NotTo(HaveOccurred())
+	Expect(mockMap.Contents).To(Equal(map[string]string{
+		string([]byte{1, 1}): string([]byte{1, 2, 4, 4}),
+	}))
+
+	// And delete.
+	mockMap.DeleteErr = ErrFail
+	cm.DeleteAll()
+	err = cm.ApplyAllChanges()
+	Expect(err).To(HaveOccurred())
```
tomastigera, line 85: deleting shouldn't happen here?

we can still have half updated map if error happens, right? But if we reload the cache, the desired state is still there and the operation that faild would be still pending. Do I get it right?

bpf/cachingmap/caching_map.go
```diff
@@ -0,0 +1,345 @@
...
+	if currentVal == nil {
+		// We don't think this value is in the dataplane so clear any pending delete.
+		logrus.Debug("DeleteDesiredState: Key not in dataplane, ignoring.")
+		c.pendingDeletions.Delete(k)
+		return
+	}
+	c.pendingDeletions.Set(k, currentVal)
+}
+
+func (c *CachingMap) DeleteAll() {
```
tomastigera, line 166: nit: the DeleteAll() name is a bit confusing. Could it be something like ResetDesiredState, I am not sure.

tomastigera (commented):
still reviewing ... :hourglass_flowing_sand:

bpf/cachingmap/caching_map_test.go
fasaxc: Yeah, if the delete fails then we don't update pendingDeletions or cacheOfDataplane so we'll try again next time.

Reloading the cache should always bring cacheOfDataplane and the pendingXXX sets into sync.

bpf/cachingmap/caching_map.go
fasaxc: Adjusted the naming to be consistent and added comments.

tomastigera (approved):
LGreatTM, I like the the spirit of the change, we have good test coverage.

I wonder if using `map[string]string` instead of the `ByteArrayToByteArrayMap` would make any perf difference.

PR #2771, opened 2021-04-14
Dont panic when we see a sctp named port

sridhartigera:
Do not panic in BPF when we see a named port other than TCP, UDP, just log the error and ignore/continue.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```


tomastigera (approved):
LGTM if we are not worries that it would pollute logs too much.

PR #2781, opened 2021-04-21
WIP: Fix aks wireguard

mikestephen:
## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```


dataplane/linux/int_dataplane.go
```diff
@@ -1245,19 +1248,34 @@ func (d *InternalDataplane) setUpIptablesBPF() {
...
+			rulesConfig.OpenStackSpecialCasesEnabled, false)...)
 
 		rpfChain := []*iptables.Chain{{
 			Name:  rules.ChainNamePrefix + "RPF",
 			Rules: rpfRules,
 		}}
 		t.UpdateChains(rpfChain)
 
+		var rawRules []iptables.Rule
+		// Set a mark on the packet if it has come from the WireGuard tunnel to ensure the RPF check allows it
```
fasaxc, line 1260: Comment isn't quite right; you're marking the encapped packets themselves, not the packets coming out of the tunnel

fv/wireguard_test.go
```diff
@@ -875,9 +876,163 @@ var _ = infrastructure.DatastoreDescribe("_BPF-SAFE_ WireGuard-Supported 3 node
...
+		}
+
+		cc.ExpectSome(wls[0], wls[1])
+		cc.ExpectSome(wls[1], wls[0])
+		cc.CheckConnectivityWithTimeout(30 * time.Second)
+
+		By("verifying packets between felix-0 and felix-1 is encrypted")
+		for i := range []int{0, 1} {
+			Eventually(func() int {
+				return tcpdumps[i].MatchCount("numTunnelPacketsFelix0toFelix1")
```
fasaxc, line 1031: Tip: we've got `tcpdump.MatchCountFn` for this case now; avoids the extra `func()`

fasaxc:
It'd be good to have a test that accesses a host from an external node to make sure we're not breaking that packet path.

fv/wireguard_test.go
```diff
@@ -875,9 +876,163 @@ var _ = infrastructure.DatastoreDescribe("_BPF-SAFE_ WireGuard-Supported 3 node
...
+			}, "10s", "100ms").Should(BeNumerically(">", 0))
+			Eventually(func() int {
+				return tcpdumps[i].MatchCount("numNonTunnelPacketsFelix0toFelix1")
+			}, "10s", "100ms").Should(BeNumerically("==", 0))
+			Eventually(func() int {
+				return tcpdumps[i].MatchCount("numNonTunnelPacketsFelix1toFelix0")
+			}, "10s", "100ms").Should(BeNumerically("==", 0))
+		}
+
+		//cc.ResetExpectations()
```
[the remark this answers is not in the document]
mikestephen: Thanks, done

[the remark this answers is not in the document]
mikestephen: Thanks, done

wireguard/wireguard.go
```diff
@@ -1679,3 +1688,57 @@ func getOnlyItemInSet(s set.Set) interface{} {
...
+}
+
+// writeProcSys writes the value to the given sysctl path
+func writeProcSys(path, value string) error {
+	f, err := os.OpenFile(path, os.O_WRONLY, 0)
+	if err != nil {
+		return err
+	}
+	n, err := f.Write([]byte(value))
+	if err == nil && n < len(value) {
```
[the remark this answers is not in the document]
mikestephen: Thanks, didn't know that!

wireguard/wireguard.go
```diff
@@ -1679,3 +1688,57 @@ func getOnlyItemInSet(s set.Set) interface{} {
 	})
 	return i
 }
+
+// compareAndSetSysctl reads the current value of the sysctl setting, and sets it to a new value iff the value to be
+// set is different.
+func compareAndSetSysctl(path, value string) error {
```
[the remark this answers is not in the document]
mikestephen: If the kernel's probably doing it I'll remove this code, less to maintain that way.

fv/wireguard_test.go
mikestephen: Thanks, updated the code to use this.

dataplane/linux/int_dataplane.go
mikestephen: I updated the comment, does it seem correct now?

dataplane/linux/wireguard_mgr.go
```diff
@@ -66,6 +69,11 @@ func (m *wireguardManager) OnUpdate(protoBufMsg interface{}) {
 			return
 		}
 		switch msg.Type {
+		case proto.RouteType_REMOTE_HOST:
+			log.Debug("RouteUpdate is a remote host update")
+			if m.dpConfig.RouteSource == "WorkloadIPs" {
+				m.wireguardRouteTable.RouteUpdate(msg.DstNodeName, cidr)
```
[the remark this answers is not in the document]
mikestephen: I added a comment to reflect this, does it seem accurate?

fv/wireguard_test.go
```diff
@@ -875,9 +880,260 @@ var _ = infrastructure.DatastoreDescribe("_BPF-SAFE_ WireGuard-Supported 3 node
...
+		cc.ExpectSome(wlsByHost[0][0], wlsByHost[1][0])
+		cc.ExpectSome(wlsByHost[1][0], wlsByHost[0][0])
+		cc.CheckConnectivityWithTimeout(30 * time.Second)
+		for i := range []int{0, 1} {
+			Eventually(tcpdumps[i].MatchCountFn("numTunnelPacketsFelix0toFelix1"), "10s", "100ms").
+				Should(BeNumerically(">", 0))
+			Eventually(tcpdumps[i].MatchCountFn("numTunnelPacketsFelix1toFelix0"), "10s", "100ms").
+				Should(BeNumerically(">", 0))
+
+			Eventually(tcpdumps[i].MatchCountFn("numNonTunnelPacketsFelix0toFelix1"), "10s", "100ms").
```
fasaxc, line 1096: `MatchCount()` doesn't reset the counter so "eventually == 0" doesn't make sense.  Can we get any stray packets before things converge; if so this could lead to flakes?

fv/wireguard_test.go
```diff
@@ -875,9 +880,260 @@ var _ = infrastructure.DatastoreDescribe("_BPF-SAFE_ WireGuard-Supported 3 node
...
+			Eventually(tcpdumps[i].MatchCountFn("numNonTunnelPacketsFelix1toFelix0"), "10s", "100ms").
+				Should(BeNumerically("==", 0))
+		}
+
+		By("checking same node pod-to-pod connectivity")
+		cc.ResetExpectations()
+		for felixIdx := 0; felixIdx < nodeCount; felixIdx++ {
+			cc.ExpectSome(wlsByHost[felixIdx][0], wlsByHost[felixIdx][1])
+		}
+		cc.CheckConnectivity()
```
fasaxc, line 1107: Any reason to do each step of these checks separately?  The connectivity checker is intended to do several checks in parallel for performance so we usually just add all the expectations and then call CheckConnectivity() once/

fasaxc (changes requested):

fv/wireguard_test.go
```diff
@@ -875,9 +876,163 @@ var _ = infrastructure.DatastoreDescribe("_BPF-SAFE_ WireGuard-Supported 3 node
...
+		By("verifying packets between felix-0 and felix-1 is encrypted")
+		for i := range []int{0, 1} {
+			Eventually(func() int {
+				return tcpdumps[i].MatchCount("numTunnelPacketsFelix0toFelix1")
+			}, "10s", "100ms").Should(BeNumerically(">", 0))
+			Eventually(func() int {
+				return tcpdumps[i].MatchCount("numTunnelPacketsFelix1toFelix0")
+			}, "10s", "100ms").Should(BeNumerically(">", 0))
+			Eventually(func() int {
+				return tcpdumps[i].MatchCount("numNonTunnelPacketsFelix0toFelix1")
```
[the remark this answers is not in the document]
mikestephen: I changed this from an Eventually to an Expect, because we expect there to be no packets caught by this pattern during the test.

fv/wireguard_test.go
mikestephen: I changed this from an `Eventually` to an `Expect`, because we expect there to be no packets caught by this pattern during the test.

fv/wireguard_test.go
mikestephen: No, I've now combined them as you suggested.

fv/wireguard_test.go
```diff
@@ -875,9 +906,252 @@ var _ = infrastructure.DatastoreDescribe("_BPF-SAFE_ WireGuard-Supported 3 node
...
+		for i := range felixes {
+			felixes[i].TriggerDelayedStart()
+		}
+
+		// Check felix Wireguard links are ready.
+		for i := range felixes {
+			Eventually(func() string {
+				out, _ := felixes[i].ExecOutput("ip", "link", "show", wireguardInterfaceNameDefault)
+				return out
+			}, "10s", "100ms").Should(Not(BeEmpty()))
```
fasaxc, line 975: Trying to think what could cause flakes; does `ip link show` actually return 0 bytes if there's no output?  If it returned just `\n` say then this test might fail.

electricjesus: yeah it's zero bytes because `out` only contains stdout and if there is no link yet, it returns an err `Device "wireguard.cali" does not exist.` in stderr instead... 

but you do bring up a good point, maybe we can use ExecCombinedOuput instead to assert that it's not the error message instead?

electricjesus: we'll leave this for now since it's also being tested this way on other tests in the same file and also in other files...

fasaxc (approved):
LGTM but need to address that flake

petercork (approved):

PR #2795, opened 2021-04-28
Retry setting AWS EC2 src-dst-check until successful

hjiawei:
Implement exponential backoff retries when setting aws-src-dst-check
config. The initial backoff interval is 30 seconds with a jitter factor
of 0.1. The maximum backoff interval is 8 minutes and will keep retrying
until the config is set successfully.

Unit test:
* add aws ec2 unit tests

Manual system test:
* deploy calico opensource on AWS
* remove `ec2:ModifyNetworkInterfaceAttribute` in AWS IAM node-role
* watch calico-node logs and verify the max backoff interval is around 8 minutes

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```


fasaxc (changes requested):
This looks pretty good.  The code looks correct from a functional point of view and the test looks correct too.

The only issue I see is that the test swaps the backoff when the API of that package makes it look like the right place to swap in a fake is the Clock abstraction.  Whenever you see a struct called `RealXXX`, there's probably a `FakeXXX` or a `MockXXX` nearby for testing.  I think it's worth making the change to use that, then you test will cover more code.

You could also move the functions to their own package `aws`?  That'd be cleaner than having them in the driver package.

caseydavenport (approved):

marvin-tigera:
OK, I will merge the pull request when it's ready, squash the commits when I merge it, and leave the branch after I've merged it.

fasaxc (approved):
LGTM

PR #2797, opened 2021-04-28
Clean shouldn't remove checked-in files

caseydavenport:
## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -130,7 +130,6 @@ clean:
 	       dist \
 	       build \
 	       fv/fv.test \
-	       $(GENERATED_FILES) \
 	       go/docs/calc.pdf \
 	       release-notes-* \
 	       fv/infrastructure/crds/ \


caseydavenport:
I originally thought this was causing a build failure but it turns out that it's not the case.

That said, it seems seems odd that `make clean` deletes code that is checked-in to git - is that intentional?

neiljerram (changes requested):
I don't think this PR should be titled "Clean shouldn't remove checked-in files", because I don't think we have any defined purpose for "make clean" per se.  So I think "make clean" is only important in the context of something else that uses it as a sub-target.

Can you reword this PR to describe what the higher purpose is here?  (Assuming that there is one.)

caseydavenport:
> So I think "make clean" is only important in the context of something else that uses it as a sub-target.

We have a bunch of scripts that call into our Makefiles for things like cutting releases, etc. I believe those generally have the assumption that `make clean` cleans files that are the output of a build stage, rather than the input to the build.

I generally expect `make clean` to not touch any files that are checked in to the repository, but my reasoning is usually because if I am developing a feature and run `make clean` it better not do something to a file I have pending changes on.

That said, this is a gray area since these are auto-generated files so 99% of the time it won't matter.

Anyway, like I said above, I thought this was causing a build failure but it turns out it was something else, so no immediate need to change this imo.

PR #2814, opened 2021-05-17
Reduce FV verbosity

fasaxc:
## Description

* Ignore infra logs during stop. (API server tends to log a few pages of "can't reach etcd", which is totally useless.)
* Log command output inline rather than buffer.  (Think the buffered approach was added before we understood we could route logrus to ginkgo's writer in a nice way...)
* Reduce k8s components log levels; I can't remember the last time we've hit a problem where those logs were useful.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/fv/named_ports_test.go
+++ b/fv/named_ports_test.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2020 Tigera, Inc. All rights reserved.
+// Copyright (c) 2020-2021 Tigera, Inc. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


fv/named_ports_test.go
```diff
@@ -156,37 +156,37 @@ func describeNamedPortTests(testSourcePorts bool, protocol string) {
 
 			profiles, err := client.Profiles().List(context.Background(), options.ListOptions{})
 			if err == nil {
-				utils.AddToTestOutput("Calico Profiles\n")
+				log.Info("DIAGS: Calico Profiles:")
 				for _, profile := range profiles.Items {
-					utils.AddToTestOutput(fmt.Sprintf("%v\n", profile))
+					log.Info(profile)
```
neiljerram, line 161: I did not know you could call Info with a single object arg....

fv/utils/utils.go
```diff
@@ -86,14 +82,16 @@ func run(input []byte, checkNoError bool, command string, args ...string) error
...
 	if err != nil {
 		log.WithFields(log.Fields{
 			"command": command,
-			"args":    args,
-			"output":  string(outputBytes)}).WithError(err).Warning("Command failed")
+			"args":    args}).WithError(err).Warningf("Command failed:\n%s", output)
+	} else {
+		log.WithFields(log.Fields{
+			"command": command,
+			"args":    args}).Infof("Command succeeded:\n%s", output)
```
neiljerram, line 94: I think it's going to be a bit confusing having ` command=kubectl whatever` appended to the last line of the output.  Can we revert to logging the command on its own first?  It's not a big deal for me, but I think that would be better.

neiljerram (changes requested):
Just one small thing...

fv/utils/utils.go
fasaxc: Yes, was just thinking the same myself after trying this out for a bit :+1:

neiljerram (approved):
Looks good, thanks.

marvin-tigera:

PR #2818, opened 2021-05-20
Fix ip rule logic to stop churning rules

mikestephen:
The netlink library has a couple of quirks when listing rules:
- it doesn't populate the family field, instead returning 0
- if a rule has fwmark of 0, netlink returns -1

These quirks were causing felix to determine that the wireguard ip
rule entry on the node was different to what should be present, and so
felix deleted the "incorrect" rule and re-created it.

To workaround these quirks:

- when listing the rules, we populate the family ourselves

- the wireguard ip rule was changed from:
  from all fwmark 0x0/0x100000 lookup 1
  to:
  not from all fwmark 0x100000/0x100000 lookup 1

## Todos
- [x] Integration tests Done

## Release Note

```release-note
Fix that, with wireguard enabled, felix would delete and re-add the wireguard routing rule every 90 seconds causing occasional dropped packets.
```

### DIFF
--- a/wireguard/wireguard.go
+++ b/wireguard/wireguard.go
@@ -1496,9 +1496,12 @@ func (w *Wireguard) ensureLinkAddressV4(netlinkClient netlinkshim.Interface) err
 
 // addRouteRule adds a routing rule to use the wireguard table.
 func (w *Wireguard) addRouteRule() {
+	// The netlink library has a bug where it returns -1 for the mark on a rule instead of 0.
+	// To work around this issue, the rule below was re-written to no longer use a mark of 0x0,
+	// instead matching the NOT of the actual wireguard mark.
 	w.routerule.SetRule(routerule.NewRule(ipVersion, w.config.RoutingRulePriority).
 		GoToTable(w.config.RoutingTableIndex).
-		MatchFWMarkWithMask(0, uint32(w.config.FirewallMark)))
+		Not().MatchFWMarkWithMask(uint32(w.config.FirewallMark), uint32(w.config.FirewallMark)))
 }
 
 // ensureDisabled ensures all calico-installed wireguard configuration is removed.
--- a/fv/wireguard_test.go
+++ b/fv/wireguard_test.go
@@ -156,7 +156,7 @@ var _ = infrastructure.DatastoreDescribe("_BPF-SAFE_ WireGuard-Supported", []api
 			for _, felix := range felixes {
 				Eventually(func() string {
 					return getWireguardRoutingRule(felix)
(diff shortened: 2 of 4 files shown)


mikestephen:
> One concern: does the negation also negate the CIDR part of the match? I think we want
> 
> ```
> 0.0.0.0/0 AND mark==0
> = TRUE AND mark == 0
> = mark == 0
> ```
> 
> But, does the negation do:
> 
> ```
> NOT(0.0.0.0/0 AND mark==1)
> = NOT(0.0.0.0/0) OR NOT(mark==1)
> = FALSE OR mark ===0
> = mark == 0
> ```
> 
> or does it do
> 
> ```
> NOT(0.0.0.0/0) AND NOT(mark==1)
> = FALSE AND mark == 0
> = FALSE
> ```
> 
> or does it end up being
> 
> ```
> 0.0.0.0/0 AND NOT(mark==1)
> = TRUE AND mark == 0
> = mark == 0
> ```



> One concern: does the negation also negate the CIDR part of the match? I think we want
> 
> ```
> 0.0.0.0/0 AND mark==0
> = TRUE AND mark == 0
> = mark == 0
> ```
> 
> But, does the negation do:
> 
> ```
> NOT(0.0.0.0/0 AND mark==1)
> = NOT(0.0.0.0/0) OR NOT(mark==1)
> = FALSE OR mark ===0
> = mark == 0
> ```
> 
> or does it do
> 
> ```
> NOT(0.0.0.0/0) AND NOT(mark==1)
> = FALSE AND mark == 0
> = FALSE
> ```
> 
> or does it end up being
> 
> ```
> 0.0.0.0/0 AND NOT(mark==1)
> = TRUE AND mark == 0
> = mark == 0
> ```

After testing this out on an AKS cluster it does appear to direct packets which don't have the mark to the wireguard routing table, and skips it for packets which do have the mark.

wireguard/wireguard.go
```diff
@@ -1498,7 +1498,7 @@ func (w *Wireguard) ensureLinkAddressV4(netlinkClient netlinkshim.Interface) err
 func (w *Wireguard) addRouteRule() {
 	w.routerule.SetRule(routerule.NewRule(ipVersion, w.config.RoutingRulePriority).
 		GoToTable(w.config.RoutingTableIndex).
-		MatchFWMarkWithMask(0, uint32(w.config.FirewallMark)))
+		Not().MatchFWMarkWithMask(uint32(w.config.FirewallMark), uint32(w.config.FirewallMark)))
```
fasaxc, line 1501: Worth a comment just to remind ourselves that there's an issue with matching 0 for the mark, I think.

fasaxc (approved):
LGTM, just a suggestion to add a comment to remind us about the netlink bug

PR #2820, opened 2021-05-20
Stop ARP traffic being dropped due to RPF check

mikestephen:
ARP request packets were being dropped on the receiving node due to the
RPF check. This was because the incoming packets arrived on eth0, but
the RPF check determined that as the packets were not marked with the
wireguard fwmark, the return route would be through the wireguard
interface.

To fix this issue, iptables rules were introduced to mark all packets
which were not from the lo, wireguard, or cali+ interfaces with the
wireguard fwmark. This has the affect that these packets will then not
be routed through the wireguard interface and will pass the RPF
check.

## Todos
- [x] Integration tests Done

## Release Note

```release-note
None required
```

### DIFF
--- a/dataplane/linux/int_dataplane.go
+++ b/dataplane/linux/int_dataplane.go
@@ -1263,11 +1263,35 @@ func (d *InternalDataplane) setUpIptablesBPF() {
 			rulesConfig.RouteSource == "WorkloadIPs" {
 			log.Debug("Adding Wireguard iptables rule")
 			rawRules = append(rawRules, iptables.Rule{
-				Match: iptables.Match().Protocol("udp").
-					DestPorts(uint16(rulesConfig.WireguardListeningPort)).
-					NotSrcAddrType(iptables.AddrTypeLocal, false),
+				Match:  nil,
+				Action: iptables.JumpAction{Target: rules.ChainSetWireguardIncomingMark},
+			})
+
+			wgRules := []iptables.Rule{
+				{
+					Match:  iptables.Match().InInterface("lo"),
+					Action: iptables.ReturnAction{}},
+				{
+					Match:  iptables.Match().InInterface(rulesConfig.WireguardInterfaceName),
+					Action: iptables.ReturnAction{}},
+			}
+
+			for _, ifacePrefix := range rulesConfig.WorkloadIfacePrefixes {
+				wgRules = append(wgRules, iptables.Rule{
+					Match:  iptables.Match().InInterface(fmt.Sprintf("%s+", ifacePrefix)),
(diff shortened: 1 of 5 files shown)


dataplane/linux/int_dataplane.go
```diff
@@ -1263,11 +1263,35 @@ func (d *InternalDataplane) setUpIptablesBPF() {
 			rulesConfig.RouteSource == "WorkloadIPs" {
 			log.Debug("Adding Wireguard iptables rule")
 			rawRules = append(rawRules, iptables.Rule{
-				Match: iptables.Match().Protocol("udp").
-					DestPorts(uint16(rulesConfig.WireguardListeningPort)).
-					NotSrcAddrType(iptables.AddrTypeLocal, false),
+				Match:  nil,
```
fasaxc, line 1266: Why is this logic duplicated here and in the rules renderer?  I think it should only be in the rules renderer.

dataplane/linux/int_dataplane.go
```diff
@@ -1263,11 +1263,35 @@ func (d *InternalDataplane) setUpIptablesBPF() {
 			rulesConfig.RouteSource == "WorkloadIPs" {
```
fasaxc, line 1263: Comment above is a bit out of date; you're now marking more packets than it says.

fasaxc (changes requested):
One question

dataplane/linux/int_dataplane.go
mikestephen: I thought the same logic is needed in both branches of this
[block](https://github.com/projectcalico/felix/pull/2820/files#diff-1f472efb904b55400c81c50c3b28ac6810f5ad14a3684ee8521f1f44371df772L1085-L1089)?

fasaxc: Ah, yep, hadn't spotted this was on the BPF codepath.  You can still use the rule renderer to calculate that chain if it's the same.

That said, I'm not sure if this fix is right for eBPF mode; we try to avoid doing things in iptables when we're in eBPF mode.  If it works it might be a good stop gap until we port the logic over to the BPF code, though.

dataplane/linux/int_dataplane.go
mikestephen: Thanks, I've updated it now.

dataplane/linux/int_dataplane.go
mikestephen: Thanks for the ruleRenderer tip, I've changed this section to use it.

fasaxc (approved):
LGTM

PR #2827, opened 2021-05-24
DebugBPFMapRepinEnabled -> false

neiljerram:
There are now no use cases where we need map repinning, so let's
disable it by default.  After running with this for a while, if there
are no issues, we can remove the related code.

## Release Note

```release-note
In eBPF mode, disable Felix's BPF map repinning logic by default. This logic was intended to avoid the need to mount the BPF filesystem into the calico/node container but it was flawed because program maps are emptied by the kernel when they are not pinned.
```


fasaxc (approved):

neiljerram:
Thanks @fasaxc for the release note!

PR #2829, opened 2021-05-25
BPF coding markups

neiljerram:


bpf-gpl/conntrack.h
```diff
@@ -504,17 +504,7 @@ static CALI_BPF_INLINE struct calico_ct_result calico_ct_v4_lookup(struct cali_t
...
 		if (!v) {
-			if (CALI_F_FROM_HOST &&
-				proto_orig == IPPROTO_TCP &&
-				(tc_ctx->skb->mark & CALI_SKB_MARK_CT_ESTABLISHED_MASK) == CALI_SKB_MARK_CT_ESTABLISHED) {
-				// Linux Conntrack has marked the packet as part of a known flow.
-				// TODO-HEP Create a tracking entry for uplifted flow so that we handle the reverse traffic more efficiently.
-				CALI_DEBUG("BPF CT related miss but have Linux CT entry: established\n");
-				result.rc = CALI_CT_ESTABLISHED;
-				return result;
-			}
```
neiljerram, line 515: Or should we perhaps keep this block as well, but correcting `proto_orig == IPPROTO_TCP` to `ct_ctx->proto == IPPROTO_TCP` ?  Is it useful to record the inner TCP flow as established, when it is presumably also now ending as a result of the ICMP error?

tomastigera: I think the proto should be treated as TCP and do the mid flow linux ct check again as it is possible that we just missed the flow and now we see the icmp response.

Iirc the established is there just to say that the packet is allowed in the opposite direction.

neiljerram: Thanks @tomastigera I will update according to your response.

tomastigera (approved):
LGTM

PR #2831, opened 2021-05-25
BPF: Simplify CT context code

neiljerram:
These code changes simplify some of our CT lookup and creation code and are a step on the path towards eliminating CT context objects where possible, instead using the overall TC context+state.

Unless I've made a mistake, this is all just equivalent transformation of the previous code, with no change in logic or behaviour.


bpf-gpl/conntrack.h
```diff
@@ -504,17 +477,7 @@ static CALI_BPF_INLINE struct calico_ct_result calico_ct_v4_lookup(struct cali_t
 		k = ct_make_key(srcLTDest, ct_ctx->proto, ct_ctx->src, ct_ctx->dst, ct_ctx->sport, ct_ctx->dport);
 		v = cali_v4_ct_lookup_elem(&k);
 		if (!v) {
-			if (CALI_F_FROM_HOST &&
```
tomastigera, line 507: If we settle on this part - which is part of a different PR, the rest LGTM, nice :+1:

PR #2832, opened 2021-05-26
Wireguard fv handshake wait

mikestephen:
## Todos
- [x] Integration tests Done

## Release Note

```release-note
None required
```

### DIFF
--- a/fv/wireguard_test.go
+++ b/fv/wireguard_test.go
@@ -19,8 +19,10 @@ package fv_test
 import (
 	"context"
 	"fmt"
+	log "github.com/sirupsen/logrus"
 	"os"
 	"regexp"
+	"strconv"
 	"strings"
 	"time"
 
@@ -28,7 +30,6 @@ import (
 
 	"github.com/onsi/gomega/types"
 
-	log "github.com/sirupsen/logrus"
 	v1 "k8s.io/api/core/v1"
 
 	"github.com/projectcalico/libcalico-go/lib/numorstring"
@@ -52,7 +53,7 @@ import (
 
 const (
 	wireguardInterfaceNameDefault       = "wireguard.cali"
-	wireguardMTUDefault                 = 1420
+	wireguardMTUDefault                 = 1440
 	wireguardRoutingRulePriorityDefault = "99"
 	wireguardListeningPortDefault       = 51820
 	defaultWorkloadPort                 = "8055"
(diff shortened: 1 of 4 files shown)


fv/wireguard_test.go
```diff
@@ -512,18 +512,9 @@ var _ = infrastructure.DatastoreDescribe("_BPF-SAFE_ WireGuard-Supported", []api
...
 
-			cc.OnFail = func(msg string) {
-				log.WithField("msg", msg).Info("Connectivity Failed after 30seconds")
-				log.Info("Checking connectivity using ping between wls0 and wls1")
-
-				err, errstr := wls[0].SendPacketsTo(wls[1].IP, 5, 100)
-				log.Infof("wls0 to wls1: err=%v, errstr=%s", err, errstr)
-				err, errstr = wls[1].SendPacketsTo(wls[0].IP, 5, 100)
-				log.Infof("wls1 to wls0: err=%v, errstr=%s", err, errstr)
-			}
```
neiljerram, line 523: I presume this is a case of removing debugging that we no longer need?

fv/wireguard_test.go
```diff
@@ -52,7 +52,7 @@ import (
 
 const (
 	wireguardInterfaceNameDefault       = "wireguard.cali"
-	wireguardMTUDefault                 = 1420
+	wireguardMTUDefault                 = 1440
```
neiljerram, line 55: Is it worth a comment to explain this value?  (And perhaps specifically why 1440 is better than 1420?)

fv/wireguard_test.go
```diff
@@ -994,6 +985,30 @@ var _ = infrastructure.DatastoreDescribe("_BPF-SAFE_ WireGuard-Supported 3-node
 		}
 
 		cc = &connectivity.Checker{}
+
+		// Check felix Wireguards have handshaked.
```
neiljerram, line 989: handshook ? :smile:

neiljerram: Or is it handshaken?

But seriously, I think "handshaked" is fine here.

fv/wireguard_test.go
```diff
@@ -994,6 +985,30 @@ var _ = infrastructure.DatastoreDescribe("_BPF-SAFE_ WireGuard-Supported 3-node
...
+		for i, felix := range felixes {
+			var matchers []types.GomegaMatcher
```
neiljerram, line 991: The preparation of `matchers` is unusual here.  I think it becomes `len(felixes)-1` copies of `BeNumerically(">", 0)` - right?  That feels surprisingly heavy, and it also feels strange that the matchers are constructed **inside** the Eventually func that you're testing.

neiljerram: But I can see that, if it fails, it's nice to see the whole set of handshake numbers together.  So maybe just move the preparation of the matchers slice before the Eventually call - would that work?

neiljerram (approved):
Basically looks fine, but I commented on a few possible small markups.

fv/wireguard_test.go
mikestephen: Yes, but also this code would mask a failing test. [Here](https://github.com/projectcalico/felix/blob/master/fv/connectivity/conncheck.go#L56) it describes how if `OnFail` is set, `gingko.Fail()` won't be called.
Looking back through the commit history, this code was added for to debug a failing test, and presumably never got removed.

fv/wireguard_test.go
mikestephen: It sounds awful, the way it is I know, I think "Check felix nodes have performed Wireguard handshakes" is better do you? :)

fv/wireguard_test.go
mikestephen: Thanks Neil, I updated to make the code clearer.

fv/wireguard_test.go
neiljerram: Yes, that last suggestion is nicer, thanks!

fv/wireguard_test.go
neiljerram: Good catch.

fv/wireguard_test.go
mikestephen: Good idea, just added one.

PR #2833, opened 2021-05-26
BPF: Two more steps towards reducing conntrack-specific context

neiljerram:
ee0cf7b36 Remove skb field from ct_create_ctx
1a48dd52c Check and enforce that conntrack_create doesn't change ct_create_ctx


neiljerram:
Not sure this is the right direction now, and it's easy enough to recreate if it is.

PR #2840, opened 2021-06-10
Wireguard MTU calculation on AKS and ANI allows for underlying network

mikestephen:
## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/config/config_params.go
+++ b/config/config_params.go
@@ -101,6 +101,18 @@ func (source Source) Local() bool {
 	}
 }
 
+// Provider represents a particular provider or flavor of Kubernetes.
+type Provider string
+
+const (
+	ProviderNone      Provider = ""
+	ProviderEKS       Provider = "EKS"
+	ProviderGKE       Provider = "GKE"
+	ProviderAKS       Provider = "AKS"
+	ProviderOpenShift Provider = "OpenShift"
+	ProviderDockerEE  Provider = "DockerEnterprise"
+)
+
 // Config contains the best, parsed config values loaded from the various sources.
 // We use tags to control the parsing and validation.
 type Config struct {
@@ -395,6 +407,27 @@ func (config *Config) OpenstackActive() bool {
 	}
 	log.Debug("No evidence this is an OpenStack deployment; disabling OpenStack special-cases")
 	return false
+}
+
+func (config *Config) KubernetesProvider() Provider {
+	parts := strings.Split(config.ClusterType, ",")
+	if len(parts) < 3 {
+		log.WithField("clusterType", config.ClusterType).Debug(
+			"failed to parse clusterType, defaulting to none")
+		return ProviderNone
+	}
+
+	p := Provider(parts[2])
+	switch p {
+	case ProviderAKS, ProviderEKS, ProviderGKE, ProviderDockerEE, ProviderOpenShift:
+		log.WithFields(log.Fields{"clusterType": config.ClusterType, "provider": p}).Debug(
+			"detected a known kubernetes provider")
+		return p
+	default:
+		log.WithField("clusterType", config.ClusterType).Debug(
+			"failed to detect a known kubernetes provider, defaulting to none")
+		return ProviderNone
+	}
 }
 
 func (config *Config) resolve() (changed bool, err error) {
--- a/config/config_params.go
+++ b/config/config_params.go
@@ -105,7 +105,7 @@ func (source Source) Local() bool {
 type Provider uint8
 
 const (
-	ProviderNone Provider = iota + 1
+	ProviderNone Provider = iota
 	ProviderEKS
 	ProviderGKE
 	ProviderAKS
--- a/config/config_params.go
+++ b/config/config_params.go
@@ -102,16 +102,54 @@ func (source Source) Local() bool {
 }
 
 // Provider represents a particular provider or flavor of Kubernetes.
-type Provider string
+type Provider uint8
 
 const (
-	ProviderNone      Provider = ""
-	ProviderEKS       Provider = "EKS"
-	ProviderGKE       Provider = "GKE"
-	ProviderAKS       Provider = "AKS"
-	ProviderOpenShift Provider = "OpenShift"
-	ProviderDockerEE  Provider = "DockerEnterprise"
+	ProviderNone Provider = iota + 1
+	ProviderEKS
+	ProviderGKE
+	ProviderAKS
+	ProviderOpenShift
+	ProviderDockerEE
 )
+
+func (p Provider) String() string {
+	switch p {
+	case ProviderNone:
+		return ""
+	case ProviderEKS:
+		return "EKS"
+	case ProviderGKE:
+		return "GKE"
+	case ProviderAKS:
+		return "AKS"
+	case ProviderOpenShift:
+		return "OpenShift"
+	case ProviderDockerEE:
+		return "DockerEnterprise"
+	default:
+		return fmt.Sprintf("<unknown-provider(%v)>", uint8(p))
+	}
+}
+
+func newProvider(s string) (Provider, error) {
+	switch strings.ToLower(s) {
+	case strings.ToLower(ProviderNone.String()):
+		return ProviderNone, nil
+	case strings.ToLower(ProviderEKS.String()):
+		return ProviderEKS, nil
+	case strings.ToLower(ProviderGKE.String()):
+		return ProviderGKE, nil
+	case strings.ToLower(ProviderAKS.String()):
+		return ProviderAKS, nil
+	case strings.ToLower(ProviderOpenShift.String()):
+		return ProviderOpenShift, nil
+	case strings.ToLower(ProviderDockerEE.String()):
+		return ProviderDockerEE, nil
+	default:
+		return 0, fmt.Errorf("unknown provider %s", s)
+	}
+}
 
 // Config contains the best, parsed config values loaded from the various sources.
 // We use tags to control the parsing and validation.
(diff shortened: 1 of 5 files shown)


dataplane/linux/int_dataplane.go
```diff
@@ -301,28 +321,39 @@ func NewIntDataplaneDriver(config Config) *InternalDataplane {
 		config.RulesConfig.IptablesMarkNonCaliEndpoint)
 
 	// Auto-detect host MTU.
-	if mtu, err := findHostMTU(config.MTUIfacePattern); err != nil {
+	hostMTU, err := findHostMTU(config.MTUIfacePattern)
+	if err != nil {
 		log.WithError(err).Fatal("Unable to detect host MTU, shutting down")
-	} else {
```
mikestephen, line 306: I refactored this a little to make it easier to read. Let me know what you think.

dataplane/linux/int_dataplane.go
```diff
@@ -301,28 +321,39 @@ func NewIntDataplaneDriver(config Config) *InternalDataplane {
...
+		if config.KubernetesProvider == ProviderAKS && config.RouteSource == "WorkloadIPs" {
+			log.Debug("Defaulting Wireguard MTU based on host and AKS with WorkloadIPs")
+			config.Wireguard.MTU = hostMTU - aksMTUOverhead - wireguardMTUOverhead
 		}
+		log.Debug("Defaulting Wireguard MTU based on host")
+		config.Wireguard.MTU = hostMTU - wireguardMTUOverhead
 	}
-	if err := writeMTUFile(config); err != nil {
+	podMTU := determinePodMTU(config)
+	if err := writeMTUFile(podMTU); err != nil {
```
mikestephen, line 356: I'm not sure on the best way to unit test the MTU calculation above. Since the `config` param is not a pointer, the test can't inspect the changes made to the config, so maybe abstract the writing of the MTU file here so I can test the `podMTU` it gets called with?

dataplane/linux/int_dataplane.go
aaaaaaaalex: Is it intentional that the code that _was_ in the `else` block will now run even when there is an error?

mikestephen: The Fatal log will exit, but maybe it's clearer what's going on if the code is back in a else block?

aaaaaaaalex: ahh I see, I'm not familiar with the fatal flag :) instead of putting back in the else-block, maybe a gratuitous `return`? Only a small thing anyway, my main worry was that it would continue executing, but since it doesnt, my worries are gone!

dataplane/linux/int_dataplane.go
```diff
@@ -301,28 +321,39 @@ func NewIntDataplaneDriver(config Config) *InternalDataplane {
...
+		// for details.
+		// Additionally, Wireguard sets the DF bit on its packets, and so if the MTU is set too high large packets will
+		// be dropped.
+		// Therefore it is necessary to allow for the difference between the MTU of the host and the underlying network.
+		if config.KubernetesProvider == ProviderAKS && config.RouteSource == "WorkloadIPs" {
+			log.Debug("Defaulting Wireguard MTU based on host and AKS with WorkloadIPs")
+			config.Wireguard.MTU = hostMTU - aksMTUOverhead - wireguardMTUOverhead
 		}
+		log.Debug("Defaulting Wireguard MTU based on host")
+		config.Wireguard.MTU = hostMTU - wireguardMTUOverhead
```
fasaxc, line 354: This code will still run and clobber the MTU even if you go through the above conditional

dataplane/linux/int_dataplane.go
mikestephen: Was thinking the same about a gratuitous `return`, I'll add that in!

dataplane/linux/int_dataplane.go
fasaxc: Why not extract out a new function that takes a pointer `detectMTUs(config *Config)`?  (This function is too big anyway!)

fasaxc (changes requested):
Think the approach should work but it does mean that the ClusterType is being upgraded from a "nice to have for logging" to a "required in AKS".  Few suggestions from me.

dataplane/linux/int_dataplane.go
mikestephen: Even with the `return nil` I just added on line 327?

mikestephen: Good catch!

petercork (approved):
Nothing to add the over other comments, LGTM

dataplane/linux/int_dataplane.go
mikestephen: That looks better, but to be able to call that new function from the `intdataplane_test` package it'll need to be public, is that ok?

fasaxc: You can do either in Go, this package has examples of both.  E.g. endpoint_mgr_test.go is in the main package.  Feel free to move int_dataplane_test.go to the non-test package if that makes it easier to test.

To expand on my thinking (since it's more nuanced than "anything goes :-P": I think the idea behind the `_test` package is several-fold:

* it allows for breaking certain kinds of import cycles (not sure I grok this but I read it somewhere!)
* it encourages blackbox testing where you test the function rather than the internals
* it lets you test from the user's point of view; verify that you actually exposed the API you were expecting.

All good things but 
* This isn't an "API" package so I'm not worried about the third point
* MTU detection feels like a nice granularity of "unit" to test (so it feels like it meets the second bullet in spirit, whether you test from the same package or `_test`).  An alternative would be to move MTU detection logic out into its own package; might be a nice cleanup but also not particularly pressing IMO.

mikestephen: Thanks for detailed explanation! I tried passing in a `MTUFileWriter` which I could then have a fake implementation of in the tests, and it worked fine, but looked a little complicated. In the end I just did as you first suggested and extracted a function which takes a pointer to the config.

dataplane/linux/int_dataplane_test.go
```diff
@@ -103,4 +108,17 @@ var _ = Describe("Constructor test", func() {
 			Expect(dp).ToNot(BeNil())
 		})
 	})
+
+	Context("with Wireguard on AKS", func() {
```
fasaxc, line 112: Worth a non-AKS test too, I think

config/config_params.go
```diff
@@ -101,6 +101,56 @@ func (source Source) Local() bool {
 	}
 }
 
+// Provider represents a particular provider or flavor of Kubernetes.
+type Provider uint8
+
+const (
+	ProviderNone Provider = iota + 1
```
fasaxc, line 108: Go idiom is to make the "zero" value useful.  Why not have `ProviderNone==0` ?

dataplane/driver.go
```diff
@@ -378,6 +381,31 @@ func StartDataplaneDriver(configParams *config.Config,
 	}
 }
 
+func kubernetesProvider(clusterType string) intdataplane.Provider {
```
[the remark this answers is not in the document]
mikestephen: Makes good sense, done.

dataplane/linux/int_dataplane.go
```diff
@@ -288,76 +291,67 @@ type InternalDataplane struct {
...
+	ipipMTUOverhead      = 20
+	vxlanMTUOverhead     = 50
+	wireguardMTUOverhead = 60
+	aksMTUOverhead       = 100
 )
 
-func NewIntDataplaneDriver(config Config) *InternalDataplane {
-	log.WithField("config", config).Info("Creating internal dataplane driver.")
-	ruleRenderer := config.RuleRendererOverride
+func NewIntDataplaneDriver(c Config) *InternalDataplane {
```
fasaxc, line 301: Not sure about the global rename.  Presumably that was to avoid a clash with the imported name?  I'd probably solve that by moving the new function to a new file. Then we can shrink this god file a bit too.

I can live with the rename but it feels like a pretty large scope to have a variable with such a short name.  It might conflict when you merge it over to the private codebase.

dataplane/driver.go
```diff
@@ -378,6 +381,31 @@ func StartDataplaneDriver(configParams *config.Config,
...
+	parts := strings.Split(clusterType, ",")
+	if len(parts) < 3 {
+		log.WithField("clusterType", clusterType).Debug(
+			"failed to parse clusterType, defaulting to none")
+		return intdataplane.ProviderNone
+	}
+
+	p := intdataplane.Provider(parts[2])
```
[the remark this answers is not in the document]
mikestephen: Thanks, this was causing problems on my AKS cluster as well. Fixed.

fasaxc (changes requested):
Looking a lot cleaner now, just a couple of thoughts.

dataplane/driver.go
```diff
@@ -378,6 +381,31 @@ func StartDataplaneDriver(configParams *config.Config,
...
+	switch p {
+	case intdataplane.ProviderAKS,
```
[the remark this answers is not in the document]
mikestephen: Good suggestion, made it case-insensitive.

config/config_params.go
mikestephen: I can't remember why I thought 0 would be a problem, tried it now and seems fine!

dataplane/linux/int_dataplane.go
mikestephen: Correct, this was due to needing to reference the `config` package inside that function, but since I extracted a function for this functionality I was able to revert this rename.

fasaxc (approved):
LGTM

PR #2848, opened 2021-06-17
Fix sem timeouts and priorities

Brian-McM:


doublek (approved):
LGTM. Double checked the run times for the jobs/blocks and the limit seems fine.

PR #2853, opened 2021-06-20
basic wireguard prometheus stats

electricjesus:
## Description
- standalone prometheus collector for basic wireguard interface statistics
- stats will be present if: 
  - wireguard is installed / if there are devices associated with wireguard
  - wireguard interfaces has peers 
- unit tests included

| metric name|metric type|metric description|remarks|
|--|--|--|--|
|wireguard_meta | gauge | device / interface information for a felix/calico node | values are in labels, gauge value is always 1 |
|wireguard_bytes_rcvd  | counter | bytes received from a peer identified by a peer public key and endpoint | peer information in labels |
|wireguard_bytes_sent | counter | bytes sent to a peer identified by a peer public key and endpoint | peer information in labels |
|wireguard_latest_handshake_seconds | gauge | last handshake with a peer, unix timestamp in seconds | peer info in labels, use deriv() to find intervals |

image

image

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```


wireguard/metrics.go
```diff
@@ -0,0 +1,258 @@
...
+	wireguardMetaHelpText = "wireguard interface and runtime metadata"
+
+	wireguardLatestHandshakeIntervalFQName   = "wireguard_latest_handshake_seconds"
+	wireguardLatestHandshakeIntervalHelpText = "wireguard interface latest handshake unix timestamp in seconds to a peer"
+
+	wireguardBytesSentFQName   = "wireguard_bytes_sent"
+	wireguardBytesSentHelpText = "wireguard interface total outgoing bytes from peer"
+
+	wireguardBytesRcvdFQName   = "wireguard_bytes_rcvd"
+	wireguardBytesRcvdHelpText = "total incoming bytes from peer"
```
mikestephen, line 49: `wireguardBytesSentHelpText` and `wireguardBytesRcvdHelpText` are slightly inconsistent with each other, one begins with "wireguard interface " and the other doesn't.

wireguard/metrics.go
```diff
@@ -0,0 +1,258 @@
...
+	}
+	return NewWireguardMetricsWithShims(hostname, netlinkshim.NewRealWireguard), nil
+}
+
+func NewWireguardMetricsWithShims(hostname string, newWireguardClient func() (netlinkshim.Wireguard, error)) *Metrics {
+	logrus.WithField("hostname", hostname).Debug("created wireguard collector for host")
+	return &Metrics{
+		hostname:           hostname,
+		newWireguardClient: newWireguardClient,
+		logCtx: logrus.WithFields(logrus.Fields{
```
mikestephen, line 94: Is the formatting slightly off here?

wireguard/metrics.go
```diff
@@ -0,0 +1,258 @@
...
+
+		peerRx: map[wgtypes.Key]int64{},
+		peerTx: map[wgtypes.Key]int64{},
+	}
+}
+
+func (collector *Metrics) refreshStats(m chan<- prometheus.Metric) {
+	wgClient, err := collector.newWireguardClient()
+	if err != nil {
+		return
```
mikestephen, line 106: Worth logging here if there was an error?

wireguard/metrics.go
```diff
@@ -0,0 +1,258 @@
...
+			prometheus.GaugeValue,
+			1,
+		)
+	}
+}
+
+func (collector *Metrics) collectPeerMetrics(wgClient netlinkshim.Wireguard, m chan<- prometheus.Metric) {
+	collector.logCtx.Debug("collecting wg peer(s) metrics")
+
+	devices, err := wgClient.Devices()
```
mikestephen, line 148: I think it's worth it for readability, but you're querying for devices here and also back in `collectDeviceMetrics`

wireguard/metrics.go
electricjesus: that's how goland decided how to format this file unfortunately :shrug:

wireguard/metrics.go
electricjesus: i moved both queries to refreshStats :)

wireguard/metrics.go
electricjesus: yup, done

wireguard/metrics.go
electricjesus: good catch! also the logic of one of those help texts are wrong.. thanks!

wireguard/metrics.go
mikestephen: You've moved to the dark-side? ;)

wireguard/metrics.go
```diff
@@ -0,0 +1,255 @@
...
+
+	rc, tc := collector.peerRx[k], collector.peerTx[k]
+	collector.logCtx.WithFields(logrus.Fields{
+		"current_rx": rc,
+		"updated_rx": rx,
+		"current_tx": tc,
+		"updated_tx": tx,
+	}).Debug("read raw values")
+
+	if delta := rx - rc; rc != 0 && delta > 0 {
```
mikestephen, line 221: Was thinking the rx and tx figures could be reset if wireguard is disabled/enabled, but you've covered it perfectly!

mikestephen:
That code was a pleasure to read, LGTM!

wireguard/metrics.go
electricjesus: yes, this took a lot of testing to get right. it does take at least two refreshStats to get a figure going from metrics. if I didn't do the check this way, it would've created a big delta spike on the metric readings over time

petercork (approved):

wireguard/metrics.go
```diff
@@ -0,0 +1,227 @@
...
+}
+
+func (collector *Metrics) refreshStats(m chan<- prometheus.Metric) {
+	wgClient, err := collector.newWireguardClient()
+	if err != nil {
+		collector.logCtx.WithError(err).Error("error initializing wireguard client devices")
+		return
+	}
+
+	devices, err := wgClient.Devices()
```
fasaxc, line 110: Do we ever have >1 wireguard tunnel device?  I thought we only had one (so you could just poll that one instead of _all_ tunnels).  Probably not worth the extra effort to make this code _less_ generic but I think it's 50:50 whether a user will want us to report non-Calico Wireguard tunnels.

wireguard/metrics.go
```diff
@@ -0,0 +1,227 @@
...
+	newWireguardClient func() (netlinkshim.Wireguard, error)
+	logCtx             *logrus.Entry
+
+	peerRx, peerTx map[wgtypes.Key]int64
+}
+
+func (collector *Metrics) Describe(chan<- *prometheus.Desc) {}
+
+func (collector *Metrics) Collect(m chan<- prometheus.Metric) {
+	collector.refreshStats(m)
```
fasaxc, line 69: Is there any rate limit on `Collect()`.  If not, might be worth rate limiting calls to the dataplane

wireguard/metrics.go
```diff
@@ -0,0 +1,227 @@
...
+	wireguardBytesRcvdHelpText = "wireguard interface total incoming bytes to peer"
+)
+
+func init() {
+	prometheus.MustRegister(
+		MustNewWireguardMetrics(),
+	)
+}
+
+type Metrics struct {
```
fasaxc, line 58: What's the reason for implementing your own metric (rather than using off-the-shelf ones from the prometheus client lib)?

fasaxc (dismissed):
Code generally looking clean; a few questions from me and one more thought: the set of peers is dynamic, does the prometheus client garbage collect "stale" stats?

fasaxc:
One more thought, I think we should put this behind a config flag.  It could produce a lot of stats and we have had users complain that we overload their finely balanced prometheus  in the past (we ended up adding a flag to disable all the go built-in stats, for example).

petercork:
> One more thought, I think we should put this behind a config flag. It could produce a lot of stats and we have had users complain that we overload their finely balanced prometheus in the past (we ended up adding a flag to disable all the go built-in stats, for example).

Hey @fasaxc do you think we should do this for this PR as a blocker? I can try to get this done shortly. I will look for that built-in go stats flag as a reference.

fasaxc:
> > One more thought, I think we should put this behind a config flag. It could produce a lot of stats and we have had users complain that we overload their finely balanced prometheus in the past (we ended up adding a flag to disable all the go built-in stats, for example).
> 
> Hey @fasaxc do you think we should do this for this PR as a blocker? I can try to get this done shortly. I will look for that built-in go stats flag as a reference.

The process for adding a flag is documented here: https://tigera.atlassian.net/wiki/spaces/ENG/pages/398295107/Adding+a+new+Felix+configuration+parameter

fasaxc:
If this is to go into v3.8 then I'm inclined _not_ to merge it unless it has a flag since we could get caught out.

fv/wireguard_test.go
```diff
@@ -1178,6 +1178,21 @@ var _ = infrastructure.DatastoreDescribe("_BPF-SAFE_ WireGuard-Supported 3-node
 		By("checking external node to pod connectivity")
 		cc.ExpectSome(externalClient, wlsByHost[0][0])
 
+		By("checking prometheus metrics render")
+		for _, felix := range felixes {
+			s, err := felix.ExecOutput("wget", "localhost:9091/metrics", "-O", "-")
```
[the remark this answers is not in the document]
electricjesus: I tried that but it seems to make an http request per metric? also I want to test for the type as well. I think we can spruce up fv/metrics/metrics.go with some sort of prometheus [expfmt](https://pkg.go.dev/github.com/prometheus/common/expfmt#TextParser.TextToMetricFamilies) parsing

wireguard/metrics.go
electricjesus: no not really.. we only have wireguard.cali (perhaps 90+% of the time). yeah I wanted to make this initial code as generic as possible.. let me know if you're happy to leave it at its generic state or make it wireguard.cali only or whatever it's set in the felix config

wireguard/metrics.go
electricjesus: I wanted _something_ to trigger stats collection. So a convenient way is just to hook it in Collect(). Otherwise, I'd have to find another way e.g. a timer or an event. Now, I could've used CounterFunc but considering the calls to the data plane that each metric requires it could get quite complicated to synchronize. I hoped that this would be way simpler!

wireguard/metrics.go
electricjesus: this is a good idea, i'll look into it

electricjesus: added in https://github.com/projectcalico/felix/pull/2853/commits/d455d8e5b9ed1ee088f6637b948964c979d25a95

wireguard/metrics.go
fasaxc: OK to leave as generic, I think

wireguard/metrics.go
```diff
@@ -0,0 +1,284 @@
...
+	devices, err := wgClient.Devices()
+	if err != nil {
+		collector.logCtx.WithError(err).Error("error listing wireguard devices")
+		return nil
+	}
+	return devices
+}
+
+func (collector *Metrics) refreshStats(m chan<- prometheus.Metric) {
+	if ct := time.Since(collector.lastCollectionTime); ct < collector.rateLimitInterval {
```
fasaxc, line 164: What does this actually do to prometheus?  Will it show as a gap in the graph?  I was imagining caching the previous value but I  guess that'd be fiddly

wireguard/metrics.go
```diff
@@ -0,0 +1,284 @@
...
+	wireguardLatestHandshakeIntervalFQName   = "wireguard_latest_handshake_seconds"
+	wireguardLatestHandshakeIntervalHelpText = "wireguard interface latest handshake unix timestamp in seconds to a peer"
+
+	wireguardBytesSentFQName   = "wireguard_bytes_sent"
+	wireguardBytesSentHelpText = "wireguard interface outgoing bytes count to peer"
+
+	wireguardBytesRcvdFQName   = "wireguard_bytes_rcvd"
+	wireguardBytesRcvdHelpText = "wireguard interface incoming bytes count from peer"
+
+	defaultCollectionRatelimit = time.Second * 5
```
fasaxc, line 52: Maybe 1s just to prevent very DoS type situations?  Can imagine someone ramping Prometheus up to 5s intervals.

wireguard/metrics.go
electricjesus: as far as my testing went it does not show as a gap since it's only 5s.. if you select a crazy query range step like 1s _maybe_ you can see a teeny gap but most tools like grafana smooth that out anyway. 

if we bring it further down to 1s (which is the smallest resolution in prom anyway), nothing will be affected then and will just prevent DDoS-type calls to dataplane

wireguard/metrics.go
electricjesus: done

PR #2854, opened 2021-06-21
Reduce organisation impact of semaphore job

fasaxc:
## Description

Proposing a few changes to the semaphore job to increase efficiency at the expense of a slightly longer job times (and some other changes):

* Coalesce the build, UT, k8sfv, image-all jobs into one job that is run first.  Previously, the jobs took an aggregate of ~32minutes but combining them reduces that to 14 minutes due to being able to make use of the warm go build cache etc.
* Enable auto-cancel for master; I think it's pretty rare that we queue up multiple master builds but it might help when we're doing a lot of pin updates etc.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -11,9 +11,9 @@ agent:
 
 auto_cancel:
   running:
-    when: "branch != 'master'"
+    when: "true"
   queued:
-    when: "branch != 'master'"
+    when: "true"
 
 promotions:
 - name: Cleanup
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -15,18 +15,6 @@ auto_cancel:
   queued:
     when: "true"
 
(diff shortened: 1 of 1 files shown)


.semaphore/semaphore.yml
```diff
@@ -47,21 +59,24 @@ blocks:
         type: e1-standard-4
         os_image: ubuntu1804
     jobs:
-    - name: Build
+    - name: Build and run UT, k8sfv
```
neiljerram, line 62: Does this name change mean that occurrences of `dependencies: ["Build"]` need to be updated?

fasaxc: No, the dependency is based on the name of the block, not the individual job.

Brian-McM (approved):
Looks great, thanks for doing that!

PR #2856, opened 2021-06-21
Combine bpf fvs

fasaxc:
## Description

Run all the "new kernel" tests in parallel from a single semaphore runner instead of one per batch.  

The tests actually run in GCP on instances with newer kernels so we were wasting a lot of Semaphore capacity just waiting for those GCP VMs to finish running tests.

- Introduce new scripts to create all the GCP VMs at once and then to run all the tests in parallel.
- Start a background thread to grep the output of each test for interesting phrases like "Failure".
- Store off all the FV logs as artifacts.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/.semaphore/collect-artifacts-from-vms
+++ b/.semaphore/collect-artifacts-from-vms
@@ -1,0 +1,37 @@
+#!/usr/bin/env bash
+
+# Copyright (c) 2019 Tigera, Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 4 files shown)


Brian-McM (approved):
This is just wonderful, thank you!

bcreane:
awesome!

Brian-McM (approved):
Re approving, I just didn't look in the right place for the artifacts.

PR #2861, opened 2021-06-22
Tweaks to new parallel FV system

fasaxc:
## Description

- Improve logging.
- Close window where we'd try to tail a log that may not exist yet (by adding `-F`).
- Skip collecting the large binaries from every test running.  They should be the same on each node so only collect them once.
- Collect diags in parallel.  (Doing them serially was taking 16-17 minutes, which was a surprise.)
- Increase slow spec threshold to avoid spam.  Some FV tests are just quite long by design.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/.semaphore/collect-artifacts-from-vms
+++ b/.semaphore/collect-artifacts-from-vms
@@ -28,10 +28,32 @@ echo "Collecting artifacts..."
 num_fv_batches=${NUM_FV_BATCHES:-8}
 batches=(ut $(seq 1 ${num_fv_batches}))
 pids=()
-monitor_pids=()
+log_files=()
 for batch in "${batches[@]}"; do
   vm_name="$vm_name_prefix$batch"
-  echo "Collecting artifacts from $vm_name..."
-  VM_NAME=$vm_name $my_dir/on-test-vm ${REPO_NAME}/.semaphore/collect-artifacts || echo "Failed to collect artifacts from ${vm_name}."
-  gcloud --quiet compute scp "--zone=${ZONE}" "ubuntu@${vm_name}:${REPO_NAME}/artifacts" ./ --recurse || echo "Failed to SCP from ${vm_name}"
+  log_file="$artifacts_dir/collect-artifacts-$batch.log"
+  log_files+=( $log_file )
+  echo "Collecting artifacts from $vm_name. Logging to $log_file."
+  (
+    args=""
+    # Only collect binaries once since they're large to copy and should be identical.
+    if [ $batch != 1 ]; then
+      args="--skip-binaries"
+    fi
+    VM_NAME=$vm_name $my_dir/on-test-vm ${REPO_NAME}/.semaphore/collect-artifacts $args || echo "Failed to collect artifacts from ${vm_name}."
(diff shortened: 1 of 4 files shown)


.semaphore/collect-artifacts-from-vms
```diff
@@ -28,10 +28,32 @@ echo "Collecting artifacts..."
...
+    VM_NAME=$vm_name $my_dir/on-test-vm ${REPO_NAME}/.semaphore/collect-artifacts $args || echo "Failed to collect artifacts from ${vm_name}."
+    gcloud --quiet compute scp "--zone=${ZONE}" "ubuntu@${vm_name}:${REPO_NAME}/artifacts" ./ --recurse || echo "Failed to SCP from ${vm_name}"
+  ) > "$log_file" &
+  pid=$!
+  pids+=( $pid )
+done
+
+echo "===== Waiting for diags collection to complete ====="
+for (( i=0; i<num_batches; i++ )); do
+  pid=${pids[$i]}
```
neiljerram, line 52: Check here that $pid is non-empty?

fv/run-batches
```diff
@@ -16,7 +16,7 @@
 
 : ${FV_NUM_BATCHES:=4}
 : ${FV_BATCHES_TO_RUN:=1 2 3 4}
-: ${FV_SLOW_SPEC_THRESH:=90}
+: ${FV_SLOW_SPEC_THRESH:=300}
```
neiljerram, line 19: I feel a little worried this will encourage people to write tests with massive sleeps in.  What's the harming in keeping the "SLOW TEST" output as we currently have it?

neiljerram (approved):
A few thoughts but LGTM

.semaphore/collect-artifacts-from-vms
```diff
@@ -28,10 +28,32 @@ echo "Collecting artifacts..."
...
+log_files=()
 for batch in "${batches[@]}"; do
   vm_name="$vm_name_prefix$batch"
-  echo "Collecting artifacts from $vm_name..."
-  VM_NAME=$vm_name $my_dir/on-test-vm ${REPO_NAME}/.semaphore/collect-artifacts || echo "Failed to collect artifacts from ${vm_name}."
-  gcloud --quiet compute scp "--zone=${ZONE}" "ubuntu@${vm_name}:${REPO_NAME}/artifacts" ./ --recurse || echo "Failed to SCP from ${vm_name}"
+  log_file="$artifacts_dir/collect-artifacts-$batch.log"
+  log_files+=( $log_file )
+  echo "Collecting artifacts from $vm_name. Logging to $log_file."
+  (
```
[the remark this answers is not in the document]
fasaxc: As discussed, we want a separate variable scope so do want the subshell

fv/run-batches
fasaxc: I was just making the output a bit messy.  When the test monitor thread is grepping for the start of the test with `-C 10` it ends up printing half a SLOW test warning half of the time.  Will revert.

.semaphore/collect-artifacts-from-vms
fasaxc: Are you thinking just an assert?  I think the code would have to be bugged to have an empty value here

neiljerram: Yes I was thinking an assert.  So maybe not worth it after all.

PR #2862, opened 2021-06-22
Reduce felix FV log severity; add env var to override for debug 

fasaxc:
## Description

Debug logging generates a lot of output when a test fails and that can cause problems:
* Ginkgo uses a lot of RAM to buffer the logs.
* The log become unwieldy for Semaphore.

Change the default to be info everywhere but add an env var to override it 

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -354,7 +354,7 @@ fv fv/latency.log fv/data-races.log: $(REMOTE_DEPS) image-test bin/iptables-lock
 	  FV_K8SIMAGE=$(FV_K8SIMAGE) \
 	  FV_NUM_BATCHES=$(FV_NUM_BATCHES) \
 	  FV_BATCHES_TO_RUN="$(FV_BATCHES_TO_RUN)" \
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -354,6 +354,7 @@ fv fv/latency.log fv/data-races.log: $(REMOTE_DEPS) image-test bin/iptables-lock
 	  FV_K8SIMAGE=$(FV_K8SIMAGE) \
 	  FV_NUM_BATCHES=$(FV_NUM_BATCHES) \
 	  FV_BATCHES_TO_RUN="$(FV_BATCHES_TO_RUN)" \
+	  FELIX_FV_FELIX_LOG_LEVEL="$(FELIX_FV_FELIX_LOG_LEVEL)" \
```
neiljerram, line 357: Don't think we need two FELIX here, and `FV_FELIX_LOG_LEVEL` would look more like other settings.

neiljerram (approved):
One naming query.

PR #2865, opened 2021-06-24
BPF: Handle window between BPF_MAP_GET_NEXT_KEY and BPF_MAP_LOOKUP_ELEM

neiljerram:
In bpf_map_load_multi, which we use for iterating through maps,
there's a window between BPF_MAP_GET_NEXT_KEY and BPF_MAP_LOOKUP_ELEM
where the entry for the next key can be deleted.  ￼

Prior to this fix, that can cause an ENOENT error that bubbles up like:

    felix-0-26959-235-felixfv[exec-err] time="2021-06-23T17:53:11Z" level=info msg="Loaded map file descriptor." fd=6 name=/sys/fs/bpf/tc/globals/cali_v4_ct3
    felix-0-26959-235-felixfv[exec-err] time="2021-06-23T17:53:11Z" level=fatal msg="Failed to iterate over conntrack entries" error="iterating the map failed: no such file or directory"

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/bpf/bpf_syscall.h
+++ b/bpf/bpf_syscall.h
@@ -141,10 +141,12 @@ int bpf_map_load_multi(__u32 map_fd,
                        void *values_out) {
    int count = 0;
    union bpf_attr attr = {};
+   __u64 last_good_key = (__u64)(unsigned long)current_key;
    attr.map_fd = map_fd;
-   attr.key = (__u64)(unsigned long)current_key;
+   attr.key = last_good_key;
    for (int i = 0; i < max_num; i++) {
      // Load the next key from the map.
+   get_next_key:
      attr.value = (__u64)(unsigned long)keys_out;
      int rc = syscall(SYS_bpf, BPF_MAP_GET_NEXT_KEY, &attr, sizeof(attr));
      if (rc != 0) {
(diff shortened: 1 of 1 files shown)


fasaxc (approved):
LGTM.  I'd be tempted to cap the retries at some large value just in case there's a corner case where we can get a systematic failure of this kind.

Does the kernel behave itself if the `last_good_key` itself is deleted?

neiljerram:
Hmm, https://man7.org/linux/man-pages/man2/bpf.2.html says:

               If key is found, the operation returns zero and sets the
              next_key pointer to the key of the next element.  If key
              is not found, the operation returns zero and sets the
              next_key pointer to the key of the first element.  If key

Which is disappointing.  I'd hope for a semantic of finding the next valid key after whatever was passed in.

neiljerram:
For the record, if/when we come back to this point...

In response to your question:
> Does the kernel behave itself if the last_good_key itself is deleted?

The answer is that it restarts in that case, i.e. returns the first key in the map.  And I've been thinking what to do about that.  
I think we could detect a restart if we can reliably check for `NEXT_KEY < PREVIOUS_KEY`, and if we detected that, we could retry the whole of `bpf_map_load_multi`.

But:
- can we perform a `NEXT_KEY < PREVIOUS_KEY` comparison ?
- there's still miss the case where the passed in `current_key` has already been deleted when `bpf_map_load_multi` is called.

If key comparison is possible, perhaps a good algorithm is to check for `NEXT_KEY > LAST_GOOD_KEY` on each iteration, and only store `NEXT_KEY` (+ advance the pointers etc) when that is true.

Anyway, we have agreed to defer this now.

PR #2866, opened 2021-06-24
Add test for delete/add during iteration of BPF map

fasaxc:
## Description

Recent flakes raised some worries about interleaved adds/deletes during BPF map iteration.  This adds a single threaded test that interleaves adds/deletes with iteration and verifies that we don't see duplicate keys or miss any keys.  It doesn't cover the case where a BPF program is adding keys while felix is iterating (the BPF programs do not delete keys by design).

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/bpf/ut/maps_test.go
+++ b/bpf/ut/maps_test.go
@@ -90,6 +90,99 @@ func TestMapIterationDeleteAfter(t *testing.T) {
 	Expect(err2).NotTo(HaveOccurred(), "Failed to delete map entry")
 }
 
+// TestDeleteDuringIter tries to validate that BPF map iteration is working as we want even when keys are being added
+// to and deleted from the map during iteration.
+func TestDeleteDuringIter(t *testing.T) {
+	RegisterTestingT(t)
+	defer cleanUpMaps()
+	logLevel := logrus.GetLevel()
+	defer func() {
+		logrus.SetLevel(logLevel)
+	}()
+	logrus.SetLevel(logrus.InfoLevel)
+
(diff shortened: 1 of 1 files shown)


bpf/ut/maps_test.go
```diff
@@ -90,6 +90,99 @@ func TestMapIterationDeleteAfter(t *testing.T) {
...
+	})
+	Expect(err).NotTo(HaveOccurred())
+	Expect(seenKeys).To(HaveLen(numEntries), "Should have seen expected num entries on deletion iteration")
+
+	// Third pass, insert key on each iteration an delete on alternate iterations.
+	numLeftAfterSecondPass := len(expectedKeys)
+	seenKeys = map[conntrack.Key]bool{}
+	deleteNextKey = false
+	insertClock := 0
+	insertIdx := numEntries
```
neiljerram, line 154: Does this mean we're only inserting at the end of the map?

neiljerram (approved):
Looks good, but two points to note:
- Inline: is it intended that insertions are only at the end of the map?
- My PR was addressing a different scenario, namely where some other process is deleting asynchronously w.r.t. the map iteration.

bpf/ut/maps_test.go
fasaxc: The maps are hashtables so I don't think they're ordered. The reason to use `insertIdx := numEntries` here is so that we get new entries rather than overwriting old ones.

PR #2877, opened 2021-06-30
XDP parser

mazdakn:
## Description

This PR adds support for parsing packets in XDP programs. Most parts are similar to the parsing in TC programs, with few exceptions. For that the original parser in TC is used and differences are separated via a new flag named "CALI_F_XDP". This allows to compile XDP programs using 
"-DCALI_COMPILE_FLAGS=64"

The PR also includes a simple XDP program in xdp.c that can be used as a skeleton for XDP programs.

## Release Note

```release-note
None required
```

### DIFF
--- a/bpf-gpl/parsing.h
+++ b/bpf-gpl/parsing.h
@@ -53,7 +53,7 @@ static CALI_BPF_INLINE int parse_packet_ip(struct cali_tc_ctx *ctx) {
 			goto deny;
 		} else {
 			CALI_DEBUG("Unknown ethertype on host interface (%x), allow\n",
-								protocol);
+									protocol);
 			goto allow_no_fib;
 		}
 	}
--- a/bpf-gpl/parsing.h
+++ b/bpf-gpl/parsing.h
@@ -67,7 +67,7 @@ static CALI_BPF_INLINE int parse_packet_ip(struct cali_tc_ctx *ctx) {
 	}
 
 	// In TC programs, parse packet and validate its size. This is
-	// already done for XDP programs at the beginning of function
+	// already done for XDP programs at the beginning of the function.
 	if (!CALI_F_XDP) {
 		if (skb_refresh_validate_ptrs(ctx, UDP_SIZE)) {
 			ctx->fwd.reason = CALI_REASON_SHORT;
--- a/bpf-gpl/parsing.h
+++ b/bpf-gpl/parsing.h
@@ -88,19 +88,9 @@ static CALI_BPF_INLINE int parse_packet_ip(struct cali_tc_ctx *ctx) {
 
 allow_no_fib:
 	fwd_fib_set(&ctx->fwd, false);
-	if (CALI_F_XDP) {
-		ctx->fwd.res = XDP_PASS;
-	} else {
-		ctx->fwd.res = TC_ACT_UNSPEC;
-	}
 	return -1;
 deny:
-	if (CALI_F_XDP) {
-		ctx->fwd.res = XDP_DROP;
-	} else {
-		ctx->fwd.res = TC_ACT_SHOT;
-	}
-	return -1;
+	return -2;
 }
 
 #endif /* __CALI_PARSING_H__ */
--- a/bpf-gpl/parsing.h
+++ b/bpf-gpl/parsing.h
@@ -97,7 +97,6 @@ static CALI_BPF_INLINE int parse_packet_ip(struct cali_tc_ctx *ctx) {
 	return 0;
 
 allow_no_fib:
-	fwd_fib_set(&ctx->fwd, false);
 	return -1;
 deny:
 	return -2;
--- a/bpf-gpl/parsing.h
+++ b/bpf-gpl/parsing.h
@@ -19,7 +19,20 @@
 #define __CALI_PARSING_H__
 
 static CALI_BPF_INLINE int parse_packet_ip(struct cali_tc_ctx *ctx) {
-	switch (bpf_htons(ctx->skb->protocol)) {
+	__u16 protocol = 0;
+
+	if (CALI_F_XDP) {
+		if (skb_refresh_validate_ptrs(ctx, UDP_SIZE)) {
+			ctx->fwd.reason = CALI_REASON_SHORT;
+			CALI_DEBUG("Too short\n");
+			goto deny;
+		}
+		protocol = bpf_ntohs(ctx->eth->h_proto);
+	} else {
+		protocol = bpf_ntohs(ctx->skb->protocol);
+	}
+
+	switch (protocol) {
 	case ETH_P_IP:
 		break;
 	case ETH_P_ARP:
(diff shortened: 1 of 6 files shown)


bpf-gpl/xdp.c
```diff
@@ -0,0 +1,93 @@
+#include <linux/if_ether.h>
```
mazdakn, line 1: This file is only for showing the structure of the XDP program. Won't be included in the final PR.

bpf-gpl/parsing.h
```diff
@@ -19,7 +19,20 @@
...
+
+	if (CALI_F_XDP) {
+		if (skb_refresh_validate_ptrs(ctx, UDP_SIZE)) {
+			ctx->fwd.reason = CALI_REASON_SHORT;
+			CALI_DEBUG("Too short\n");
+			goto deny;
+		}
+		protocol = bpf_ntohs(ctx->eth->h_proto);
+	} else {
+		protocol = bpf_ntohs(ctx->skb->protocol);
```
mazdakn, line 32: Originally, this line was bpf_htons(ctx->skb->protocol), however, skb.protocol type (in Kernel source: include/linux/skbuff.h) is __be16. So changed it to bpf_ntohs(ctx->skb->protocol).
Also note that protocol type in /usr/include/linux/bpf.h is __u32 (like all other fields), but I assume this is the case for alignment, and fields are mapped to kernel definition. Is that correct?

bpf-gpl/parsing.h
```diff
@@ -36,19 +49,21 @@ static CALI_BPF_INLINE int parse_packet_ip(struct cali_tc_ctx *ctx) {
...
-	if (skb_refresh_validate_ptrs(ctx, UDP_SIZE)) {
-		ctx->fwd.reason = CALI_REASON_SHORT;
-		CALI_DEBUG("Too short\n");
-		goto deny;
+	if (!CALI_F_XDP) {
+		if (skb_refresh_validate_ptrs(ctx, UDP_SIZE)) {
+			ctx->fwd.reason = CALI_REASON_SHORT;
+			CALI_DEBUG("Too short\n");
+			goto deny;
+		}
```
mazdakn, line 66: This branch is similar to the one at line 24, and can be merged together. The downside is that for we will use more cpu cycle on some switch cases. What do you think?

neiljerram:
@mazdakn Please can you add something to the Description field to explain the purpose of the PR?

Also, is it really still a draft, or could it potentially be merged now?  If the latter, please remove the "Draft" status.

bpf-gpl/parsing.h
neiljerram: The bpf_ntohs change LGTM.  I'm afraid I don't know why the fields in struct __sk_buff are all __u32 - @fasaxc ?

bpf-gpl/parsing.h
```diff
@@ -19,7 +19,20 @@
 #define __CALI_PARSING_H__
 
 static CALI_BPF_INLINE int parse_packet_ip(struct cali_tc_ctx *ctx) {
-	switch (bpf_htons(ctx->skb->protocol)) {
+	__u16 protocol = 0;
+
+	if (CALI_F_XDP) {
+		if (skb_refresh_validate_ptrs(ctx, UDP_SIZE)) {
```
neiljerram, line 25: It feels odd to need an skb_refresh_validate_ptrs call in the XDP case but not for TC.  Could the refresh call be moved up into the caller, so that we can have the same assumption - i.e. that ctx pointers are valid - for both XDP and TC here?

neiljerram: Oh I see, you have addressed this below - please ignore this comment.

bpf-gpl/parsing.h
neiljerram: I think it's best to leave like this, to save those CPU cycles in the TC case.

bpf-gpl/parsing.h
```diff
@@ -73,10 +88,18 @@ static CALI_BPF_INLINE int parse_packet_ip(struct cali_tc_ctx *ctx) {
 
 allow_no_fib:
 	fwd_fib_set(&ctx->fwd, false);
-	ctx->fwd.res = TC_ACT_UNSPEC;
+	if (CALI_F_XDP) {
+		ctx->fwd.res = XDP_PASS;
```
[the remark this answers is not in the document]
neiljerram: I think fwd.res is needed for the TC cases, because the code will call forward_or_drop, which will immediately check ctx->fwd.res.  For XDP case it depends what happens after parse_packet_ip returns.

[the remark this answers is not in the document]
neiljerram: Might be nicer to use a common set of `res` values here, and translate to XDP values higher up.

bpf-gpl/skb.h
```diff
@@ -118,12 +123,17 @@ static CALI_BPF_INLINE void skb_refresh_hdr_ptrs(struct cali_tc_ctx *ctx)
...
  * - ctx->ip_header
- * - ctx->nh/tcp_header/udp_header/icp_header.
+ * - ctx->nh/tcp_header/udp_header/icmp_header.
  */
 static CALI_BPF_INLINE int skb_refresh_validate_ptrs(struct cali_tc_ctx *ctx, long nh_len) {
 	int min_size = skb_iphdr_offset(ctx->skb) + IP_SIZE;
 	skb_refresh_start_end(ctx);
 	if (ctx->data_start + (min_size + nh_len) > ctx->data_end) {
+		// This is an XDP program and there is not enough data for next header.
+		if (CALI_F_XDP) {
```
neiljerram, line 133: Would it be useful to add a CALI_DEBUG here?  Can we get tracing for XDP programs?

bpf-gpl/xdp.c
neiljerram: I think it could be a good idea to include this.  For reference, and so that we can evolve it later to what we really want.  But please add a comment at the top to say that it is not yet in active use.

neiljerram: Should also add similar copyright as we have in other files in this directory.

bpf-gpl/xdp.c
```diff
@@ -0,0 +1,93 @@
...
+};
+
+SEC("prog")
+static CALI_BPF_INLINE int calico_xdp(struct xdp_md *xdp_ctx) {
+
+	struct cali_tc_ctx ctx = {
+		.state = state_get(),
+		.xdp = xdp_ctx,
+		.fwd = {
+			.res = XDP_PASS, // or XDP_DROP?
```
neiljerram, line 49: Hmm, I think it depends on the wider design.  Given that an XDP program can be installed at several levels, and some hardware may have differing capabilities, what's our plan for ensuring that a program can work at the place where it is installed?

neiljerram (changes requested):
Looks good; just a few points to consider

bpf-gpl/parsing.h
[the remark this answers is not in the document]
mazdakn: Absolutely, using a common set of `res` makes the code nicer and allows us to re-use more code from TC program. The only challenge is that the current code heavily uses TC return codes like "TC_ACT_SHOT". If you agree, I can try to make the return code (at least) in parsing part generic and translate it later on in the callee functions. Does it make sense?

bpf-gpl/skb.h
mazdakn: We do. Added a CALI_DEBUG here.

bpf-gpl/xdp.c
mazdakn: I think that should be check in the control plane. For example, if we use "ip" command to load the XDP programs (like TC programs), we can try to load with native support and when fails fall back to the generic mode. 
As you said, the default action depends on the wider design. I'll keep it as "XDP_PASS" since it is safest at the moment, and we should change it later on based on the final design.

mazdakn:
> @mazdakn Please can you add something to the Description field to explain the purpose of the PR?
> 
> Also, is it really still a draft, or could it potentially be merged now? If the latter, please remove the "Draft" status.

Sure, I'll remove the "Draft" status. The main thing (besides your comments) is what UT are needed? 
but these PR is passing all test since executing these codes need a special compilation flag.

neiljerram (approved):
LGTM

PR #2879, opened 2021-07-02
Fix XDP FV tests and add them to Semaphore CI

neiljerram:
Our XDP FV tests had fallen through the cracks, as regards running in Semaphore CI.
- They don't run in the regular FV, because the default Semaphore kernel is <4.16 and doesn't support XDP acceleration.
- They don't run in the "new kernel" FV, because there we focus on "BPF-SAFE".

Also, if you run them manually, some of them fail.  Turns out this is because behaviour of `ip link set dev <ifname> <xdpmode> off` has changed, and now reports **success** if there **wasn't** any XDP program attached with mode `<xdpmode>`.

### DIFF
--- a/.semaphore/collect-artifacts-from-vms
+++ b/.semaphore/collect-artifacts-from-vms
@@ -16,6 +16,8 @@
 
 set -e
 
+source .semaphore/new-kernel-common.sh
+
 vm_name_prefix=$1
 project=unique-caldron-775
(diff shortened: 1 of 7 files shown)


fasaxc (approved):
LGTM

marvin-tigera:
I'm sorry but I failed delete the branch after merging the pull request.

PR #2881, opened 2021-07-02
Convert XDP callback usage to mainline Felix architecture

neiljerram:
This is part 1 of work to remove the XDP callbacks architecture and use mainline Felix patterns to achieve equivalent function.  Callbacks for policy and IP sets are replaced by xdp_state.go OnUpdate handling for the relevant proto messages.


fasaxc (approved):
LGTM

Issue #2619, opened 2020-12-07
Can I disable felix-xdp functionalities?


sjh5205:
I'm using XDP for modify packet payload directly on datapath.
But also Calico use XDP for failsafe ports and CIDR map.

When each Calico's XDP resync period, my XDP program has been wiped out, our system is not working properly.

## Context
Felix wiped out attached XDP program on each XDP resync period.
Can I shutdown XDP functionalities in Calico?

## Your Environment
I'm using Kubernetes from Kubespray.
Calico version is v3.15.2

## Problematic Code

Here is code sinppet from Felix.
https://github.com/projectcalico/felix/blob/bbbd53863935730011636e6fa9652def9dc48586/dataplane/linux/int_dataplane.go#L449-L476

xdpState is initialize, XDP program has been wiped out.

https://github.com/projectcalico/felix/blob/bbbd53863935730011636e6fa9652def9dc48586/dataplane/linux/int_dataplane.go#L1569-L1595

Again, on resync.

lwr20:
It should be possible to disable XDP in Calico, either by setting `XDPEnabled=false` in the felixconfig default object or by setting the `FELIX_XDPENABLED=false` in the env vars passed to Felix (or calico-node).  As you've already discovered, this setting defaults to `true`.

I'm afraid I can't tell you how to do that in kubespray (they may or may not have plumbed through that configuration setting), but its certainly possible in Calico.

https://docs.projectcalico.org/reference/felix/configuration#general-configuration has more information on Felix config options.

sjh5205:
> It should be possible to disable XDP in Calico, either by setting `XDPEnabled=false` in the felixconfig default object or by setting the `FELIX_XDPENABLED=false` in the env vars passed to Felix (or calico-node). As you've already discovered, this setting defaults to `true`.
> 
> I'm afraid I can't tell you how to do that in kubespray (they may or may not have plumbed through that configuration setting), but its certainly possible in Calico.
> 
> https://docs.projectcalico.org/reference/felix/configuration#general-configuration has more information on Felix config options.

I'm afraid it doesn't work for me.
Actually, Felix's XDPEnabled option cotrol XDP Acceleration on Felix.
Regardless of whether the XDP Enabled option is Enabled or not,  XDP Refresh function still active.

lwr20:
@fasaxc as the likely writer of that code, does that ^ match your expectations?

fasaxc:
[...]

PR #2882, opened 2021-07-04
Fix XDP Cleanup issue

sjh5205:
## Description
This patch fix issue #2619 'Felix clears XDP programs it doesn't own'.
I Added conditional statement to check XDPEnabled option is true, before initializing XDPState.
It works on my system, but I think more tests will be needed.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/dataplane/linux/int_dataplane.go
+++ b/dataplane/linux/int_dataplane.go
@@ -467,7 +467,7 @@ func NewIntDataplaneDriver(config Config) *InternalDataplane {
 	}
 
 	// TODO Integrate XDP and BPF infra.
-	if !config.BPFEnabled && dp.xdpState == nil {
+	if !config.BPFEnabled && config.XDPEnabled && dp.xdpState == nil {
 		xdpState, err := NewXDPState(config.XDPAllowGeneric)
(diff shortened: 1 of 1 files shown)


dataplane/linux/int_dataplane.go
```diff
@@ -467,7 +467,7 @@ func NewIntDataplaneDriver(config Config) *InternalDataplane {
 	}
 
 	// TODO Integrate XDP and BPF infra.
-	if !config.BPFEnabled && dp.xdpState == nil {
+	if !config.BPFEnabled && config.XDPEnabled && dp.xdpState == nil {
```
neiljerram, line 470: We would like this code to run, even when `config.XDPEnabled` is false, so that Felix can clean up its own XDP state after a restart.

neiljerram (changes requested):

dataplane/linux/int_dataplane.go
sjh5205: Okay, I'll revert this pull request.

PR #2888, opened 2021-07-12
XDP program for untracked policies

mazdakn:
## Description

XDP program for untracked policies including passing metadata between XDP and TC programs. In addition, some parsing code is moved from tc.c to parsing.h to be used by XDP programs

## Todos

## Release Note

```release-note
None required
```

### DIFF
--- a/bpf-gpl/jump.h
+++ b/bpf-gpl/jump.h
@@ -43,14 +43,6 @@ struct bpf_map_def_extended __attribute__((section("maps"))) cali_jump = {
 #endif
 };
 
-static CALI_BPF_INLINE void tc_state_fill_from_iphdr(struct cali_tc_state *state, struct iphdr *ip)
-{
-	state->ip_src = ip->saddr;
-	state->ip_dst = ip->daddr;
-	state->ip_proto = ip->protocol;
-	state->ip_size = ip->tot_len;
-}
-
 /* Add new values to the end as these are program indices */
 enum cali_jump_index {
 	PROG_INDEX_POLICY,
--- a/bpf-gpl/nat.h
+++ b/bpf-gpl/nat.h
@@ -428,7 +428,6 @@ auto_allow:
 	return -2;
 
 deny:
-	ctx->fwd.res = TC_ACT_SHOT;
 	return -1;
 }
 
--- a/bpf-gpl/nat.h
+++ b/bpf-gpl/nat.h
@@ -428,6 +428,7 @@ auto_allow:
 	return -2;
 
 deny:
+	ctx->fwd.res = TC_ACT_SHOT;
 	return -1;
 }
 
--- a/bpf-gpl/parsing.h
+++ b/bpf-gpl/parsing.h
@@ -136,6 +136,8 @@ static CALI_BPF_INLINE int parse_packet_nextheader(struct cali_tc_ctx *ctx) {
 		}
 		break;
 	case IPPROTO_ICMP:
+		ctx->state->icmp_type = ctx->icmp_header->type;
+		ctx->state->icmp_code = ctx->icmp_header->code;
 		CALI_DEBUG("ICMP; type=%d code=%d\n",
 				ctx->icmp_header->type, ctx->icmp_header->code);
 		break;
--- a/bpf-gpl/parsing.h
+++ b/bpf-gpl/parsing.h
@@ -1,5 +1,5 @@
 // Project Calico BPF dataplane programs.
-// Copyright (c) 2020 Tigera, Inc. All rights reserved.
+// Copyright (c) 2021 Tigera, Inc. All rights reserved.
 //
 // This program is free software; you can redistribute it and/or modify
 // it under the terms of the GNU General Public License as published by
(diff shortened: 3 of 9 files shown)


bpf-gpl/nat.h
```diff
@@ -428,7 +428,6 @@ static CALI_BPF_INLINE int vxlan_attempt_decap(struct cali_tc_ctx *ctx) {
 	return -2;
 
 deny:
-	ctx->fwd.res = TC_ACT_SHOT;
```
mazdakn, line 431: This is also done in tc program for the return value of -1. Removed it to make it generic even for XDP programs.

bpf-gpl/jump.h
```diff
@@ -43,14 +43,6 @@ struct bpf_map_def_extended __attribute__((section("maps"))) cali_jump = {
...
 };
 
-static CALI_BPF_INLINE void tc_state_fill_from_iphdr(struct cali_tc_state *state, struct iphdr *ip)
-{
-	state->ip_src = ip->saddr;
-	state->ip_dst = ip->daddr;
-	state->ip_proto = ip->protocol;
-	state->ip_size = ip->tot_len;
-}
-
```
mazdakn, line 53: Moved this function to parsing.h and updated its signature.

bpf-gpl/tc.c
```diff
@@ -168,79 +169,14 @@ static CALI_BPF_INLINE int calico_tc(struct __sk_buff *skb)
 	}
 
 	/* Copy fields that are needed by downstream programs from the packet to the state. */
-	tc_state_fill_from_iphdr(ctx.state, ctx.ip_header);
+	tc_state_fill_from_iphdr(&ctx);
 
 	/* Parse out the source/dest ports (or type/code for ICMP). */
-	switch (ctx.state->ip_proto) {
```
mazdakn, line 174: Moved this switch statement to a separate function called "tc_state_fill_from_nextheader(&ctx)" to use it for XDP.

bpf-gpl/tc.c
```diff
@@ -379,25 +315,6 @@ static CALI_BPF_INLINE int calico_tc(struct __sk_buff *skb)
...
-		ctx.fwd.reason = CALI_REASON_SHORT;
-		CALI_DEBUG("Too short\n");
-		goto deny;
-	}
-
-	/* icmp_type and icmp_code share storage with the ports; now we've used
-	 * the ports set to 0 to do the conntrack lookup, we can set the ICMP fields
-	 * for policy.
-	 */
-	if (ctx.state->ip_proto == IPPROTO_ICMP) {
```
mazdakn, line 395: Merged this if statement with tc_state_fill_from_nextheader(&ctx).

bpf-gpl/xdp.c
```diff
@@ -65,11 +66,63 @@ int calico_xdp(struct xdp_md *xdp_ctx) {
 	}
 
 	// Parse packets and drop malformed and unsupported ones
-	if (parse_packet_ip(&ctx) == -2) {
-		return XDP_DROP;
+	switch (parse_packet_ip(&ctx)) {
+	case -1:
```
neiljerram, line 70: I think we should #define constants for these -1 and -2 values, so it is easier to see what they mean.

neiljerram: Looks like -1 = DROP, -2 = ALLOW_WITHOUT_ENFORCING_POLICY , is that right?

bpf-gpl/xdp.c
```diff
@@ -65,11 +66,63 @@ int calico_xdp(struct xdp_md *xdp_ctx) {
...
+		ctx.fwd.res = XDP_DROP;
+		goto deny;
+	case -2:
+		ctx.fwd.res = XDP_PASS;
+		goto allow;
+	}
+
+	// Allow a packet if it hits an entry in the failsafe map
+	if (is_failsafe_in(ctx.state->ip_proto, ctx.state->dport, ctx.state->ip_src)) {
+		CALI_DEBUG("Inbound failsafe port: %d. Skip policy\n", ctx.state->post_nat_dport);
```
neiljerram, line 91: I guess it's wrong to have `dport` on the previous line and `post_nat_dport` here.  I think it should be `dport` both times, as "post-NAT" doesn't make sense here.

bpf-gpl/xdp.c
```diff
@@ -65,11 +66,63 @@ int calico_xdp(struct xdp_md *xdp_ctx) {
...
+		ctx.state->pol_rc = CALI_POL_ALLOW;
+		goto allow;
+	}
+
+	// Jump to the policy program
+	CALI_DEBUG("About to jump to policy program.\n");
+	bpf_tail_call(xdp_ctx, &cali_jump, PROG_INDEX_POLICY);
+	if (CALI_F_HEP) {
```
neiljerram, line 99: We only use this program for HEP ingress, so no need to check this.  Return from the tail call => no untracked policy => allow.

bpf-gpl/parsing.h
```diff
@@ -97,9 +97,104 @@ static CALI_BPF_INLINE int parse_packet_ip(struct cali_tc_ctx *ctx) {
 	return 0;
 
 allow_no_fib:
-	return -1;
+	return -2;
```
neiljerram, line 100: As also noted below, can we #define constants for these -1 and -2 numbers?

bpf-gpl/tc.c
neiljerram: Nice.

bpf-gpl/tc.c
```diff
@@ -379,25 +315,6 @@ static CALI_BPF_INLINE int calico_tc(struct __sk_buff *skb)
...
 	}
 
-	/* [SMC] I had to add this revalidation when refactoring the conntrack code to use the context and
-	 * adding possible packet pulls in the VXLAN logic.  I believe it is spurious but the verifier is
-	 * not clever enough to spot that we'd have already bailed out if one of the pulls failed. */
-	if (skb_refresh_validate_ptrs(&ctx, UDP_SIZE)) {
-		ctx.fwd.reason = CALI_REASON_SHORT;
-		CALI_DEBUG("Too short\n");
-		goto deny;
-	}
```
neiljerram, line 389: It's good to remove this if we can, but have you done any analysis or experimenting to check that it is OK?

neiljerram (changes requested):
Looking great; just a few points and questions.

bpf-gpl/parsing.h
mazdakn: Yes, sure. I wanted to do that, but noticed in other functions return codes are used like this. But it would be much cleaner with constants. Will fix it.

bpf-gpl/xdp.c
mazdakn: Exactly, will define the constants.

bpf-gpl/xdp.c
mazdakn: right.

bpf-gpl/tc.c
mazdakn: All tests are passing. I also loaded the prog manually. Will do more tests.

bpf-gpl/nat.h
mazdakn: It is not needed by XDP. I was trying to make some functions generic. I'll revert this and raise it in another PR.

bpf-gpl/tc.c
mazdakn: I reverted the change. It seems to work fine. I just think maybe it's better to make this change separately out of this PR.

bpf-gpl/types.h
```diff
@@ -101,6 +101,8 @@ enum cali_state_flags {
 	/* CALI_ST_SRC_IS_HOST is set if the packet is heading away from the host namespace and the source
 	 * belongs to the host. */
 	CALI_ST_SRC_IS_HOST	  = 0x08,
+	/* CALI_ST_ACCEPTED_BY_XDP is set if the packet is already accepted by XDP */
+	CALI_ST_ACCEPTED_BY_XDP 	= 0x10,
```
neiljerram, line 105: I think this should be a standalone #define, rather than here, because this enum is for possible values for the cali_tc_state.flags field

neiljerram (approved):
Just one last nit to move the definition of the ACCEPTED_BY_XDP constant, but otherwise this looks good - lovely work!

PR #2897, opened 2021-07-14
Propagate WireguardHostEncryptionEnabled config_param

aaaaaaaalex:
## Description

We'd like to expose an option to the user for toggling pod-to-hostnetwork-pod wireguard encryption, ontop of the standard pod-to-pod encryption. This flag is critical for getting full encryption working on EKS w/ CalicoCNI. We plan to also make an operator change that will set this flag on for AKS CNI, since wireguard encryption is broken without it on that provider

Please ask if more detail is needed!

Related Operator PR: https://github.com/tigera/operator/pull/1391

Co-author @mikestephen :raised_hands: 

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```


dataplane/driver.go
```diff
@@ -297,6 +297,9 @@ func StartDataplaneDriver(configParams *config.Config,
 				InterfaceName:       configParams.WireguardInterfaceName,
 				MTU:                 configParams.WireguardMTU,
 				RouteSource:         configParams.RouteSource,
+				// currently, AKS *must* encrypt hostnetwork traffic, while it is optional on EKS
+				HostEncryptionEnabled: configParams.KubernetesProvider() == config.ProviderAKS ||
```
mikestephen, line 301: I would consider extracting this logic in to a helper function to make it a bit more readable, but that's just my opinion.

mikestephen, line 301: I'm thinking `HostEncryptionEnabled` sounds too similar to the name of the flag, but struggling to come up with a better name. Maybe `EncryptHostTraffic` or `HostEncryptionSupported`? The reason is that someone might enable the `WireguardHostEncryptionEnabled` flag on say GKE which we don't support.

aaaaaaaalex: Had a hunch about the same - I think `EncryptHostTraffic` sounds the clearest, will rewrite that name in a sec :+1:

dataplane/driver.go
```diff
@@ -297,6 +298,8 @@ func StartDataplaneDriver(configParams *config.Config,
 				InterfaceName:       configParams.WireguardInterfaceName,
 				MTU:                 configParams.WireguardMTU,
 				RouteSource:         configParams.RouteSource,
+				// currently, AKS *must* encrypt hostnetwork traffic, while it is optional on EKS
```
mikestephen, line 301: This comment line probably should be moved inside `ShouldEncryptWireguardHostTraffic()`

dataplane/linux/int_dataplane.go
```diff
@@ -942,7 +942,7 @@ func ConfigureDefaultMTUs(hostMTU int, c *Config) {
 		c.VXLANMTU = hostMTU - vxlanMTUOverhead
 	}
 	if c.Wireguard.MTU == 0 {
-		if c.KubernetesProvider == config.ProviderAKS && c.RouteSource == "WorkloadIPs" {
+		if c.Wireguard.EncryptHostTraffic {
```
mikestephen, line 945: This line should actually be:
`if c.KubernetesProvider == config.ProviderAKS &&c.Wireguard.EncryptHostTraffic {`
because we only need to tweak the MTU like this on AKS.

aaaaaaaalex:
@mikestephen does the operator currently auto-set `routeSource="WorkloadIPs"` when running on AKS? If so, I'll probably have to do something similar for the new config option will I?

mikestephen:
> @mikestephen does the operator currently auto-set `routeSource="WorkloadIPs"` when running on AKS? If so, I'll probably have to do something similar for the new config option will I?

[This line](https://github.com/tigera/operator/blob/93b3459d5d9504cac12955a158017ea930237226/pkg/render/node.go#L1261) in operator is the one that does it, and it looks like it'll be fine for any managed cluster (AKS, EKS, GKE etc), but good you thought of it, I hadn't :)

mikestephen (approved):
LGTM Alex! Good job.

aaaaaaaalex:
ah looks good so, nice one :smile: is there anything to be said for moving [this decision](https://github.com/projectcalico/felix/pull/2897/files#diff-6d3b325b8dfb02061fd60c6641b100ad40c798ceaa2a27105b29e2af5cfb8847R471) to the operator, and having felix blindly follow whatever the `WireguardEncryptHostTraffic` param tells it? My (probably naive) view of the operator/felix currently, is that the operator makes more situational config decisions, but I'm not sure how well-defined that line needs to be

aaaaaaaalex:
and thanks!

config/config_params.go
```diff
@@ -467,6 +467,13 @@ func (config *Config) KubernetesProvider() Provider {
 	return ProviderNone
 }
 
+// ShouldEncryptWireguardHostTraffic factors in the kubernetes provider and config params to decide if host-traffic encryption should happen
+func (config *Config) ShouldEncryptWireguardHostTraffic() bool{
+	// currently, AKS *must* encrypt hostnetwork traffic, while it is optional on EKS
+	return config.KubernetesProvider() == ProviderAKS ||
+		(config.WireguardHostEncryptionEnabled && config.KubernetesProvider() == ProviderEKS)
```
mikestephen, line 474: According to the [docs](https://docs.tigera.io/compliance/encrypt-cluster-pod-traffic), we support Wireguard on AKS with CalicoCNI (without host encryption), so I _think_ the if statement needs to be:
`return config.WireguardHostEncryptionEnabled &&
    (config.KubernetesProvider() == ProviderEKS || config.KubernetesProvider() == ProviderAKS)`

config/config_params.go
```diff
@@ -467,6 +467,13 @@ func (config *Config) KubernetesProvider() Provider {
 	return ProviderNone
 }
 
+// ShouldEncryptWireguardHostTraffic factors in the kubernetes provider and config params to decide if host-traffic encryption should happen
+func (config *Config) ShouldEncryptWireguardHostTraffic() bool{
```
mikestephen, line 471: Also, might be worth adding a unit test for this function.

dataplane/linux/int_dataplane.go
```diff
@@ -1283,7 +1283,7 @@ func (d *InternalDataplane) setUpIptablesBPF() {
 
 		var rawRules []iptables.Rule
 		if t.IPVersion == 4 && rulesConfig.WireguardEnabled && len(rulesConfig.WireguardInterfaceName) > 0 &&
-			rulesConfig.RouteSource == "WorkloadIPs" {
```
petercork, line 1286: I think another option here would be to have it like 

`rulesConfig.RouteSource == "WorkloadIPs" && d.config.Wireguard.EncryptHostTraffic`

which should limit the feature to just workloadIPs i.e. not calicoCNI using calicoIPAM. Host-to-host might work with calicoCNI and IPAM perfectly but I guess it just widens the blast radius for testing for 3.9 if we say we have to now test essentially all platforms host-to-host. What do you think?

aaaaaaaalex: I'd agree with you about the blast radius, how about I add workload IP to [this](https://github.com/projectcalico/felix/pull/2897/files#diff-6d3b325b8dfb02061fd60c6641b100ad40c798ceaa2a27105b29e2af5cfb8847R473) check here, would that be functionally the same?

config/config_params.go
mikestephen: Maybe we should just really lock host to host encryption down to the platforms we know we can support, i.e.
```
p := config.KubernetesProvider()
r := configParams.RouteSource
return config.WireguardHostEncryptionEnabled &&
  (p == ProviderAKS && r == "WorkloadIPs" ||
  p == ProviderEKS && r == "CalicoIPAM")`

**Review by mikestephen — APPROVED** [review]
Looking good @aaaaaaaalex, nice job.

File: wireguard/wireguard.go [hunk 7]
```diff
@@ -1490,7 +1490,7 @@ func (w *Wireguard) ensureLinkAddressV4(netlinkClient netlinkshim.Interface) err
 		}
 		logCxt.Debug("Address set on wireguard device")
 	}
-
+	logCxt.Debug("Address not set on wireguard device")
```
fasaxc, line 1493: This log will fire even if the "Address set" log fires

dataplane/linux/int_dataplane.go
fasaxc: If it defaults to disabled then I think it'd be OK to not require WorkloadIPs, especially if you can kick the tires in CalicoIPAM mode and veryify that it's not totally broken.  You could document that it's only been tested with WorkloadIPs so should be considered "tech preview" in CalicoIPAM mode.

Most users won't switch away from the defaults so if we document it reasonably we might get some community testers who want that feature to work.

fasaxc (changes requested):
One minor bug and added my 2p on whether we should require WorkloadIPs.

wireguard/wireguard.go
aaaaaaaalex: good catch, rookie mistake on my part :laughing:

fasaxc (approved):
LGTM

aaaaaaaalex:
I can't edit the tags on this PR so can't add docs-pr-done, docs PR is here though: https://github.com/projectcalico/calico/pull/5041

PR #2910, opened 2021-07-22
BPF mode support for general untracked policy

neiljerram:
So far we have support in BPF mode for untracked ingress Deny policies.  This PR adds untracked ingress Allow policies and untracked egress policy, with the result that we should then support any untracked policy.


bpf-gpl/xdp.c
```diff
@@ -84,10 +84,18 @@ static CALI_BPF_INLINE int calico_xdp(struct xdp_md *xdp)
...
-		goto allow_with_metadata;
+		ctx.state->pol_rc = CALI_POL_ALLOW;
+		goto allow;
+	}
+
+	// Allow a packet if it hits an entry in the outbound ports failsafe map
+	if (is_failsafe_out(ctx.state->ip_proto, ctx.state->sport, ctx.state->ip_src)) {
+		CALI_DEBUG("Outbound failsafe port: %d. Skip policy\n", ctx.state->sport);
+		ctx.state->pol_rc = CALI_POL_ALLOW;
+		goto allow;
```
mazdakn, line 98: the XDP program runs only in RX path. I am not totally sure about how the failsafe works, but is checking "is_failsafe_out" valid for RX path?

mazdakn: one more thing, since after matching against failsafe map we return XDP_PASS and leave XDP execution, do we need to set "state.pol_rc" field?

dataplane/linux/policy_mgr.go
```diff
@@ -47,11 +52,35 @@ func newPolicyManager(rawTable, mangleTable, filterTable iptablesTable, ruleRend
...
+			neededIPSets := set.New()
+			filteredChains := []*iptables.Chain(nil)
+			for _, chain := range chains {
+				if strings.Contains(chain.Name, string(rules.PolicyOutboundPfx)) {
+					filteredChains = append(filteredChains, chain)
+					neededIPSets.AddAll(chain.IPSetIDs())
+				}
+			}
+			chains = filteredChains
+			m.ipSetsCallback(neededIPSets)
```
fasaxc, line 82: Won't this only cover one policy?

fv/bpf_test.go
```diff
@@ -266,6 +266,8 @@ func describeBPFTests(opts ...bpfTestOpt) bool {
 				options.WireguardEnabled = true
 				// Enable Wireguard.
 				options.ExtraEnvVars["FELIX_WIREGUARDENABLED"] = "true"
+				// With Wireguard the default IptablesMarkMask of 0xffff0000 isn't enough.
```
fasaxc, line 269: Think we'll need to address that; maybe we ditch the high-order fixed bits in the BPF program marks?

ipsets/ipsets.go
```diff
@@ -626,7 +630,19 @@ func (s *IPSets) tryResync() (numProblems int, err error) {
 // 'ipset restore' session in order to minimise process forking overhead.  Note: unlike
 // 'iptables-restore', 'ipset restore' is not atomic, updates are applied individually.
 func (s *IPSets) tryUpdates() error {
```
fasaxc, line 632: I think it's worth deleting IP sets that are no longer needed too; otherwise when we debug from `ipset list` we'll see stale data and that'll be confusing.

iptables/rules.go
```diff
@@ -166,3 +166,20 @@ func (c *Chain) RuleHashes(features *Features) []string {
 	}
 	return hashes
 }
+
+func (c *Chain) IPSetIDs() (ipSetIDs []string) {
+	if c == nil {
+		return nil
+	}
+	for _, rule := range c.Rules {
+		for _, matchString := range []string(rule.Match) {
```
fasaxc, line 175: I think this for loop should be in a method on the `MatchCriteria` type so that the string parsing is alongside the code that generates the strings.  Then this loop can be simply:
```
for _, rule := range c.Rules {
  ipSetIDs = append(ipSetIDs, r.Match.IPSetIDs())
}
```

ipsets/ipsets.go
```diff
@@ -670,6 +686,9 @@ func (s *IPSets) tryUpdates() error {
 	// Ask each dirty IP set to write its updates to the stream.
 	var writeErr error
 	s.dirtyIPSetIDs.Iter(func(item interface{}) error {
+		if s.neededIPSetNames != nil && !s.neededIPSetNames.Contains(s.IPVersionConfig.NameForMainIPSet(item.(string))) {
```
fasaxc, line 689: How about a utility method `ipSetNeeded(name)` that did the nil check

fasaxc (changes requested):
Could be mistaken but I think there's a bug in the setting of the IP set filter (it only covers the most recently applied policy?)

bpf-gpl/xdp.c
neiljerram: Thanks @mazdakn , it definitely makes sense to add a comment about why this code is here, so I will do that.

neiljerram: @fasaxc The "goto allow" here is in fact just a fall through to TC processing.  It does not do the marking that would cause the TC program to exit early.

bpf/ut/xdp_test.go
```diff
@@ -162,7 +162,7 @@ var xdpTestCases = []xdpTest{
 			SrcPort: 54321,
 		},
 		Drop:     false,
-		Metadata: true,
+		Metadata: false,
```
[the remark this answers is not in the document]
neiljerram: It's the TC fallback case as discussed in other comments here: matches a failsafe so should drop through so TC can check against conntrack state.  I've fixed the test name.

dataplane/linux/policy_mgr.go
neiljerram: Good catch.  Added UT and fixing coding.

bpf-gpl/xdp.c
mazdakn: Ah, got it. This is a tricky one!

fasaxc (approved):
LGTM

neiljerram:
Docs at https://github.com/projectcalico/calico/pull/4858

PR #2911, opened 2021-07-23
Add policy and icmp reply programs to XDP

mazdakn:
## Description

Adding the default policy program to XDP to implement the behaviour for the window that Felix has not yet updated the policy program. In addition, adding the ICMP reply program which is needed by unit tests.

## Todos

## Release Note

```release-note
None required
```


neiljerram (changes requested):
Not sure but I don't think we want to call the epilogue program...

neiljerram (approved):
LGTM

PR #2918, opened 2021-07-27
XDP UT framework

mazdakn:
## Description

This PR allows to run unit tests in XDP mode. List of changes are:
- In bpf/ut directory a flag called "forXDP" is added to differentiate tc and xdp test setup. For example, to use correct arguments in bpftool.

## Todos

## Release Note

```release-note
None required
```


bpf/ut/ip_dec_ttl_test.go
```diff
@@ -23,10 +23,38 @@ import (
...
+
+		ip35 := *ipv4Default
+		ip35.TTL = 127
+		_, _, _, _, pktBytes, err = testPacket(nil, &ip35, nil, nil)
+		Expect(err).NotTo(HaveOccurred())
+
+		Expect(res.dataOut).To(Equal(pktBytes))
+	})
+}
+
```
mazdakn, line 53: This test does not make sense for XDP. It is just here to test the UT framework. Will be removed in the end.

bpf-gpl/ut/ip_dec_ttl.xdp.c
```diff
@@ -0,0 +1,31 @@
...
+	if (skb_refresh_validate_ptrs(ctx, UDP_SIZE)) {
+		ctx->fwd.reason = CALI_REASON_SHORT;
+		CALI_DEBUG("Too short\n");
+		return -1;
+	}
+
+	ip_dec_ttl(ctx->ip_header);
+
+	return 0;
+}
```
mazdakn, line 31: This file is only to test ip_dec_ttl in xdp mode, just to test the xdp ut framework. Will be removed in the end.

bpf/ut/bpf_prog_test.go
```diff
@@ -309,12 +311,13 @@ func initMapsOnce() {
 		ipsMap = ipsets.Map(mc)
 		stateMap = state.Map(mc)
 		testStateMap = state.MapForTest(mc)
-		jumpMap = jump.MapForTest(mc)
+		xdpJumpMap = MapForTest(mc)
```
neiljerram, line 314: nit: maybe move into its own package, e.g. "xdpjump" ?

bpf/ut/icmp_port_unreachable_test.go
```diff
@@ -31,7 +31,7 @@ func TestICMPPortUnreachable(t *testing.T) {
 	_, ipv4, _, _, pktBytes, err := testPacketUDPDefault()
 	Expect(err).NotTo(HaveOccurred())
 
-	runBpfUnitTest(t, "icmp_port_unreachable.c", func(bpfrun bpfProgRunFn) {
+	runBpfUnitTest(t, "icmp_port_unreachable.c", false, func(bpfrun bpfProgRunFn) {
```
neiljerram, line 34: Do we need the forXDP parameter in runBpfUnitTest?  If not, I think better to revert in order to save a few changes.

bpf/ut/main_test.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2020 Tigera, Inc. All rights reserved.
+// Copyright (c) 2020-2021 Tigera, Inc. All rights reserved.
```
neiljerram, line 1: Should revert this copyright change, when the file isn't changing in any other way.

neiljerram (changes requested):
Looking good, thanks.  I think there are just a couple more kinds of reversion needed.

- I don't think we need `forXDP` parameter in `runBpfUnitTest` calls, so please can we revert those changes?

- After other reversions, when a file is not changing in any way apart from the copyright line, we should remove that file completely from the PR (aka leave its copyright line as it was before).  You may need the `git commit` `--no-verify` option for this, to skip our hooks when removing those files from the commit.

Then one nit: maybe move the XDP jump map definition into its own package?  I don't feel strongly about this though.

Finally: is it intended that you haven't added new XDP tests yet?  I guess they might come in the next PR, but it's also possible that you missed `git add` for that file!

bpf/ut/bpf_prog_test.go
mazdakn: The only thing is that the xdpjump is only needed for ut. We don't need anywhere else. For that, I felt maybe it is better to define it somewhere else. Does it make sense?

neiljerram: OK, no worries, that's absolutely fine.

mazdakn: Well, just realised that the extra map is only needed for "runBpfUnitTest", so it is also removed now.

mazdakn:
> Finally: is it intended that you haven't added new XDP tests yet? I guess they might come in the next PR, but it's also possible that you missed `git add` for that file!

Yes, I'll use another PR for the actual tests. The thing is that initially I worked on the "runBpfUnitTest" which later on looked useless for XDP. Then I focused on "runBpfTest", but that needs metadata to work. After this PR, I'll start to make "runBpfTest" tests like failsafe ones to work.

neiljerram (approved):
LGTM - thanks @mazdakn !

PR #2929, opened 2021-07-28
Add UT for the control plane half of BPF mode XDP function

neiljerram:


mazdakn (approved):
After reading two times, I believe I understand it now. LGTM.

neiljerram:
Thanks @mazdakn

PR #2945, opened 2021-08-02
ease up wireguard metrics log levels

electricjesus:
## Description

`logrus.Error` seems too harsh overall for the metrics collection functionality for WireGuard. At best, should be more or less informational even if it encounters program errors

## Todos
None required

## Release Note

```release-note
None required
```

### DIFF
--- a/wireguard/metrics.go
+++ b/wireguard/metrics.go
@@ -124,7 +124,6 @@ func MustNewWireguardMetrics() *Metrics {
 func NewWireguardMetrics() (*Metrics, error) {
 	hostname, err := os.Hostname()
 	if err != nil {
-		logrus.WithError(err).Warn("cannot register wireguard metrics stats")
 		return nil, err
 	}
 	return NewWireguardMetricsWithShims(hostname, netlinkshim.NewRealWireguard, defaultCollectionRatelimit), nil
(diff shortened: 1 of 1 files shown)


wireguard/metrics.go
```diff
@@ -78,7 +78,7 @@ func (collector *Metrics) Describe(d chan<- *prometheus.Desc) {
 
 func (collector *Metrics) descsByDevice(d chan<- *prometheus.Desc, device *wgtypes.Device) {
 	if device == nil {
-		collector.logCtx.Error("BUG: called descsByDevice with nil device")
+		collector.logCtx.Debug("BUG: called descsByDevice with nil device")
```
caseydavenport, line 81: This one seems like a real error to me

wireguard/metrics.go
```diff
@@ -95,7 +95,7 @@ func (collector *Metrics) descsByDevice(d chan<- *prometheus.Desc, device *wgtyp
 
 func (collector *Metrics) descByPeer(d chan<- *prometheus.Desc, peer *wgtypes.Peer) {
 	if peer == nil {
-		collector.logCtx.Error("BUG: called descByPeer with nil peer")
+		collector.logCtx.Debug("BUG: called descByPeer with nil peer")
```
caseydavenport, line 98: same here - we should never hit this unless something is actually very wrong, so seems good to yell about it.

wireguard/metrics.go
```diff
@@ -149,13 +149,13 @@ func NewWireguardMetricsWithShims(hostname string, newWireguardClient func() (ne
 func (collector *Metrics) getDevices() []*wgtypes.Device {
 	wgClient, err := collector.newWireguardClient()
 	if err != nil {
-		collector.logCtx.WithError(err).Error("error initializing wireguard client devices")
+		collector.logCtx.WithError(err).Warn("error initializing wireguard client devices")
```
caseydavenport, line 152: nit: I'd also remove the word error from this. It leads users to think the log is an error, even if the severity is warning.

caseydavenport (approved):

PR #2946, opened 2021-08-03
Update to AWS API v2.

fasaxc:
## Description

The v2 API uses far fewer pointers in its data and it drops the non-Context API variants, slimming the API a lot.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```


go.mod
```diff
@@ -33,10 +37,11 @@ require (
 	github.com/sirupsen/logrus v1.7.0
 	github.com/spf13/cobra v1.1.1
 	github.com/spf13/viper v1.7.0
+	github.com/tigera/api v0.0.0-20210803092542-1ba9f4618d05
```
neiljerram, line 40: Is that right?  I'd expect only projectcalico/api in OSS Felix.

go.sum
```diff
@@ -687,6 +712,8 @@ github.com/syndtr/gocapability v0.0.0-20200815063812-42c35b437635/go.mod h1:hkRG
 github.com/tchap/go-patricia/v2 v2.3.1/go.mod h1:VZRHKAb53DLaG+nA9EaYYiaEx6YztwDlLElMsnSHD4k=
 github.com/thecodeteam/goscaleio v0.1.0/go.mod h1:68sdkZAsK8bvEwBlbQnlLS+xU+hvLYM/iQ8KXej1AwM=
 github.com/tidwall/pretty v1.0.0/go.mod h1:XNkn88O1ChpSDQmQeStsy+sBenx6DDtFZJxhVysOjyk=
+github.com/tigera/api v0.0.0-20210803092542-1ba9f4618d05 h1:96aLjQjs4qXlvtemv3I52HpgPcAKbPPZdplv0Y0xrDY=
+github.com/tigera/api v0.0.0-20210803092542-1ba9f4618d05/go.mod h1:KZ+Ck27bbY+L0sJap9f9eJy1wMEWrssdY0oOOSb0bnM=
```
neiljerram, line 716: Can revert these if tigera/api shouldn't actually be imported.

neiljerram (commented):
Generally looks reasonable, but one query if it's right for go.mod to pull in tigera/api

go.mod
fasaxc: Good catch.  Copied the code out of felix-private and forgot that I needed to change that

neiljerram (approved):

PR #2949, opened 2021-08-05
Align setting of BPF policy result with Enterprise code

neiljerram:


bpf/polprog/pol_prog_builder.go
```diff
@@ -265,7 +265,7 @@ func (p *Builder) writeProgramHeader() {
 	p.b.LoadMapFD(R1, uint32(p.stateMapFD)) // R1 = 0 (64-bit immediate)
 	p.b.Call(HelperMapLookupElem)           // Call helper
 	// Check return value for NULL.
-	p.b.JumpEqImm64(R0, 0, "deny")
+	p.b.JumpEqImm64(R0, 0, "exit")
```
fasaxc, line 268: Feel like "exit" doesn't convey that the packet will be dropped. "drop-and-exit" or "error-exit" maybe?

fasaxc (approved):
One suggestion but LGTM

bpf/polprog/pol_prog_builder.go
neiljerram: This is the naming that already exists in Enterprise - does that change the equation at all?

PR #2944, opened 2021-08-02
XDP UT and fixes to the XDP program

mazdakn:
## Description

This PR includes the following related to XDP:
- Some changes to support running XDP UT
- Some XDP UT test cases
- Some fixes to the XDP program and shared functions with TC

## Todos
- [x] Unit tests (full coverage)
- [ ] Integration tests (delete as appropriate) In plan/Not needed/Done

## Release Note

```release-note
None required
```

PR #2950, opened 2021-08-05
Correction to polprog building for XDP

neiljerram:
Just spotted a small error that I missed when reviewing #2944 

Also think it's good to apply a couple of commits that I pushed to the UT infrastructure code in https://github.com/tigera/felix-private/pull/1385, so that we have the same code in OSS and Enterprise.


mazdakn:
LGTM:-)

mazdakn (approved):
LGTM:-)

PR #2963, opened 2021-08-17
Load TC programs using libbpf

sridhartigera:
## Description
TC programs are loaded using 'tc' utility. Moving this to use libbpf
## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```


bpf-gpl/tc.c
```diff
@@ -436,7 +437,7 @@ static CALI_BPF_INLINE int calico_tc(struct __sk_buff *skb)
 	goto finalize;
 }
 
-__attribute__((section("1/1")))
+SEC("classifier/1")
```
tomastigera, line 440: Iirc the 1/1 convention was for `tc` to be able to load it. Since we are loading it ourselves now, can we have more meaningful section names now? Like `tc/main` or `tc/icmp` etc. ?

bpf/tc/attach.go
```diff
@@ -73,6 +82,9 @@ func (ap AttachPoint) Log() *log.Entry {
 func (ap AttachPoint) AttachProgram() error {
 	logCxt := log.WithField("attachPoint", ap)
 
+	if optsMap == nil {
```
tomastigera, line 85: The map could be pre-created at init time and the tests could be removed.

bpf/tc/attach.go
```diff
@@ -512,3 +537,9 @@ func (ap *AttachPoint) JumpMapFDMapKey() string {
 func (ap *AttachPoint) IfaceName() string {
 	return ap.Iface
 }
+
+func GetTCOpts(ifaceName, hook string) bool {
+	key := ifaceName + "_" + hook
```
tomastigera, line 542: even if just for testing now, might be worth using the RLock to prevent future burns if anybody uses it for non-test code :sweat_smile:

bpf/tc/attach.go
```diff
@@ -472,34 +496,35 @@ func RemoveQdisc(ifaceName string) error {
...
-		if strings.Contains(line, progName) {
-			re := regexp.MustCompile(`id (\d+)`)
-			m := re.FindStringSubmatch(line)
-			if len(m) > 0 {
-				return m[1], nil
+	key := ap.Iface + "_" + string(ap.Hook)
+	if optsMap != nil {
+		optsLock.RLock()
+		defer optsLock.RUnlock()
+		if val, ok := optsMap[key]; ok {
```
tomastigera, line 520: If the access to the optsMap was wrapped in functions like the `GetTCOpts` this could read just `if ExistsTCOpts() { ... }` or `if TCOptsExists(...)` ... (also you would not hold the lock for the rest of the function - i guess it does not matter here though)

tomastigera (commented):
Great to see this happening. LGTM in general, my comments are purely coding nits.

bpf-gpl/Makefile
```diff
@@ -26,7 +26,8 @@ CFLAGS +=  \
 	-fno-stack-protector \
 	-O2 \
 	-target bpf \
-	-emit-llvm
+	-emit-llvm \
+	-g
```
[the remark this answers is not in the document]
sridhartigera: Yes -g is a requirement for BTF

bpf-gpl/jump.h
```diff
@@ -32,6 +32,14 @@ static CALI_BPF_INLINE struct cali_tc_state *state_get(void)
 	return cali_v4_state_lookup_elem(&key);
 }
 
+#ifdef __LIBBPF_LOADER__
+struct bpf_map_def_extended __attribute__((section("maps"))) cali_jump = {
```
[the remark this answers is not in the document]
sridhartigera: I dont think its possible now. Since CALI_MAP has versioning

bpf/tc/attach.go
```diff
@@ -512,3 +549,39 @@ func (ap *AttachPoint) JumpMapFDMapKey() string {
...
 }
+
+func GetTCOpts(key string) (*libbpf.TCOpts, bool) {
+	optsLock.RLock()
+	defer optsLock.RUnlock()
+	val, ok := optsMap[key]
+	return val, ok
+}
+
+func InitTcOpts() {
```
tomastigera, line 560:
```suggestion
func InitTCOpts() {
```

I think you can just do vvv without the need of actually calling this function anywhere

`var optsMap = map[string]*libbpf.TCOpts{}`

https://play.golang.org/p/scOfvfYwxSG

bpf/tc/attach.go
```diff
@@ -512,3 +549,39 @@ func (ap *AttachPoint) JumpMapFDMapKey() string {
...
+        optsMap = make(map[string]*libbpf.TCOpts)
+}
+
+// nolint
+func updateTcOpts(key string, opts *libbpf.TCOpts) {
```
tomastigera, line 565:
```suggestion
func updateTCOpts(key string, opts *libbpf.TCOpts) {
```

bpf/tc/attach.go
```diff
@@ -512,3 +549,39 @@ func (ap *AttachPoint) JumpMapFDMapKey() string {
...
+        optsLock.Lock()
+        defer optsLock.Unlock()
+        optsMap[key] = opts
+}
+
+func UpdateJumpMap(obj *libbpf.Obj, isHost bool) error {
```
tomastigera, line 571: any need for this to be exported?

bpf/libbpf/libbpf_api.h
```diff
@@ -0,0 +1,90 @@
...
+#include "libbpf.h"
+#include <linux/limits.h>
+#include <net/if.h>
+#include <bpf.h>
+#include <stdlib.h>
+#include <errno.h>
+
+struct bpf_object* bpf_obj_open_load(char *filename) {
+	errno = 0;
+	return bpf_object__open(filename);
```
fasaxc, line 24: Are you sure that `bpf_object__open` sets `errno` in all failure cases?  For a case where the file doesn't exist, I'd expect errno to be set as a side effect of `bpf_object__open` calling `open` under the covers but for other errors, I'm not sure if `bpf_object__open` would set `errno` or if it's return a fake error pointer.

Sadly, in C, it's hard to be sure without code reading/documentation.

bpf/libbpf/libbpf_api.h
```diff
@@ -0,0 +1,90 @@
...
+	DECLARE_LIBBPF_OPTS(bpf_tc_hook, hook, .attach_point = BPF_TC_EGRESS);
+	DECLARE_LIBBPF_OPTS(bpf_tc_opts, attach);
+
+	if (isIngress) {
+		hook.attach_point = BPF_TC_INGRESS;
+	}
+
+	attach.prog_fd = bpf_program__fd(bpf_object__find_program_by_name(obj, secName));
+	if (attach.prog_fd < 0) {
+		opts.prog_fd = attach.prog_fd;
```
fasaxc, line 40: This looks like an error case did you mean to set errno?

fasaxc (changes requested):

bpf/libbpf/libbpf.go
```diff
@@ -0,0 +1,200 @@
...
+
+func (o *Obj) Load() error {
+	err := C.bpf_object__load(o.obj)
+	if err != 0 {
+		return fmt.Errorf("error loading object %v", err)
+	}
+	return nil
+}
+
+func (o *Obj) FirstMap() (*Map, error) {
```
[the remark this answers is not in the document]
sridhartigera: NextMap will return ```nil``` if its the last map and that is not an error. Am I right?

bpf/libbpf/libbpf.go
```diff
@@ -0,0 +1,200 @@
...
+	QdiskIngress QdiskHook = "ingress"
+	QdiskEgress  QdiskHook = "egress"
+)
+
+func (m *Map) Name() string {
+	name, err := C.bpf_map__name(m.bpfMap)
+	if err != nil {
+		return ""
+	}
+	return C.GoString(name)
```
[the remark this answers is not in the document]
sridhartigera: Free resulted in a crash. Its a double free

bpf/libbpf/libbpf.go
```diff
@@ -0,0 +1,207 @@
...
+	cPath := C.CString(path)
+	defer C.free(unsafe.Pointer(cPath))
+	err := C.bpf_map__set_pin_path(m.bpfMap, cPath)
+	if err != 0 {
+		return fmt.Errorf("pinning map failed %v", err)
+	}
+	return nil
+}
+
+// bpf_obj__open does not set errno. Errno is returned in the obj.
```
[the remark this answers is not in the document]
sridhartigera: I think this is an older version of the file. I have removed this comment.

bpf/libbpf/libbpf_api.h
sridhartigera: I overlooked the libbpf version. I had libbpf version 0.5 in my laptop and that had proper error handling. Guess will be better to move to libbpf v0.5.0. WDYT?

sridhartigera: Tried with v0.5.0, but looks like relying on errno set by libbpf may not help

tomastigera: libbpf's convention is to return `0` on success and `-errno` (in 0.4.0) changed to `+errno` later on. It does not set `errno`. So as @sridhartigera and I discussed, imo the solution is to call into libbpf and than set the `errno`. It is not clear whether you need to return a negative return value. From @fasaxc 's description it seems like you do not need to.

I'd say that to be compatible with future upgrades of libbpf, you should do

```
void myfunc(void)
{
    int ret = libbpf_some_functio();
    errno = ret >= 0 ? : -ret;
}
```

bpf/libbpf/libbpf_api.h
```diff
@@ -0,0 +1,96 @@
...
+
+#include "libbpf.h"
+#include <linux/limits.h>
+#include <net/if.h>
+#include <bpf.h>
+#include <stdlib.h>
+#include <errno.h>
+
+static void set_errno(int ret) {
+	errno = ret >=0 ? 0 : -ret;
```
tomastigera, line 23:
```suggestion
	errno = ret >= 0 ? ret : -ret;
```
as non-zero is always an error and the polarity depends on libbpf version?

bpf/libbpf/libbpf.go
```diff
@@ -0,0 +1,200 @@
...
+	if err != nil {
+		return -1
+	}
+	return int(mapType)
+}
+
+func (m *Map) SetPinPath(path string) error {
+	cPath := C.CString(path)
+	defer C.free(unsafe.Pointer(cPath))
+	err := C.bpf_map__set_pin_path(m.bpfMap, cPath)
```
fasaxc, line 68: I think this error handling will work, comparing with the libbpf code.  It looks like it returns `-errno` on failure.

bpf/libbpf/libbpf.go
```diff
@@ -0,0 +1,200 @@
...
+	}
+	return nil
+}
+
+func OpenObject(filename string) (*Obj, error) {
+	bpf.IncreaseLockedMemoryQuota()
+	cFilename := C.CString(filename)
+	defer C.free(unsafe.Pointer(cFilename))
+	obj, err := C.bpf_obj_open(cFilename)
+	if obj == nil || err != nil {
```
fasaxc, line 80: I think you should make sure the C code only returns `obj==nil` if `err!=nil`

bpf/libbpf/libbpf_api.h
```diff
@@ -0,0 +1,96 @@
...
+	DECLARE_LIBBPF_OPTS(bpf_tc_opts, attach);
+
+	if (isIngress) {
+		hook.attach_point = BPF_TC_INGRESS;
+	}
+
+	attach.prog_fd = bpf_program__fd(bpf_object__find_program_by_name(obj, secName));
+	if (attach.prog_fd < 0) {
+		errno = -attach.prog_fd;
+		return opts;
```
fasaxc, line 50: `opts` is uninitialised here.  Returning uninitialised memory can cause problems for CGO's validation logic.  Suggest zeroing it out before returning.

bpf/libbpf/libbpf_api.h
```diff
@@ -0,0 +1,96 @@
...
+
+void bpf_tc_remove_qdisc (int ifIndex) {
+        DECLARE_LIBBPF_OPTS(bpf_tc_hook, hook, .attach_point = BPF_TC_EGRESS | BPF_TC_INGRESS);
+        hook.ifindex = ifIndex;
+        set_errno(bpf_tc_hook_destroy(&hook));
+        return;
+}
+
+int bpf_tc_update_jump_map(struct bpf_object *obj, char* mapName, char *progName, int progIndex) {
+	int prog_fd = bpf_program__fd(bpf_object__find_program_by_name(obj, progName));
```
fasaxc, line 84: Can `bpf_object__find_program_by_name` fail?  Should do that operation first and return early if it fails, I think.

bpf/tc/attach.go
```diff
@@ -286,7 +322,7 @@ func (ap AttachPoint) IsAttached() (bool, error) {
 
 // tcDirRegex matches tc's auto-created directory names so we can clean them up when removing maps without accidentally
 // removing other user-created dirs..
-var tcDirRegex = regexp.MustCompile(`[0-9a-f]{40}`)
+var tcDirRegex = regexp.MustCompile(`[0-9a-f]`)
```
fasaxc, line 325: Do we need a regex that covers your new directory names?

bpf/tc/attach.go
```diff
@@ -514,3 +528,22 @@ func (ap *AttachPoint) JumpMapFDMapKey() string {
 func (ap *AttachPoint) IfaceName() string {
 	return ap.Iface
 }
+
+// nolint
+func updateJumpMap(obj *libbpf.Obj, isHost bool) error {
```
fasaxc, line 533: So this function is setting up the default contents of the jump map?  (Because libbpf doesn't do that for us?)

bpf/tc/attach.go
```diff
@@ -100,20 +100,57 @@ func (ap AttachPoint) AttachProgram() error {
...
-		return err
+		return nil, fmt.Errorf("error loading program %v", err)
+	}
+
+	isHost := false
+	if ap.Type == "host" {
+		isHost = true
+	}
+
+	err = updateJumpMap(obj, isHost)
```
fasaxc, line 141: Since we can modify the map before attaching, would be great to install the right policy program at this stage rather than a temp one.  Feels like a follow-on piece of work though.

bpf/tc/compiler.go
```diff
@@ -48,6 +48,20 @@ const (
...
+type ProgName string
+
+const (
+	policyProgram ProgName = "calico_tc_norm_pol_tail"
+	allowProgram  ProgName = "calico_tc_skb_accepted_entrypoint"
+	icmpProgram   ProgName = "calico_tc_skb_send_icmp_replies"
+)
+
+const (
+	POLICY_PROGRAM_INDEX = iota
```
fasaxc, line 60: Nit: golang naming convention is to name constants like normal variables. PolicyProgramIndex, for example.

bpf/ut/attach_test.go
```diff
@@ -56,26 +56,19 @@ func TestJumpMapCleanup(t *testing.T) {
...
 		startingJumpMaps := countJumpMaps()
 		startingTCDirs := countTCDirs()
 		ap.HostIP = net.ParseIP("10.0.0.1")
 		ap.IntfIP = net.ParseIP("10.0.0.2")
 		err := tc.EnsureQdisc(ap.Iface)
 		Expect(err).NotTo(HaveOccurred())
-		err = ap.AttachProgram()
+		opts, err := ap.AttachProgram()
 		Expect(err).NotTo(HaveOccurred())
+		Expect(opts).NotTo(Equal(nil))
```
fasaxc, line 68: Should be `NotTo(BeNil())`

bpf/ut/attach_test.go
```diff
@@ -56,26 +56,19 @@ func TestJumpMapCleanup(t *testing.T) {
...
 		Expect(countJumpMaps()).To(BeNumerically("==", startingJumpMaps+1), "unexpected number of jump maps")
 		Expect(countTCDirs()).To(BeNumerically("==", startingTCDirs+1), "unexpected number of TC dirs")
 
-		t.Log("Replacing program should add another map and dir.")
-		ap.HostIP = net.ParseIP("10.0.0.2")
-		err = ap.AttachProgram()
```
fasaxc, line 74: You should keep the code that does a second attach but remove the lines that expect 2 jump maps; there should only be one even though you reattached the program.

bpf/libbpf/libbpf.go
```diff
@@ -0,0 +1,200 @@
...
+
+	if hook == string(QdiskIngress) {
+		isIngress = 1
+	}
+
+	opts, err := C.bpf_tc_program_attach(o.obj, cSecName, C.int(ifIndex), C.int(isIngress))
+	if err != nil {
+		return nil, fmt.Errorf("Error attaching tc program %w", err)
+	}
+	return &TCOpts{opts: opts}, nil
```
fasaxc, line 132: You could do the prog ID lookup here and return that as an `int`; would that be a nicer API?

bpf/libbpf/libbpf.go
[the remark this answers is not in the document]
fasaxc: Worth commenting to that effect with a `godoc`, I think

fasaxc (changes requested):

bpf/tc/attach.go
sridhartigera: Right. libbpf doesn't load the jump maps with program indices

bpf/libbpf/libbpf_api.h
```diff
@@ -26,7 +26,12 @@ static void set_errno(int ret) {
 struct bpf_object* bpf_obj_open(char *filename) {
 	struct bpf_object *obj;
 	obj = bpf_object__open(filename);
-	set_errno(libbpf_get_error(obj));
+	int err = libbpf_get_error(obj);
+	if (err) {
+		free(obj);
```
fasaxc, line 31: If err is non-zero I think you don't need to free anything because it won't be a valid pointer?

bpf/tc/attach.go
```diff
@@ -441,7 +434,7 @@ func CleanUpJumpMaps() {
 		if err != nil {
 			return err
 		}
-		if info.IsDir() && tcDirRegex.MatchString(info.Name()) {
+		if info.IsDir() && info.Name() != "globals" {
```
fasaxc, line 437: Can we restrict to only the dirs that we expect?  I.e. the old regex unioned with whatever new dirs you're expecting?  don't want to break other apps that might have a dir in here.

bpf/tc/attach.go
```diff
@@ -509,14 +502,8 @@ func RemoveQdisc(ifaceName string) error {
...
 
-func (ap *AttachPoint) ProgramID(tcOpts *libbpf.TCOpts) (string, error) {
-	logCtx := log.WithField("iface", ap.Iface)
-	logCtx.Info("Finding TC program ID")
-	progId, err := libbpf.GetProgID(ap.Iface, string(ap.Hook), tcOpts)
-	if err != nil {
-		return "", errors.New("failed to find TC program")
-	}
-	return strconv.Itoa(progId), nil
+func (ap *AttachPoint) ProgramID() (string, error) {
```
fasaxc, line 505: Is this no longer used?

sridhartigera: Yeah. Since we are returning the program ID in Attach. This can be removed. Need to do some cleanup in xdp as well.

fasaxc (approved):
LGTM

PR #2964, opened 2021-08-17
Fix slow performance of updating a namespace.

fasaxc:
## Description

Previously we assumed that namespace updates were rare so
we did a full scan of the endpoint index to find any endpoints
matching a particular namespace.  That assumption doesn't hold
on clusters with many namespaces because we need to process
the namespaces at start of day.

Add an index that tracks the enpdoints that have a given
parent (i.e. namespace) so that we can do a fast lookup.

Add benchmarks to verify the change.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
Fix slow performance when updating a Kubernetes namespace when there are many Pods (and in turn, slow startup performance when there are many namespaces).
```

### DIFF
--- a/labelindex/named_port_bench_test.go
+++ b/labelindex/named_port_bench_test.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2020 Tigera, Inc. All rights reserved.
+// Copyright (c) 2020-2021 Tigera, Inc. All rights reserved.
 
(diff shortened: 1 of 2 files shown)


robbrockbank (approved):
LGTM

PR #2966, opened 2021-08-18
remove unions in cali_tc_ctx/state structure

mazdakn:
## Description

This PR corrects the type punning in bpf code by removing the unions in the `cai_tc_ctx` structure.

## Todos

## Release Note

```release-note
None required
```

### DIFF
--- a/bpf-gpl/parsing.h
+++ b/bpf-gpl/parsing.h
@@ -39,7 +39,7 @@ static CALI_BPF_INLINE int parse_packet_ip(struct cali_tc_ctx *ctx) {
 			CALI_DEBUG("Too short\n");
 			goto deny;
 		}
-		protocol = bpf_ntohs(parsing_ethhdr(ctx)->h_proto);
+		protocol = bpf_ntohs(tc_ethhdr(ctx)->h_proto);
 	} else {
 		protocol = bpf_ntohs(ctx->skb->protocol);
 	}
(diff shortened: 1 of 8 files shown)


bpf-gpl/parsing.h
```diff
@@ -22,6 +22,11 @@
 #define PARSING_ERROR -1
 #define PARSING_ALLOW_WITHOUT_ENFORCING_POLICY -2
 
+#define ETHHDR(x) ((struct ethhdr *)x->data_start)
```
tomastigera, line 25: nit: these only work on tc context, what if we prefix them with TC_ ? What if they were functions that return the typecasted type and take typed cali_tc_ctx input? Would it work?

Why not to define them in types.h where the cali_tc_ctx is defined?

tomastigera (commented):
looks OK to me

bpf-gpl/parsing.h
mazdakn: Just noticed that in types.h only the structures are defined, and the rest of functions are either in skb.h or parsing.h. It looks like those 3 files has a lot in common. The functions are little scattered, for example `ip_dec_ttl()` is in bpf.h. Maybe we should sort them out?

mazdakn: > these only work on tc context, what if we prefix them with TC_ ? What if they were functions that return the typecasted type and take typed cali_tc_ctx input? Would it work?
That would definitely work. Honestly between macros and functions, I prefer functions:-D I'll switch to function.

mazdakn: @tomastigera I think you right. it is better to move these to types.h.

tomastigera (approved):
LGTM thanks

PR #2980, opened 2021-08-23
Abolish BPF mode's fixed Calico mark pattern

neiljerram:


tomastigera:
What is the rational? To free up a bit or two? (:+1: )

neiljerram:
@tomastigera Yes, exactly.

neiljerram:
@tomastigera
- Default IptablesMarkMask (0xFFFF0000) is 16 bits.
- BPF mode did use 11 bits (0x7FF00000)
- Because we use iptables for untracked egress policy, we need more of the iptables-mode bits, even in BPF mode
- And then Enterprise has various further bits that it tries to allocate for various features.

Net, we think it's really worthwhile to save the 2 Calico BPF fixed pattern bits now.

tomastigera (approved):
:cool:

PR #2981, opened 2021-08-23
Move IP sets manager to common location

neiljerram:
This reflects the fact that ipsets_mgr really is OS-independent, and is to ease future merges against the Calico Enterprise code.


neiljerram:
@song-jiang I just took a quick look at your additional changes, and FWIW they look good to me.

song-jiang (approved):
LGTM

PR #2983, opened 2021-08-26
bpf: source port collision resolution

tomastigera:
## Problem
A single client can connect to a NodePort via any node. The client is free to use the same source ip:port to connect to two different nodes. BPF NodePorts forward the connection using a VXLAN tunnel to the node which hosts a backend pod. The source ip:port is preserved when the node decaps the traffic from the tunnel and when the packets are forwarded to the backend pod. If 2 such connections that were originally directed to different nodes end up being forwarded to the same backend pod, BPF conntrack and the pod cannot distinguish between them as they have the exact same 5-tuple.

In such a case, the packets look like they belong to the same connection, however, they have different sequence numbers, unexpected SYN is seen on the connection, etc. and thus it results in retransmissions and/or the backed pod sending RST to break the connections. This is observed as hanging connections, slow connections and breaking connections.

## Solution
We already stop any packet if we detected that the source node
changed - part of our RPF check.

We now also stop any packet that would have a conflicting tracking
record after NAT. This means that it has the same 5-tuple after NAT,
which means that the source is the same and the destination is the
same but the destination was different before NAT.

The detection is race free in respect to any concurrent connection as we
first try to create the tracking entry atomically and if that fails, no
fwd entry it created.

If there is a collision on the source due to forwarding via multiple
nodes, we pick a new source port, we remember the source port in
conntrack and we SNAT the port on the way towards the pod and we
translate it back for the return traffic. That port is not seen outside
of the node.

We remember the new port in the FWD entry an we also remember the 
original source port in the reverse entry. We use currenty unsed space
in the entries, hence there is no need to version the conntrack table.

refs https://github.com/projectcalico/calico/issues/4661
refs https://github.com/projectcalico/calico/issues/4431

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```


fasaxc (dismissed):
I think this will stop the old conntrack entry from being clobbered but what stops the new packet from matching the existing conntrack entry.  Do we already have something in place?

bpf-gpl/conntrack.h
```diff
@@ -26,6 +26,14 @@
 
 // Connection tracking.
 
+#define PSNAT_START	20000
+#define PSNAT_NUM	10000
```
tomastigera, line 30: I'd like t make this configurable

tomastigera, line 30: I'd like t make this configurable

bpf-gpl/tc.c
```diff
@@ -820,9 +825,19 @@ static CALI_BPF_INLINE struct fwd calico_tc_skb_accepted(struct cali_tc_ctx *ctx
 
 		switch (ctx->ip_header->protocol) {
 		case IPPROTO_TCP:
+			if (state->ct_result.nat_sport) {
+				CALI_DEBUG("Fixing source port from %d to %d\n",
```
neiljerram, line 829: Add in "TCP" here, and "UDP" just below?

bpf-gpl/tc.c
```diff
@@ -904,19 +922,30 @@ static CALI_BPF_INLINE struct fwd calico_tc_skb_accepted(struct cali_tc_ctx *ctx
 		switch (ctx->ip_header->protocol) {
 		case IPPROTO_TCP:
 			tc_tcphdr(ctx)->source = bpf_htons(state->ct_result.nat_port);
+			if (state->ct_result.nat_sport) {
+				CALI_DEBUG("Fixing dest port from %d to %d\n",
```
neiljerram, line 926: Perhaps "TCP" and "UDP" again here, and also something to indicate that it's a reverse SNAT as opposed to a forwards SNAT.

bpf-gpl/conntrack.h
```diff
@@ -200,7 +208,40 @@ static CALI_BPF_INLINE int calico_ct_v4_create_tracking(struct ct_create_ctx *ct
...
+			*k = (struct calico_ct_key) {
+				.protocol = ct_ctx->proto,
+					.addr_a = ip_dst, .port_a = dport,
+					.addr_b = ip_src, .port_b = sport,
+			};
+		}
+
+		dump_ct_key(k);
+
+		if ((err = cali_v4_ct_update_elem(k, &ct_value, BPF_NOEXIST))) {
```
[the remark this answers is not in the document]
neiljerram: I wonder if we should try calling psnat_get_port again?  Maybe we were just unlucky.

bpf/ut/bpf_prog_test.go
```diff
@@ -680,6 +681,34 @@ func udpResposeRaw(in []byte) []byte {
 	return out.Bytes()
 }
 
+func tcpResposeRaw(in []byte) []byte {
```
neiljerram, line 684: s/respose/response/

neiljerram (commented):
I think I understand this now, and that it's generally looking good.  Have dropped a few minor thoughts.  I think probably the only bigger one is that we should probably be cleverer about calling psnat_get_sport, repeatedly if necessary, to make sure that we get an sport that isn't yet in use for the rest of the tuple.

bpf-gpl/conntrack.h
```diff
@@ -200,7 +208,40 @@ static CALI_BPF_INLINE int calico_ct_v4_create_tracking(struct ct_create_ctx *ct
 		CALI_DEBUG("CT-ALL Whitelisted dest side - to EP\n");
 	}
 
-	err = cali_v4_ct_update_elem(k, &ct_value, 0);
+	err = cali_v4_ct_update_elem(k, &ct_value, BPF_NOEXIST);
+
+	if (err == -17 /* EEXIST */) {
+		CALI_DEBUG("Source collision for 0x%x:%d\n", bpf_htonl(ip_src), sport);
```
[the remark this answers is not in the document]
tomastigera: It is possible, that there is a connection with the same tuple that is already expired but not collected by the cleaner yet. We do not treat expiration in BPF yet. It is racy with the cleaner. Nevertheless, it is a known issue that is tcp_tw_reuse is set, we can hit this problem (tracked as a separate issue). Afaict this would also (partially) resolve that tcp_tw_reuse issue as it would pick a different port. Not sure if that is the best resolution. We may need to do the expiration in BPF and be more clever about the race.

bpf-gpl/conntrack.h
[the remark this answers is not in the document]
tomastigera: yes, we can be unlucky, then we drop the packet. However, it will try to send another SYN (up to 3?) so it may pick a new port then. The connection may time out, but hat is still better than breaking the existing connection.

I am not sure if we can easily keep calling it - that is what I would do outside of BPF :sweat_smile: 

We were considering sending back and RST to avoid the timeout. I just did not get to the RST part yet. TBH I would like to see how the random pick behaves in the wild firts - but there is no harm in adding the RST, except that middle-boxes shouldn't send one :sweat_smile:

bpf-gpl/conntrack.h
```diff
@@ -200,7 +208,40 @@ static CALI_BPF_INLINE int calico_ct_v4_create_tracking(struct ct_create_ctx *ct
...
+
+		ct_ctx->orig_sport = ct_value.orig_sport = sport;
```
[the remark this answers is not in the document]
tomastigera: the first assignment here is duplicate of line 76 (probably 76 added later), good catch!

bpf-gpl/conntrack.h
```diff
@@ -200,7 +208,40 @@ static CALI_BPF_INLINE int calico_ct_v4_create_tracking(struct ct_create_ctx *ct
...
+					.addr_a = ip_src, .port_a = sport,
+					.addr_b = ip_dst, .port_b = dport,
+			};
+		} else  {
+			*k = (struct calico_ct_key) {
+				.protocol = ct_ctx->proto,
+					.addr_a = ip_dst, .port_a = dport,
+					.addr_b = ip_src, .port_b = sport,
+			};
+		}
```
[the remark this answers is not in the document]
tomastigera: yeah, I added `ct_make_key()` while back :sweat_smile:

bpf-gpl/conntrack.h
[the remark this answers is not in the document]
tomastigera: I added a test that demonstrates that another collision is possible and that a retransmission is likely to fix this.

bpf-gpl/conntrack.h
```diff
@@ -243,6 +284,9 @@ static CALI_BPF_INLINE int calico_ct_v4_create_nat_fwd(struct ct_create_ctx *ct_
 
 	dump_ct_key(&k);
 	ct_value.nat_rev_key = *rk;
+	if (ct_ctx->orig_sport) {
```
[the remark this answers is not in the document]
tomastigera: Yeah, the intention of this if is to set the entry so that it keeps changing the port  for subsequent packets only if there was a collision. It is meaningless as is, since it is always set (great catch) and it sets it for all entries. I will fix that.

bpf/ut/nat_test.go
```diff
@@ -1902,4 +1910,79 @@ func TestNATSourceCollision(t *testing.T) {
 
 		recvPkt = res.dataOut
 	})
+
+	// Test random port conflict by sending another SYN packet. To avoid the
+	// complexity of VXLAN encap in the test, send it with a different node IP,
+	// which in fact mimics as if this node had 2 IPs. That is a realistic
+	// scenario, but the main reason is to excercise the conflict after the
```
neiljerram, line 1917: nit: exercise

bpf/ut/nat_test.go
```diff
@@ -1902,4 +1910,79 @@ func TestNATSourceCollision(t *testing.T) {
...
+		Expect(tcpL).NotTo(BeNil())
+
+		tcp := tcpL.(*layers.TCP)
+		newSPort = uint16(tcp.SrcPort)
+		if newSPort != uint16(22223) {
+			return fmt.Errorf("Unexpected resolution port")
+		}
+
+		return nil
+	}, "120s").Should(Succeed())
```
neiljerram, line 1987: I think we should write this more deterministically so that it cannot possibly flake.  What we really care about is that a packet can go through even after the previous N retransmissions were dropped.  So how about testing twice with (22222, 22222) and then once with (22223, 22223), and checking that the last one is successful?

neiljerram (approved):
LGTM.  There's one typo and one little bit of test restructuring that I've suggested, but I'm happy to leave those to your judgment.

bpf/ut/nat_test.go
tomastigera: I was thinking about doing it this way, but it would not really test what I want to test. I would change it if it proves flaky, but (a) with the 3x loop in BPF it rarely does the `Eventually` loop and (b) 2m is imo so many tries that it should at least once pick the other value from the 2 available :sweat_smile: :crossed_fingers:

PR #2986, opened 2021-08-31
bugfix: vxlan noencap blackhole contention

electricjesus:
## Description
bugfix for `fullResyncRoutesForLink` criteria on `programmedRoutes`. If your `RouteTable` has 1.) `InterfaceNone`/No-OIF 2.)  Has to operate with another `RouteTable` on the same route table number (e.g. blackhole routing and noencap routing in this example), `netlink.RouteListFiltered` will only have a single, valid filter (route table number) left to operate with and will select more routes than needed for the resync operation (clear and re-add routes, if needed / not correct)

vxlan_mgr.go:
- blackhole routing table and noencap routing table were cleaning each other's routes because they share a protocol number, table index. link index for blackhole routes are 0 so it is unusable by the programmed routes filter in `fullResyncRoutesForLink`. prevented that by adding an additional, and optional route filter: route type

route_table.go
- can now provide additional options for logging context 
- can now provide additional options for route filters in `fullResyncRoutesForLink` / `netlink.RouteListFiltered`

## Todos
- [ ] Backport (v.3.20)
- [ ] Release note

## Release Note

```release-note
bugfix: blackhole routing table with No-OIF / InterfaceNone-only is clobbering all other routes in the same routing table because if netlink.RT_FILTER_OIF is specified with a netlink.Route{LinkIndex: 0}, it will return all routes using the remaining applicable filter (netlink.RT_FILTER_TABLE / Table 254) link routes.
```

## Related Issues

Fixes https://github.com/projectcalico/calico/issues/4866

### DIFF
--- a/dataplane/linux/vxlan_mgr.go
+++ b/dataplane/linux/vxlan_mgr.go
@@ -118,8 +118,8 @@ func newVXLANManager(
 		false,
 		0,
 		opRecorder,
-		routetable.WithAddiionalLogFields(log.Fields{"route_table": "vxlan_blackhole"}),
-		routetable.WithAdditionalRouteFilters(&netlink.Route{Type: bhTarget.RouteType()}),
+		routetable.WithAdditionalLogFields(log.Fields{"route_table": "vxlan_blackhole"}),
+		routetable.WithAdditionalRouteFilter(&netlink.Route{Type: bhTarget.RouteType()}),
 	)
 
 	return newVXLANManagerWithShims(
@@ -133,8 +133,8 @@ func newVXLANManager(
 			return routetable.New(interfaceRegexes, ipVersion, vxlan, netlinkTimeout,
 				deviceRouteSourceAddress, deviceRouteProtocol, removeExternalRoutes, 0,
 				opRecorder,
-				routetable.WithAddiionalLogFields(log.Fields{"route_table": "vxlan_noencap"}),
-				routetable.WithAdditionalRouteFilters(&netlink.Route{Type: noencTarget.RouteType()}),
+				routetable.WithAdditionalLogFields(log.Fields{"route_table": "vxlan_noencap"}),
+				routetable.WithAdditionalRouteFilter(&netlink.Route{Type: noencTarget.RouteType()}),
 			)
 		},
 	)
--- a/dataplane/linux/vxlan_mgr.go
+++ b/dataplane/linux/vxlan_mgr.go
@@ -105,9 +105,6 @@ func newVXLANManager(
 		blackHoleProto = dpConfig.DeviceRouteProtocol
 	}
 
-	noencTarget := routetable.Target{Type: routetable.TargetTypeNoEncap}
-	bhTarget := routetable.Target{Type: routetable.TargetTypeBlackhole}
-
 	brt := routetable.New(
 		[]string{routetable.InterfaceNone},
 		4,
@@ -119,7 +116,7 @@ func newVXLANManager(
 		0,
 		opRecorder,
 		routetable.WithAdditionalLogFields(log.Fields{"route_table": "vxlan_blackhole"}),
-		routetable.WithAdditionalRouteFilter(&netlink.Route{Type: bhTarget.RouteType()}),
+		routetable.WithAdditionalRouteFilter(&netlink.Route{Type: syscall.RTN_BLACKHOLE}),
 	)
 
 	return newVXLANManagerWithShims(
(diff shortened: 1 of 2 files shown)


mumbley:
Any idea when this will go to release?  Trying to find the source of this issue has cost me a significant amount of time already and we really could do with the fix

neiljerram:
@electricjesus Please can you fill in the release note field in the Description, and change docs required label to docs not required?

routetable/route_table.go
```diff
@@ -188,9 +188,34 @@ type RouteTable struct {
 	conntrack         conntrackIface
 	time              timeshim.Interface
 
+	additionalRouteFilters *netlink.Route
+
 	opReporter logutils.OpRecorder
 }
 
+type RouteTableOptions func(rt *RouteTable)
```
neiljerram, line 196: One object of this type is a single option, so please s/RouteTableOptions/RouteTableOption/

routetable/route_table.go
```diff
@@ -766,6 +801,35 @@ func (r *RouteTable) createL3Route(linkAttrs *netlink.LinkAttrs, target Target)
 	return route
 }
 
+func (r *RouteTable) routeListFilterParams(linkAttrs *netlink.LinkAttrs) (*netlink.Route, uint64) {
+	var routeFilter *netlink.Route
+
+	if r.additionalRouteFilters == nil {
```
neiljerram, line 807: IIUC this is a single route filter, so please remove "s" from the name.

dataplane/linux/vxlan_mgr.go
```diff
@@ -105,6 +105,9 @@ func newVXLANManager(
 		blackHoleProto = dpConfig.DeviceRouteProtocol
 	}
 
+	noencTarget := routetable.Target{Type: routetable.TargetTypeNoEncap}
+	bhTarget := routetable.Target{Type: routetable.TargetTypeBlackhole}
```
neiljerram, line 109: I don't think we need these.  Simpler just to put `routetable.TargetType...` inline below.

dataplane/linux/vxlan_mgr.go
```diff
@@ -127,7 +132,10 @@ func newVXLANManager(
 			deviceRouteSourceAddress net.IP, deviceRouteProtocol netlink.RouteProtocol, removeExternalRoutes bool) routeTable {
 			return routetable.New(interfaceRegexes, ipVersion, vxlan, netlinkTimeout,
 				deviceRouteSourceAddress, deviceRouteProtocol, removeExternalRoutes, 0,
-				opRecorder)
+				opRecorder,
+				routetable.WithAddiionalLogFields(log.Fields{"route_table": "vxlan_noencap"}),
+				routetable.WithAdditionalRouteFilters(&netlink.Route{Type: noencTarget.RouteType()}),
```
neiljerram, line 137: Use `routetable.TargetTypeNoEncap` instead of `noencTarget.RouteType()`.

dataplane/linux/vxlan_mgr.go
```diff
@@ -115,6 +118,8 @@ func newVXLANManager(
 		false,
 		0,
 		opRecorder,
+		routetable.WithAddiionalLogFields(log.Fields{"route_table": "vxlan_blackhole"}),
+		routetable.WithAdditionalRouteFilters(&netlink.Route{Type: bhTarget.RouteType()}),
```
neiljerram, line 122: Use `routetable.TargetTypeBlackhole` instead of `bhTarget.RouteType()`

neiljerram (changes requested):
I don't yet understand why the blackhole and noencap routing tables were overwriting each other's routes.  Please can you add a detailed explanation somewhere, probably in a commit message or in the PR description?

IIUC
- the blackhole table is created with `InterfaceNone`, so I guess it would only touch routes with no OIF
- the noencap table is created with the parent of the VXLAN device, e.g. "eth0", so should only touch routes via eth0
- the vxlan table is created with the vxlan.calico device.

So why were any of these overwriting the other's routes?

dataplane/linux/vxlan_mgr.go
electricjesus: they're not the same types since Type will need something like `syscall.RTN_BLACKHOLE`... should I just hardcode the type here?

dataplane/linux/vxlan_mgr.go
electricjesus: related to my comment above https://github.com/projectcalico/felix/pull/2986#discussion_r703402564

electricjesus:
>     * the blackhole table is created with `InterfaceNone`, so I guess it would only touch routes with no OIF

unfortunately we base the old filtering through gathering link attributes by link name. and since InterfaceNone routing tables will always have LinkIndex value of 0 for both queried route and route filter -- this is unfortunate as it will select all routes on eth0 (which has LinkIndex 0). fullResyncRoutesForLink then deletes these routes before rebuilding them

>     * the noencap table is created with the parent of the VXLAN device, e.g. "eth0", so should only touch routes via eth0

So the reason why this is flapping is that when the above blackhole routing table fullResyncRoutesForLink logic selects all eth0 routes and deletes them, this noencap table later on also finds its own expected routes not present so it attempts to reprogram them in. And then we're back in the beginning.

>     * the vxlan table is created with the vxlan.calico device.

The vxlan routing table is mostly untouched so it's ok, we're not considering it as relevant for this bug/PR

neiljerram (approved):
Thanks @electricjesus , LGTM now.  I think CI should pass if you merge the latest master.

PR #3001, opened 2021-09-14
bpf: Do not write services with no endpoints

salanki:
## Description

https://github.com/projectcalico/felix/pull/2686 introduces a behavior where only node local endpoints are written to the BPF NAT table. The service entry itself is however still written, which leads to undesirable behavior if someone is trying to access the `LoadBalancer` endpoint from a Pod running on a node which does not have an endpoint for that service. The traffic is black-holed.

The simple workaround is to use the `ClusterIP` of the service when working inside the cluster. However, for many use cases it is not always possible to make special considerations for traffic originating from inside the cluster. It does make sense to allow traffic from inside the cluster to be treated the same as if it came from the internet for these.

This PR avoids programming service entries for services with:
1) No endpoints at all
2) No local endpoints in the case of an `externalTrafficPolicy: Local` service

This mitigates the black-holing behavior, letting the network route traffic to the local policy pods like if it came from outside the cluster. It also doesn't write service entries for services that have no endpoints at all, which should save some space in the map.

## Todos
- [ ] Documentation
- [ ] Backport
- [X] Release note

## Release Note

```release-note
Services with `externalTrafficPolicy: Local` are not written to the forwarding table in case they have no local endpoints. This avoids black-holing of traffic coming from inside the cluster to a LoadBalancer IP.
```

### DIFF
--- a/bpf/proxy/syncer.go
+++ b/bpf/proxy/syncer.go
@@ -544,6 +544,11 @@ func (s *Syncer) apply(state DPSyncerState) error {
 		skey := getSvcKey(sname, "")
 		eps := state.EpsMap[sname]
 
+		if len(eps) == 0 {
+			log.Debugf("Service %s not written due to no Endpoints", sname)
+			continue
+		}
+
 		err := s.applySvc(skey, sinfo, eps)
 		if err != nil {
 			return err
--- a/bpf/proxy/syncer.go
+++ b/bpf/proxy/syncer.go
@@ -487,8 +487,11 @@ func (s *Syncer) applyDerived(
 	skey = getSvcKey(sname, getSvcKeyExtra(t, sinfo.ClusterIP().String()))
 	switch t {
 	case svcTypeLoadBalancer:
-		// Handle LB services the same as NodePort type.
-		fallthrough
+		// Handle LB services the same as NodePort if it has 1 or more local endpoints
+		// otherwise fall back to cluster behavior to avoid traffic blackholeing
+		if sinfo.NodeLocalExternal() && local > 0 {
+			count = local // use only local eps
+		}
 	case svcTypeNodePort:
 		if sinfo.NodeLocalExternal() {
 			count = local // use only local eps


tomastigera (commented):
> The service entry itself is however still written, which leads to undesirable behavior if someone is trying to access the LoadBalancer endpoint from a Pod running on a node which does not have an endpoint for that service. The traffic is black-holed.

This sounds like an issue for the connect time load balancer, in which case, we still want the pod to reach the LB, right?

Otoh, this fix also affect NodePorts, where I believe the behaviour is correct.

The bug here is that we need to treat LB as NP and ignore the external IP of the LB so that it is reachable - and not to do the LB decision ourselves - which triggers the bug if there is no local backend - clearly wrong :sweat:

tomastigera:
There are probably a few ways how to deal with this

(a) do not point the external IP to the backends at all
(b) ignore the `Local` part in the in CTLB by the pods

(b) would be imo better then (a) and equally correct as it would avoid the hop to LB and back to a nodeport and would void any NAT and any possibility of forwarding (thus `Local` would be meaningless). Simply, we would always pick a backend as if it went to `ClusterIP` instead of the LB IP. Would that be a behaviour that you @salanki would expect / prefer?

salanki:
Initially, I thought that `(a)` was the best implementation. This works well for us, as our load-balancing is network based. LB IPs are allocated using `MetalLB` and announced by Calico via BGP to top of rack switches. Traffic finds it's way directly to the nodes where the Pods are running via our routing infrastructure. This has the added benefit that traffic is always routed to the node closest to the origin. Calico does not currently support [Service Topology](https://kubernetes.io/docs/concepts/services-networking/service-topology/) or [Topology Aware Hints](https://kubernetes.io/docs/concepts/services-networking/topology-aware-hints/) so it is impossible to get calico to route traffic only to endpoints within the same zone/region.

There are some drawbacks with `(a)` however.
- Egress `NetworkPolicies` break, as traffic is not directed towards a Pod IP but a Service IP when it hits the policy layer.
- For people using any other method of load balancing except for Layer 3 networking, ie ELB, it will lead to extra hops for the traffic. This might be desirable in some cases, ie. if additional policy and accounting is conducted in the LB, but generally it is likely not desired.

For general consumption `(b)` is likely the best path. I have updated this PR to follow that. It would be really neat to trigger the `(a)` behavior on a service based on an annotation.

My knowledge of the Calico codebase is not good enough to know how to cleanly implement that. I have implemented a very naive version of it by matching for a specific pattern in the service name for our use case (a hack), a proper implementation would be nice so it can be merged upstream however.

tomastigera:
> For general consumption `(b)` is likely the best path. I have updated this PR to follow that. It would be really neat to trigger the `(a)` behavior on a service based on an annotation.

Unfortunately, all we get are `HintsAnnotation returns the value of the v1.AnnotationTopologyAwareHints annotation.` as we currently use k8s proxy pkg as the "front end" for our kube-proxy.

salanki:
The new Topology Hints implementation seems significantly less flexible than Service Topology. How they could deprecate Service Topology for that beats me. We could repurpose `HintsAnnotation` to control the behavior I'm describing. If empty default to `(b)` and if set to `direct-on-no-local` `(a)` is used. This leaves the path open to implement `auto` as it is implement in `kube-proxy` if we would ever want to do that, even though I personally find that functionality rather useless.

tomastigera:
If we could set the hints, we can do anything on calico's end. It is a bummer that we cannot get service annotations or something like that with the current k8s code. Note that the hints are feature-gated and the kube-proxy code needs to explicitly enable it.

caseydavenport:
The Calico team is reorganizing our code and repository structure in order to make development easier, and as such all of the code in this repository is now located at https://github.com/projectcalico/calico. We appreciate your PR and your patience as we work to improve our developer experience, and welcome you to re-open your changes against the new location. Please reach out if you have any questions!

PR #3003, opened 2021-09-22
Fix vxlan logspam

electricjesus:
## Description

- `filterErrorByIfaceState` error could be just `nil` to prevent full resyncs at all times for a `NoOIF` interface
- `routeFilters` / `nl.RouteListFiltered` parameter construction shouldn't be using pointers

## Todos
- [ ] Backport v3.20

## Release Note

```release-note
None required
```

## Related Issues
- https://github.com/projectcalico/calico/issues/4944

### DIFF
--- a/routetable/route_table.go
+++ b/routetable/route_table.go
@@ -1140,7 +1140,7 @@ func (r *RouteTable) filterErrorByIfaceState(ifaceName string, currentErr, defau
 	if ifaceName == InterfaceNone {
 		// Short circuit the no-OIF interface name.
 		logCxt.Info("No interface on route.")
-		return defaultErr
+		return nil
 	}
 
 	if strings.Contains(currentErr.Error(), "not found") {
--- a/routetable/route_table.go
+++ b/routetable/route_table.go
@@ -188,7 +188,7 @@ type RouteTable struct {
 	conntrack         conntrackIface
 	time              timeshim.Interface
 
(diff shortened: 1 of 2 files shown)


routetable/route_table.go
```diff
@@ -1140,7 +1134,7 @@ func (r *RouteTable) filterErrorByIfaceState(ifaceName string, currentErr, defau
 	if ifaceName == InterfaceNone {
 		// Short circuit the no-OIF interface name.
 		logCxt.Info("No interface on route.")
-		return defaultErr
+		return nil
```
robbrockbank, line 1137: Shouldn't we still be returning the error here?  I assume it is the error that ensures we keep the state dirty so that we retry again.

electricjesus: It will always error/always be dirty for a NoOIF interface though as netlink will never be able to find it, i think. I had to make this change because the original log spam was every few hundred ms so this helped me see the rest of the log. 

perhaps let's log it as a warning? or something

robbrockbank: As discussed, netlink should be able to find routes with no-OIF.  An interface index of 0 implies unassigned since the indexing of interfaces starts from 0.

It is important that these errors are not swallowed for the the no-OIF case because that means Felix wouldn't know to retry the netlink request if it failed.

robbrockbank:
@electricjesus  - I think this can be closed out now.

PR #3009, opened 2021-09-23
Use v1 EndpointSlice API

caseydavenport:
## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```


neiljerram:
I can understand the "v1beta1" -> "v1" changes here. :smile:   But what is the reason for the removal of `TopologyKeys` ?

caseydavenport:
> But what is the reason for the removal of TopologyKeys ?

@neiljerram TopologyKeys was present in v1beta but removed from the API in v1 :)

For now we need to hold off on merging this because it breaks compat with k8s v1.20, but we'll merge this after a release goes by to give a suitable version skew.

PR #3013, opened 2021-09-27
bpf/proxy: exclude local workloads from NodePortRemotes

tomastigera:
## Description

Fixes the problem, where bogus entries were written to NAT table due to the fact that local workloads do not have NextHop as an IP, but as the index of the cali interface. That lead to programming bogus entries like 3.0.0.0, 4.0.0.0 etc in the NAT table.

As a side effect, those entries each had just one backend (the pod with the cali iface) and due to another bug overwrote the backends for the cleaner that thought that there is just a single backend which leads to stalls if there is more then a single local backend https://github.com/projectcalico/calico/issues/4431. This PR fixes only programming the NAT table and does not address the cleaner.

    To work around the fact that CTLB does not resolve nodeports at the
    node, but immediately at connect time, we need to know if there are any
    local pods at the destination node if ExternalTrafficPolicy is set to
    Local. Otherwise we would not be compliant with legacy kube-proxy if we
    make connetion to a non-local pod of the destination node.
    
    Note that CTLB would make such a connection with source preservation and
    no extra hop anyway.

refs https://github.com/projectcalico/calico/issues/4431
refs https://github.com/projectcalico/calico/issues/4712
refs https://github.com/projectcalico/calico/issues/4817

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/bpf/proxy/syncer.go
+++ b/bpf/proxy/syncer.go
@@ -368,8 +368,6 @@ func (s *Syncer) applySvc(skey svcKey, sinfo k8sp.ServicePort, eps []k8sp.Endpoi
 		svc:        sinfo,
 	}
 
-	s.newEpsMap[skey.sname] = eps
-
 	if log.GetLevel() >= log.DebugLevel {
 		log.Debugf("applied a service %s update: sinfo=%+v", skey, s.newSvcMap[skey])
 	}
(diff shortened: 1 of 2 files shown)


salanki:
Does this relate to https://github.com/projectcalico/felix/pull/3001 at all?

fasaxc (approved):
LGTM, maybe add a comment to explain why you check those flags?

marvin-tigera:
I'm sorry but I failed delete the branch after merging the pull request.

PR #3015, opened 2021-09-29
bpf/proxy: fix conn cleaning when ExternalTrafficPolicy=Local

tomastigera:
Our desire is to clean up connections from conntrack if we detect that a
given frontend no longer has a given backend. As a result, clients do
not need to wait for timeouts as the nodes would reply with RST for the
clients retransmits.

Because of wrong tracking which frontend may have which backend, when
ExternalTrafficPolicy=Local, we were incorrectly breaking perfectly good
connections.

This makes our decisions more conservative as we consider all currently
known backends as valid. There is no harm in not cleaning up connections
that couldn't have been started and letting them timeout if they exist.

fixes https://github.com/projectcalico/calico/issues/4431
refs https://github.com/projectcalico/calico/issues/4712
refs https://github.com/projectcalico/calico/issues/4817

## Todos
- [ ] Documentation
- [ ] Backport
- [x] Release note

## Release Note

```release-note
BPF: Fixed a bug in ExternalTrafficPolicy=Local that lead to connection stalling.
```

### DIFF
--- a/bpf/proxy/syncer.go
+++ b/bpf/proxy/syncer.go
@@ -91,7 +91,9 @@ func hasSvcKeyExtra(skey svcKey, t svcType) bool {
 }
 
 func isSvcKeyDerived(skey svcKey) bool {
-	return hasSvcKeyExtra(skey, svcTypeExternalIP) || hasSvcKeyExtra(skey, svcTypeNodePort) || hasSvcKeyExtra(skey, svcTypeLoadBalancer)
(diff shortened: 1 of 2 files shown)


salanki:
Knock on wood I think you can change that refs to fixes for https://github.com/projectcalico/calico/issues/4431 based on my preliminary testing!!

fasaxc (approved):

PR #3016, opened 2021-09-30
Replace hyperkube, update k8s version

caseydavenport:
## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```


fv/certs/admin-key.pem
```diff
@@ -0,0 +1,27 @@
+-----BEGIN RSA PRIVATE KEY-----
```
fasaxc, line 1: Github might flag a security warning on our repo for checking in private keys?

caseydavenport: We're actually already doing this for the FV tests: https://github.com/projectcalico/felix/blob/master/fv/private.key

I just added a few more. We can probably remove the old one, though. 

And of course this doesn't change the fact that GH might warn about it in the future, though....

fasaxc (approved):
LGTM

PR #3020, opened 2021-10-04
Cleanup of map struct and compilation flags

sridhartigera:
## Description
compilation flags are not right for test files as a result, both __LIBBPF_LOADER__ and __BPFTOOL_LOADER__ gets enabled, which is contradictory. This results in ut failures in ubuntu 20.04.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/bpf-gpl/arp.h
+++ b/bpf-gpl/arp.h
@@ -28,6 +28,6 @@ struct arp_value {
 	char mac_dst[6];
 };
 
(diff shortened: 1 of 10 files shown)


tomastigera (approved):
yup, that is a good clean up!

PR #3028, opened 2021-10-12
[release-v3.17] Semaphore Auto Pin Update

marvin-tigera:


marvin-tigera:
Removing "merge-when-ready" label due to new commits

PR #3046, opened 2021-10-22
bpf: optionally disable CTLB for connected udp

tomastigera:
If the selected backend disappears, client is stuck with the selection
for ever. Kube-proxy resolves the issue by cleaning up old UDP entries.
If we avoid CTLD and resort to the regular NAT, and we clean up
conntrack entry because the backend is gone (which we do), a new backend
is selected by the next packed due to conntrack miss.

This may break connectivity from host to local pods using connected udp.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```


tomastigera:
this isn't the way forward

PR #3055, opened 2021-10-29
Fix Windows FV test assertion

caseydavenport:
## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/dataplane/windows/ipsets/ipsets.go
+++ b/dataplane/windows/ipsets/ipsets.go
@@ -19,6 +19,7 @@ import (
 
 	log "github.com/sirupsen/logrus"
 
+	"github.com/projectcalico/felix/ipsets"
 	"github.com/projectcalico/libcalico-go/lib/set"
 )
 
@@ -56,7 +57,7 @@ func (s *IPSets) AddOrReplaceIPSet(setMetadata IPSetMetadata, members []string) 
 		"setID":   setMetadata.SetID,
 		"setType": setMetadata.Type,
 	}).Info("Creating IP set")
-	filteredMembers := s.filterMembers(members)
+	filteredMembers := s.filterMembers(members, setMetadata.Type)
 
 	// Create the IP set struct and stores it by id
 	setID := setMetadata.SetID
@@ -82,7 +83,7 @@ func (s *IPSets) AddMembers(setID string, newMembers []string) {
 	}
 
 	ipSet := s.ipSetIDToIPSet[setID]
-	filteredMembers := s.filterMembers(newMembers)
+	filteredMembers := s.filterMembers(newMembers, ipSet.Type)
 	if filteredMembers.Len() == 0 {
 		return
 	}
@@ -104,7 +105,7 @@ func (s *IPSets) RemoveMembers(setID string, removedMembers []string) {
 	}
 
 	ipSet := s.ipSetIDToIPSet[setID]
-	filteredMembers := s.filterMembers(removedMembers)
+	filteredMembers := s.filterMembers(removedMembers, ipSet.Type)
 	if filteredMembers.Len() == 0 {
 		return
 	}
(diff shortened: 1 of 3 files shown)


dataplane/windows/ipsets/ipsets.go
```diff
@@ -19,6 +19,7 @@ import (
 
 	log "github.com/sirupsen/logrus"
 
+	"github.com/projectcalico/felix/ipsets"
```
song-jiang, line 22: I would avoid importing `felix/ipsets` package because this package is the windows equivalent and should be at the same level of `felix/ipsets`. We could add linux specific dependencies into `felix/ipsets` later and it will break Windows build.

dataplane/windows/ipsets/ipsets.go
```diff
@@ -142,12 +143,26 @@ func (s *IPSets) GetIPSetMembers(setID string) []string {
...
+func (s *IPSets) filterMembers(members []string, setType ipsets.IPSetType) set.Set {
 	filtered := set.New()
 	wantIPV6 := s.IPVersionConfig.Family == IPFamilyV6
+
+	// IPSet members can come in two forms: IP, or IP and port.
+	// To determine the address family for an IP set member, we must first
+	// determine which type of IP set this is.
+	memberIsIPv6 := func(m string) bool {
+		addr := m
+		if setType == ipsets.IPSetTypeHashIPPort {
```
song-jiang, line 155: Can we just detect `,` and avoid using `ipsets.IPSetTypeHashIPPort`? or Just copy over the definition from `felix/ipsets`?

fv/winfv/policy_test.go
```diff
@@ -159,7 +159,7 @@ var _ = Describe("Windows policy test", func() {
 
 			// Assert that it's now reachable.
 			err = kubectlExec(`-t porter -- powershell -Command 'Invoke-WebRequest -UseBasicParsing -SkipCertificateCheck -TimeoutSec 5 https://kubernetes.default.svc.cluster.local'`)
```
song-jiang, line 161: It may require more than `-SkipCertificateCheck` for Invoke-WebRequest to return good status. I've used this script manually
https://gist.github.com/jmassardo/2e0dd7cce292f16ff8f6945b8b3752b5#file-invoke-webrequest_ignore_ssl-ps1-L2

and still hit error. However, I believe connection is successful. 

```
PS C:\> Invoke-WebRequest -UseBasicParsing -TimeoutSec 1 https://kubernetes.default.svc.cluster.local
Invoke-WebRequest : {
  "kind": "Status",
  "apiVersion": "v1",
  "metadata": {

  },
  "status": "Failure",
  "message": "forbidden: User \"system:anonymous\" cannot get path \"/\"",
  "reason": "Forbidden",
  "details": {

  },
  "code": 403
}
At line:1 char:1
+ Invoke-WebRequest -UseBasicParsing -TimeoutSec 1 https://kubernetes.d ...
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidOperation: (System.Net.HttpWebRequest:HttpWebRequest) [Invoke-WebRequest], WebExc
   eption
    + FullyQualifiedErrorId : WebCmdletWebResponseException,Microsoft.PowerShell.Commands.InvokeWebRequestCommand
PS C:\>
```

song-jiang: Maybe we should just use `nginx` service (created by infra) for WinFV test.

song-jiang (changes requested):
A few comments.

dataplane/windows/ipsets/ipsets.go
caseydavenport: I suppose we need to move the metadata constants out of that package then, although personally I'd recommend doing that at the time a breaking change is added rather than forcing an additional package now.

dataplane/windows/ipsets/ipsets.go
caseydavenport: Splitting on "," would probably work, but IMO is unnecessarily fragile - we hit the current bug because we were splitting based on ":", which isn't truly an indicator of whether a member is IPv4 or IPv6. Would rather avoid using string parsing to determine the type of IP set, when we have a `Type` field available to us!

lmm (approved):

PR #3059, opened 2021-11-02
Load CTLB using libbpf

tomastigera:
## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/bpf/nat/connecttime.go
+++ b/bpf/nat/connecttime.go
@@ -27,6 +27,7 @@ import (
 	log "github.com/sirupsen/logrus"
 
 	"github.com/projectcalico/felix/bpf"
+	"github.com/projectcalico/felix/bpf/libbpf"
 )
 
 type cgroupProgs struct {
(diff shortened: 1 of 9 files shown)


sridhartigera:
Should we modify calculate-flags to remove __BPFTOOL_LOADER__ for ctlb files?

bpf/nat/connecttime.go
```diff
@@ -84,50 +85,61 @@ func RemoveConnectTimeLoadBalancer(cgroupv2 string) error {
 	return nil
 }
```
sridhartigera, line 86: Should ```RemoveConnectTimeLoadBalancer``` as well be moved to ```libbpf``` APIs?

bpf/nat/connecttime.go
```diff
@@ -84,50 +85,61 @@ func RemoveConnectTimeLoadBalancer(cgroupv2 string) error {
...
-	if err != nil {
-		err = errors.Wrapf(err, "failed to attach program %s", progName)
-		goto out
+	baseDir := "/sys/fs/bpf/tc/globals/"
+	for m, err := obj.FirstMap(); m != nil && err == nil; m, err = m.NextMap() {
+		// In case of global variables, libbpf creates an internal map <prog_name>.rodata
+		// The values are read only for the BPF programs, but can be set to a value from
+		// userspace before the program is loaded.
+		if m.IsMapInternal() {
+			log.WithField("prog", progName).Warn("Load-time configuration not supported, using defaults.")
```
fasaxc, line 116: When do we hit this?

tomastigera: When the program has globals. I am adding the globals in a follow up PR

bpf/nat/connecttime.go
tomastigera: After some discussion, this is not straight forward and we will leave it out for now. It is only executed when felix restarts and we cannot remember links.

bpf/nat/connecttime.go
tomastigera: This is the follow up https://github.com/projectcalico/felix/pull/3046

sridhartigera (approved):

PR #3067, opened 2021-11-04
bpf: enforce affinity for unconnected UDP with CTLB

tomastigera:
Without CTLB NAT decision is made by the first packet and the
decision is then inderited by the follow up packets from conntrack until
the conntrack entry expires. Connected CTLB makes the decision only once
at connect time. CTLB for unconnected protocols was making a NAT
decision for each sendmsg() call, therefore packets from a single socket
were sprayed across all possible backends of a service. That is not
desirable as applications usually expect that subsequent packets sent to
the same destination end up at the same destination. Breaking that
assumption  may have performance or correctness impact.

Since CTLB does not store NAT information in conntrack (as it sees only
the final destination), we mimic the same behaviour by enforcing
affinity. We set the affinity timeout to 60s which is the default for
UDP not seen when cleaning conntrack. Since we update the affinity
timer for each sendmsg, its expiration behaves the same as conntrack
cleaning.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/bpf-gpl/nat.h
+++ b/bpf-gpl/nat.h
@@ -59,12 +59,14 @@ static CALI_BPF_INLINE int skb_nat_l4_csum_ipv4(struct __sk_buff *skb, size_t of
 	return ret;
 }
 
-static CALI_BPF_INLINE struct calico_nat_dest* calico_v4_nat_lookup2(__be32 ip_src,
+static CALI_BPF_INLINE struct calico_nat_dest* calico_v4_nat_lookup3(__be32 ip_src,
 								     __be32 ip_dst,
 								     __u8 ip_proto,
 								     __u16 dport,
 								     bool from_tun,
-								     nat_lookup_result *res)
+								     nat_lookup_result *res,
+								     bool affinity_always,
+								     bool affinity_tmr_update)
 {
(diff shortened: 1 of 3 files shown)


bpf-gpl/nat.h
```diff
@@ -72,12 +72,14 @@ static CALI_BPF_INLINE int skb_nat_l4_csum_ipv4(struct __sk_buff *skb, size_t of
...
-static CALI_BPF_INLINE struct calico_nat_dest* calico_v4_nat_lookup2(__be32 ip_src,
+static CALI_BPF_INLINE struct calico_nat_dest* calico_v4_nat_lookup3(__be32 ip_src,
 								     __be32 ip_dst,
 								     __u8 ip_proto,
 								     __u16 dport,
 								     bool from_tun,
-								     nat_lookup_result *res)
+								     nat_lookup_result *res,
+								     bool affinity_always,
+								     bool affinity_to_update)
```
fasaxc, line 82: `affinity_to_update` reads as "pointer to the affinity to update", I'd probabaly call this `update_affinity`

fasaxc (approved):
One naming suggestion but looks good.

PR #3069, opened 2021-11-05
bpf/ctlb: fail connect if there is no backend

tomastigera:
We may hit a race when backends are not programmed yet and if we let
TCP connect succeed, a SYN from pod will hit NAT in tc program, which
does not work in all cases. It would not work for host at all until the
TCP session expired and application retries.

This also makes it aligned with tc programs that generate ICMP response
host-unreachable in such a situation.

Connect returns EPERM in this case and application may retry immediately.

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
connect() fails for apps EPERM if there is no backed for the target service. This avoids mixing of CTLB and regular NAT in eBPF
```

### DIFF
--- a/bpf-gpl/connect_balancer.c
+++ b/bpf-gpl/connect_balancer.c
@@ -18,8 +18,9 @@
 
 #include "sendrecv.h"
 
-static CALI_BPF_INLINE void do_nat_common(struct bpf_sock_addr *ctx, __u8 proto, bool connect)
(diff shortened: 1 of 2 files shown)


fasaxc (commented):
Looks good; could do with a regression test.

fasaxc (approved):

PR #3076, opened 2021-11-16
bpf: fix ExternalTrafficPolicy=Local

tomastigera:
## Description

ExternalTrafficPolicy=Local should only apply to external traffic.

As per https://github.com/kubernetes-sigs/kpng/blob/master/doc/service-proxy.md :

> If a service has .spec.internalTrafficPolicy set to "Local", then "internal" traffic to the service (that is, traffic which comes from a pod or node in the cluster) should only be redirected to endpoint IPs on the node that the connection originated from. (If there are no endpoint IPs on that node then internal connections to the service from that node should be rejected.)
> 
> If a service has .spec.externalTrafficPolicy set to "Local", then "external" traffic to the service (that is, traffic which comes from a source outside the cluster) should only be redirected to endpoint IPs on the node where the connection first arrives in the cluster. (If there are no local endpoint IPs on that node then external connections to the service via that node should be dropped, not rejected.)

And also:

> (There is a subtle interaction here with the "short circuit" behavior of pod-to-load-balancer traffic: the behavior when a pod connects to a load balancer IP should logically be the same whether the service proxy short circuits it or not. Thus, when a pod connects to an externalTrafficPolicy: Local service via its load balancer IP, the connection must succeed even if the pod is on a node with no endpoints for that service.)

That is the primary thing that was broken. Since we treat pod traffic as cluster local traffic now, we would ignore the `Local` and pick any backend.

We do not handle `ProxyTerminatingEndpoints` yet, it is alpha in 1.22

> If the ProxyTerminatingEndpoints feature is enabled, then when generating rules for an externalTrafficPolicy: Local service, if there are no local endpoints for the service which are "ready" (ie, .endpoints[].conditions.ready is either unset or true), then the proxy should fall back to using any endpoints which are both "Serving" and "Terminating" (ie, .endpoints[].conditions.serving is either unset or true, and .endpoints[].conditions.terminating is true (not unset)).

## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
Align handling ExternalTrafficPolicy=Local with legacy kube-proxy - apply it only to external traffic. Enable InternalTrafficPolicy.
```


bpf-gpl/nat_types.h
```diff
@@ -53,9 +53,13 @@ struct calico_nat_v4_value {
...
 	__u32 local;
 	__u32 affinity_timeo;
+	__u32 flags;
 };
 
-CALI_MAP(cali_v4_nat_fe, 2,
+#define NAT_FLG_EXTERNAL_LOCAL	0x1
+#define NAT_FLG_INTERNAL_LOCAL	0x2
+
+CALI_MAP(cali_v4_nat_fe, 3,
```
fasaxc, line 62: Remembered to bump the version :+1:

fasaxc (approved):
LGTM

PR #3088, opened 2021-12-01
Write the name of the policy into iptables as a comment.

fasaxc:
## Description

Makes it easier to debug policy.
## Todos
- [ ] Documentation
- [ ] Backport
- [ ] Release note

## Release Note

```release-note
When generating iptables chains for network policies, Felix now adds a comment to the first rule in the chain with the name of the policy.
```

### DIFF
--- a/rules/policy.go
+++ b/rules/policy.go
@@ -31,11 +31,11 @@ import (
 func (r *DefaultRuleRenderer) PolicyToIptablesChains(policyID *proto.PolicyID, policy *proto.Policy, ipVersion uint8) []*iptables.Chain {
 	inbound := iptables.Chain{
 		Name:  PolicyChainName(PolicyInboundPfx, policyID),
-		Rules: r.ProtoRulesToIptablesRules(policy.InboundRules, ipVersion, fmt.Sprintf("Policy %s (ingress)", policyID.Name)),
+		Rules: r.ProtoRulesToIptablesRules(policy.InboundRules, ipVersion, fmt.Sprintf("Policy %s ingress", policyID.Name)),
 	}
 	outbound := iptables.Chain{
 		Name:  PolicyChainName(PolicyOutboundPfx, policyID),
-		Rules: r.ProtoRulesToIptablesRules(policy.OutboundRules, ipVersion, fmt.Sprintf("Policy %s (egress)", policyID.Name)),
+		Rules: r.ProtoRulesToIptablesRules(policy.OutboundRules, ipVersion, fmt.Sprintf("Policy %s egress", policyID.Name)),
 	}
 	return []*iptables.Chain{&inbound, &outbound}
 }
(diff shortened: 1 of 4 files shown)


rules/policy.go
```diff
@@ -31,32 +31,38 @@ import (
...
 	outbound = &iptables.Chain{
 		Name:  ProfileChainName(ProfileOutboundPfx, profileID),
-		Rules: r.ProtoRulesToIptablesRules(profile.OutboundRules, ipVersion),
+		Rules: r.ProtoRulesToIptablesRules(profile.OutboundRules, ipVersion, fmt.Sprintf("Profile %s (egress)", profileID.Name)),
 	}
 	return
 }
 
-func (r *DefaultRuleRenderer) ProtoRulesToIptablesRules(protoRules []*proto.Rule, ipVersion uint8) []iptables.Rule {
+func (r *DefaultRuleRenderer) ProtoRulesToIptablesRules(protoRules []*proto.Rule, ipVersion uint8, chainComments ...string) []iptables.Rule {
```
neiljerram, line 55: Why allowing for multiple comments, since it looks like callers only provide one.

rules/policy.go
```diff
@@ -31,32 +31,38 @@ import (
...
 	var rules []iptables.Rule
 	for _, protoRule := range protoRules {
 		rules = append(rules, r.ProtoRuleToIptablesRules(protoRule, ipVersion)...)
 	}
+	if len(chainComments) > 0 {
+		if len(rules) == 0 {
```
neiljerram, line 61: Does this case arise in practice?  I'd want to double check that adding the following comment-only rule doesn't change any behaviour.

neiljerram (commented):
Nice!  Couple of queries.

rules/policy.go
fasaxc: It was mainly to make it an optional argument so that the existing tests could stay as-is and avoid conflicts when merging over to private.

rules/policy.go
fasaxc: Yes, it does arise. A rule with just a comment is a no-op, packets just fall through to whatever would happen next.

Having said that, my change causes a panic due to having a nil action on the rule.  I'll fix that up.

neiljerram (approved):
LGTM

marvin-tigera:

fasaxc:
Haha, guess that FV isn't actually BPF safe :grin:

alban:
Thanks! Does it also add a comment for GlobalNetworkPolicies?

marvin-tigera:
Removing "merge-when-ready" label due to new commits

marvin-tigera:

fasaxc:
> Thanks! Does it also add a comment for GlobalNetworkPolicies?

Yes, there's only one type of policy by this point in the processing.

